PR #96, opened 2026-01-12
feat(cli): bake runtime directory into binary at build time

DorianZheng:
## Summary
- Eliminates the need for `DYLD_LIBRARY_PATH=./target/boxlite-runtime` when running boxlite-cli
- Build.rs copies runtime files to `~/.local/share/boxlite/` and bakes path into binary
- Users can still override with `BOXLITE_RUNTIME_DIR` env var at runtime

## Test plan
- [x] Build boxlite-cli: `cargo build -p boxlite-cli`
- [x] Run without env var: `./target/debug/boxlite run alpine -- echo "test"`
- [x] Verify runtime copied to `~/.local/share/boxlite/`

### DIFF
--- a/boxlite-cli/build.rs
+++ b/boxlite-cli/build.rs
@@ -10,11 +10,23 @@ fn main() {
     let manifest_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
     let project_root = manifest_dir.parent().unwrap();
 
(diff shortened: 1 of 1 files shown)


DorianZheng:
@shayne-snap Hi, would you like to review this PR

shayne-snap:
LGTM

PR #113, opened 2026-01-13
fix(cli): use ditto instead of fs::copy to preserve macOS code signing

DorianZheng:
## Summary
- Replace Rust's `fs::copy()` with macOS `ditto` command in `build.rs` to preserve code signing metadata when copying runtime files
- Use `cp -R -T` on Linux for similar behavior

## Problem
CLI and shim binaries were killed with SIGKILL (exit 137) on macOS because `fs::copy()` creates new files that lose code signing extended attributes, causing "Invalid Page" code signature validation failures at dyld load time.

The crash report showed:
```
"signal":"SIGKILL (Code Signature Invalid)"
"termination":{"namespace":"CODESIGNING","indicator":"Invalid Page"}
```

## Root Cause
- `fs::copy()` creates a new file that doesn't preserve macOS code signing metadata
- When binaries are copied from `target/boxlite-runtime/` to `~/.local/share/boxlite/`, the code signatures become invalid
- macOS kills the process at dyld load time before any code executes

## Solution
Use `ditto` on macOS which preserves extended attributes, resource forks, and code signing metadata.

## Test plan
- [x] Clean rebuild: `rm -rf ~/.local/share/boxlite && make cli`
- [x] Test CLI: `./target/debug/boxlite -h` shows help
- [x] Test shim: `~/.local/share/boxlite/boxlite-shim --help` shows help

### DIFF
--- a/boxlite-cli/build.rs
+++ b/boxlite-cli/build.rs
@@ -76,15 +76,27 @@ fn set_rpath(dest: &Path) {
 }
 
 fn copy_dir_all(src: &Path, dst: &Path) -> std::io::Result<()> {
+    use std::process::Command;
+
     fs::create_dir_all(dst)?;
-    for entry in fs::read_dir(src)? {
-        let entry = entry?;
-        let dst_path = dst.join(entry.file_name());
-        if entry.file_type()?.is_dir() {
-            copy_dir_all(&entry.path(), &dst_path)?;
-        } else {
-            fs::copy(entry.path(), &dst_path)?;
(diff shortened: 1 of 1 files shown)


DorianZheng:
@shayne-snap Hi, I found a bug when use rust std::fs::copy didn't preserve signing. Would you like to review?

boxlite-cli/build.rs
henklokote187-design: Donde esta

PR #121, opened 2026-01-14
feat: use PID file as single source of truth for process tracking

### LINKED ISSUE #118: BoxLite PID not persisted to database in detach mode
## Description

https://github.com/boxlite-ai/boxlite/pull/81
https://github.com/boxlite-ai/boxlite/pull/81#discussion_r2672518458

Hi! I was testing detach mode and noticed that `boxlite rm --force` doesn't actually kill the shim process.  The shim process becomes orphan.

After some digging, I found that the PID isn't  being saved to the database.

## Steps to reproduce

```
// Create box with detach=true
let options = BoxOptions {
    rootfs: RootfsSpec::Image("alpine:latest".into()),
    detach: true,
    auto_remove: false,
    ..Default::default()
};
let box = runtime.create(options, None).await?;
box.exec(cmd).await?;
```

```
# Check database
sqlite3 ~/.boxlite/db/boxlite.db "SELECT id, status, pid FROM box_state;"
# Output: 01KEX4BGR4ZWXEZM621T2CR1E5|running|
#                                         ^ PID is empty!
```

## Possible Root cause

In `vmm_spawn.rs`,  BoxLite PID is saved to database. But then in `box_impl.rs init_live_state()`, it reads from in-memory `self.state `(no PID), and saves to database - **overwriting** the PID.

```
// box_impl.rs line ~370
{
    let mut state = self.state.write();  // in-memory, no PID
    state.set_status(BoxStatus::Running);
    self.runtime.box_manager.save_box(&self.config.id, &state)?;  // overwrites PID
}
```

## Impact

- ` runtime.remove(id, force=true)` can't kill the shim process (no PID)
- Recovery logic sees "Running but no PID", marks as Stopped
- Shim process becomes orphan

## Suggested fix

Read latest state from database before saving:

```
{
    let mut state = self.runtime.box_manager.update_box(&self.config.id)?;
    state.set_status(BoxStatus::Running);
    self.runtime.box_manager.save_box(&self.config.id, &state)?;
    *self.state.write() = state;
}
```

This solution works on my local branch. No shim process orphan anymore.

@DorianZheng Could you help confirm if my analysis is correct?

DorianZheng:
## Summary
- Fix race condition where PID saved to database could be lost
- Write PID file in pre_exec (after fork, before exec) as single source of truth
- Database PID becomes a cache, rebuilt from file on recovery

Fixes #118

## Changes
- New `boxlite/src/jailer/common/pid.rs` - Async-signal-safe PID file writing
- New `boxlite/tests/pid_file.rs` - 27 comprehensive E2E tests
- Modified recovery in `rt_impl.rs` to use PID file
- Added `pid_file_path()` to layout

## Test plan
- [x] Unit tests pass (read_pid_file, is_process_alive, is_same_process)
- [x] Cargo check passes
- [ ] Integration tests (require VM setup)

### DIFF
--- a/boxlite/src/jailer/command.rs
+++ b/boxlite/src/jailer/command.rs
@@ -129,8 +129,9 @@ impl Jailer {
 
         let resource_limits = self.security.resource_limits.clone();
         let cgroup_procs_path = cgroup::build_cgroup_procs_path(&self.box_id);
+        let pid_file_path = self.build_pid_file_path();
 
-        pre_exec::add_pre_exec_hook(&mut cmd, resource_limits, cgroup_procs_path);
+        pre_exec::add_pre_exec_hook(&mut cmd, resource_limits, cgroup_procs_path, pid_file_path);
         cmd
     }
 
@@ -293,7 +294,8 @@ impl Jailer {
         };
 
         let resource_limits = self.security.resource_limits.clone();
-        pre_exec::add_pre_exec_hook(&mut cmd, resource_limits, None);
+        let pid_file_path = self.build_pid_file_path();
+        pre_exec::add_pre_exec_hook(&mut cmd, resource_limits, None, pid_file_path);
         cmd
     }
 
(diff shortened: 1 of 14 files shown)


shayne-snap:
LGTM

Issue #37, opened 2025-12-23
Unable to use local Docker images


liuyingduo:
Unable to use local Docker images
my image is my-python-slim:latest

<img width="830" height="183" alt="Image" src="https://github.com/user-attachments/assets/69a46eec-de90-4c79-bdd1-a7814b41b226" />

```
import asyncio
import os
from boxlite import CodeBox

# 自动获取当前项目下 project/001 的绝对路径
HOST_PROJECT_DIR = os.path.abspath("./project/001")
GUEST_WORKDIR = "/workspace"

async def main():
    # 使用 CodeBox 简化配置
    # volumes 和 working_dir 会自动传递给底层的 BoxOptions
    async with CodeBox(
        image="my-python-slim:latest",
        volumes=[{"host": HOST_PROJECT_DIR, "guest": GUEST_WORKDIR, "read_only": False}],
        working_dir=GUEST_WORKDIR
    ) as codebox:
        print(f"Box 已启动，挂载目录: {HOST_PROJECT_DIR}")

        # 在沙盒里运行 myscript.py
        print("\n正在运行 myscript.py...")
        res = await codebox.exec("python", "/workspace/myscript.py")
        
        print(f"脚本退出码: {res.exit_code}")
        if res.stdout:
            print(f"脚本输出 (stdout):\n{res.stdout}")
        if res.stderr:
            print(f"脚本错误 (stderr):\n{res.stderr}")

if __name__ == "__main__":
    asyncio.run(main())
```

but It still pulls image 
```
(excel-report) liuyingduo@ubuntu:~/study/excel_report$ python run_in_box.py 
Box 已启动，挂载目录: /home/liuyingduo/study/excel_report/project/001

正在运行 myscript.py...
Traceback (most recent call last):
  File "/home/liuyingduo/study/excel_report/run_in_box.py", line 30, in <module>
    asyncio.run(main())
  File "/usr/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
[...]

PR #134, opened 2026-01-20
feat(rootfs): Supports loading rootfs from local

uran0sH:
Loading rootfs from local paths is currently not supported. Using a local rootfs will result in the error "Direct rootfs paths not yet supported".
This commit supports loading rootfs from local paths. Users can save Docker images using the following command.
```
docker save <image_name> -o <image_name>.tar
mkdir <image_name>
tar xvf <image_name>.tar -C <image_name>
```
Then pass the folder path to Boxlite.

fix: #37

### DIFF
--- a/boxlite/src/images/archive/tar.rs
+++ b/boxlite/src/images/archive/tar.rs
@@ -30,8 +30,46 @@ pub fn extract_layer_tarball_streaming(tarball_path: &Path, dest: &Path) -> Boxl
         ))
     })?;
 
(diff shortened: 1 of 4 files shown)


DorianZheng:
@uran0sH Nice work! I have a few points would like to discuess:
1. the `RootfsSpec::RootfsPath` should expect the merged rootfs directory

DorianZheng:
@uran0sH FYI, it seems we can use `oci_spec::image::OciLayout` for this

Issue #35, opened 2025-12-21
Add CLI tool for standalone usage


DorianZheng:
## Summary
Provide a CLI tool for using BoxLite without writing code, offering a familiar Docker/Podman-like experience.

## Motivation
- Quick testing without writing code
- Shell scripting and automation
- Familiar UX for Docker/Podman users
- Debugging and development
- Demo and evaluation

### Run (create + start + exec + rm)
```bash
# Run command and exit
boxlite run alpine:latest -- echo "Hello from BoxLite"

# Interactive shell
boxlite run -it python:slim

# With options
boxlite run --cpus 2 --memory 1024 -e FOO=bar alpine:latest -- sh
```

### Box Lifecycle
```bash
# Create without starting
boxlite create --name mybox alpine:latest

# Start/stop/restart
boxlite start mybox
boxlite stop mybox
boxlite restart mybox

# Remove
boxlite rm mybox
boxlite rm -f mybox  # force remove running box
```

### Inspect and Monitor
```bash
# List boxes
boxlite ls
boxlite ls -a  # include stopped

# Inspect details
boxlite inspect mybox
boxlite inspect --format '{{.State}}' mybox

# Logs
boxlite logs mybox
boxlite logs -f mybox  # follow

# Metrics
boxlite stats mybox
```

### Execute in Running Box
```bash
boxlite exec mybox -- python script.py
boxlite exec -it mybox -- /bin/sh
```

### Images
```bash
boxlite pull python:slim
boxlite images
boxlite rmi python:slim
```

### System
```bash
boxlite info          # Runtime info
boxlite system prune  # Cleanup unused resources
```

## Output Formats
```bash
boxlite ls --format json
boxlite ls --format table  # default
boxlite inspect mybox --format yaml
```

## Implementation Notes
- Use `clap` for argument parsing
- Reuse existing BoxliteRuntime
- Consider `boxlite` binary name (or `bx` for short?)
- Support both sync and async operations

## Tasks
- [ ] Design CLI structure with clap
- [ ] Implement `run` command
- [ ] Implement `create/start/stop/rm` commands
- [ ] Implement `exec` command with TTY support
- [ ] Implement `ls/inspect/logs` commands
- [ ] Implement `pull/images` commands
- [ ] Add output format options (json, table, yaml)
- [ ] Write documentation
- [ ] Add shell completions (bash, zsh, fish)

shayne-snap:
[...]

PR #169, opened 2026-01-26
feat(cli): implement rmi command and add core integration

shayne-snap:
## Info

issue:https://github.com/boxlite-ai/boxlite/issues/35

## Key Changes
- Added `boxlite rmi `command to CLI for removing one or more images from local store.
- Implemented core removal logic in `ImageStore` with protection against:
  - Deleting images used by active or stopped boxes.
  - Deleting images with multiple names/tags (without `--force`).
  - Deleting layers shared across multiple images.
- Added unit tests and integration tests for `rmi`
- Added Isolated test context for destruction tests, sometimes it maybe cause error like this `OCI API error: You have reached your unauthenticated pull rate limit`. Pls make sure you don't reach the limit.
- added `ImageInUse`  and `ImageMultipleTags` to show better error for  cli and c SDK

## Testing

- [x] make test:rust passed
- [x] make test:cli passed

## Notes

- It's a  **DESTRUTIVE** operation. Need a lot of attention.
- Most code is core library changes and integration tests for cli `rmi`, so I didn't split it up to two PRs.


boxlite/src/images/store.rs
```diff
@@ -922,6 +955,241 @@ impl ImageStore {
...
+        RemovalPlan {
+            tags_to_remove,
+            manifest_to_delete,
+            config_to_delete,
+            layers_to_delete,
+        }
+    }
+
+    /// Check if any boxes are currently basing on the target image.
+    async fn check_box_occupancy(&self, target_digest: &str) -> BoxliteResult<()> {
```
DorianZheng: I'm afraid this might introduce data race condition. The possible execution path:

```
  Thread A: create() starts
  Thread B: rmi() → check_box_occupancy() → list_all() reads DB (snapshot at t1)
  Thread A: persist to DB (t2)
  Thread B: proceeds with deletion (didn't see Thread A's box) ❌
  Thread A: start() → ERROR: layers missing!
```

boxlite/src/images/store.rs
```diff
@@ -922,6 +955,241 @@ impl ImageStore {
...
+        let box_store = BoxStore::new(self.db.clone());
```
DorianZheng: It seems really odd that the image component has a coupling with the box component.

PR #171, opened 2026-01-26
fix(deps): macOS libkrun cross-compile toolchain

IANTHEREAL:
- Use LLVM clang/lld for libkrun init cross-compile on macOS (with BOXLITE_LIBKRUN_CC_LINUX override)
- Re-run build when BOXLITE_LIBKRUN_CC_LINUX changes and validate ld.lld --version

Developed with Codex CLI.


DorianZheng (approved):
LGTM

PR #200, opened 2026-02-03
allow the node sdk to support persistent disks

tobowers:
It looks like the rust FFI already supports the persistent disk option disk_size_gb but the node SDK didn't expose the option. It was difficult to actually manually test this so I'm not 100% sure I didn't miss something.


DorianZheng:
> It looks like the rust FFI already supports the persistent disk option disk_size_gb but the node SDK didn't expose the option. It was difficult to actually manually test this so I'm not 100% sure I didn't miss something.

you could use `make setup` && `make dev:node` to setup local build environment and build node sdk locally

PR #249, opened 2026-02-12
fix(python): clarify run_script() reads from host, add run_guest_script()

### LINKED ISSUE #235: CodeBox.run_script() reads from host filesystem, not from inside VM
## Description

`run_script(path)` reads the script file from the **host** filesystem, not from inside the VM. This is counter-intuitive and undocumented — the docs show a pattern of creating a file inside the VM and then running it with `run_script()`, which fails.

## Reproduction

```python
import asyncio
import boxlite

async def main():
    async with boxlite.CodeBox() as codebox:
        # Create a script inside the VM
        await codebox.run("""
with open('/tmp/analysis.py', 'w') as f:
    f.write('import json\\nprint(json.dumps({"status": "ok"}))\\n')
""")
        # Try to run the script from VM path
        result = await codebox.run_script("/tmp/analysis.py")
        print(result)

asyncio.run(main())
```

## Expected

`run_script("/tmp/analysis.py")` runs the script that was created inside the VM.

## Actual

```
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/analysis.py'
```

The error occurs on the **host** side, confirming `run_script()` tries to read from host filesystem.

## Environment

- boxlite 0.5.10
- macOS Apple Silicon
- Python 3.12

## Suggested fix

Either:
1. Make `run_script()` read from inside the VM (consistent with the name), or
2. Clearly document that `run_script()` reads from the host and sends content to the VM for execution

yingjunwu:
## Summary
- Clarify `run_script()` docstring: it reads from the **host** filesystem, not from inside the container
- Add `run_guest_script()` method for executing scripts that exist inside the VM
- Mirror both changes in `SyncCodeBox` to maintain API parity

Closes #235

## Dependency
**After PR #248 is merged** (which changes `run()` to return stdout only), `run_guest_script()` and `SyncCodeBox.run_guest_script()` should be updated to also return stdout only for consistency. Currently they return `stdout + stderr` to match the current `run()` behavior.

## Test plan
- [ ] Verify `run_script()` still works with host-side scripts
- [ ] Verify `run_guest_script()` works with container-side scripts
- [ ] Verify SyncCodeBox mirrors the same methods

### DIFF
--- a/sdks/python/boxlite/codebox.py
+++ b/sdks/python/boxlite/codebox.py
@@ -72,17 +72,34 @@ class CodeBox(SimpleBox):
 
     async def run_script(self, script_path: str) -> str:
         """
-        Execute a Python script file in the container.
+        Read a Python script from the **host** filesystem and execute it in the container.
+
+        The script file is read on the host side and its contents are sent to
+        the container for execution via ``python -c``. To run a script that
+        already exists inside the VM, use :meth:`run_guest_script` instead.
 
         Args:
-            script_path: Path to the Python script on the host
+            script_path: Path to the Python script on the **host** filesystem
 
         Returns:
-            Execution output as a string
+            Execution output as a string (stdout + stderr)
         """
         with open(script_path, "r") as f:
             code = f.read()
         return await self.run(code)
+
+    async def run_guest_script(self, guest_script_path: str) -> str:
+        """
+        Execute a Python script that exists inside the container.
+
+        Args:
+            guest_script_path: Path to the script inside the container
+
+        Returns:
+            Execution output as a string (stdout + stderr)
+        """
+        result = await self.exec("/usr/local/bin/python", guest_script_path)
+        return result.stdout + result.stderr
 
     async def install_package(self, package: str) -> str:
         """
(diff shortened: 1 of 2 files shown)


yingjunwu:
Self-review after revision: **This PR is now correct, with one documented dependency.**

The initial version had 3 bugs that have been fixed:

**Bug 1 — Wrong docstring.** `run_script()` docstring claimed "Returns: Execution stdout as a string" but `self.run()` on this branch returns `stdout + stderr`. Fixed to say "Execution output as a string (stdout + stderr)".

**Bug 2 — Inconsistent run_guest_script() return.** `run_guest_script()` originally returned only `result.stdout` while `run()` returns `stdout + stderr`. Fixed to return `result.stdout + result.stderr`.

**Bug 3 — SyncCodeBox not updated.** `SyncCodeBox` didn't get `run_guest_script()` or the updated `run_script()` docstring, breaking the "API mirrors async CodeBox exactly" contract. Fixed: both methods are now mirrored.

**Dependency on #248:** This PR is self-consistent against `main`. However, after PR #248 is merged (which changes `run()` to return stdout only), `run_guest_script()` in both async and sync APIs should be updated to also return stdout only. This is documented in the PR description.

PR #263, opened 2026-02-14
feat: add log and stat command

fatelei:
```
 $ boxlite logs --help
  Show logs from a box
  Usage: boxlite logs [OPTIONS] <BOX>
  Options:
    -n, --tail <TAIL>    Number of lines to show from the end [default: 0]
    -f, --follow         Follow log output
```

```
$ boxlite stats --help
  Display live metrics for a box
  Usage: boxlite stats [OPTIONS] <BOX>
  Options:
    --format <FORMAT>  Output format (table, json, yaml) [default: table]
```

### DIFF
--- a/boxlite-cli/src/cli.rs
+++ b/boxlite-cli/src/cli.rs
@@ -90,7 +90,7 @@ pub enum Commands {
     /// Show logs from a box
     Logs(crate::commands::logs::LogsArgs),
 
-    /// Display live metrics for a box
+    /// Display resource usage statistics for a box
     Stats(crate::commands::stats::StatsArgs),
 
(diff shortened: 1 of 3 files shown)


fatelei:
> Hi, great work!
> 
> Quick confirmation: `boxlite logs` currently reads `console.log `(as I know it's guest/runtime console output), not application stdout/stderr like `docker logs/podman logs`. Is this the intended behavior for now, or should `boxlite logs `eventually represent persisted app logs semantics?

The current behavior is intended for debugging the box environment itself (kernel + agent), while persisted application logging is a feature that would need to be added to the runtime.

Issue #33, opened 2025-12-21
Support health checks


DorianZheng:
## Summary
Add health check support to monitor box health and detect failures.

## Motivation
- Detect unresponsive applications inside boxes
- Enable automatic recovery when combined with restart policies
- Standard container feature expected by users

### Rust
```rust
let options = BoxOptions {
    healthcheck: Some(Healthcheck {
        command: vec!["curl", "-f", "http://localhost:8080/health"],
        interval: Duration::from_secs(30),
        timeout: Duration::from_secs(10),
        retries: 3,
        start_period: Duration::from_secs(60),
    }),
    ..Default::default()
};

// Query health status
let info = box.info()?;
match info.health_status {
    HealthStatus::Healthy => println!("Box is healthy"),
    HealthStatus::Unhealthy => println!("Box failed health check"),
    HealthStatus::Starting => println!("Health check starting..."),
    HealthStatus::None => println!("No health check configured"),
}
```

### Python
```python
box = SimpleBox(
    image="python:slim",
    healthcheck={
        "command": ["python", "-c", "import requests; requests.get('http://localhost:8080/health')"],
        "interval": 30,
        "timeout": 10,
        "retries": 3,
    }
)

info = await box.info()
print(f"Health: {info.health_status}")
```

## Health States
| State | Description |
|-------|-------------|
| `None` | No health check configured |
| `Starting` | Within start_period, not yet checked |
| `Healthy` | Last health check passed |
| `Unhealthy` | Failed `retries` consecutive checks |

## Integration
- When unhealthy + restart policy set → trigger restart
- Expose health status in `box.info()` and metrics
- Log health check failures

## Tasks
- [ ] Add Healthcheck struct and HealthStatus enum
- [ ] Implement health check executor (runs command in box)
- [ ] Add background health check loop
- [ ] Store health status in BoxState
- [ ] Integrate with restart policies
- [ ] Update Python SDK
- [ ] Add tests and documentation

PR #266, opened 2026-02-14
feat: add health check support

uran0sH:
- Add HealthCheckConfig to BoxOptions for configuring health checks
- Add HealthState and HealthStatus types to track box health
- Implement health check task with start_period grace period
- Skip health checks during start_period to avoid false failures
- Expose health status through BoxInfo.health_status field
- Add health status state management (starting/healthy/unhealthy)
- Track consecutive failures and mark box as unhealthy after retries

This is an initial framework with basic functionality:
- Health check configuration and execution
- Start period grace period logic
- Health state tracking and status reporting

issue: #33

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -438,6 +438,7 @@ name = "boxlite-go-bridge"
 version = "0.1.0"
 dependencies = [
  "boxlite",
+ "boxlite-ffi",
  "boxlite-shared",
  "libc",
  "once_cell",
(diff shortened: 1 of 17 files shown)


uran0sH:
Hi @DorianZheng, if you have time, plz review this pr. Thanks

boxlite/src/runtime/options.rs
```diff
@@ -292,6 +302,67 @@ pub struct ExportOptions {}
 #[derive(Debug, Clone, Default)]
 pub struct CloneOptions {}
 
+/// Health check configuration for boxes.
+///
+/// Defines how to periodically check if a box's guest agent is responsive.
+/// Similar to Docker's HEALTHCHECK directive.
+#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
+pub struct HealthCheckConfig {
```
DorianZheng: how about name as `HealthCheckOptions` and move to `AdvancedBoxOptions` since the naming style of user-facing arguments should be consistent and normally user should not modify the default value of this option

DorianZheng (approved):

PR #281, opened 2026-02-18
fix: restore runtime backend compile path and binary resolution

DorianZheng:
## Summary
- fix local-only clone/export/snapshot/portability paths after backend trait abstraction
- add backend/runtime helpers to fail explicitly on non-local backends
- fix runtime binary lookup priority to prefer `target/boxlite-runtime` in local dev/tests (prevents unsigned shim selection on macOS)
- harden fd-closing tests by running destructive fd operations in forked child processes
- align jailer path-access tests with current shared/mounts layout behavior
- update lifecycle integration tests to explicitly `start()` before asserting active/stopped states

## Verification
- `make runtime-debug`
- `cargo check -p boxlite`
- `make test:rust`
- `cargo test -p boxlite --test execution_shutdown test_all_waits_return_on_stop -- --test-threads=1 --nocapture`


DorianZheng:
> The command to run integration tests now is: `BOXLITE_RUNTIME_DIR=$(pwd)/target/boxlite-runtime \ LD_LIBRARY_PATH=$(pwd)/target/boxlite-runtime:$LD_LIBRARY_PATH \ cargo test -p boxlite --test execution_shutdown test_wait_behavior_on_box_stop -- --test-threads=1 --nocapture`. After merging this PR, is it no longer necessary to configure environment variables?

Hi @uran0sH. The integration tests issue will be fixed ASAP. The root cause is I accidentally set `BOXLITE_RUNTIME_DIR` to `OUT_DIR=runtime`. However, `OUT_DIR=runtime` doesn't contain `boxlite-guest` which leads to binary not found issue. This PR haven't fixed it yet. Will file another PR for the test

Issue #358, opened 2026-03-09
/dev/vdb missing inside VM — guest init does not mount devtmpfs


yan5xu:
## Symptom

`/dev/vdb` (the 10 GB container disk) does not exist inside the VM even though the kernel sees both disks in `/proc/partitions`. Container.Init fails to mount the container disk, so `apt-get install` and other container writes land on the tiny 256 MB guest rootfs (`/dev/vda`) instead.

## Root Cause

The guest init process (`boxlite-guest`) mounts tmpfs on `/tmp`, `/var/tmp`, and `/run` via `mount_essential_tmpfs()` in `guest/src/mounts.rs`, but **never mounts devtmpfs at `/dev`**. Without devtmpfs, the kernel has no mechanism to auto-populate block device nodes — `/dev/vda` exists only because the rootfs image has it baked in, but `/dev/vdb` has no pre-existing node.

## Impact

- **Container disk is never mounted.** `BlockDeviceMount::mount()` in `guest/src/storage/block_device.rs` correctly checks `device.exists()` and returns `Storage("Block device not found: /dev/vdb")`.
- **Container writes go to guest rootfs.** The 256 MB vda fills up quickly, causing `apt-get` and similar tools to fail with ENOSPC.
- **Affects all boxes** that use a separate container disk (which is the default configuration).

## Fix

Mount devtmpfs at `/dev` early in guest startup (before the gRPC server starts). The kernel will then auto-populate all block device nodes discovered via virtio-blk.

```rust
mount(Some("devtmpfs"), "/dev", Some("devtmpfs"), MS_NOSUID, Some("mode=0755"))
```

This should run before the existing tmpfs mounts and before `GuestServer::run()`.

yan5xu:
> Hi [@yan5xu](https://github.com/yan5xu). From my test, the `/dev/vdb` exists inside the vm
> 
> the command:
> 
> ```
> ./target/debug/boxlite create --name "cli-test" "alpine:latest"
> ./target/debug/boxlite exec -e BOXLITE_EXECUTOR=guest "cli-test" -- ls -alrt /dev
> ```
> 
> the output:
> 
> ```
> total 4
> crw-rw-rw-  1 root root   1,   5 Jan  1  1970 zero
> crw-------  1 root root  10, 257 Jan  1  1970 vtpmx
> crw-------  1 root root 245,   0 Jan  1  1970 vport2p0
> brw-------  1 root root 254,  16 Jan  1  1970 vdb
[...]

PR #360, opened 2026-03-09
fix(serve,guest): volumes support, debugfs quoting, devtmpfs mount

yan5xu:
## Summary

Three fixes for BoxLite serve mode:

1. **serve: add volumes to CreateBoxRequest** — Without this, virtiofs mounts passed in the create request were silently dropped. Containers couldn't see host-mounted directories (e.g. `/clip`).

2. **disk/ext4: quote host path in debugfs write** — On macOS, the embedded runtime extracts to `~/Library/Application Support/boxlite/...`. The space in "Application Support" caused `debugfs write` to silently fail, leaving the guest binary missing from the rootfs ext4 image.

3. **guest: mount devtmpfs at /dev** — The kernel sees block devices in `/proc/partitions` but `/dev/vd*` device nodes were never created. This prevented Container.Init from mounting the container disk.

## Test plan

- [x] `boxlite serve` → create box with volumes → exec `/bin/echo hello` → stdout received
- [x] Guest binary correctly injected into rootfs (no "No such file or directory")
- [x] Container sees virtiofs volumes at expected mount points
- [x] `apt-get install curl git python3` works inside container (10GB disk)
- [x] HTTPS connectivity verified (`curl https://www.google.com`)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -198,14 +198,14 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "edca88bc138befd0323b20752846e6587272d3b03b0343c8ea28a6f819e6e71f"
 dependencies = [
  "async-trait",
- "axum-core",
+ "axum-core 0.4.5",
  "bytes",
  "futures-util",
  "http",
  "http-body",
  "http-body-util",
  "itoa",
- "matchit",
+ "matchit 0.7.3",
  "memchr",
  "mime",
  "percent-encoding",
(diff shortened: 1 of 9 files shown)


DorianZheng:
Two blockers:

1. The volume support is implemented in the old boxlite-cli/src/commands/serve.rs rather than the current src/cli server. It needs to be ported to the right location before we can merge.

2. The devtmpfs change in guest/src/mounts.rs:44-90 now hard-fails guest boot on mount error with no recovery. That's a risky change without an integration test proving a box with volumes actually boots and finds /dev/vdb.

Also the debugfs quoting fix only wraps the host path — embedded quotes are still fragile.

G4614:
Hi @yan5xu — checking in on this since it's been ~38 days since the last update and the devtmpfs hunk inside is blocking #358 (every box with a separate container disk silently never mounts vdb, guest writes fall through to the 256 MB rootfs, ENOSPC on the first `apt-get install`).

Would you be open to splitting the devtmpfs change back into its own focused PR — basically restoring the shape of #359 — so it can land at its own pace while the serve `volumes` plumbing and the macOS debugfs path-quoting parts continue review here in #360? Each of the three fixes has independent review surface (guest mounts vs CLI serve vs build/disk tooling), so unbundling shouldn't lose anything substantive.

Happy to do the mechanical split into a separate branch + PR off of this one if it saves you time — just say the word. Otherwise great if you want to drive it yourself; mainly flagging #358 is a real user-visible bug and the fix has been sitting here ready for over a month.

PR #361, opened 2026-03-11
fix: use Vec<u8> for exec stdout/stderr to prevent UTF-8 corruption

yan5xu:
## Summary

- `String::from_utf8_lossy()` in `portal/interfaces/exec.rs` was replacing incomplete UTF-8 byte sequences at gRPC chunk boundaries with U+FFFD replacement characters
- This caused Chinese text corruption when stdout output was large enough to be split across multiple gRPC chunks (e.g. a 130KB JSON response)
- Changed stdout/stderr channels from `String` to `Vec<u8>` throughout the pipeline to preserve raw bytes

## Changes

- `portal/interfaces/exec.rs`: `route_output` sends raw `chunk.data` (Vec<u8>) instead of `String::from_utf8_lossy` conversion
- `litebox/exec.rs`: `ExecStdout`/`ExecStderr` stream items changed from `String` to `Vec<u8>`
- `serve.rs`: base64-encodes raw bytes directly (was `.as_bytes()` on String)
- `terminal/mod.rs`: writes raw bytes to stdout/stderr (was `.as_bytes()` on String)

## Root cause analysis

```
Process stdout (valid UTF-8) 
  → gRPC chunk splits at arbitrary byte boundary (mid-character)
  → String::from_utf8_lossy replaces trailing incomplete bytes with U+FFFD
  → downstream receives corrupted text
```

Verified: same Go binary producing 130KB JSON output had 6 U+FFFD in DB (pre-existing) but 11 after passing through BoxLite serve — the extra 5 were introduced by `from_utf8_lossy`.

## Test plan

- [x] `cargo build --release -p boxlite-cli` passes
- [x] Verified: `pinix invoke get-topic` now returns exactly 6 U+FFFD (matching DB), down from 11

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -198,14 +198,14 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "edca88bc138befd0323b20752846e6587272d3b03b0343c8ea28a6f819e6e71f"
 dependencies = [
  "async-trait",
- "axum-core",
+ "axum-core 0.4.5",
  "bytes",
  "futures-util",
  "http",
  "http-body",
  "http-body-util",
  "itoa",
- "matchit",
+ "matchit 0.7.3",
  "memchr",
  "mime",
  "percent-encoding",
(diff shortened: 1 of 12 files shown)


DorianZheng:
Two blockers:

1. Changing ExecStdout/ExecStderr from String to Vec<u8> in litebox/exec.rs:321-355 is a breaking public API change. There's no migration plan or version bump here — we need to figure out that story before merging.

2. The SSE format change (base64-wrapped JSON events) in boxlite-cli/src/commands/serve.rs:433-489 is implemented against the old pre-refactor server stack. It's not mergeable with the current src/cli server and will silently break old clients without versioning. This probably needs to be split into separate PRs or rebased onto the current stack.

PR #406, opened 2026-03-26
fix(jailer): replace hardcoded FD limits with dynamic closure strategies

lilongen:
## Summary

- Replace hardcoded FD upper bounds (1024 on Linux, 4096 on macOS) with dynamic limits from `getrlimit(RLIMIT_NOFILE)`, fixing FD leakage on systems with raised `ulimit -n`
- Add `/proc/self/fd` enumeration via raw `getdents64` syscall as an efficient middle strategy on Linux (zero heap allocation, async-signal-safe)
- Restructure `close_fds_from()` into a 3-strategy cascade: `close_range` → `/proc/self/fd` → brute-force with dynamic limit

## Problem

The FD cleanup in the `pre_exec` hook used hardcoded upper bounds:
- Linux: `for fd in first_fd..1024`
- macOS: `for fd in first_fd..4096`

On production systems with `ulimit -n 65536` or higher, any FDs above these limits leaked into the jailed process, potentially exposing credentials, database connections, or network sockets inherited from the parent.

### Linux — 3-strategy cascade

| Strategy | Condition | Complexity | Description |
|----------|-----------|------------|-------------|
| 1. `close_range` | Linux 5.9+ | O(1) | Kernel closes all FDs in range (already existed) |
| 2. `/proc/self/fd` | `/proc` mounted | O(open FDs) | Enumerate via raw `getdents64`, close only open FDs |
| 3. Brute-force | Fallback | O(ulimit) | Close `first_fd..getrlimit(RLIMIT_NOFILE)` |

### macOS

Brute-force close with dynamic limit from `getrlimit(RLIMIT_NOFILE)` (replaces hardcoded 4096).

### Key constraints

All operations are **async-signal-safe** (required for `pre_exec` context):
- No heap allocation — stack buffer for `getdents64`, byte literals for paths
- No `io::Error`, `String`, `Vec`, `CString` — raw syscalls only
- `getrlimit` is POSIX async-signal-safe
- `RLIM_INFINITY` handled by capping to 1,048,576 (Linux `nr_open` default)

## Test plan

- [x] `test_close_high_numbered_fd` — creates FD 2000 (above old limits), verifies closure
- [x] `test_get_max_fd_returns_positive` — validates dynamic limit query
- [x] `test_parse_fd_from_name` (Linux) — covers valid FDs, `.`/`..`, empty, i32 overflow
- [x] All 4 existing tests continue to pass
- [x] `cargo clippy -D warnings` — zero warnings on both platforms
- [x] `cargo fmt --check` — clean
- [x] Tested on macOS ARM64 (6/6 passed) and Linux ARM64 via Lima (7/7 passed)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file descriptor cleanup on Linux and macOS.
  * Fixed cleanup for high-numbered file descriptors beyond previous limits.
  * Added more reliable fallback handling when optimized cleanup methods are unavailable.
  * Improved compatibility with newer Linux systems through more efficient descriptor cleanup.

### DIFF
--- a/src/boxlite/src/jailer/common/fd.rs
+++ b/src/boxlite/src/jailer/common/fd.rs
@@ -6,6 +6,160 @@
 //!
 //! Only the async-signal-safe `close_inherited_fds_raw()` is used,
 //! called from the `pre_exec` hook before exec().
+//!
+//! ## Strategy (in order of preference)
+//!
+//! 1. **Linux 5.9+**: `close_range(first_fd, ~0U, 0)` — O(1), kernel closes all.
+//! 2. **Linux < 5.9**: Enumerate `/proc/self/fd` via raw `getdents64` syscall
+//!    (no memory allocation) and close only open FDs. Falls back to brute-force
+//!    with a dynamic upper bound from `getrlimit(RLIMIT_NOFILE)`.
+//! 3. **macOS**: Brute-force close with dynamic upper bound from
+//!    `getrlimit(RLIMIT_NOFILE)` instead of a hardcoded limit.
+
+/// Query the soft limit for `RLIMIT_NOFILE`. Async-signal-safe.
+///
+/// Returns the soft limit (current max open files), or a safe default
+/// if the query fails. `getrlimit` is async-signal-safe per POSIX.
+///
+/// Note: Cannot reuse `rlimit::get_rlimit()` because it returns `io::Error`
+/// which heap-allocates (not async-signal-safe for pre_exec context).
+fn get_max_fd() -> i32 {
+    let mut rlim = libc::rlimit {
+        rlim_cur: 0,
+        rlim_max: 0,
+    };
+    // SAFETY: getrlimit is async-signal-safe per POSIX. rlim is a valid stack-allocated struct.
+    let result = unsafe { libc::getrlimit(libc::RLIMIT_NOFILE, &mut rlim) };
+    if result == 0 && rlim.rlim_cur > 0 {
+        if rlim.rlim_cur < i32::MAX as u64 {
+            rlim.rlim_cur as i32
+        } else {
+            // rlim_cur is RLIM_INFINITY or very large; cap to a safe maximum.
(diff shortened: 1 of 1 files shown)


DorianZheng:
Same getdents64 bug as #407: in fd.rs:71-118, nread < 0 is treated as success so `close_fds_via_proc()` returns `true` on failure and the getrlimit-based fallback never kicks in. Fix is to return `false` when getdents64 fails. Would also be good to add a test that exercises this error path.

lilongen:
Addressed all review feedback in commit f90c941:

**@Copilot + @DorianZheng** — `close_fds_via_proc` returns `false` on `getdents64` failure
- Split `nread <= 0` into two checks: `nread < 0` (error → return `false`, triggers brute-force fallback) and `nread == 0` (normal EOF)

**@Copilot** — Off-by-one in `get_max_fd()`
- Changed `< i32::MAX` to `<= i32::MAX` so exact `i32::MAX` soft limit is handled correctly

**@Copilot** — Fallback too low when `getrlimit` fails
- Raised from `1024` to `1_048_576` to preserve security invariant

**@DorianZheng** — Added test for fallback path
- New `test_brute_force_fallback_closes_fds` exercises `brute_force_close_fds()` directly

All tests pass, clippy clean.

lilongen:
@DorianZheng rebased onto current main and added a test for the `getdents64` error path you asked about (933814cd):

- `test_proc_enumeration_reports_getdents64_failure` — in a forked child, a seccomp filter makes `getdents64` return EPERM; `close_fds_via_proc` must return `false`.
- `test_close_fds_from_falls_back_when_getdents64_fails` — denies both `close_range` and `getdents64`; `close_fds_from(3)` must still close an inherited fd via the brute-force fallback.

Checked both directions: with the pre-fix `nread <= 0 → break / return true` loop restored, both fail (child exit 2: enumeration reported success; exit 4: fd left open); with the fix they pass (`make test:unit:rust FILTER=fd::` → 10 passed). Also switched the fd-name parser tests to `c"…"` literals for the newer `clippy::manual_c_str_literals`.

PR #407, opened 2026-03-26
feat(vmm): replace ProcessMonitor polling with pidfd/kqueue event-driven detection

lilongen:
## Summary

- Replace `ProcessMonitor::wait_for_exit()` 500ms sleep-based polling loop with platform-native event-driven process exit detection
- **Linux**: `pidfd_open()` (kernel 5.3+) wrapped in `tokio::io::unix::AsyncFd` for near-zero latency
- **macOS**: `kqueue` + `EVFILT_PROC` + `NOTE_EXIT` wrapped in `AsyncFd`
- **Fallback**: 100ms polling for older kernels (Linux < 5.3) when event-driven setup fails
- Fixes violation of project Rule #15: "No Sleep for Events"

## Motivation

`ProcessMonitor::wait_for_exit()` is used in production at `guest_connect.rs` inside a `tokio::select!` race to detect VM subprocess death during startup. The 500ms polling loop added up to 500ms latency to crash detection. With `pidfd`/`kqueue`, detection is near-instantaneous.

## Design

```
wait_for_exit()
  ├── Fast path: try_wait() → already dead? return immediately
  ├── Linux: pidfd_open() + AsyncFd::readable().await
  ├── macOS: kqueue(EVFILT_PROC, NOTE_EXIT) + AsyncFd::readable().await
  └── Fallback: 100ms polling loop (for kernel < 5.3)
```

Key safety decisions:
- `OwnedFd` wraps raw FDs immediately after creation — all error paths are leak-free by construction
- `fcntl(O_NONBLOCK)` return checked; falls back to polling on failure (required by `AsyncFd`)
- Block scope for `kevent` struct (contains `*mut c_void`, not `Send`) — dropped before `.await`
- Best-effort race guard via `is_alive()` after FD setup, before `.await`
- No new crate dependencies — uses existing `libc` + `tokio` (already in `Cargo.toml`)

## Test plan

- [x] New async test: `test_wait_for_exit_captures_exit_code` — verifies exit code through event-driven path
- [x] New async test: `test_wait_for_exit_detects_delayed_exit` — spawns child with 200ms delay, verifies prompt detection
- [x] New async test: `test_wait_for_exit_already_dead_pid` — non-existent PID returns `Unknown` immediately
- [x] All 598 existing tests pass on macOS ARM64
- [x] All process tests (31/31) pass on Linux ARM64 (via Lima VM)
- [x] Clippy clean (zero warnings) on both platforms
- [x] `cargo fmt --check` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved process exit detection for faster and more reliable status updates.
  * Added resilient fallback handling when native exit notifications are unavailable.
  * Improved cleanup of inherited file descriptors, including high-numbered descriptors, reducing resource leaks during process startup.
  * Enhanced handling of processes that have already exited or were terminated by signals.
  * Improved reliability when monitoring multiple processes concurrently.

### DIFF
--- a/src/boxlite/src/jailer/common/fd.rs
+++ b/src/boxlite/src/jailer/common/fd.rs
@@ -12,35 +12,76 @@
 //! 1. **Linux 5.9+**: `close_range(first_fd, ~0U, 0)` — O(1), kernel closes all.
 //! 2. **Linux < 5.9**: Enumerate `/proc/self/fd` via raw `getdents64` syscall
 //!    (no memory allocation) and close only open FDs. Falls back to brute-force
-//!    with a dynamic upper bound from `getrlimit(RLIMIT_NOFILE)`.
+//!    with a dynamic upper bound queried via raw `prlimit64` syscall.
 //! 3. **macOS**: Brute-force close with dynamic upper bound from
-//!    `getrlimit(RLIMIT_NOFILE)` instead of a hardcoded limit.
+//!    `getrlimit(RLIMIT_NOFILE)`.
 
-/// Query the soft limit for `RLIMIT_NOFILE`. Async-signal-safe.
+/// Safe cap for brute-force FD closure (2^20), matches Linux /proc/sys/fs/nr_open.
+/// Used when the rlimit query fails or returns an excessively large value.
+const MAX_FD_CAP: i32 = 1_048_576;
+
+/// Convert rlimit soft/hard values to a capped i32 upper bound for FD iteration.
 ///
-/// Returns the soft limit (current max open files), or a safe default
-/// if the query fails. `getrlimit` is async-signal-safe per POSIX.
-///
-/// Note: Cannot reuse `rlimit::get_rlimit()` because it returns `io::Error`
-/// which heap-allocates (not async-signal-safe for pre_exec context).
-fn get_max_fd() -> i32 {
-    let mut rlim = libc::rlimit {
-        rlim_cur: 0,
-        rlim_max: 0,
-    };
-    // SAFETY: getrlimit is async-signal-safe per POSIX. rlim is a valid stack-allocated struct.
-    let result = unsafe { libc::getrlimit(libc::RLIMIT_NOFILE, &mut rlim) };
-    if result == 0 && rlim.rlim_cur > 0 {
-        if rlim.rlim_cur < i32::MAX as u64 {
-            rlim.rlim_cur as i32
-        } else {
-            // rlim_cur is RLIM_INFINITY or very large; cap to a safe maximum.
-            // 1048576 (2^20) matches Linux's default /proc/sys/fs/nr_open.
-            1_048_576
-        }
+/// Uses `max(soft, hard)` to cover FDs opened before a soft limit decrease.
+/// Caps excessively large values (including RLIM_INFINITY) to [`MAX_FD_CAP`].
+/// Async-signal-safe: pure arithmetic, no allocation.
+fn rlimit_to_max_fd(soft: u64, hard: u64) -> i32 {
+    let limit = soft.max(hard);
+    if limit > 0 && limit <= i32::MAX as u64 {
+        limit as i32
     } else {
-        // getrlimit failed or returned 0; safe default per POSIX OPEN_MAX
-        1024
+        // limit is 0, RLIM_INFINITY, or exceeds i32::MAX
+        MAX_FD_CAP
+    }
+}
+
+/// Query the upper bound for open FDs. Async-signal-safe.
+///
+/// On Linux, uses a raw `prlimit64` syscall to avoid libc wrappers that may
+/// not be async-signal-safe in all implementations. On other platforms, uses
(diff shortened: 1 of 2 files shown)


DorianZheng:
Hi @lilongen. Would you mind add some test cases?

lilongen:
> Hi @lilongen. Would you mind add some test cases?

no problem!

lilongen:
add some test cases ...

lilongen:
## Added 3 test cases per review feedback

**Commit:** 43d1a98

### 1. `test_wait_for_exit_signal_kill` (HIGH value)
Spawns `sh -c "sleep 60"`, sends SIGKILL via `libc::kill()`, verifies `ProcessExit::Code(128 + 9)`. Covers the `WIFSIGNALED` decode path through event-driven detection — existing tests only covered normal exits.

### 2. `test_wait_for_exit_concurrent_monitors` (HIGH value)
Spawns 3 children with different exit codes (1, 2, 3). Creates 3 `ProcessMonitor`s, `tokio::join!` all three `wait_for_exit()` calls. Verifies each gets the correct exit code. Tests concurrent pidfd/kqueue FD registrations.

### 3. `test_wait_polling_captures_exit_code` (MEDIUM value)
Calls `wait_polling()` directly (private `async fn`, accessible from same-module tests). Spawns `sh -c "exit 5"`, verifies exit code. Validates the 100ms polling fallback path independently of event-driven setup.

### Verification
- macOS ARM64: 26/26 process tests pass, clippy clean, fmt clean
- Linux ARM64 (Lima): 26/26 process tests pass

DorianZheng:
One blocker: in jailer/common/fd.rs:71-118, when `getdents64` returns a negative value we're still returning `true` from `close_fds_via_proc()`. That makes the /proc strategy look like it succeeded, so the `getrlimit(RLIMIT_NOFILE)` fallback never runs and we silently leak inherited FDs. Need to return `false` on negative nread.

lilongen:
Addressed all review feedback in commit 671bc0b:

### fd.rs (6 fixes)

**@DorianZheng (blocker) + @Copilot** — `close_fds_via_proc` returns `false` on `getdents64` failure
- Split `nread <= 0` into: `nread == 0` (EOF), `nread < 0` with EINTR retry, other errors → `return false`

**@Copilot** — Use raw syscalls for async-signal-safety
- `get_max_fd()`: replaced `libc::getrlimit` with raw `SYS_prlimit64` on Linux
- `close_fds_via_proc()`: replaced `libc::open` with raw `SYS_openat`

**@Copilot** — `brute_force_close_fds` may miss high FDs after soft limit decrease
- Now uses `max(rlim_cur, rlim_max)` via `rlimit_to_max_fd()` helper

**@Copilot** — Fallback too low / off-by-one
- Fallback raised from `1024` to `1_048_576` (MAX_FD_CAP)
- Fixed off-by-one: `<= i32::MAX` via `rlimit_to_max_fd()`

**Tests**: Added `test_rlimit_to_max_fd` and `test_brute_force_fallback_closes_fds`

### process.rs (5 fixes)

**@Copilot** — `fcntl(F_SETFD, FD_CLOEXEC)` return unchecked
- Now checked on both pidfd and kqueue FDs, returns `None` on failure

**@Copilot** — `fcntl(F_SETFL, O_NONBLOCK)` overwrites existing flags
- Changed to `F_GETFL` + OR pattern on both pidfd and kqueue FDs

**@Copilot** — `libc::kill` return unchecked in test
- Added `assert_eq!` on kill return value in `test_wait_for_exit_signal_kill`

All 631 tests pass, clippy clean, fmt clean.

uran0sH:
`wait_for_exit` is only be used in `wait_for_guest_ready`. So I think it can use `tokio::process::Child`, like #451. We don't need to encapsulate it ourselves.

lilongen:
Closing this in favour of #451.

@boscowy you pointed here at #451, and on reflection that is the right call for
the paths it covers. This PR replaces 100 ms polling with pidfd/kqueue, which is
an optimisation — nobody has reported exit-detection latency as a problem, so it
is not worth competing with your own implementation for it.

One note worth leaving on the record, in case #451 later reaches the attach path:
`wait_for_exit` has a second caller besides `wait_for_guest_ready` — the box
watcher awaits it in both arms of its `select!` (`litebox/watcher.rs:149`,
`:161`). That caller cannot use `tokio::process::Child`, because it never
spawned the process it watches: `ProcessMonitor` is built from a bare pid
(`util/process.rs:59`), and `arm_watcher` takes that pid out of persisted state
(`litebox/box_impl.rs:213`) precisely so the runtime can adopt a box that is
already running — the `boxlite serve`-after-a-restart case its doc comment
describes. `try_wait` already encodes the same fact, treating `ECHILD` as "dead,
but we are not the parent" (`util/process.rs:88`). That matches your own note in
#451 that the attach path keeps PID-based polling.

The `jailer/common/fd.rs` change carried here belongs to #406, which stands on
its own and is unaffected by this.

PR #409, opened 2026-03-27
feat(portal): stream file uploads to reduce memory from O(file_size) to O(5 MiB)

lilongen:
## Summary

- Replace full-file `Vec<UploadChunk>` buffering in `upload_tar` with a bounded `mpsc::channel(4)` + spawned reader task, capping peak memory at ~5 MiB regardless of file size
- Extract `stream_file_chunks` helper (accepts `impl AsyncRead`) for testability and single-responsibility
- Use `std::mem::take` for `dest_path`/`container_id` (first chunk only, zero-copy on subsequent chunks)
- Always await reader `JoinHandle` before checking gRPC result to surface root-cause errors first
- Return error on receiver drop instead of silent `Ok(())` to detect incomplete uploads

## Motivation

`upload_tar` previously read the entire tar file into a `Vec<UploadChunk>` before streaming to the guest. For a 500 MB file, this consumed ~500 MB of heap. The `download_tar` path in the same file already used streaming correctly — this PR closes the asymmetry.

| Scenario | Before | After |
|----------|--------|-------|
| 500 MB upload | ~500 MB heap | ~5 MiB |
| 1 GB upload | ~1 GB heap | ~5 MiB |

## Test plan

- [x] `cargo fmt -- --check` clean
- [x] `cargo clippy -p boxlite --no-default-features --lib -- -D warnings` clean (0 warnings)
- [x] `cargo test -p boxlite --no-default-features --lib` — 608 passed, 0 failed
- [x] 8 new unit tests covering all branches of `stream_file_chunks`:
  - Multi-chunk (3 MiB), single-chunk (500 B), partial last chunk, empty file
  - Receiver dropped (cancellation) returns error
  - I/O read error propagation via `FailingReader` mock
  - Byte-level data integrity verification
  - `upload_tar` file-not-found error path

### DIFF
--- a/src/boxlite/src/portal/interfaces/files.rs
+++ b/src/boxlite/src/portal/interfaces/files.rs
@@ -4,7 +4,8 @@
 
 use boxlite_shared::{BoxliteError, BoxliteResult, DownloadRequest, FilesClient, UploadChunk};
 use tokio::fs::File;
-use tokio::io::{AsyncReadExt, AsyncWriteExt};
+use tokio::io::{AsyncRead, AsyncReadExt, AsyncWriteExt};
+use tokio_stream::wrappers::ReceiverStream;
 use tonic::transport::Channel;
 
 const CHUNK_SIZE: usize = 1 << 20; // 1 MiB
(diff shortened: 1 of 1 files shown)


lilongen:
## PR #409 Review Fix

### Copilot Comment #1: Reader error masks gRPC error — **Fixed**

When the gRPC server rejects an upload (e.g. permission denied) and drops the receiver, `stream_file_chunks` returns a generic `Internal("Upload stream closed...")` that masks the real `tonic::Status`.

**Fix:** Distinguish real I/O errors (`Storage`) from stream-closed errors:
- `Storage` error → return directly (reader I/O failure is the root cause)
- Other error (e.g. "stream closed") → check `upload_result` first; if gRPC also failed, return the gRPC error (it's the real root cause)

### Copilot Comment #2: Reader may block on stalled filesystem — **Not fixing**

When gRPC fails, the `ReceiverStream` is dropped, which closes the channel. `stream_file_chunks` will see `tx.send()` fail and exit immediately in the vast majority of cases. The only scenario where the reader truly blocks is a stalled filesystem (system-level fault), which adding a timeout/abort wouldn't meaningfully help with. Low risk, high complexity trade-off.

### Additional
- Fixed `clippy::io_other_error` lint (Rust 1.94: `Error::new(ErrorKind::Other, ...)` → `Error::other(...)`)
- All 620 tests pass, clippy clean, fmt clean

lilongen:
Closing: #1409 added a streaming `upload_stream` to the portal files interface, which supersedes the helper in this PR.

One buffering path remains on main: `upload_tar` still collects the whole tar into memory (`src/boxlite/src/portal/interfaces/files.rs:64,76`) and the path-based `BoxImpl::copy_into` still calls it. That's better fixed as a small follow-up that routes `copy_into` through `upload_stream`.

PR #413, opened 2026-03-30
feat(litebox): add pause/resume API for zero-CPU VM freezing

lilongen:
## Summary

Add `pause()` and `resume()` methods to `LiteBox` that freeze/thaw a running VM using SIGSTOP/SIGCONT with optional guest filesystem quiesce (FIFREEZE/FITHAW), enabling zero-CPU idle states while preserving full memory and process state.

- **pause()**: guest I/O quiesce (best-effort FIFREEZE) → SIGSTOP → state → Paused
- **resume()**: SIGCONT → liveness check → guest thaw (FITHAW if quiesced) → state → Running
- **Idempotent**: pause on Paused = no-op, resume on Running = no-op
- **Exec/copy gated**: operations rejected with `InvalidState` while paused
- **stop() from Paused**: works directly (SIGCONT → teardown), no resume needed

## How It Works

```
Running ──pause()──► Paused ──resume()──► Running
                       │
                       └──stop()──► Stopped
```

The pause path:
1. Guest quiesce via gRPC (`FIFREEZE`) — freezes guest filesystems for consistency (best-effort, 5s timeout)
2. `SIGSTOP` on shim process — freezes all vCPUs, virtio backends, and I/O threads atomically
3. State transition to `Paused`, track `quiesced` flag for later snapshot awareness

The resume path:
1. `SIGCONT` on shim process — unfreezes everything
2. `kill(pid, 0)` liveness check — detect if process died while paused
3. Guest thaw via gRPC (`FITHAW`) if quiesced — only if FIFREEZE succeeded during pause
4. State transition back to `Running`

## Key Design Decisions

1. **SIGSTOP over cgroup freezer**: Works on both macOS and Linux without cgroup setup; freezes the entire shim process tree atomically including all virtio backends
2. **Guest quiesce before SIGSTOP**: FIFREEZE flushes dirty pages and freezes filesystems for point-in-time consistency, making paused state safe for snapshots/exports
3. **`quiesced` flag (runtime-only, `#[serde(skip)]`)**: Tracks whether FIFREEZE succeeded so `with_quiesce_async` (export/snapshot) can skip redundant quiesce or warn about crash-consistent degradation
4. **ESRCH handling**: If shim dies between status check and signal, transitions to `Stopped` instead of returning confusing errors
5. **stop() race handling**: If `stop()` races with `pause()`, undo SIGSTOP and yield to stop's teardown path

### Core Runtime (boxlite/)
- `litebox/box_impl.rs` — `pause()`, `resume()`, exec/copy guards, `with_quiesce_async` awareness, ESRCH + stop-race handling
- `litebox/state.rs` — `BoxStatus::Paused`, transition rules, `quiesced` field, `can_pause()`/`can_resume()` predicates
- `litebox/mod.rs` — Public `pause()`/`resume()` on `LiteBox` handle
- `runtime/backend.rs` — `pause()`/`resume()` on `BoxBackend` trait
- `rest/litebox.rs` — REST endpoints for pause/resume

### Event System
- `event_listener/listener.rs` — `on_box_paused()`/`on_box_resumed()` callbacks
- `event_listener/event.rs` — `BoxPaused`/`BoxResumed` event kinds
- `event_listener/audit_event_listener.rs` — Records pause/resume events

### CLI & HTTP
- `boxlite-cli/src/commands/serve/handlers/boxes.rs` — POST `/boxes/:id/pause`, `/boxes/:id/resume`
- `boxlite-cli/src/commands/serve/mod.rs` — Pattern-match error classification (replaces string matching)

### SDKs
- `sdks/python/src/box_handle.rs` — `pause()`/`resume()` with idempotency docs
- `sdks/node/src/box_handle.rs` — `pause()`/`resume()` via napi-rs
- `sdks/node/lib/simplebox.ts` — `pause()`/`resume()` on SimpleBox

### Tests
- `boxlite/tests/pause_resume.rs` — 10 integration tests (pause/resume, idempotency, exec/copy rejection, stop-from-paused, multi-cycle, error cases)
- `boxlite/src/litebox/state.rs` — 14 unit tests for state machine transitions + quiesced tracking
- `boxlite/src/event_listener/audit_event_listener.rs` — 2 event recording tests
- `boxlite/tests/audit.rs` — Integration test for pause/resume event emission

### Example
- `examples/python/03_lifecycle/pause_and_resume.py` — 4 demos: basic pause/resume, exec-blocked-while-paused, multi-cycle, stop-from-paused

## Test Plan

- [x] 608/608 unit tests pass (macOS)
- [x] 583/583 unit tests pass on Linux/Lima (25 pre-existing KVM failures unrelated)
- [x] Clippy clean (`cargo clippy -p boxlite --no-default-features --lib -- -D warnings`)
- [x] Format clean (`cargo fmt -- --check`)
- [x] 26 pause/resume specific tests pass (14 state machine + 2 audit + 10 integration)
- [x] All 47 state.rs tests pass (existing + new)

## Summary by CodeRabbit

* **New Features**
  * Added pause and resume controls for boxes through the Python and Node.js SDKs and HTTP API.
  * Paused boxes can be resumed or stopped, while command execution and file transfers are blocked until resumed.
  * Added lifecycle state tracking and audit events for pause and resume operations.
* **Documentation**
  * Added a Python example demonstrating pause/resume workflows, repeated cycles, and stopping paused boxes.
* **Bug Fixes**
  * Health checks now handle paused boxes without incorrectly reporting them as unhealthy.
  * Improved recovery and reliability when pausing, resuming, or stopping boxes.

### DIFF
--- a/examples/python/03_lifecycle/pause_and_resume.py
+++ b/examples/python/03_lifecycle/pause_and_resume.py
@@ -1,0 +1,192 @@
+#!/usr/bin/env python3
+"""
+Pause and Resume Example - Zero-CPU VM Freezing
+
+Demonstrates the pause/resume API:
+- pause(): Freezes VM (SIGSTOP) — zero CPU, memory preserved
+- resume(): Thaws VM (SIGCONT) — continues from exact point
+- Idempotent: pause on paused = no-op, resume on running = no-op
+- Exec rejected while paused (InvalidState)
+- Stop works directly from paused state
+"""
+
+import asyncio
+
+import boxlite
+
+
+async def basic_pause_resume():
+    """Pause a box, then resume and verify it still works."""
+    print("\n=== Basic Pause/Resume ===")
+
+    runtime = boxlite.Boxlite.default()
+
+    box = await runtime.create(boxlite.BoxOptions(
+        image="alpine:latest",
+        auto_remove=False,
+    ))
+    print(f"Created box: {box.id}")
+
+    # Run a command to verify box is working
+    execution = await box.exec("echo", ["Box is running"])
+    stdout = execution.stdout()
+    async for line in stdout:
+        print(f"  {line.strip()}")
+    await execution.wait()
+
+    info = box.info()
+    print(f"State: {info.state}")
+
+    # Pause — VM frozen, zero CPU usage
+    print("\nPausing box...")
+    await box.pause()
+    info = box.info()
+    print(f"State after pause: {info.state}")
+
+    # Resume — VM continues from exact point
+    print("\nResuming box...")
+    await box.resume()
+    info = box.info()
+    print(f"State after resume: {info.state}")
+
+    # Verify box still works
+    execution = await box.exec("echo", ["Still alive after pause/resume!"])
+    stdout = execution.stdout()
+    async for line in stdout:
+        print(f"  {line.strip()}")
+    await execution.wait()
+
+    await box.stop()
+    await runtime.remove(box.id, force=False)
+    print("\nBox stopped and removed")
+
+
+async def exec_blocked_while_paused():
+    """Show that exec is rejected while the box is paused."""
+    print("\n\n=== Exec Blocked While Paused ===")
+
+    runtime = boxlite.Boxlite.default()
+
+    box = await runtime.create(boxlite.BoxOptions(
+        image="alpine:latest",
+        auto_remove=False,
+    ))
+    print(f"Created box: {box.id}")
+
+    execution = await box.exec("echo", ["ready"])
+    await execution.wait()
+
+    await box.pause()
+    print("Box paused")
+
+    # Attempt exec while paused
+    print("Attempting exec while paused...")
+    try:
+        await box.exec("echo", ["should fail"])
+        print("  Unexpected: exec succeeded")
+    except Exception as e:
+        print(f"  Expected error: {e}")
+
+    # Resume and exec works again
+    await box.resume()
+    print("Box resumed")
+
+    execution = await box.exec("echo", ["works again!"])
+    stdout = execution.stdout()
+    async for line in stdout:
+        print(f"  {line.strip()}")
+    await execution.wait()
+
+    await box.stop()
+    await runtime.remove(box.id, force=False)
+
+
+async def pause_resume_cycles():
+    """Multiple pause/resume cycles without corruption."""
+    print("\n\n=== Multiple Pause/Resume Cycles ===")
+
+    runtime = boxlite.Boxlite.default()
+
+    box = await runtime.create(boxlite.BoxOptions(
+        image="alpine:latest",
+        auto_remove=False,
+    ))
+    print(f"Created box: {box.id}")
+
+    execution = await box.exec("echo", ["init"])
+    await execution.wait()
+
+    for i in range(3):
+        await box.pause()
+        info = box.info()
+        print(f"  Cycle {i}: paused (state={info.state})")
+
+        await box.resume()
+        execution = await box.exec("echo", [f"cycle-{i}"])
+        stdout = execution.stdout()
+        async for line in stdout:
+            print(f"  Cycle {i}: {line.strip()}")
+        await execution.wait()
+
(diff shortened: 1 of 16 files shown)


DorianZheng:
Two blockers before I can merge this:

1. We're reusing the same `Paused` state for both the external pause/resume API and internal snapshot/export/clone quiesce brackets (state.rs:25-27, box_impl.rs:1225-1294). An external `resume()` during an internal quiesce window will break snapshot consistency — needs a separate internal state.

2. In box_impl.rs:525-543, we clear `state.quiesced` and return success before we actually know the thaw succeeded. And in the shutdown path (582-627) we resume a paused box without thawing filesystems first — that can leave the guest running with frozen filesystems.

Also worth documenting explicitly that health checks freeze during pause and the last result is held indefinitely.

lilongen:
## Copilot Review Fixes

Addressed 5 of 6 Copilot review comments:

### Fixed

1. **`guest_quiesce()` returns `count > 0`** (Comment #1) — Previously always returned `true` on RPC success, even when `frozen_count = 0` (no freezable mounts). Now correctly returns `false` when nothing was actually frozen, so `state.quiesced` reflects reality.

2. **Conditional `guest_thaw()` on resume** (Comment #2) — `resume()` now captures `was_quiesced` before clearing `state.quiesced`, and only calls `guest_thaw()` when filesystems were actually frozen during pause. Avoids unnecessary RPC/log noise.

3. **Unique temp files in tests** (Comment #4) — Replaced fixed `std::env::temp_dir().join("boxlite-test-...")` paths with `tempfile::NamedTempFile` to prevent parallel test collisions.

4. **Error variant matching** (Comment #5) — Replaced brittle `msg.contains("Paused")` string assertions with `matches!(&err, BoxliteError::InvalidState(_))` and `matches!(&err, BoxliteError::Stopped(_))`.

5. **`can_pause()`/`can_resume()` docs** (Comment #6) — Clarified these are "meaningful transition allowed" predicates, not "API call will succeed" checks, since the public API is idempotent.

### Not Fixed

- **UFCS for pause/resume in BoxBackend** (Comment #3) — Standard Rust inherent-method-over-trait resolution. Same pattern used for `stop()`, `metrics()`, `copy_into()`, `copy_out()` in the same impl block. Changing only pause/resume would be inconsistent.

**Verification:** 652 tests pass, clippy clean, fmt clean.

the-mentor:
Is this pr abandoned?

lilongen:
Rebased onto current main (not abandoned). Notes for review:

- **`stop()` on a paused box**: main now only does graceful shutdown for `Running` via `ensure_booted()`, which refuses other statuses. To avoid leaking the shim of a paused box, `stop()` sends SIGCONT and then uses the cached `self.live.get()` for `Paused` (`Running` still goes through `ensure_booted()`). Please take a look at this hunk.
- `spawn_health_check` was removed upstream; the "skip ping while paused" behavior moved to `BoxWatcher::on_health_tick`.
- REST routes follow main's `/v1/boxes/{id}/pause|resume`; the Node native contract now declares `pause()`/`resume()`.

@DorianZheng your two blockers from April (shared `Paused` state with the internal quiesce bracket; clearing `quiesced` before thaw succeeds / resume-without-thaw on shutdown) are **not** addressed by this rebase — they're still open.

lilongen:
Closing this — not because the feature is wrong, but because the cost no longer
justifies it against everything else open.

@DorianZheng your two blockers from 2026-04-17 still stand, and the second one
has grown: the external pause/resume API and the internal snapshot/export/clone
quiesce bracket share the same `Paused` state, and `main` has since built
`clone_export.rs` on top of that state. Splitting it correctly is a larger piece
of work now than when you raised it, and it needs a design agreed up front
rather than another round on the code.

Meanwhile AutoStop has landed and covers the "an idle box should not cost
anything" case that motivated much of this.

The branch stays, so nothing is lost. If there is real demand for freeze/thaw
with in-memory state preserved — faster than stop/start, which is the part
AutoStop cannot give — I am happy to reopen and start from a design proposal
that addresses both blockers.

Thanks for the review time on it.

PR #415, opened 2026-04-01
fix(box_impl): offload blocking handler.stop() and metrics() to spawn_blocking

lilongen:
## Summary

- `ShimHandler::stop()` uses a `std::thread::sleep(50ms)` polling loop (up to 2 seconds).
  Called directly from async `BoxImpl::stop()`, this blocks a Tokio worker thread.
  Similarly, `handler.metrics()` performs sysinfo syscalls (~5ms), also a blocking operation.
- Wrap `LiveState.handler` from `std::sync::Mutex<...>` to `Arc<std::sync::Mutex<...>>`
  so it can be cloned and moved into `spawn_blocking` closures.
- Both `handler.stop()` and `handler.metrics()` now run on Tokio's dedicated blocking
  thread pool via `tokio::task::spawn_blocking`, keeping worker threads free.
- Lock poisoning uses asymmetric handling: `stop()` swallows it (shutdown must not be
  blocked by a poisoned lock), `metrics()` propagates it (monitoring should surface
  anomalies).

## Files changed

| File | Change |
|------|--------|
| `boxlite/src/litebox/box_impl.rs` | Arc-wrap handler field; `spawn_blocking` for `stop()` and `metrics()` |

## Test plan

- [x] `cargo test -p boxlite --no-default-features --lib` — 593 passed (main branch)
- [x] `cargo clippy -p boxlite --no-default-features --lib -- -D warnings` — clean
- [x] `cargo fmt -- --check` — clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved responsiveness when collecting metrics or stopping a running box.
  * Added more reliable handling for internal locking and background task errors.
  * Ensured boxes can still be stopped safely when a lock encounters an error.
  * Improved cleanup of temporary files used during copy operations.

### DIFF
--- a/src/boxlite/src/litebox/box_impl.rs
+++ b/src/boxlite/src/litebox/box_impl.rs
@@ -659,7 +659,12 @@ impl BoxImpl {
             h.metrics()
         })
         .await
-        .map_err(|e| BoxliteError::Internal(format!("spawn_blocking join: {}", e)))??;
+        .map_err(|e| {
+            BoxliteError::Internal(format!(
+                "metrics spawn_blocking join (box_id={}): {}",
+                self.config.id, e
+            ))
+        })??;
 
         Ok(BoxMetrics::from_storage(
             &live.metrics,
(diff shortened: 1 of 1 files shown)


src/boxlite/src/litebox/box_impl.rs
```diff
@@ -294,11 +294,15 @@ impl BoxImpl {
         }
 
         let live = self.live_state().await?;
-        let handler = live
-            .handler
-            .lock()
-            .map_err(|e| BoxliteError::Internal(format!("handler lock poisoned: {}", e)))?;
-        let raw = handler.metrics()?;
+        let handler = Arc::clone(&live.handler);
+        let raw = tokio::task::spawn_blocking(move || {
```
uran0sH, line 655: Is this raw not being used?

uran0sH (commented):
I made a mistake before; I think there's no problem moving them into spawn_blocking. Overall LGTM

lilongen:
## Review Fixes

Addressed 3 of 4 review comments:

### Fixed

1. **Descriptive `spawn_blocking` join errors** (Copilot #1 & #2) — Added operation name (`metrics`/`stop`) and `box_id` to join error messages for both `metrics()` and `stop()` paths.

2. **Recover poisoned mutex during shutdown** (Copilot #3) — Instead of silently returning `Ok(())` when handler mutex is poisoned (which leaks the VM/shim), now recovers the guard via `PoisonError::into_inner()` and still calls `handler.stop()`.

### Not Fixed

3. **"Is this raw not being used?"** (uran0sH) — `raw` is used on lines 309-310: `raw.cpu_percent` and `raw.memory_bytes` are passed to `BoxMetrics::from_storage()`. This was a false positive.

**Verification:** 637 tests pass, clippy clean, fmt clean.

src/boxlite/src/litebox/box_impl.rs
lilongen: `raw` is used: it is bound at `box_impl.rs:655` and read at `:671`-`:672`, where
`raw.cpu_percent` and `raw.memory_bytes` go into `BoxMetrics::from_storage`.

Sorry this sat unanswered in the thread — I replied in a separate comment back in
April, which is why it never reached you here.

That was the last open point; you said "Overall LGTM" on the rest. Could you
convert that into an approval when you get a chance?

PR #492, opened 2026-05-05
fix: recover detached boxes after stdin shim config

### LINKED ISSUE #434: `listInfo()` reports detached boxes as "stopped" while VM is still running
# Bug: `listInfo()` reports detached boxes as "stopped" while VM is still running

## Summary

When a box is created with `detach: true`, the VM continues running after the parent process exits (confirmed via active port forwarding), but `listInfo()` incorrectly reports `"status": "stopped"` and `"running": false`.

## Environment

- `@boxlite-ai/boxlite`: latest
- Runtime: Bun v1.3.6 (also reproducible with Node.js)
- OS: Linux x86_64 (WSL2 6.6.87.2-microsoft-standard-WSL2)

### 1. Create a detached box with an HTTP server

Save as `start-detached.ts`:

```typescript

function sleep(ms: number) {
  return new Promise((resolve) => setTimeout(resolve, ms));
}

async function main() {
  const box = new SimpleBox({
    image: "python:slim",
    name: "detach-bug-repro",
    ports: [{ hostPort: 8080, guestPort: 8080, protocol: "tcp" }],
    detach: true,
    autoRemove: false,
  });

  await box.exec(
    "sh", "-c",
    "python -m http.server 8080 --directory /tmp > /dev/null 2>&1 &"
  );
  await sleep(2000);

  const response = await fetch("http://localhost:8080");
  console.log(`Server responding inside process: ${response.status}`);
  console.log(`Box ID: ${box.id}`);
  console.log("Exiting without calling box.stop()...");
}

main();
```

```bash
bun run start-detached.ts
```

### 2. Verify the VM is still running after process exit

```bash
curl http://localhost:8080
# Returns directory listing — the VM is alive and serving requests
```

### 3. Check what listInfo() reports

```bash
bun -e "import { JsBoxlite } from '@boxlite-ai/boxlite'; const r = JsBoxlite.withDefaultConfig(); console.log(JSON.stringify(await r.listInfo(), null, 2))"
```

## Expected Result

```json
{
  "id": "...",
  "state": {
    "status": "running",
    "running": true
  },
  "name": "detach-bug-repro"
}
```

## Actual Result

```json
{
  "id": "...",
  "state": {
    "status": "stopped",
    "running": false
  },
  "name": "detach-bug-repro"
}
```

The box reports `"stopped"` even though the VM is actively serving HTTP requests on port 8080.

One additional note it looks like even the CLI reports stopped even though the vm is running and responding to curl requests.

```bash
2026-04-08T04:50:37.198197Z  INFO boxlite::runtime::rt_impl: Recovering 1 boxes from database
2026-04-08T04:50:37.198284Z  INFO boxlite::rootfs::guest: guest_binary_hash: using compile-time hash hash_prefix="6d62325392a9"
2026-04-08T04:50:37.198409Z  INFO boxlite::rootfs::guest: gc_inner: scanned boxes for references referenced_count=1 total_records=1
2026-04-08T04:50:37.198436Z  INFO boxlite::rootfs::guest: gc_inner: summary total_entries=1 preserved_current=0 preserved_referenced=1 removed=0
2026-04-08T04:50:37.198443Z  INFO boxlite::rootfs::guest: GC completed elapsed_ms=0 suffix=-6d62325392a9
2026-04-08T04:50:37.198473Z  INFO boxlite::runtime::rt_impl: Box recovery complete
┌──────────────┬─────────────┬─────────┬─────────────────────────┬──────────────────┐
│ ID           │ IMAGE       │ STATUS  │ CREATED                 │ NAMES            │
├──────────────┼─────────────┼─────────┼─────────────────────────┼──────────────────┤
│ CG4TEgntpzAp │ python:slim │ Stopped │ 2026-04-08 04:49:46 UTC │ detach-bug-repro │
└──────────────┴─────────────┴─────────┴─────────────────────────┴──────────────────┘

```

zxyasfas:
## Summary
- tag spawned shim processes with `BOXLITE_BOX_ID` so recovery can validate the owning box without putting the InstanceSpec back in argv
- update Linux PID validation to require a `boxlite-shim` process with the matching env marker instead of checking for the box id in `/proc/<pid>/cmdline`
- add regression coverage for the Linux env-based process check

Fixes #434.

## Testing
- `git diff --check`
- Not run: `cargo fmt --check` / `cargo test -p boxlite util::process` because `cargo` is not installed in this Windows environment and WSL commands timed out before reaching the repo.

### DIFF
--- a/src/boxlite/src/litebox/box_impl.rs
+++ b/src/boxlite/src/litebox/box_impl.rs
@@ -359,6 +359,28 @@ impl BoxImpl {
             if let Ok(mut handler) = live.handler.lock() {
                 handler.stop()?;
             }
+        } else {
+            let (status, pid) = {
+                let state = self.state.read();
+                (state.status, state.pid)
+            };
+
+            if let Some(pid) = pid
+                && !matches!(status, BoxStatus::Configured | BoxStatus::Stopped)
+            {
+                tracing::warn!(
+                    box_id = %self.config.id,
+                    pid = pid,
+                    status = ?status,
+                    "Stopping recovered box process without live state"
+                );
+                if !crate::util::kill_process(pid) {
+                    return Err(BoxliteError::InvalidState(format!(
+                        "failed to kill box process {}",
+                        pid
+                    )));
+                }
+            }
         }
 
         // Clean up PID file (single source of truth)
@@ -622,6 +644,16 @@ impl BoxImpl {
 
         let state = self.state.read().clone();
         let is_first_start = state.status == BoxStatus::Configured;
+
+        if !matches!(
+            state.status,
+            BoxStatus::Configured | BoxStatus::Stopped | BoxStatus::Running
+        ) {
+            return Err(BoxliteError::InvalidState(format!(
+                "Cannot initialize box {} in {} state",
+                self.config.id, state.status
+            )));
+        }
 
         // Retrieve the lock (allocated in create())
         let lock_id = state.lock_id.ok_or_else(|| {
--- a/src/boxlite/src/runtime/rt_impl.rs
+++ b/src/boxlite/src/runtime/rt_impl.rs
@@ -780,11 +780,11 @@ impl RuntimeImpl {
         if let Some((config, state)) = self.box_manager.box_by_id(id)? {
             // Box exists in database - handle as before
             let mut state = state;
-            if state.status.is_active() {
+            if state.status.is_active() || state.pid.is_some() {
                 if force {
                     // Force mode: kill the process directly
                     if let Some(pid) = state.pid {
-                        tracing::info!(box_id = %id, pid = pid, "Force killing active box");
+                        tracing::info!(box_id = %id, pid = pid, "Force killing box process");
                         crate::util::kill_process(pid);
                     }
                     // Update status to stopped and save
@@ -794,8 +794,8 @@ impl RuntimeImpl {
                 } else {
                     // Non-force mode: error on active box
                     return Err(BoxliteError::InvalidState(format!(
-                        "cannot remove active box {} (status: {:?}). Use force=true to stop first",
-                        id, state.status
+                        "cannot remove box {} with live pid {:?} (status: {:?}). Use force=true to stop first",
+                        id, state.pid, state.status
                     )));
                 }
             }
@@ -1164,6 +1164,7 @@ impl RuntimeImpl {
         for (config, mut state) in persisted {
             let box_id = &config.id;
             let original_status = state.status;
+            let original_pid = state.pid;
 
             // Reclaim the lock for this box if one was allocated
             if let Some(lock_id) = state.lock_id {
@@ -1211,13 +1212,14 @@ impl RuntimeImpl {
                         }
                         (true, ProcessIdentity::Unverified) => {
                             // Older detached shims may predate BOXLITE_BOX_ID. Preserve
-                            // the PID file and avoid deleting a live VM we cannot prove.
+                            // the PID file and recover them as Running so lifecycle ops
+                            // keep treating the live VM as active.
                             state.set_pid(Some(pid));
-                            state.set_status(BoxStatus::Unknown);
+                            state.set_status(BoxStatus::Running);
                             tracing::warn!(
                                 box_id = %box_id,
                                 pid = pid,
-                                "Box shim is live but unverified, marking as Unknown"
+                                "Recovered legacy running box from unmarked shim PID file"
                             );
                         }
                         _ => {
@@ -1255,7 +1257,7 @@ impl RuntimeImpl {
             }
 
             // Save updated state to database if changed
-            if state.status != original_status {
+            if state.status != original_status || state.pid != original_pid {
                 self.box_manager.save_box(box_id, &state)?;
             }
         }
@@ -2098,7 +2100,7 @@ mod tests {
 
     #[cfg(target_os = "linux")]
     #[test]
-    fn test_recovery_marks_unverified_live_shim_unknown() {
+    fn test_recovery_marks_unverified_live_shim_running() {
         use std::os::unix::process::CommandExt;
 
         let (runtime, _dir) = create_test_runtime();
@@ -2126,7 +2128,7 @@ mod tests {
         runtime.recover_boxes().expect("Failed to recover boxes");
 
         let (_, db_state) = runtime.box_manager.box_by_id(&config.id).unwrap().unwrap();
-        assert_eq!(db_state.status, BoxStatus::Unknown);
+        assert_eq!(db_state.status, BoxStatus::Running);
         assert_eq!(db_state.pid, Some(pid));
         assert!(
             pid_file.exists(),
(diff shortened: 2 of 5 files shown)


zxyasfas:
Status update: the PR is still mergeable and the relevant CI is green. Rust formatting, Rust tests across darwin/linux targets, Clippy, config checks, and codecov patch have all passed. The diff remains scoped to shim process tagging and detached box recovery validation.

DorianZheng:
Hi @zxyasfas. Could you check why there are lots of irrelevant diff?

zxyasfas:
Thanks for catching this. I rebuilt the branch on top of the current `main` and force-pushed the same scoped change, so the irrelevant diff is gone now.

The PR diff is back to the intended files only:
- `src/boxlite/src/util/process.rs`
- `src/boxlite/src/vmm/controller/spawn.rs`

Current compare is 1 commit ahead / 0 behind, with the diff limited to the shim env marker and PID recovery validation changes. CI is re-running on the cleaned branch.

DorianZheng:
two things look off to me:

1. `read_to_string` on `/proc/<pid>/environ` will fail if any inherited env var isn't valid utf-8. one bad byte and we false-negative → live box gets marked stopped. should read as bytes and split on `\0`.
2. boxes already detached from before this PR don't have `BOXLITE_BOX_ID` set. after upgrade, recovery will think they're dead and delete their pid file. need a fallback or treat unmarked live shims as Unknown instead of stale.

direction is right though, just want these two covered with tests before we merge.

zxyasfas:
Thanks, pushed a follow-up for both cases in b0f3ca2.

- /proc/<pid>/environ is now read as bytes and split on NUL, so an unrelated non-UTF-8 env var does not make a correctly tagged shim look stale.
- PID validation now distinguishes match, mismatch, and unverified. Recovery keeps an unmarked live shim's PID file and marks the box Unknown instead of Stopped, which preserves detached shims created before BOXLITE_BOX_ID existed.
- Added Linux coverage for tagged shims with non-UTF-8 env entries, unmarked legacy shims, and the recovery Unknown path.

zxyasfas:
Fixed the export/format follow-up in 35c4648. CI is green now.

DorianZheng:
looked at this more closely and the Unknown + live pid model has issues across the lifecycle, not just `rm`:

1. **rm orphans the vm** — `can_remove(Unknown) = true` and the kill branch in `remove_box` only runs when `is_active()`. so on a legacy detached shim recovered as Unknown, `boxlite rm <id>` (no force) deletes the db row + box dir while the vm keeps running. (`rt_impl.rs:776-801`, `state.rs:91-96`, `state.rs:52-54`)

2. **stop also orphans it** — `box_impl.rs:314-442`: for a recovered Unknown box, `self.live.get()` returns None (recovery only writes the db, never rebuilds LiveState), so the VM-shutdown block at 344-362 is skipped. then it removes the pid file, clears `state.pid`, marks Stopped, returns Ok. user sees success, runtime drops the handle, vm keeps running.

3. **exec panics** — `live_state` → `init_live_state` → `get_execution_plan(Unknown)` hits `_ => panic!("Invalid BoxStatus for initialization: {:?}", status)`. crashes the runtime.

small thing separate from the above: `process_identity_linux` returns `Unverified` when `fs::read(environ_path)` errors. legacy shims have a readable environ (just no marker), so that Err branch is only reached on different-uid pid reuse — should be `Mismatch`.

1/2/3 share a root: `Unknown` historically meant "weird db state, no live process", and the lifecycle ops never check `pid.is_some()`.

zxyasfas:
Addressed the lifecycle issues from the latest review in 88b30dc.

- Legacy unmarked shims now recover as Running instead of Unknown, so rm/exec keep the box in the active lifecycle path.
- Non-force rm now refuses any persisted live PID, including unusual recovered states; force rm still kills and removes it.
- stop now kills a recovered PID even when no LiveState has been rebuilt, so it cannot remove the PID file and return success while leaving the VM process alive.
- Unknown initialization now returns InvalidState instead of panicking, and unreadable /proc environ is treated as a mismatch.

CI is green across formatting, Clippy, Rust tests, and codecov patch.

nieyy:
@zxyasfas could you add several test cases?

law-chain-hot:
Really appreciate the iteration here. The env-marker, 3-state `ProcessIdentity`, legacy-shim recovery path, and lifecycle fallbacks are all coming together nicely. Three things we'd like to fold in before merging, if you have cycles.

---

## 1. Identity check: cross-platform alignment

Linux side is solid. macOS is weaker: `process_identity` only checks the process name there, can't tell which box, no PID-reuse defense.

```
┌──────────┬─────────────────────────────┬───────────────────────┐
│ Platform │ Current check                │ What it gives us       │
├──────────┼─────────────────────────────┼───────────────────────┤
│ Linux    │ /proc/<pid>/environ for      │ Identifies the box;    │
│          │ BOXLITE_BOX_ID               │ implicitly PID-reuse-  │
│          │                              │ safe ✓                 │
├──────────┼─────────────────────────────┼───────────────────────┤
│ macOS    │ sysinfo, process-name only   │ Only confirms "this is │
│          │ (`util/process.rs`:          │ a shim", can't tell    │
│          │  `let _ = box_id; //         │ which box, no PID-     │
│          │   Unused on macOS`)          │ reuse defense ⚠️       │
└──────────┴─────────────────────────────┴───────────────────────┘
```

Since macOS is a first-class platform, we'd like to align both sides on `(pid, start_time)`, the standard pattern (systemd `PidRef`, postgres `postmaster.pid`, runc; `pidfd_open` is the kernel version). Store start_time at spawn via sysinfo, compare on recovery; mismatch = PID reused.

```
# On spawn
──────────────────────────────────────────────
cmd.spawn() → pid
     ↓
sysinfo::Process::start_time(pid) → t0    (seconds, UTC epoch)
     ↓                                    (same API on Linux & macOS)
BoxState { pid, shim_start_time = Some(t0) } → DB


# On recovery
──────────────────────────────────────────────
DB → (pid, expected = t0)
     ↓
sysinfo::Process::start_time(pid) → current = t1
     ↓
match (expected, current):
  (Some(t0), Some(t1)) if t0 == t1  → Matches     (same process)
  (Some(_),  Some(_))               → Mismatch    (PID was reused)
  (None,     Some(_))               → Unverified  (legacy box)
  (_,        None)                  → Mismatch    (process gone)
```

Landing plan:

- `BoxState` gets `shim_start_time: Option<u64>` with `#[serde(default)]`, same trick `health_status` uses, no SQL migration.
- `process_identity` collapses into one cross-platform fn; Linux environ parsing goes away.
- `BOXLITE_BOX_ID` env can stay as a debug tag, just not part of validation.
- Legacy rows (no start_time) → `Unverified`, backfill on first recovery.

---

## 2. Graceful shutdown: use the existing ShimHandler

The new `else` branch calls `kill_process(pid)` = straight SIGKILL. That skips the shim's own SIGTERM handler (`src/shim/src/main.rs:243-310`, does `Guest.Shutdown` RPC + qcow2 flush), which is exactly the qcow2 corruption case that handler exists to prevent. Recovered boxes hitting `stop()` are effectively getting a power-cord pull.

`ShimHandler` already covers this: `from_pid` (`shim.rs:70`) + `stop()` (line 86) does SIGTERM → waitpid poll → SIGKILL fallback. The `if Some(live)` branch right above already uses it (owned-mode). Both can converge:

```
Today (three paths, each on its own)
────────────────────────────────────────────────────────────

box_impl.rs::stop()
    ├── if Some(live)  ──► live.handler.stop()     (= ShimHandler ✓)
    └── else           ──► kill_process(pid)        (SIGKILL ✗ new in PR)

rt_impl.rs::shutdown_sync (lines 693-714)
                       ──► 20+ lines of inline SIGTERM loop + kill_process
                           (correct, but duplicates ShimHandler::stop)

Proposed (everything converges on ShimHandler)
────────────────────────────────────────────────────────────

box_impl.rs::stop()
    ├── if Some(live)  ──► live.handler.stop()                   ┐
    └── else           ──► ShimHandler::from_pid(pid).stop()     ├──► shared
                                                                 │
rt_impl.rs::shutdown_sync                                        │
                       ──► ShimHandler::from_pid(pid).stop()     ┘

                           Internals (unified):
                           SIGTERM → waitpid poll (with reap) → SIGKILL fallback
```

Both spots become:

```rust
let mut h = ShimHandler::from_pid(pid, self.config.id.clone());
h.stop()?;
```

> `rt_impl.rs:788`'s `remove_box(force=true)` we'd leave alone, respecting the "force = kill directly" intent.

Also: `GRACEFUL_SHUTDOWN_TIMEOUT_MS = 2000` (`shim.rs:90`) < shim's own 8s budget (`main.rs:237-241`), so parent SIGKILLs mid-`Guest.Shutdown`. Bump to `10_000` with a one-line comment (parent wait ≥ shim budget). Different things, no shared constant needed.

---

## 3. Tests

Lining up with @nieyy, we'd like the test cases to follow:

```
(a) the test reproduces the bug   ── red before the fix
(b) the test passes after the fix ── green after
```

---

A possible follow-up we're kicking around (not for this PR): could `recover_boxes` rebuild `LiveState` via `ShimHandler::from_pid` at recovery? That way `stop` / `exec` / etc. would all flow through the normal `if Some(live)` path and the "no LiveState" fallbacks could be retired over time. Either way `recover_boxes` doesn't need to be touched in this PR — flagging it mainly so the picture is clear. Curious what you and @DorianZheng think.

Push back on anything that looks off.

DorianZheng:
Hi @zxyasfas. I have make some refinement on this PR. If you think it's fine, I will push the commits in this PR

DorianZheng:
Closed for new PR #590

PR #502, opened 2026-05-11
fix: make Start() idempotent when box is already running

nieyy:
## Problem

Start() returns ErrInvalidState when the box was already started by Create(), even though the box is actually running. Affects all users following the standard pattern of calling create() then start().

## Root Cause

Create() was updated to auto-start the box, but Start() was not made idempotent. When the box is already Running, Start() returns ErrInvalidState, which propagates as an error.

## Change

Extract startIdempotent() in the runner client: when ErrInvalidState is received, call Info() to confirm the box is actually Running before suppressing the error. All other states (Stopping, etc.) propagate the error unchanged.

Unconditionally suppressing ErrInvalidState would be wrong — the error code also covers Stopping/Paused states that are genuine failures.

Also adds api_id() to BoxBackend for REST API ID handling and fixes CLI linker issue on Linux.

## Verification

```bash
cd apps/runner
go test -tags boxlite_dev ./pkg/boxlite/ -run TestStartIdempotent -v
```

All 4 tests pass.

### DIFF
--- a/apps/runner/pkg/boxlite/client.go
+++ b/apps/runner/pkg/boxlite/client.go
@@ -205,7 +205,33 @@ func (c *Client) Create(ctx context.Context, sandboxDto dto.CreateSandboxDTO) (s
 	return bx.ID(), "boxlite", nil
 }
 
+// boxStarter is the subset of *boxlite.Box needed by startIdempotent.
+type boxStarter interface {
+	Start(context.Context) error
+	Info(context.Context) (*boxlite.BoxInfo, error)
+}
+
+// startIdempotent calls bx.Start and suppresses the error only when the box is
+// confirmed to be already running. ErrInvalidState from any other state
+// (Stopping/Paused/Unknown) is propagated unchanged.
+func startIdempotent(ctx context.Context, bx boxStarter, sandboxId string, log *slog.Logger) error {
+	if err := bx.Start(ctx); err != nil {
+		if boxlite.IsInvalidState(err) {
+			// Rust start() returns Ok(()) for Running state, so ErrInvalidState means
+			// the box is in a non-startable state. Verify before suppressing.
+			info, infoErr := bx.Info(ctx)
+			if infoErr == nil && info.State == boxlite.StateRunning {
+				log.DebugContext(ctx, "start called on already running box, treating as no-op", "sandbox_id", sandboxId)
+				return nil
+			}
+		}
+		return err
+	}
+	return nil
+}
+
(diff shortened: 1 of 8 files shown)


apps/runner/pkg/boxlite/client.go
```diff
@@ -205,7 +205,33 @@ func (c *Client) Create(ctx context.Context, sandboxDto dto.CreateSandboxDTO) (s
...
+// boxStarter is the subset of *boxlite.Box needed by startIdempotent.
+type boxStarter interface {
+	Start(context.Context) error
+	Info(context.Context) (*boxlite.BoxInfo, error)
+}
+
+// startIdempotent calls bx.Start and suppresses the error only when the box is
+// confirmed to be already running. ErrInvalidState from any other state
+// (Stopping/Paused/Unknown) is propagated unchanged.
+func startIdempotent(ctx context.Context, bx boxStarter, sandboxId string, log *slog.Logger) error {
```
DorianZheng, line 217: why not just update `boxlite.Start()` behaviour or error code?

src/boxlite/src/runtime/backend.rs
```diff
@@ -70,6 +70,15 @@ pub(crate) trait RuntimeBackend: Send + Sync {
 pub(crate) trait BoxBackend: Send + Sync {
     fn id(&self) -> &BoxID;
 
+    /// Returns the identifier string to use in API/URL calls.
+    ///
+    /// For local boxes this is identical to `id().to_string()`.
+    /// For REST-backed boxes this preserves the raw server-returned ID
+    /// (e.g., a UUID), which may not be a valid `BoxID`.
+    fn api_id(&self) -> String {
```
DorianZheng, line 78: why do we have to add external id in the box?

src/cli/build.rs
```diff
@@ -1 +1,7 @@
-fn main() {}
+fn main() {
```
DorianZheng, line 1: fixed in main

apps/runner/pkg/boxlite/client.go
law-chain-hot: https://github.com/boxlite-ai/boxlite/blob/be75a4a61dec4c91361ec5b543d8739d15db28fa/src/boxlite/src/litebox/box_impl.rs#L193-L195

From box_impl.rs L193, `StateRunning` already return `Ok(())`, so ErrInvalidState should only be reachable from `Stopping / Paused / Unknown`. Which state actually triggers it when calling create() then start()

law-chain-hot (approved):

PR #506, opened 2026-05-12
feat(release): curl|sh installer + SHA256SUMS + build provenance

DorianZheng:
## Summary

End-to-end story for installing the `boxlite` CLI with `curl | sh`, plus the supply-chain integrity tooling around it.

**Release artifacts (`build-runtime.yml`):**
- Per-file `.sha256` sidecars next to every tarball (matches cargo-dist / uv / ripgrep convention; unlocks `cargo-binstall` out of the box).
- Combined `SHA256SUMS` for batch verification.
- GitHub Artifact Attestations via `actions/attest-build-provenance@v2` — sigstore-backed, OIDC, no secrets. Verify with `gh attestation verify <file> --repo boxlite-ai/boxlite`.
- A rendered, version-locked `install.sh` published as a release asset.

**`scripts/release/install.sh.template`** — POSIX-sh installer modeled on mise (`mise.run`) and uv (`astral.sh/uv/install.sh`). CI substitutes `__VERSION__` and the three `__SHA_*__` placeholders from the per-file sidecars at release time, then uploads the rendered `install.sh`.

**User-facing one-liner:**
```bash
curl -fsSL https://github.com/boxlite-ai/boxlite/releases/latest/download/install.sh | sh

# Or pin a version / install dir (env vars MUST sit on the sh side of the pipe):
curl -fsSL https://github.com/boxlite-ai/boxlite/releases/latest/download/install.sh \
  | BOXLITE_VERSION=v0.9.4 BOXLITE_INSTALL_DIR=/usr/local/bin sh
```

## How install.sh behaves
- Defaults `BOXLITE_INSTALL_DIR=$HOME/.local/bin` (no sudo needed).
- Detects target via `uname -s` / `uname -m`; rejects macOS Intel and any other arch.
- Fast path: verifies against the checksum embedded at render time. Falls back to fetching `<artifact>.sha256` from the release for `BOXLITE_VERSION` pinning.
- Atomic install: download to `mktemp -d`, verify, extract with `tar --no-same-owner`, then `install -m 0755` into place. Trap cleans the tmpdir.

## Adversarial review history
Codex adversarial review surfaced three findings on an earlier revision; all were validated and fixed in this branch:
[...]

PR #507, opened 2026-05-12
feat(install): ship curl|sh installer with embedded checksums

DorianZheng:
> Stacked on top of #506 (`ci/cli-release-tag-checksums`). Merge that first.

## Summary
- Add `scripts/install.sh.template`, a POSIX-sh installer modeled on mise (`mise.run`) and uv (`astral.sh/uv/install.sh`).
- CI substitutes `__VERSION__` and the three `__SHA_*__` placeholders from the per-file `.sha256` sidecars generated in PR #506, then uploads the rendered `install.sh` to the release. The substitution step has a sanity check that fails the workflow if any placeholder slipped through.
- README's REST API quick-start install becomes a single `curl … | sh` line. The manual `sha256sum` + `gh attestation verify` recipe stays below for users who download tarballs directly.

## How it works
1. CI on tag push (or `release: published`) builds the CLI on three targets.
2. `upload_to_release` job downloads the artifacts, writes per-file `.sha256` sidecars, then renders `dist/install.sh` from `scripts/install.sh.template`.
3. Rendered script has the current version's checksum embedded for the fast path. Non-current versions (`BOXLITE_VERSION=v...`) fall through to fetching `<artifact>.sha256` from the release.

## Behavior
- Default install dir: `\$HOME/.local/bin` (overridable with `BOXLITE_INSTALL_DIR`).
- Target detection: `Darwin-arm64` → `aarch64-apple-darwin`; `Linux-x86_64` → `x86_64-unknown-linux-gnu`; `Linux-aarch64|arm64` → `aarch64-unknown-linux-gnu`.
- Hard error on macOS Intel and any other arch.
- Uses `curl` or `wget`; uses `sha256sum` or `shasum -a 256`. POSIX `sh` only (no bashisms).
- Atomic install: download to `mktemp -d`, verify, then `mv` the binary into place.
- Trap cleans up tmpdir on success or interrupt.

## Test plan
- [x] `shellcheck scripts/install.sh.template` — zero warnings.
- [x] `sh -n` on the template and on a locally-rendered copy — clean.
- [x] Manual `sed` dry-run with fake checksums — placeholders correctly replaced.
- [ ] Push a throwaway `v0.0.0-rc1` tag to a fork; confirm `install.sh` appears in the release with real checksums substituted in.
- [ ] `curl … install.sh | sh` on macOS arm64: `~/.local/bin/boxlite --version` matches.
- [ ] `curl … install.sh | sh` inside `docker run -it ubuntu:22.04` (with `curl ca-certificates`): same.
- [ ] Sad path: `BOXLITE_VERSION=v9.9.9 curl … | sh` — must fail clearly on 404.
- [ ] Sad path: simulate macOS Intel — must reject with the documented error message.


DorianZheng:
Addressed all three Codex adversarial-review findings:

- **Finding 1** (high) — fixed in PR #506 (`fix(ci): revert v* tag trigger`, f095516). This branch rebased onto the corrected base.
- **Finding 2** (high, root install ownership) — `tar --no-same-owner` + `install -m 0755` (61a8d07). starship has the same bug; we now do better.
- **Finding 3** (medium, README env-var pipe) — env vars moved to the `sh` side of the pipe (61a8d07).

PR #512, opened 2026-05-13
fix(exec): compat /output endpoint for SDK < 0.9.4 (#512)

nieyy:
Old SDK callers hitting the removed SSE /output endpoint received an
opaque "Result channel closed" error. This adds a compat handler that
kills the running execution before emitting a single SSE exit event
with an actionable upgrade message, preventing split-brain state where
the caller sees immediate failure while the process continues running.

On kill failure the handler returns 503 instead of emitting a false
terminal event, preserving the no-split-brain invariant.

### DIFF
--- a/apps/runner/pkg/api/controllers/boxlite_exec_output_compat.go
+++ b/apps/runner/pkg/api/controllers/boxlite_exec_output_compat.go
@@ -1,0 +1,87 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+// Copyright (c) 2026 BoxLite AI
+
+package controllers
+
+import (
+	"encoding/json"
+	"fmt"
+	"log/slog"
+	"net/http"
+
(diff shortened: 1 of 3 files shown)


nieyy:
This compat shim was intended to surface a user-friendly upgrade message
when an old SDK (< 0.9.4) connects to a runner that has removed the SSE
`/output` endpoint — instead of the opaque "Result channel closed" error.

However, testing showed that SDK 0.9.3 is already incompatible with the
current orchestrator API at the create/start level, so old SDK versions
never reach the exec phase where this shim would be triggered. The shim
can only be hit during a narrow window where the runner has been upgraded
but the orchestrator API is still compatible with the old SDK — a window
that either never existed in practice or has already closed.

Since no real user traffic will reach `GET /executions/:id/output`,
this endpoint is effectively dead code. Closing.

PR #525, opened 2026-05-14
fix qcow2 backing and container status checks

GatewayJ:
## Summary

- Stop qcow2 backing chain walks cleanly when they reach a raw/ext4 backing file instead of logging an invalid qcow2 magic warning.
- Refresh libcontainer status before deciding whether a newly started container init process is running.
- Add focused coverage for raw backing files as terminal backing-chain nodes.

## Root Cause

Raw backing files are valid terminal nodes for qcow2 COW children, but the backing-chain walker tried to parse every backing file as qcow2. Separately, the container init health check read the persisted libcontainer state immediately after start without refreshing it from the process state, which could treat a stale `Created` state as a failed init.

## Validation

- `make fmt`
- `git diff --check`
- `env BOXLITE_DEPS_STUB=1 make test:unit:rust`

`test:unit:rust` ran the full unit suite with native dependency stubs. The boxlite crate had one environment failure unrelated to this change: `jailer::tests::test_jailer_full_flow_with_real_tempdir` failed because AppArmor disallows bwrap unprivileged user namespaces on this host. All other boxlite tests passed, including the new qcow2 tests; `boxlite-shared` passed 36/36.

## Release Notes

* **Bug Fixes**
  * Enhanced QCOW2 disk image handling to properly traverse backing chains containing mixed file types
  * Fixed container status reporting to ensure current state is always reflected instead of returning cached values


src/guest/src/container/start.rs
```diff
@@ -274,13 +274,21 @@ pub(crate) fn cleanup_bundle_directory(bundle_path: &std::path::Path) {
...
-    let container = LibContainer::load(container_state_path.to_path_buf()).map_err(|e| {
+    let mut container = LibContainer::load(container_state_path.to_path_buf()).map_err(|e| {
         BoxliteError::Internal(format!(
             "Failed to load container from {}: {}",
             container_state_path.display(),
             e
         ))
     })?;
 
+    container.refresh_status().map_err(|e| {
```
DorianZheng, line 285: Please add a UT to reproduce the issue

GatewayJ: Added a focused UT in `src/guest/src/container/start.rs` covering this regression: it writes stale libcontainer state with a live PID, calls `load_container_status`, and asserts the refreshed status is `Running`.

Validation run:
- `make fmt`
- `make fmt:check`
- `git diff --check`
- `BOXLITE_DEPS_STUB=1 cargo check -p boxlite-guest --tests`

I also tried the focused `cargo test -p boxlite-guest load_container_status_refreshes_stale_persisted_status -- --test-threads=1`, but this runner cannot link `boxlite-guest` because the host is missing system `libseccomp` (`unable to find library -lseccomp`).

DorianZheng (approved):
LGTM

PR #533, opened 2026-05-15
fix: move test cache under workspace target

uran0sH:
The test cache was stored under src/target/boxlite-test, outside Cargo's workspace target directory. cargo clean therefore missed it, so large integration-test artifacts could remain after normal cleanup.

Compute the cache path from the repository root and store it under target/boxlite-test. Update the cache_dir unit test to assert the workspace target location explicitly.


DorianZheng:
Good catch!

DorianZheng (approved):

PR #535, opened 2026-05-16
fix(node): wrap native rest static via Proxy instead of mutating it

DorianZheng:
## Summary
- `sdks/node/lib/index.ts` adapted the native positional `rest(url, credential?, prefix?)` to the public `BoxliteRestOptions` bag by **reassigning** the `rest` static on the native napi-rs class. napi-rs registers class statics as non-writable **and** non-configurable, so the assignment threw `TypeError: Cannot assign to read only property 'rest'` **at module import** — breaking every consumer of `../lib/index.js`. Three integration suites (`credential`, `images`, `snapshot-clone-export`) failed on import; `main` was red on Node integration.
- A Proxy *over the native class* is also rejected by the JS spec: a `get` trap may not substitute a non-configurable, non-writable own data property. The fix proxies a **fresh function** that owns no such property — `rest` resolves to the bag adapter, while construction (`new`) and every other static forward to the native class untouched.
- Adds `tests/rest-binding.integration.test.ts` (reproducer: import + bag adaptation).

Regression introduced by #532 (bearer auth). Surfaced while bumping SDK patch versions, whose pre-push `make test` runs the Node integration suite.

## Test plan
- [x] New reproducer fails before fix (exact `TypeError` at `lib/index.ts:60`), passes after
- [x] Previously-failing suites pass: `credential`, `images`, `snapshot-clone-export`
- [x] Full `make test:integration:node`: 9 passed / 1 skipped, 44 tests, exit 0
- [x] `npm run build` (tsc) clean; unit project 69/69
- [x] pre-push `make test` green on push

### DIFF
--- a/sdks/node/lib/index.ts
+++ b/sdks/node/lib/index.ts
@@ -38,9 +38,9 @@ export type {
 
 // The native REST binding is positional `(url, credential?, prefix?)`.
 // The public surface is the `BoxliteRestOptions` bag — identical in
-// name and shape across the C/Go/Node/Python SDKs. This adapter
-// destructures the bag into the native binding once at module load;
-// every other static/instance member is the native one unchanged.
+// name and shape across the C/Go/Node/Python SDKs. A Proxy substitutes
+// the bag-taking `rest`; every other static/instance member is the
+// native one unchanged.
 // `Omit` over the constructor interface preserves the `withDefaultConfig`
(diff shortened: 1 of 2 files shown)


DorianZheng:
Superseded by #536. The #536 branch contains this Proxy hotfix commit plus the binding-layer refactor that removes it; consolidating into one PR to main (net diff = the idiomatic binding-layer solution). Not merged separately.

PR #541, opened 2026-05-16
feat(ssh): real-SSH access via in-VM sshd + gvproxy port forwarding

nieyy:
Adds full SSH (SFTP, scp, port forwarding, non-PTY exec) to BoxLite sandboxes by routing SSH-access tokens through a per-sandbox port on the Runner EC2 directly to a real sshd running inside the container namespace, bypassing the Rust exec bridge that blocks binary streams. The external SSH Gateway detects real-SSH tokens (unix_user non-null) and TCP-proxies to the allocated port; legacy exec-bridge tokens fall through unchanged.

Security contract:
- Only the gateway public key is installed in guest sshd authorized_keys
- Real-SSH tokens require unix_user; empty/absent → null (exec-bridge) at controller boundary
- Gateway is fail-closed on every error path: lookup failure, degraded state, unix_user mismatch, missing RUNNER_API_TOKEN, empty unix_user string
- Per-sandbox Redis lock (90s TTL, compare-and-delete) guards concurrent enable/disable
- Expired row cleanup covers all three cases: runner-attached, runner-less, null sandbox
- Non-expired token with null sandbox (concurrent delete) returns invalid cleanly
- Revocation is fail-closed (disable-before-delete): runner unreachable → error propagated, DB row preserved for retry; legacy exec-bridge tokens use best-effort runner cleanup
- PermitRootLogin set to prohibit-password; default SSH user is root

Implementation:
- CI statically compiles OpenSSH 9.7p1 (sshd + ssh-keygen) via Alpine musl Docker;
   guest.rs injects binaries + management scripts into the guest ext4 rootfs at build time
- SSH infra files (sshd, ssh-keygen, management scripts) bind-mounted read-only into
   container from guest rootfs, gated on /boxlite/bin/sshd presence with per-file checks
- gvproxy admin HTTP server on gvproxy-admin.sock handles /services/forwarder/expose
   for dynamic per-sandbox port forwarding to guest sshd (port 22222)
- boxlite-enable-ssh: UID collision check covers /etc/group; sudoers written before sshd
   starts; passwordless sudo scoped to default 'boxlite' user only; root home dir handled
- boxlite-disable-ssh: mirrors enable-side sudoers guard on disable
- sst.config.ts: auto-derive SSH_GATEWAY_PUBLIC_KEY from SSH_PRIVATE_KEY_B64; quote key
   in systemd Environment= (spaces); restrict port 22100–22199 ingress to ECS gateway SG
- runner-update-binary.sh: guard SSH_GATEWAY_PUBLIC_KEY injection so running without the
   secret does not erase the live value
- runnerAdapter.v2.ts: implement enableSSHAccess / disableSSHAccess natively;
   503 surfaces as error, 404 is no-op on disable; remove stale apiVersion gate
- sandbox.service.ts: fix remainingRealSsh to count only unixUser≠null rows; add
   lock-lost fencing before runner disable; rollback on create failure skips if prior
   real-SSH session existed; real→legacy rotation calls disableSSHAccess fail-closed
- client_ssh.go: internalBoxID() helper with three-level fallback (stored state, runtime
   query, external UUID) for correct gvproxy socket path resolution; hostPort=0 case
   handled in forward recovery (degraded mode when gvproxy was unavailable at enable time)
- ssh-gateway: CloseWrite() after each io.Copy direction so scp and commands waiting for
   server EOF exit cleanly instead of hanging
- 60 regression tests covering all fixed invariants

## Summary by CodeRabbit

* **New Features**
  * Added box-level SSH access management, including creating, checking, and revoking access with optional per-user SSH sessions.
  * Introduced support for real SSH connections alongside the existing terminal flow, with better handling for interactive sessions and command execution.
  * Added a build target for producing static SSH server binaries used by the platform.

* **Bug Fixes**
  * Improved reliability of SSH access validation, routing, and session handling.
  * Tightened runtime checks so SSH-related features fail safely when required keys or services are unavailable.


.github/workflows/build-runtime.yml
```diff
@@ -84,6 +84,44 @@ jobs:
...
+          docker run --rm \
+            -v "$(pwd)/dist:/output" \
+            alpine:3.19 sh -c '
+              set -ex
+              apk add --no-cache \
+                build-base linux-headers \
+                openssl-dev openssl-libs-static \
+                zlib-dev zlib-static
+              OPENSSH_VERSION=9.7p1
+              wget -q "https://cdn.openbsd.org/pub/OpenBSD/OpenSSH/portable/openssh-${OPENSSH_VERSION}.tar.gz"
```
law-chain-hot: Not a blocker, but cheap to add. What do you think?

The "Build static sshd" step in `.github/workflows/build-runtime.yml` downloads the OpenSSH source and builds it, but does not verify the tarball. Since this `boxlite-sshd` runs in every box for SSH auth, I think we should check it.

Maybe pin the SHA256 and fail the build on mismatch:

```sh
# update OPENSSH_VERSION and the hash together
OPENSSH_SHA256=490426f766d82a2763fcacd8d83ea3d70798750c7bd2aff2e57dc5660f773ffd  # 9.7p1, from openssh.com
wget -q "https://cdn.openbsd.org/.../openssh-${OPENSSH_VERSION}.tar.gz"
echo "${OPENSSH_SHA256}  openssh-${OPENSSH_VERSION}.tar.gz" | sha256sum -c -
```

law-chain-hot (commented):
Left one comment on the OpenSSH download step.

lilongen:
### Code review

Found 4 issues:

1. `boxlite-disable-ssh` hard-codes `HOME_DIR="/home/$UNIX_USER"` while `boxlite-enable-ssh` branches on `root`. When SSH was enabled for `root` (the controller default), enable writes the marker to `/root/.ssh/.ssh_enabled`, disable then deletes a non-existent `/home/root/.ssh/.ssh_enabled` (no-op), and the surviving marker causes `boxlite-ensure-ssh` to silently re-enable sshd on the next VM restart — defeating the API-level revoke.

https://github.com/boxlite-ai/boxlite/blob/11a15a36ce5fba97bef090bbe7d22cd398eb9d57/src/guest/scripts/boxlite-disable-ssh#L5-L10

Compare with the enable-side branch:

https://github.com/boxlite-ai/boxlite/blob/11a15a36ce5fba97bef090bbe7d22cd398eb9d57/src/guest/scripts/boxlite-enable-ssh#L6-L12

2. `sshUserOrDefault()` doc says it returns `\"boxlite\"` and the surrounding comment claims it exists to *prevent* inheriting root from the image, but the implementation returns `\"root\"`. The struct field doc (`// Empty means \"boxlite\"`) and the test name `TestSSHUserDefaultsToBoxlite` confirm the intended default. CLAUDE.md: \"Update nearby comments when behavior changes.\"

https://github.com/boxlite-ai/boxlite/blob/11a15a36ce5fba97bef090bbe7d22cd398eb9d57/apps/runner/pkg/sshgateway/service.go#L192-L201

3. Per-box mutex entries in `boxSSHMu` are inserted on first `EnableSSHAccess` but never deleted. `cleanupSSHOnDestroy` removes `sshStates` and `sshBoxes` entries but leaves `boxSSHMu` to grow without bound across the runner process lifetime. CLAUDE.md: \"No unbounded queues/concurrency/memory.\"

https://github.com/boxlite-ai/boxlite/blob/11a15a36ce5fba97bef090bbe7d22cd398eb9d57/apps/runner/pkg/boxlite/client_ssh.go#L810-L836

4. New log line prints the raw SSH bearer token in cleartext on the fail-closed path. The token is a live revocable credential. CLAUDE.md: \"Mask secrets in errors and logs.\" Consider redacting (prefix + `…`) consistently with how other credentials are handled.

https://github.com/boxlite-ai/boxlite/blob/11a15a36ce5fba97bef090bbe7d22cd398eb9d57/apps/ssh-gateway/main.go#L258-L264

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

lilongen (changes requested):
See detailed review: https://github.com/boxlite-ai/boxlite/pull/541#issuecomment-4488649329

Blocking on issue #1 (`boxlite-disable-ssh` root-home asymmetry) — when the controller's default `unix_user=root` is used, disable-ssh deletes a non-existent path and the surviving `/root/.ssh/.ssh_enabled` marker causes `boxlite-ensure-ssh` to silently re-enable sshd on the next VM restart, defeating the API-level revoke. This is a security-relevant revoke bypass on the default code path.

Issues #2 (`sshUserOrDefault` comment/impl mismatch), #3 (`boxSSHMu` unbounded growth — CLAUDE.md "No unbounded queues/concurrency/memory"), and #4 (raw SSH bearer token logged in cleartext — CLAUDE.md "Mask secrets in errors and logs") should also be addressed before merge.

🤖 Generated with [Claude Code](https://claude.ai/code)

src/deps/libgvproxy-sys/gvproxy-bridge/main.go
```diff
@@ -433,6 +435,32 @@ func gvproxy_create(configJSON *C.char) C.longlong {
 		instance.vn = vn
 		instance.vnMu.Unlock()
 
+		// Start admin HTTP server for dynamic port forwarding.
+		// The runner calls POST /services/forwarder/expose on this socket to
+		// forward a host port to the guest sshd (port 22222) for real SSH access.
+		// Socket path is co-located with the tap socket in the same sockets/ directory.
+		adminSockPath := filepath.Join(filepath.Dir(config.SocketPath), "gvproxy-admin.sock")
```
DorianZheng: should create `gvproxy-admin.sock` from caller and pass to config to align with `config.SocketPath`. create the socket in the dir of `config.SocketPath` is a little bit hack

DorianZheng:
**Why does the runner compute the gvproxy admin socket path itself?**

`apps/runner/pkg/boxlite/client_ssh.go`:

```go
filepath.Join(homeDir, "boxes", boxId, "sockets", "gvproxy-admin.sock")
```

This makes `~/boxes/<id>/sockets/` an implicit contract between the runner and boxlite — if the layout ever changes (or the id space; cf. the three-level fallback in `internalBoxID`), the runner breaks at runtime with no compile-time signal. Is this coupling intentional, or should the path come from boxlite instead?

DorianZheng:
**Local builds silently skip sshd injection, so real-SSH can't be exercised end-to-end on dev machines.**

The static `boxlite-sshd` is only produced by the new CI step in `.github/workflows/build-runtime.yml` (Alpine Docker → `dist/boxlite-sshd`). On a developer's machine there's no equivalent `make` target, so each layer falls back to a warn-and-skip:

- `src/boxlite/build.rs::find_prebuilt_sshd` returns `None` → `copy_prebuilt_binary` warns and skips.
- `src/boxlite/src/rootfs/guest.rs::build_and_install` logs `boxlite-sshd not found, SSH access will be unavailable` and proceeds.
- `src/guest/src/container/spec.rs` short-circuits the bind mounts on `!Path::new("/boxlite/bin/sshd").exists()`.

Each layer fails open, so the gap doesn't surface until someone enables SSH on a locally-built sandbox and gets `enable_ssh script exited 127`. Would it make sense to add a `make sshd` target (extracting the workflow's Alpine-Docker step into a shared `scripts/build-static-sshd.sh`) so `make runtime-debug` can produce a fully SSH-capable guest locally?

DorianZheng:
should update `box.openapi.yaml` if we modify SDK API

.github/workflows/build-runtime.yml
nieyy: Good call, done — pinned OPENSSH_SHA256 for 9.7p1 and added sha256sum -c - before tar. Build fails immediately if the download is tampered. Hash is 490426f... (from openssh.com).

nieyy:
> ### Code review
> Found 4 issues:
> 
> 1. `boxlite-disable-ssh` hard-codes `HOME_DIR="/home/$UNIX_USER"` while `boxlite-enable-ssh` branches on `root`. When SSH was enabled for `root` (the controller default), enable writes the marker to `/root/.ssh/.ssh_enabled`, disable then deletes a non-existent `/home/root/.ssh/.ssh_enabled` (no-op), and the surviving marker causes `boxlite-ensure-ssh` to silently re-enable sshd on the next VM restart — defeating the API-level revoke.
> 
> https://github.com/boxlite-ai/boxlite/blob/11a15a36ce5fba97bef090bbe7d22cd398eb9d57/src/guest/scripts/boxlite-disable-ssh#L5-L10
> 
> Compare with the enable-side branch:
> 
> https://github.com/boxlite-ai/boxlite/blob/11a15a36ce5fba97bef090bbe7d22cd398eb9d57/src/guest/scripts/boxlite-enable-ssh#L6-L12
> 
> 2. `sshUserOrDefault()` doc says it returns `\"boxlite\"` and the surrounding comment claims it exists to _prevent_ inheriting root from the image, but the implementation returns `\"root\"`. The struct field doc (`// Empty means \"boxlite\"`) and the test name `TestSSHUserDefaultsToBoxlite` confirm the intended default. CLAUDE.md: "Update nearby comments when behavior changes."
> 
> https://github.com/boxlite-ai/boxlite/blob/11a15a36ce5fba97bef090bbe7d22cd398eb9d57/apps/runner/pkg/sshgateway/service.go#L192-L201
> 
> 3. Per-box mutex entries in `boxSSHMu` are inserted on first `EnableSSHAccess` but never deleted. `cleanupSSHOnDestroy` removes `sshStates` and `sshBoxes` entries but leaves `boxSSHMu` to grow without bound across the runner process lifetime. CLAUDE.md: "No unbounded queues/concurrency/memory."
> 
> https://github.com/boxlite-ai/boxlite/blob/11a15a36ce5fba97bef090bbe7d22cd398eb9d57/apps/runner/pkg/boxlite/client_ssh.go#L810-L836
> 
> 4. New log line prints the raw SSH bearer token in cleartext on the fail-closed path. The token is a live revocable credential. CLAUDE.md: "Mask secrets in errors and logs." Consider redacting (prefix + `…`) consistently with how other credentials are handled.
> 
> https://github.com/boxlite-ai/boxlite/blob/11a15a36ce5fba97bef090bbe7d22cd398eb9d57/apps/ssh-gateway/main.go#L258-L264
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.

Fixed all 4, thanks for the thorough review:

boxlite-disable-ssh root-home asymmetry — added the root branch so disable targets /root/.ssh/ consistently with enable. Revoke now works on the default code path.
sshUserOrDefault comment/test mismatch — updated the doc comment and renamed TestSSHUserDefaultsToBoxlite → TestSSHUserDefaultsToRoot, both now match the "root" implementation.
boxSSHMu unbounded growth — added delete(c.boxSSHMu, boxId) in cleanupSSHOnDestroy after releasing the mutex.
Raw token in logs — added redactToken() (keeps 8-char prefix + …), applied to both log sites.

nieyy:
> **Local builds silently skip sshd injection, so real-SSH can't be exercised end-to-end on dev machines.**
> 
> The static `boxlite-sshd` is only produced by the new CI step in `.github/workflows/build-runtime.yml` (Alpine Docker → `dist/boxlite-sshd`). On a developer's machine there's no equivalent `make` target, so each layer falls back to a warn-and-skip:
> 
> * `src/boxlite/build.rs::find_prebuilt_sshd` returns `None` → `copy_prebuilt_binary` warns and skips.
> * `src/boxlite/src/rootfs/guest.rs::build_and_install` logs `boxlite-sshd not found, SSH access will be unavailable` and proceeds.
> * `src/guest/src/container/spec.rs` short-circuits the bind mounts on `!Path::new("/boxlite/bin/sshd").exists()`.
> 
> Each layer fails open, so the gap doesn't surface until someone enables SSH on a locally-built sandbox and gets `enable_ssh script exited 127`. Would it make sense to add a `make sshd` target (extracting the workflow's Alpine-Docker step into a shared `scripts/build-static-sshd.sh`) so `make runtime-debug` can produce a fully SSH-capable guest locally?

Good suggestion — added make sshd that calls scripts/build/build-static-sshd.sh (requires Docker). CI now uses the same script so the build logic isn't duplicated. Devs can run make sshd to get a fully SSH-capable guest locally.

Added make sshd target that builds static boxlite-sshd / boxlite-ssh-keygen via Alpine Docker (same as CI). Tested on the dev machine — output is statically linked OpenSSH 9.7p1. Requires Docker Desktop locally.

nieyy:
> should update `box.openapi.yaml` if we modify SDK API

Thanks — added the SSH endpoints (`createSshAccess`, `revokeSshAccess`, `validateSshAccess`) and their schemas (`SshAccess`, `SshAccessValidationResult`, `CreateSshAccessRequest`) to `box.openapi.yaml`.

src/deps/libgvproxy-sys/gvproxy-bridge/main.go
nieyy: Fixed — added `admin_sock_path` to `GvproxyConfig` (derived from `socket_path` at construction in Rust, passed via JSON). The Go bridge now reads `config.AdminSockPath` directly, no more `filepath.Dir` arithmetic. Also removed the now-unused `path/filepath` import.

nieyy:
> **Why does the runner compute the gvproxy admin socket path itself?**
> 
> `apps/runner/pkg/boxlite/client_ssh.go`:
> 
> ```go
> filepath.Join(homeDir, "boxes", boxId, "sockets", "gvproxy-admin.sock")
> ```
> 
> This makes `~/boxes/<id>/sockets/` an implicit contract between the runner and boxlite — if the layout ever changes (or the id space; cf. the three-level fallback in `internalBoxID`), the runner breaks at runtime with no compile-time signal. Is this coupling intentional, or should the path come from boxlite instead?

Added `boxlite_box_admin_sock_path(CBoxHandle*)` to the C SDK — it constructs the path from the `home_dir` already stored in `RuntimeHandle` at runtime creation time. The Go SDK exposes this as `(*Box).AdminSockPath()`.

`EnableSSHAccess` now calls `bx.AdminSockPath()` on first use and stores the result in SSHState.AdminSockPath (persisted to `ssh-state.json`). Subsequent SSH operations (Disable, Reapply, Cleanup) retrieve it from state via `adminSockForState()`, which falls back to the old local derivation only for state files written before this change.

One related note for the future: if a VM is ever migrated to a different runner machine, the gvproxy port forward will be lost along with the socket. `EnableSSHAccess` would need to be re-called after migration to re-establish the forward and refresh the stored path. Worth keeping in mind when designing the migration path.

DorianZheng:
> > **Why does the runner compute the gvproxy admin socket path itself?**
> > `apps/runner/pkg/boxlite/client_ssh.go`:
> > ```go
> > filepath.Join(homeDir, "boxes", boxId, "sockets", "gvproxy-admin.sock")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This makes `~/boxes/<id>/sockets/` an implicit contract between the runner and boxlite — if the layout ever changes (or the id space; cf. the three-level fallback in `internalBoxID`), the runner breaks at runtime with no compile-time signal. Is this coupling intentional, or should the path come from boxlite instead?
> 
> Added `boxlite_box_admin_sock_path(CBoxHandle*)` to the C SDK — it constructs the path from the `home_dir` already stored in `RuntimeHandle` at runtime creation time. The Go SDK exposes this as `(*Box).AdminSockPath()`.
> 
> `EnableSSHAccess` now calls `bx.AdminSockPath()` on first use and stores the result in SSHState.AdminSockPath (persisted to `ssh-state.json`). Subsequent SSH operations (Disable, Reapply, Cleanup) retrieve it from state via `adminSockForState()`, which falls back to the old local derivation only for state files written before this change.
> 
> One related note for the future: if a VM is ever migrated to a different runner machine, the gvproxy port forward will be lost along with the socket. `EnableSSHAccess` would need to be re-called after migration to re-establish the forward and refresh the stored path. Worth keeping in mind when designing the migration path.

I think it's too broad to expose the admin socket. The user will not know how to use it. A set of abstraction APIs is required

nieyy:
> > > **Why does the runner compute the gvproxy admin socket path itself?**
> > > `apps/runner/pkg/boxlite/client_ssh.go`:
> > > ```go
> > > filepath.Join(homeDir, "boxes", boxId, "sockets", "gvproxy-admin.sock")
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This makes `~/boxes/<id>/sockets/` an implicit contract between the runner and boxlite — if the layout ever changes (or the id space; cf. the three-level fallback in `internalBoxID`), the runner breaks at runtime with no compile-time signal. Is this coupling intentional, or should the path come from boxlite instead?
> > 
> > 
> > Added `boxlite_box_admin_sock_path(CBoxHandle*)` to the C SDK — it constructs the path from the `home_dir` already stored in `RuntimeHandle` at runtime creation time. The Go SDK exposes this as `(*Box).AdminSockPath()`.
> > `EnableSSHAccess` now calls `bx.AdminSockPath()` on first use and stores the result in SSHState.AdminSockPath (persisted to `ssh-state.json`). Subsequent SSH operations (Disable, Reapply, Cleanup) retrieve it from state via `adminSockForState()`, which falls back to the old local derivation only for state files written before this change.
> > One related note for the future: if a VM is ever migrated to a different runner machine, the gvproxy port forward will be lost along with the socket. `EnableSSHAccess` would need to be re-called after migration to re-establish the forward and refresh the stored path. Worth keeping in mind when designing the migration path.
> 
> I think it's too broad to expose the admin socket. The user will not know how to use it. A set of abstraction APIs is required

Moved `boxlite_box_admin_sock_path` to `boxlite_internal.h` — it's now only included by first-party consumers (the Go SDK's `bridge.h`), not part of the public boxlite.h surface. The pattern is similar to MySQL's internal header split.

apps/api/src/sandbox/services/sandbox.service.ts
```diff
@@ -2101,78 +2101,346 @@ export class SandboxService {
...
+          }
+        }
+      }
+
+      // Runner is disabled (or not applicable). Now delete the DB token(s).
+      if (token) {
+        await this.sshAccessRepository.delete({ sandboxId: sandbox.id, token })
+      } else {
+        await this.sshAccessRepository.delete({ sandboxId: sandbox.id })
+      }
```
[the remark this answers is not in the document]
nieyy: Superseded — this file was renamed to `apps/api/src/box/services/box.service.ts` earlier in the PR's history (sandbox→box terminology rename). The current `revokeSshAccess` there already has the fail-closed fix this finding asks for (see line ~1740: throws when the runner record is missing but real-SSH state exists), added in commit e9ea0c1.

apps/libs/api-client/src/api/sandbox-api.ts
```diff
@@ -1819,8 +1821,8 @@ export const SandboxApiFp = function(configuration?: Configuration) {
          * @param {*} [options] Override http request option.
          * @throws {RequiredError}
          */
-        async createSshAccess(sandboxIdOrName: string, xBoxLiteOrganizationID?: string, expiresInMinutes?: number, options?: RawAxiosRequestConfig): Promise<(axios?: AxiosInstance, basePath?: string) => AxiosPromise<SshAccessDto>> {
-            const localVarAxiosArgs = await localVarAxiosParamCreator.createSshAccess(sandboxIdOrName, xBoxLiteOrganizationID, expiresInMinutes, options);
+        async createSshAccess(sandboxIdOrName: string, xBoxLiteOrganizationID?: string, expiresInMinutes?: number, createSshAccessBodyDto?: { unix_user?: string }, options?: RawAxiosRequestConfig): Promise<(axios?: AxiosInstance, basePath?: string) => AxiosPromise<SshAccessDto>> {
+            const localVarAxiosArgs = await localVarAxiosParamCreator.createSshAccess(sandboxIdOrName, xBoxLiteOrganizationID, expiresInMinutes, createSshAccessBodyDto, options);
```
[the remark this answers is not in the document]
nieyy: Superseded — this file was renamed to `apps/api/libs/api-client/src/api/box-api.ts` earlier in the PR's history. The current generated client places `createSshAccessBodyDto` after the other optional params (not before, as this finding's reproduction described) since the request body itself is optional (`required: false`) in the regenerated spec — verified no caller uses positional args that would break; the one internal caller (`useCreateSshAccessMutation.ts`) was updated to match in commit c929a08.

DorianZheng:
a design note on the sshd choice: since the guest agent is rust, the tempting alternative was embedding an ssh library in it — russh is mature (warpgate, a production ssh gateway, is built on it), and that route would delete the whole openssh supply line: the alpine/musl ci build, dist/ bundling, ext4 injection, bind mounts, even ssh-keygen.

why real openssh is still the right call here: with an embedded library, our own code becomes the security guard — "run this session as alice, with alice's permissions, nobody else's" turns into hand-rolled fork/setuid with no AllowUsers and no privilege separation. that's the known weak spot of the embedded approach (daytona's in-sandbox daemon takes it: gliderlabs server accepting any pubkey, a hardcoded password, one shared uid, exit codes flattened to 127). since unix_user enforcement is this pr's core security contract, a 25-year-audited sshd at the login boundary beats a library plus homemade login logic.

rule of thumb: embed the library where we're a broker (the external gateway could be russh one day), exec the real sshd where we're the login target. this pr lands on the right side of that line.

nieyy:
> a design note on the sshd choice...

Yeah, I hadn't looked into russh before — but I was already convinced this had to be built on real sshd from the start. Appreciate the validation!

nieyy:
Thanks for the detailed review — all 4 are fixed now:

1. `boxlite-disable-ssh` root-home asymmetry: it now branches on `root` the same way `boxlite-enable-ssh` does (`HOME_DIR="/root"` vs `/home/$UNIX_USER`), so disable correctly removes the marker for the controller's default `root` user.
   src/guest/scripts/boxlite-disable-ssh#L6-L7

2. `sshUserOrDefault()` docs now match the actual `root` default (updated the `startExecFn` doc, the `sshUser` field doc, and the function doc — all previously said "boxlite").
   apps/runner/pkg/sshgateway/service.go#L158-L195

3. `boxSSHMu` entries are now deleted in `cleanupSSHOnDestroy`, so the map no longer grows unbounded across the runner process lifetime.
   apps/runner/pkg/boxlite/client_ssh.go#L850-L852

4. SSH bearer tokens are redacted before logging (8-char prefix + `…`) on the fail-closed path.
   apps/ssh-gateway/main.go#L253, #L669-L677

Appreciate the catches — #3 and #4 in particular would've been nasty to track down in prod.

nieyy:
Need to reimplement via russh, will open another PR.

PR #543, opened 2026-05-18
fix(guest): exec timeout watcher must SIGKILL, not SIGALRM by GHSA-xjhv-pp2r-6f82

G4614:
The timeout watcher in src/guest/src/service/exec/timeout.rs sent Signal::SIGALRM (signal 14, catchable) when an exec exceeded its deadline. Any workload that installed `signal.signal(SIGALRM, SIG_IGN)` in Python or `trap '' ALRM` in shell absorbed the signal and ran to completion past the configured timeout, returning exit_code=0 — bypassing the deadline entirely.

The inline comment already said "Kill process with SIGKILL"; only the code passed the wrong Signal. Use Signal::SIGKILL (signal 9, uncatchable) so the watcher guarantees termination regardless of in-process signal handlers.

An audit of catchable-signal usage across the codebase found no other instances of this bug class. All graceful-shutdown paths (container/lifecycle.rs, exec/registry.rs, vmm/controller/shim.rs, runtime/rt_impl.rs, cli/serve/mod.rs) already escalate SIGTERM → wait → SIGKILL correctly.

Two regression tests guard the fix:

- sdks/python/tests/test_exec_timeout_sigalrm.py — Python pytest port of the reporter PoC: workload installs SIG_IGN for SIGALRM, sleeps 8s with timeout=3s. Asserts result.exit_code != 0 AND elapsed < TIMEOUT_S + 2.0.
- src/boxlite/tests/exec_options.rs::test_timeout_kills_sigalrm_ignoring_process — canonical Rust integration test (counterpart to security_enforcement.rs). Shell workload uses `trap '' ALRM; sleep 15` with timeout=2s; same assertion shape.

Verified via controlled experiment (reverting timeout.rs to SIGALRM): both new tests FAIL in the bug state with rich diagnostics (`exit_code=0`, elapsed near WORKLOAD_S), both PASS with the fix. Three independent control tests (test_api_key_credential_get_token, test_oci_symlink_escape_blocked, test_timeout_kills_long_command) PASS in both states — confirming the fix is precisely scoped and the new tests aren't trivially-passing or trivially-failing.

---

## Follow-up: Two-stage termination (commit `5eef684`)

The original commit (`19131e2`) used single-step SIGKILL. This follow-up
upgrades the watcher to the standard **SIGTERM → grace → SIGKILL** pattern
already used by `ExecRegistry::shutdown_all`, `Container::shutdown`, and
`runtime/rt_impl.rs`.

### Why

- SIGKILL alone is correct but harsh — cooperative workloads have no
  chance to flush buffers, close files, or commit transactions.
- Industry default (Docker `--stop-timeout`, K8s
  `terminationGracePeriodSeconds`, systemd `TimeoutStopSec`) is exactly
  this pattern: ask politely first, force after grace expires.
- **Hard-deadline guarantee from the original commit is preserved**:
  any workload that ignores SIGTERM via `SIG_IGN` / signal trap is
  still killed by SIGKILL after the 2-second grace.

### New stage-3 regression test

[`test_exec_timeout_sigkill_fallback_when_sigterm_ignored`](sdks/python/tests/test_exec_timeout_sigalrm.py)
exercises the SIGKILL fallback specifically: the workload installs
`SIG_IGN` for **both** SIGTERM and SIGALRM, so stages 1+2 of the
watcher are absorbed and only the uncatchable SIGKILL after grace can
stop it.

Together with the existing
`test_exec_timeout_kills_sigalrm_ignoring_process` (stage-1
verification), the suite now covers all three stages.

### Controlled experiments proving both tests catch real regressions

| `timeout.rs` state | stage-1 test | **stage-3 test** | Other 4 tests |
|---|---|---|---|
| `SIGALRM` (original bug) | **FAIL** `exit=0, 8s` | n/a | PASS |
| `SIGKILL` single-step (commit 1) | PASS | n/a | PASS |
| `SIGTERM → 2s → SIGKILL` (this commit) | PASS | **PASS** `~5s` | PASS |
| `SIGTERM`-only (synthetic regression) | PASS | **FAIL** `exit=0, 8.08s` | PASS |

Each test independently catches a distinct class of regression and
**neither is always-pass**.

### DIFF
--- a/sdks/python/tests/test_exec_timeout_sigalrm.py
+++ b/sdks/python/tests/test_exec_timeout_sigalrm.py
@@ -40,8 +40,25 @@ for _ in range(seconds):
     time.sleep(1)
 """
 
+# Stage-3 workload: ignore every catchable signal the watcher might send
+# in stages 1/2 (SIGTERM is the cooperative ask; SIGALRM is the historical
+# bug we already fixed). Only SIGKILL — which is uncatchable — can stop
+# this process. Used to verify the SIGKILL fallback after the grace window.
+IGNORE_TERM_AND_ALRM = """
+import sys, time, signal
+seconds = int(sys.argv[1]) if len(sys.argv) > 1 else 8
+signal.signal(signal.SIGTERM, signal.SIG_IGN)
+signal.signal(signal.SIGALRM, signal.SIG_IGN)
+for _ in range(seconds):
+    time.sleep(1)
+"""
+
 TIMEOUT_S = 3.0
 WORKLOAD_S = 8
+
+# Mirrors TIMEOUT_GRACE in src/guest/src/service/exec/timeout.rs.
+# Keep in sync if the guest grace constant changes.
+GRACE_S = 2.0
 
 
 async def test_exec_timeout_kills_sigalrm_ignoring_process():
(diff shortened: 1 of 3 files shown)


G4614:
@DorianZheng request for a review for the fix about [https://github.com/boxlite-ai/boxlite/security/advisories/GHSA-xjhv-pp2r-6f82](url), along with 2 new test parts, thanks

DorianZheng (approved):
LGTM!

PR #551, opened 2026-05-19
fix(jailer): propagate child errno from userns probe

G4614:
in Ubuntu 26.04, /etc/appaarmor.d/disable/unprivileged_userns closed so that the Child clone returns pid=-1 with EACCES, so EACCES should be added to the exit-code assert range

### DIFF
--- a/src/boxlite/src/jailer/bwrap.rs
+++ b/src/boxlite/src/jailer/bwrap.rs
@@ -95,7 +95,9 @@ pub fn is_available() -> bool {
 ///
 /// Performs two checks:
 /// 1. **Chrome-style raw probe** — `clone(CLONE_NEWUSER)` for kernel-level
-///    diagnosis (captures errno: EPERM, EUSERS, EINVAL, ENOSPC)
+///    diagnosis (captures errno: EPERM, EUSERS, EINVAL, ENOSPC, EACCES
+///    — last is AppArmor `userns_create` denial on Linux 6.7+, see
+///    [`credentials::check_clone_new_user_errno`](super::credentials::check_clone_new_user_errno))
 /// 2. **bwrap probe** — `bwrap --unshare-user` to test actual bwrap capability
 ///    (handles AppArmor per-binary profiles where bwrap may work even if
 ///    our process's clone fails)
--- a/src/boxlite/src/jailer/bwrap.rs
+++ b/src/boxlite/src/jailer/bwrap.rs
@@ -113,14 +113,14 @@ pub fn can_create_user_namespace() -> Result<(), String> {
     // Chrome-style raw probe for kernel-level diagnosis.
     // This may fail on AppArmor systems even when system bwrap works
     // (bwrap has its own AppArmor profile with userns permission).
-    let clone_errno = match super::credentials::can_create_process_in_new_user_ns() {
+    let clone_failure = match super::credentials::can_create_process_in_new_user_ns() {
         Ok(()) => None,
-        Err(errno) => {
+        Err(failure) => {
             tracing::debug!(
-                errno = errno,
+                failure = %failure,
                 "clone(CLONE_NEWUSER) failed — will still try bwrap (may have AppArmor profile)"
             );
-            Some(errno)
+            Some(failure)
         }
     };
 
@@ -139,7 +139,7 @@ pub fn can_create_user_namespace() -> Result<(), String> {
                 "bundled"
             };
             Err(build_diagnostic(
-                clone_errno,
+                clone_failure,
                 bwrap_source,
                 bwrap_path,
                 &stderr,
@@ -159,7 +159,7 @@ fn is_system_bwrap(path: &Path) -> bool {
 /// Reads sysctl files to detect the specific restriction and provides
 /// targeted fix commands for each scenario.
 fn build_diagnostic(
-    clone_errno: Option<i32>,
+    clone_failure: Option<super::credentials::ProbeFailure>,
     bwrap_source: &str,
     bwrap_path: &Path,
     bwrap_stderr: &str,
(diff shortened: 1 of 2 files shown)


lilongen:
### Code review

Found 1 issue:

1. `check_clone_new_user_errno` was not updated to include `EACCES` in its expected-errno set, so the very case this PR is designed to surface (AppArmor's `userns_create` denial on Ubuntu 24.04+ / Linux 6.7+) will fall through to the `_` arm and log `"clone(CLONE_NEWUSER) failed with UNEXPECTED errno"` at `tracing::error!` level on every probe. The test allowlist in `test_can_create_process_in_new_user_ns` was updated to include `EACCES`, but the production routing function was not — the two now disagree about whether `EACCES` is expected, partially defeating the fix's intent.

https://github.com/boxlite-ai/boxlite/blob/bfcc5a7d56f47c68f13099da9ed0a0402b2024af/src/boxlite/src/jailer/credentials.rs#L19-L41

Called via:

https://github.com/boxlite-ai/boxlite/blob/bfcc5a7d56f47c68f13099da9ed0a0402b2024af/src/boxlite/src/jailer/credentials.rs#L126-L130

Suggest adding `libc::EACCES` to the `EPERM | EUSERS | EINVAL | ENOSPC` arm and updating the doc-comment to mention the AppArmor case (which the rest of the PR already documents nicely on `can_create_process_in_new_user_ns`).

A few other lower-confidence observations worth a glance but not blocking:
- `decode_probe_status` returns `Err(libc::EINTR)` for a signal-killed child. `EINTR` carries retry semantics and will also trip the same `UNEXPECTED errno` log path. A dedicated sentinel (or surfacing as a distinct `Err` variant) would be clearer.
- `write_proc_file` calls `libc::close(fd)` between the failing `write` and returning `Err(())`. POSIX guarantees errno is preserved on a successful `close`, so this works today, but the classic "save errno across cleanup" pattern would make this robust against future maintenance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

lilongen (commented):
See review: https://github.com/boxlite-ai/boxlite/pull/551#issuecomment-4494027806

Non-blocking. The fix achieves its goal — the real errno propagates back to callers, so the probe is now correct. The remaining gap is that `check_clone_new_user_errno` still treats `EACCES` as "UNEXPECTED" and logs at `tracing::error!` level, so every Ubuntu 24.04+ host will emit a spurious error-level log line on every probe. The match-arm update is a one-line fix-up; happy to leave the escalation choice to the author.

🤖 Generated with [Claude Code](https://claude.ai/code)

src/boxlite/src/jailer/bwrap.rs
```diff
@@ -113,14 +113,14 @@ pub fn can_create_user_namespace() -> Result<(), String> {
     // Chrome-style raw probe for kernel-level diagnosis.
```
nieyy, line 115: Non-blocking: the function-level comment above still lists the raw probe as capturing `EPERM`, `EUSERS`, `EINVAL`, and `ENOSPC`, but this PR also makes `EACCES` an expected AppArmor `userns_create` result. Could we add `EACCES` there too so the documentation matches the behavior?

src/boxlite/src/jailer/credentials.rs
```diff
@@ -74,7 +147,7 @@ pub(crate) fn can_create_process_in_new_user_ns() -> Result<(), i32> {
 
         if pid == -1 {
```
G4614, line 148: **Where `pid == -1, errno == EACCES` comes from — and what the 24.04 vs 26.04 split actually looks like**

This `EACCES` is the AppArmor `userns_create` LSM hook's deny path. I diffed `security/apparmor/lsm.c` and `security/apparmor/task.c` between noble and resolute end to end — the **C code is byte-identical** (one blank line apart). Both releases reach the same branch:

```c
// security/apparmor/task.c :: aa_profile_ns_perm()
if (profile_unconfined(profile) && profile == profiles_ns(profile)->unconfined) {
    if (!aa_unprivileged_userns_restricted ||
        ns_capable_noaudit(current_user_ns(), CAP_SYS_ADMIN))
        return aa_get_newest_label(&profile->label);     // ALLOW (sysctl=0 or has CAP_SYS_ADMIN)

    // restricted: try to transition into a profile named "unprivileged_userns"
    new = aa_label_parse(&profile->label, "unprivileged_userns", ...);
    if (IS_ERR(new)) {
        perms.deny |= request;     // ← clone() returns -EACCES, pid == -1, *no child created*
        goto hard_coded;
    }
    perms.allow = request;         // ← clone() succeeds; child runs in lobotomized profile
    goto hard_coded;
}
```

So `pid == -1 + EACCES` requires all three of:

1. `kernel.apparmor_restrict_unprivileged_userns = 1` (sysctl, default-on since Ubuntu 23.10)
2. Caller lacks `CAP_SYS_ADMIN` in `current_user_ns()`
3. The AppArmor `unprivileged_userns` profile is **not loaded** (either masked by an `/etc/apparmor.d/disable/unprivileged_userns` symlink, or never installed)

What we'd been calling "26 always fails, 24 always works" is **not a distro difference** — it's the per-host state of (3):

- **profile loaded** → takes the `perms.allow=request` branch. `clone()` returns `pid > 0`; the child runs in the lobotomized `unprivileged_userns` profile (`audit deny capability, change_profile`), and the later `setgroups` / `gid_map` writes to `/proc` get blocked by that profile → child `_exit(errno)`. This is exactly the failure mode this PR now surfaces via `WEXITSTATUS`.
- **profile not loaded** → takes the `perms.deny` branch. `clone()` returns `-1`/`EACCES` directly, **no fork happens**, and `waitpid(-1, ..., WNOHANG)` returns `ECHILD`. This is why adding `EACCES` to the expected set and dropping the `EPERM` fallback is the right call.

Both failure modes are now covered and orthogonal — no need to special-case AppArmor here. Just worth keeping in mind when reading logs later: an `EACCES` line and a child-errno line look related but come from different kernel branches.

PR #563, opened 2026-05-20
fix(go-sdk): fold stream drain into Execution.Wait (os/exec io.Writer case)

G4614:
## Bug

Go SDK's `box.Exec` / `Cmd.Run` / `Execution.Wait` returns with stdout chunks still in flight — the caller reads a buffered Stdout sink before the trailing stream callbacks have delivered them.

- Stock pre-fix: 18-26 / 50 truncated (also `test_p0_6_exec_stdout_race.py`).
- This commit: 50/50 complete.

## Fix (Go-only, no new API, no C FFI changes)

Mirror `os/exec.Cmd.Wait`'s `Process.Wait` → `awaitGoroutines` sequence inside `Execution.Wait`. BoxLite's streams are always pushed into a user-supplied `Writer` / callback — there is no `StdoutPipe`-style user-read pipe — so every `Execution` IS `os/exec`'s **io.Writer case**, and `Wait` is the single terminal that must guarantee output completeness on return.

| Layer | Behavior |
|---|---|
| `Execution.Wait` | Reap (`e.reap(ctx)`) → block on `<-streamState.drained` (the `awaitGoroutines` analog) → return. Non-cancelable by ctx during drain (os/exec parity); only runtime shutdown breaks it. |
| Callback ordering | C-side `exit_pump` gates `on_exit` on every stream pump's `done_tx`. Go-side `deliverExit` closes the private `executionStreamState.drained` channel. So by the time `Wait` unblocks, every `OnStdout` / `OnStderr` callback (and every `io.Writer.Write`) has fired. |
| `Cmd.Run` / `box.Exec` | Just call `Execution.Wait` — no extra drain. The strong contract lives in one place. |

C FFI untouched.

## Two-side verification

`sdks/go/exec_integration_test.go::TestIntegrationExecStdoutRace` runs `echo marker-line` 20× via `Cmd.Run` without the `&& sleep 0.1` drainPad workaround that the existing `TestIntegrationExecEnvWorkingDirTimeout` uses.

| Side | `Execution.Wait` drain | Result |
|---|---|---|
| A | removed (Wait = pure reap) | 4/20 truncated to `""`, **FAIL** |
| B (this PR) | present | 20/20 complete, **PASS** |

§5 no-regression suite green: `HandleDeletedOnExit`, `NoDoubleDeleteRaceDuringClose`, `AbandonAsync*` (×6), `DrainAndDelete*` (×3), `ClaimOrFreePayload_*` (×2).

E2E (`test_p0_6_exec_stdout_race.py`) passes both sides — the Python path goes through the runner's `streamBus`, which buffers stdout in backlog regardless of close ordering, so it doesn't surface the race. The Go SDK direct test path does.

## Supersedes

#685 — same bug, C FFI `streams_pending` approach (couples Wait with streams at FFI). Closed in favor of this Go-only fix.

### DIFF
--- a/sdks/go/exec.go
+++ b/sdks/go/exec.go
@@ -92,6 +92,16 @@ type executionStreamState struct {
 	onStderr func([]byte)
 
 	released atomic.Bool
+
+	// drained is the os/exec `goroutineErr` analog: closed by
+	// deliverExit when the C-side on_exit callback fires. C's exit_pump
+	// gates Exit on every stream pump's done_tx, so by the time on_exit
+	// runs, the drain goroutine has already dispatched every
+	// Stdout/Stderr callback for this execution. Execution.Wait blocks
+	// on this after the reap so a caller using buffered Stdout/Stderr
+	// sinks never sees a truncated buffer. The fold happens at the SDK
+	// boundary; the C-side Wait task stays decoupled from streams.
+	drained chan struct{}
 }
 
 func newExecutionStreamState(opts ExecutionOptions) *executionStreamState {
@@ -100,6 +110,7 @@ func newExecutionStreamState(opts ExecutionOptions) *executionStreamState {
 		stderr:   opts.Stderr,
 		onStdout: opts.OnStdout,
 		onStderr: opts.OnStderr,
+		drained:  make(chan struct{}),
 	}
 }
 
@@ -137,6 +148,7 @@ func (s *executionStreamState) deliverStderr(data []byte) {
 
 func (s *executionStreamState) deliverExit(_ int) {
 	s.released.Store(true)
+	close(s.drained)
 }
 
 func (s *executionStreamState) markReleased() {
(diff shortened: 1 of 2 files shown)


DorianZheng (dismissed):
LGTM

DorianZheng:
clean, glad it's just go now.

small thing: the drain only happens in Cmd.Run and box.Exec, so Execution.Wait on its own still returns before the pumps finish. if someone passes their own Stdout writer and calls Wait directly, they still get a short read:

```go
exec, _ := box.StartExecution(ctx, ExecutionOptions{Command: ..., Stdout: &buf})
exec.Wait(ctx)
buf.String() // still truncated
```

our streams are always pushed into a writer (theres no StdoutPipe to read yourself), so Wait is really the io.Writer case and should block until the bytes are delivered, the same way os/exec's Wait does. could you move the `<-drained` select into Execution.Wait right after the reap, and drop it from the two wrappers? that fixes the direct path too and gets rid of the duplicated select.

also the title still says fix(ffi) but theres no rust left in here now — fix(go-sdk) fits better.

DorianZheng (approved):

PR #567, opened 2026-05-21
feat: opt-in --support-docker for running dockerd inside a box

G4614:
## Summary

Adds opt-in dind support to `boxlite run` via `--support-docker`, so you can run `dockerd` (and `docker build` / `docker run`) inside a box. Default behaviour is byte-for-byte unchanged — every existing user keeps the lean, non-docker flow.

Stack:

- **`feat(box): opt-in --support-docker`** — wires the flag through CLI → proto → guest; gates extra caps + cgroup rw + the dind kernel behind the explicit opt-in.
- **`feat(cli): --entrypoint`** — override an image's ENTRYPOINT (needed for early dind bring-up; kept as a general-purpose flag).
- **`feat(box): Phase B — dind-capable libkrunfw blob + runtime selection`** — adds `scripts/build/build-libkrunfw-dind.sh` + `make libkrunfw-dind`. The dind kernel (mqueue, netfilter, cgroup2, overlay) is built once (~10–20 min, cached) and selected per-box only when `--support-docker` is set; lean boxes still boot on the stock libkrunfw.
- **`feat(cli): --cmd`** — pass args to the image's real ENTRYPOINT without replacing it. Gated on `--support-docker` so the lean flow's `[COMMAND...]` semantics (secondary-exec attach) are unchanged. This drops the previous `--entrypoint sh` dind bypass; the image's own `dockerd-entrypoint.sh` now runs as PID 1.
- **`test(cli): dind end-to-end`** — `src/cli/tests/dind_build.rs` boots `docker:dind` under `--support-docker` and asserts `docker build` succeeds end-to-end against the real VM.

## Current limitation: `network=host` only

`dockerd` inside the box is launched with `--bridge=none --iptables=false --storage-driver=vfs`. That means:

- **`docker build` must use `--network=host`** (the e2e test does exactly this).
- `docker run` inside the box also effectively runs in the box's host network — there is no docker0 bridge and no NAT/iptables.
- Container-to-container networking via a user-defined bridge is **not** supported in this PR.

Why: libcontainer 0.5.7 still bind-remounts `/proc/sys` read-only inside the box, which breaks dockerd's default bridge + iptables setup. Disabling them is what lets dind boot cleanly today; restoring full bridged networking is follow-up work (needs either a libcontainer bump or a guest-side `/proc/sys` shim).

## One-line test

```
make libkrunfw-dind && make test:integration:dind
```

- `make libkrunfw-dind` is the one-time ~10–20 min kernel build (cached at `target/dind-kernel/lib64/libkrunfw-dind.so.5` afterwards).
- `make test:integration:dind` runs just the dind end-to-end (`dind_supports_docker_build`). Under the hood it's a thin alias for `make test:integration:cli FILTER=dind_supports_docker_build`, which exports `BOXLITE_LIBKRUNFW_DIND_PATH` so the CLI build staples the dind blob into the embedded runtime.
- On hosts that genuinely can't run dind (no nested virt, restrictive security profile): `BOXLITE_SKIP_DIND_TEST=1 make test:integration:cli` skips just that test.

## Test plan

- [x] `make test:unit:core` (rust + ffi + cli unit suites) green
- [x] `make libkrunfw-dind && make test:integration:dind` green on a host with nested virt
- [x] `make test:integration:cli` green (full CLI matrix incl. dind)
- [ ] Spot-check on a host without nested virt: `BOXLITE_SKIP_DIND_TEST=1 make test:integration:cli` skips dind, rest passes
- [ ] Verify `boxlite run` without `--support-docker` is byte-for-byte unchanged on a representative image

### DIFF
--- a/make/build.mk
+++ b/make/build.mk
@@ -1,4 +1,4 @@
-PHONY_TARGETS += guest shim runtime cli cli\:release skillbox-image build\:apps
+PHONY_TARGETS += guest shim runtime cli cli\:release skillbox-image build\:apps libkrunfw-dind
 
 guest:
 	@bash $(SCRIPT_DIR)/build/build-guest.sh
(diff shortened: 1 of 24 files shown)


G4614:
Superseded by #568. Same diff, but the flag is renamed from `--support-docker` to `--privileged` (operation-named: capability release + cgroup rw + dind-capable kernel; matches the `docker run --privileged` convention so the flag's semantics aren't tied to dockerd specifically), and a clippy regression on macOS (dead `stage_optional_dind_blob` stub) is fixed.

PR #568, opened 2026-05-21
feat: opt-in --privileged for in-box privileged execution (dind support)

G4614:
## Summary

Adds opt-in privileged execution to `boxlite run` via `--privileged`, so you can run `dockerd` (and `docker build` / `docker run`) inside a box. Default behaviour is byte-for-byte unchanged — every existing user keeps the lean, sandboxed flow.

The `--privileged` flag bundles the four guest-side knobs an in-box dockerd needs:

1. **Capability release.** Container runs with `docker run --privileged`-equivalent caps (`src/guest/src/container/capabilities.rs`).
2. **Cgroup v2 rw mount.** Guest mounts `/sys/fs/cgroup` rw so dockerd / runc can move processes into subtrees they own.
3. **Dind-capable libkrunfw selection.** Per-box runtime picks between the lean libkrunfw (default) and a "fat" libkrunfw with mqueue / netfilter / cgroup2 / overlay modules. The fat blob is built out-of-band via `make libkrunfw-dind` and staged in by `BOXLITE_LIBKRUNFW_DIND_PATH`; a normal build stays byte-identical to today.
4. **Tenant `readonly_paths` / `masked_paths` leak mitigation.** `libcontainer::container::tenant_builder::adapt_spec_for_tenant` (libcontainer-0.5.7/src/container/tenant_builder.rs:442) rebuilds the Linux block from `LinuxBuilder::default()`, dropping the init spec's empty `readonly_paths` / `masked_paths` and re-injecting the OCI defaults via `Linux::default()`. Every tenant exec (the foreground exec, every subsequent `boxlite exec`) then ro-bind-mounts `/proc/sys` etc. into the **shared** mount namespace, which breaks dockerd's default bridge (write to `/proc/sys/net/ipv4/ip_forward` → EROFS). Bug present in libcontainer 0.5.7, 0.6.0, and youki upstream `main`. Mitigated by `BoxliteExecutor` (`src/guest/src/container/executor.rs`): a custom libcontainer `Executor` that, when the spec carries the `io.boxlite.privileged=true` annotation, `umount2(MNT_DETACH)`s the OCI default ro/masked paths right before execvp. Non-privileged boxes are byte-for-byte unaffected — the executor delegates to libcontainer's `DefaultExecutor` whenever the annotation is absent.

Commits:

- **`feat(box): opt-in --privileged`** — caps + cgroup rw + dind kernel selection; proto + SDK plumbing; the dind libkrunfw config and build script.
- **`feat(cli): wire --privileged, --entrypoint, --cmd through ManagementFlags`** — CLI surface, with `--cmd` gated on `--privileged` so the lean flow's `[COMMAND...]` semantics (secondary-exec attach) are unchanged.
- **`test(cli): dind end-to-end + make targets`** — `src/cli/tests/dind_build.rs` boots `docker:dind` under `--privileged` and asserts `docker build` succeeds end-to-end; new `test:unit:cli` and `test:integration:dind` make targets.
- **`test(privileged): pin OCI spec readonly_paths contract`** — guest unit test asserting `create_oci_spec(privileged=true)` emits `readonly_paths`/`masked_paths` as `Some([])` and sets the `io.boxlite.privileged` annotation, both surviving `Spec::save`/`Spec::load`.
- **`fix(box): drop dockerd dind workaround flags via BoxliteExecutor`** — root-causes the `/proc/sys` ro symptom to the libcontainer tenant-builder leak; introduces `BoxliteExecutor` and the `io.boxlite.privileged` annotation; drops `--cmd --bridge=none --cmd --iptables=false --cmd --storage-driver=vfs` from the dind test and stops forcing `docker build --network=host`.
- **`refactor(executor): delegate to libcontainer DefaultExecutor`** — `BoxliteExecutor::validate` now delegates to `DefaultExecutor::validate` so the non-privileged code path (where the `io.boxlite.privileged` annotation is absent) is byte-for-byte equivalent to a stock libcontainer run.

Naming note: an earlier draft of this PR called the flag `--support-docker`. That framed the flag in terms of one use case; `--privileged` describes the actual operation (capability release + cgroup rw + privileged-capable kernel + tenant-builder leak mitigation) and matches the existing `docker run --privileged` convention. Supersedes #567.

## What `docker run` / `docker build` look like in-box

With `BoxliteExecutor` in place, dockerd boots under default flags:

- **Default bridge.** `docker0` comes up; per-container veth pairs and bridge networking work.
- **Default iptables.** `iptables-nft` rules install cleanly (the dind kernel carries `NF_TABLES` + `NFT_*`).
- **Default storage driver.** `overlayfs` works (the dind kernel carries `CONFIG_OVERLAY_FS`).
- **`docker build` without `--network=host`.** Builds pull base images and run `RUN` steps over the default bridge.

Verified by `src/cli/tests/dind_build.rs` — see the end-to-end log in CI (boots `docker:dind`, runs `docker build` against an Alpine `FROM` + `RUN` Dockerfile, asserts `naming to docker.io/library/boxlite-dind-test:1` in the build output).

## One-line test

```
make libkrunfw-dind && make test:integration:dind
```

- `make libkrunfw-dind` is the one-time ~10–20 min kernel build (cached at `target/dind-kernel/lib64/libkrunfw-dind.so.5` afterwards).
- `make test:integration:dind` runs just the dind end-to-end (`dind_supports_docker_build`). Under the hood it's a thin alias for `make test:integration:cli FILTER=dind_supports_docker_build`, which exports `BOXLITE_LIBKRUNFW_DIND_PATH` so the CLI build staples the dind blob into the embedded runtime.
- On hosts that genuinely can't run dind (no nested virt, restrictive security profile): `BOXLITE_SKIP_DIND_TEST=1 make test:integration:cli` skips just that test.

## One-line manual reproducer (no test harness)

Same end-to-end path the test exercises, condensed to a single `boxlite run` invocation — boots `docker:dind`, waits for `dockerd` to come up, runs `docker build` against a Dockerfile piped on stdin. Default dockerd flags, default bridge, no `--network=host`:

```
boxlite run --rm --privileged --memory 2048 docker:dind \
  sh -c 'until [ -S /var/run/docker.sock ]; do sleep 1; done; \
         printf "FROM alpine:3.19\nRUN echo built\n" | \
         docker build -t boxlite-dind-test -'
```

Prereq: same as the test — `make libkrunfw-dind` must have been run once, and the CLI must have been built with `BOXLITE_LIBKRUNFW_DIND_PATH` exported so the dind blob is staged into the embedded runtime (the `test:integration:cli` target does this for you; for an out-of-band build, export it before `make cli`). Success looks like exit 0 plus `naming to docker.io/library/boxlite-dind-test done` in the output.

## Follow-up

The clean upstream fix is a youki PR that teaches `adapt_spec_for_tenant` to preserve the init spec's `readonly_paths` / `masked_paths`. Once that ships in a libcontainer release we can bump the dep and delete `BoxliteExecutor`. Not blocking this PR.

## Test plan

- [x] `make fmt:check` clean
- [x] `make clippy` clean
- [x] `make test:unit:core` (rust + ffi + cli unit suites) — 97 CLI tests, 76 guest container tests green
- [x] `make libkrunfw-dind && make test:integration:dind` green on a host with nested virt (default bridge, default iptables, default storage, no `--network=host`)
- [x] Representative non-privileged `boxlite run` integration tests (`test_run_exit_code_success`, `test_run_command_not_found`, `test_run_single_env_var`, `test_run_stdout_capture`) — 4/4 pass
- [x] Representative non-privileged `boxlite exec` tenant-path integration tests (`test_exec_on_running_box`, `test_exec_multiple_times_same_box`) — 2/2 pass
- [ ] Spot-check on a host without nested virt: `BOXLITE_SKIP_DIND_TEST=1 make test:integration:cli` skips dind, rest passes
- [ ] Verify `boxlite run` without `--privileged` is byte-for-byte unchanged on a representative image


nieyy:
One thing I think we should handle before merging: the REST/remote path seems to drop `--privileged` today.

Local `boxlite run --privileged ...` sets `BoxOptions.privileged`, but when the CLI is using `--url` / `BOXLITE_REST_URL`, `CreateBoxRequest::from_options` does not include a `privileged` field in the JSON request. So a user could ask for `--privileged` against dev.boxlite.ai and silently get a normal box instead.

I think we should either plumb this through the REST DTO/API path, or fail fast in remote mode with a clear “--privileged is not supported remotely yet” error. Silently ignoring it would be pretty confusing.

nieyy:
Another thing I would prefer to tighten: if `--privileged` needs the dind-capable libkrunfw for the advertised docker-in-box behavior, I think missing `libkrunfw-dind.so.5` should fail fast instead of warning and falling back to the lean kernel.

Otherwise the box starts in a half-working state: caps/cgroup are relaxed, but bridge/iptables/overlayfs can still fail later inside dockerd, which will be much harder for users to understand. A clearer error like “dind libkrunfw not found; run `make libkrunfw-dind` and rebuild with `BOXLITE_LIBKRUNFW_DIND_PATH`” would be better than letting dockerd fail downstream.

nieyy:
Small runtime-safety point on `BoxliteExecutor`: I think the `umount2(MNT_DETACH)` loop should only ignore the expected errors, not every error.

`EINVAL` / `ENOENT` make sense to ignore because some paths may not be mount points or may not exist on a given kernel. But if we get something unexpected like `EPERM` or another mount-namespace error, continuing into `exec` will likely just make dockerd fail later with a much less useful “/proc/sys is read-only” style error.

Could we match the errno and fail clearly for unexpected unmount failures, ideally including the path + errno?

G4614:
Rename pass landed in `b4fdf3a`: `libkrunfw-dind` → `libkrunfw-privileged` everywhere on the internal surface, because the kernel blob is the **privileged** kernel; dind is just one client of it. Future privileged use cases (overlayfs-in-box, k8s-in-box, etc.) will load the same blob, so naming it after one workload was misleading.

**Renamed**
- make target: `libkrunfw-dind` → `libkrunfw-privileged`
- build script: `scripts/build/build-libkrunfw-dind.sh` → `build-libkrunfw-privileged.sh`
- env var (build-time): `BOXLITE_LIBKRUNFW_DIND_PATH` → `BOXLITE_LIBKRUNFW_PRIVILEGED_PATH`
- staged blob: `libkrunfw-dind.so.5` → `libkrunfw-privileged.so.5`
- kconfig overlay dir: `src/deps/libkrun-sys/dind-configs/` → `privileged-configs/`
- overlay files: `overlay-dind_<arch>` → `overlay-privileged_<arch>`
- `libkrun-sys/build.rs::stage_optional_dind_blob` → `stage_optional_privileged_blob`
- `boxlite/src/vmm/controller/spawn.rs::stage_dind_libkrunfw` (+ all local vars: `dind_libs` / `dind_blob`) → `stage_privileged_libkrunfw` / `privileged_libs` / `privileged_blob`
- all log/error/comment strings updated

**Intentionally kept** (use-case / external surface):
- `test:integration:dind` make target — names the use case, not the kernel
- `dind_build.rs` / `dind_compose.rs` filenames + `dind_supports_docker_build` / `dind_compose_multi_service_network` test fn names
- `docker:dind` image tag in the tests — that's the upstream Docker image name
- `BOXLITE_SKIP_DIND_TEST` env var — user-facing opt-out for the dind integration tests; renaming would break any CI gates set on it

**Verification** (`--privileged` kernel rebuilt; all suites run single-threaded for dind to avoid the known concurrent gvproxy/conntrack flake documented on `dind_compose_multi_service_network`):

| Stage | Result |
| --- | --- |
| `cargo check` + `cargo build -p boxlite-cli --tests` | clean |
| `make libkrunfw-privileged` | 21.4 MB blob, BuildID `f75efd9d…` (kconfig unchanged, only filenames moved) |
| `spawn::tests` + `util::process` unit tests | **23/23 PASS** in 0.56s |
| `guest::container::executor` (Review #3 EPERM path) | **3/3 PASS** in 5ms |
| `dind_supports_docker_build` (single nextest invoke) | **PASS** in 28.16s |
| `dind_compose_multi_service_network` (single nextest invoke) | **PASS** in 36.33s |
| dind stacked under `--test-threads=1` | **2/2 PASS** in 68.22s |

G4614:
Followup landed in \`bafe537\`: dropped the \`BOXLITE_LIBKRUNFW_PRIVILEGED_PATH\` env var as a required step, and fixed a stale-artifact bug it had been masking.

**The env var was load-bearing for one thing: detecting that \`make libkrunfw-privileged\` had been run.** The only path it ever carried was the canonical output of that target (\`<workspace>/target/privileged-kernel/lib64/libkrunfw-privileged.so.5\`), and \`build.rs\` can check that path itself. The variable created an easy-to-miss step — run the make target but forget the \`export\`, and the build silently produces a lean-only \`boxlite\` whose \`--privileged\` flag fails confusingly far from the root cause. Review #2's fail-fast guard plastered the symptom; this removes the source.

**After this commit, the developer workflow is one button:**

\`\`\`sh
make libkrunfw-privileged    # one-time ~10–20 min kernel build, cached after
make runtime:debug           # build.rs auto-detects the blob; no export step
boxlite run --privileged ... # uses the privileged kernel
\`\`\`

\`BOXLITE_LIBKRUNFW_PRIVILEGED_PATH\` is preserved as an explicit override for CI caches / distro packaging that stage the blob outside the workspace tree, and \`cargo:rerun-if-changed=<canonical>\` ensures cargo re-stages when the blob appears, disappears, or changes mtime.

**Stale artifact fix bundled in:**

While verifying the rename, the build log showed both \`libkrunfw-privileged.so.5\` (new) AND a 21 MB \`libkrunfw-dind.so.5\` (pre-rename) being bundled into \`boxlite\`'s embedded runtime. \`libkrun-sys\`'s \`OUT_DIR\` is persistent across cargo runs and only writes — never deletes — so a privileged variant staged by a prior build lingers forever. Added two targeted sweeps:

- \`libkrun-sys/build.rs::clean_stale_libkrunfw_variants\` — runs before \`LibFixup::fix\` walks the dir, removes any \`libkrunfw-*.so*\` variant (lean \`libkrunfw.so*\` has no dash and is preserved). The current run's privileged blob gets re-staged immediately after.
- \`boxlite/build.rs::prune_stale_libkrunfw_variants\` — when bundling libkrunfw into the embedded \`runtime/\` dir, drops any \`libkrunfw-<variant>.so*\` already in the destination that the current source no longer produces.

**Verification** (\`BOXLITE_LIBKRUNFW_PRIVILEGED_PATH\` deliberately unset; canonical-path detection only):

| Stage | Result |
| --- | --- |
| \`make runtime:debug\` w/o env var | auto-staged canonical privileged blob |
| Embedded runtime size | **8 files / 423.9 MB → 7 files / 403.5 MB** (stale dind blob gone) |
| \`find target -name "libkrunfw-dind*" -newer <baseline>\` | empty (no new dind blob in any fresh OUT_DIR) |
| spawn + util unit tests | **23/23 PASS** in 4.58s |
| guest \`container::executor\` (Review #3 EPERM path) | **3/3 PASS** in 0.013s |
| \`dind_supports_docker_build\` (single nextest invoke) | PASS in 32.76s |
| \`dind_compose_multi_service_network\` (single nextest invoke) | PASS in 41.74s — DNS + ICMP + TCP service-to-service all green |
| dind stacked under \`--test-threads=1\` | **2/2 PASS** in 67.44s |

DorianZheng:
Can you split these PR into multiple small PRs and create a issue to list all the related PRs for tracking?

G4614:
👌

src/guest/src/container/capabilities.rs
```diff
@@ -58,6 +58,122 @@ pub fn capability_names() -> Vec<String> {
...
+///
+/// SECURITY: This is the full Linux capability set. Code running as root
+/// inside a box with this set can mount arbitrary filesystems, modify
+/// network configuration, write to all of /proc and /sys, and perform every
+/// container-escape primitive that requires capabilities. The microVM
+/// boundary still contains the process — host stays protected by KVM/libkrun
+/// isolation — but the in-VM blast radius is effectively `root`. Only use
+/// this for trusted workloads or where you would otherwise use
+/// `docker run --privileged`.
+pub fn docker_capabilities() -> HashSet<Capability> {
```
DorianZheng, line 76: internal code shouldn't understand the concept of docker. we need a general way to let the caller who understand docker pass the caps all the way to here

PR #574, opened 2026-05-22
fix(quality): lint:python tomllib check works on Python 3.10

G4614:
## Summary
- `make lint:python` imported `tomllib` unconditionally, but that module is 3.11+ stdlib — the SDK declares `requires-python = ">=3.10"`, so the lint step blew up on 3.10 envs.
- Fall back to the `tomli` backport when running on 3.10, and declare `tomli>=2.0; python_version < '3.11'` as a dev dep so it's actually present instead of relying on transitive pulls from `maturin`/`pytest`.

## Test plan
- [ ] `make lint:python` passes on Python 3.10
- [ ] `make lint:python` passes on Python 3.11+


DorianZheng (approved):
LGTM!

Issue #544, opened 2026-05-19
Question: should `test_can_create_process_in_new_user_ns` treat AppArmor’s EACCES as an expected errno?


G4614:
### Context

While running `make test:unit:rust` on Ubuntu (with `kernel.apparmor_restrict_unprivileged_userns=1`, the default on 23.10+), I hit a single non-environmental failure that I want to ask about before assuming it's a bug:

```
thread 'jailer::credentials::tests::test_can_create_process_in_new_user_ns' panicked at
src/boxlite/src/jailer/credentials.rs:259:17:
Unexpected errno: 13 (Permission denied (os error 13))
```

### What I think is happening

The probe `can_create_process_in_new_user_ns()` is a port of Chrome's
`Credentials::CanCreateProcessInNewUserNS()` and the test asserts that any
`Err(errno)` returned by it falls into Chrome's documented set:

```rust
errno == libc::EPERM
    || errno == libc::EUSERS
    || errno == libc::EINVAL
    || errno == libc::ENOSPC,
```

(`src/boxlite/src/jailer/credentials.rs:259-263`)

On AppArmor-restricted hosts (Ubuntu 23.10+, the sysctl `kernel.apparmor_restrict_unprivileged_userns=1`), the kernel's AppArmor module denies `clone(CLONE_NEWUSER)` for processes whose binary has no profile granting `userns,`, and the syscall returns `EACCES` (13) instead of the older `EPERM`. This is the regular behavior on a stock Ubuntu 24.04 install.

The companion production helper `check_clone_new_user_errno` already handles this case forgivingly — it only logs a `tracing::error!("...UNEXPECTED errno")` and otherwise returns the errno without crashing (the inline comment says: *"Chrome PCHECK crashes here; we log error instead — we're a library"*). So production code tolerates the new errno; only the test asserts strictly.

### My question

Is this test intentionally a **canary** for "Chrome's documented errno set is incomplete on modern kernels" — i.e., is the panic on `EACCES` exactly the signal the test is supposed to give, prompting a maintainer to extend the set?

If yes, then the resolution is presumably:

- Add `libc::EACCES` to the match arm in `check_clone_new_user_errno` (line 22) so the production log stays at `debug` level instead of `error`.
- Add `libc::EACCES` to the test's assertion set.
- Maybe note "AppArmor userns restriction (Ubuntu 23.10+)" in a comment near the match for future readers.

[...]

Issue #468, opened 2026-05-03
Jailer integration tests fail on Ubuntu 24+ default kernel/AppArmor (unprivileged userns blocked)


DorianZheng:
## Summary

Two integration tests in `src/boxlite/tests/jailer.rs` fail on hosts where unprivileged user namespaces are blocked by the kernel or AppArmor:

- `jailer::jailer_creates_isolated_mount_namespace`
- `jailer::jailer_enabled_box_starts_and_executes`

The failure surfaces during the box start preflight as:

```
called `Result::unwrap()` on an `Err` value:
Config(\"Sandbox preflight failed: bwrap cannot create user namespaces.

failed to run bwrap: Permission denied (os error 13)\")
```

Kernel audit log on a fresh Ubuntu 24.04 dev host:

```
audit: type=1400 apparmor=\"DENIED\" operation=\"capable\"
profile=\"unprivileged_userns\" capability=21 capname=\"sys_admin\"
```

## Reproduction

```bash
# fresh Ubuntu 24.04 EC2 host (kernel 7.0.0-1004-aws)
git clone https://github.com/boxlite-ai/boxlite
cd boxlite
make setup
make runtime:debug
make test:integration:rust FILTER=jailer_enabled_box
```

`bwrap --unshare-user --bind / / true` succeeds in isolation, but the libcontainer-driven preflight in the jailer requires `CAP_SYS_ADMIN` inside the userns and that is denied by the `unprivileged_userns` AppArmor profile that ships in `apparmor` ≥ 4.0 (Ubuntu 24+ default).

Setting:
- `kernel.unprivileged_userns_clone=1`
- `kernel.apparmor_restrict_unprivileged_userns=0`
- `apparmor_parser -R /etc/apparmor.d/unprivileged_userns`

does not unblock the preflight on this host. The kernel/AppArmor stack appears to require additional unconfinement for `bwrap`.

## Decision (this PR)

These two tests are marked `#[ignore]` until a portable fix lands. The skip reason is encoded in the `#[ignore]` attribute so it shows up in test output. Run them explicitly with:

```bash
cargo nextest run -p boxlite --features krun,gvproxy --test jailer --run-ignored only
```

## Options for a real fix

1. **Test-side preflight** — detect userns availability at test start, skip with clear logging on unsupported hosts. Keeps the signal in CI runners with permissive kernels.
2. **Run jailer tests in a privileged container** in CI (the runc / youki convention).
[...]

PR #575, opened 2026-05-22
fix(security): paired redaction — stop CA key/secret leaks via trace! + Debug

G4614:
## Summary

Paired secret-redaction fixes against the same threat model — CA private keys and user-provided secrets must not bleed into log files or Debug output. Originally opened as two PRs (#575 + the now-superseded #576); merged here because each is incomplete without the other.

- **Commit a4c4172** (this PR's original commit): stop `tracing::trace!(config = %config_json, ...)` in `vmm::controller::shim::ShimController::start`. That line dumped the serialized `InstanceSpec` to the box log, leaking `NetworkBackendConfig.secrets` and (when MITM proxy is configured) the PKCS8 CA private key whenever `RUST_LOG=trace` is set and any operator reads the box log. Replaces the dump with a redacted summary (`engine` / `box_id` / `json_bytes=<len>`) emitted from a single audited helper `emit_redacted_box_config_trace`, so the trace site has exactly one auditable callsite. The whole reason the config travels via stdin rather than CLI args was to keep these bytes out of `/proc/<pid>/cmdline` (see `spawn.rs:97-99`) — trace-logging the same blob defeated that mitigation.

- **Commit ada6b59** (was the standalone PR #576): implement `Debug` manually on `GvproxyConfig` and `NetworkBackendConfig` so `ca_key_pem` / `ca_cert_pem` render as `Some("[REDACTED]")` instead of the full PEM. Defense-in-depth: a grep of the codebase finds **zero** production `{:?}` callers on either type today, but a future `tracing::debug!(?config, ...)` or `format!("{:?}", config)` would silently regress the trace-leak class fixed by the first commit. `Serialize` / `Deserialize` are unchanged (wire format byte-identical). The companion test for `NetworkBackendConfig::Debug` lives in `src/net/mod.rs` as a unit test (visible to `make test:unit:rust`). `GvproxyConfig` lives behind `#[cfg(feature = "gvproxy")]` and `make test:unit:rust` uses `--no-default-features`, so the assertion for it is placed in `tests/gvproxy_debug_redaction.rs` (picked up by the existing `make test:integration:rust` target with `--features krun,gvproxy --test '*'`).

## Why neither half is complete alone

- Only the trace! fix → any future `Debug`-format call on these structs still leaks
- Only the `Debug` redaction → the current `trace!(config = %config_json)` site still leaks today

Together they form a closed loop: current leak source fixed, future regression class blocked.

Closes boxlite-ai/boxlite#576.

## Test plan

- [x] **PR-introduced tests pass** — `cargo nextest run -p boxlite --lib -E 'test(/redacted_box_config|debug.*redact|network_backend_config_debug|gvproxy.*debug/)'` → **4/4 PASS**.
- [x] **Gvproxy-gated integration test** — `cargo nextest run -p boxlite --features krun,gvproxy --test gvproxy_debug_redaction` → **1/1 PASS** (the `GvproxyConfig` Debug redact case, compiles + runs only under the gvproxy feature).
- [x] **Broader regression sweep** — `cargo nextest run -p boxlite --lib --no-fail-fast` → 707 tests run, **705 PASS** + 2 pre-existing failures (`jailer::credentials::test_can_create_process_in_new_user_ns`, `jailer::tests::test_jailer_full_flow_with_real_tempdir`) — both are AppArmor host-environment issues unrelated to this change, tracked upstream as #468 / #544 (verified via `git stash` against `main`).
- [x] `cargo nextest run -p boxlite-shared --lib` → **36/36 PASS**.
- [x] `cargo clippy -p boxlite --features krun,gvproxy --no-deps` and `cargo clippy -p boxlite-shared --no-deps` → both exit 0, no warnings.
- [x] **Reverse verification** per the original PRs:
  - Re-introducing `config = %config_json` inside `emit_redacted_box_config_trace` → the trace test fails with CA-key / secret sentinel bytes echoed in captured subscriber output.
  - Re-deriving `Debug` on `GvproxyConfig` (regression injection) → the integration test fails with the CA-key sentinel echoed.
  - Both restore green on revert. The assertions actually catch the bug they claim to.

## Out of scope (flagged for follow-up — not addressed here)

The `net::gvproxy` submodule has ~28 in-crate `#[test]` items that are likewise invisible to `make test:unit:rust` (it uses `--no-default-features`). Fixing that broader gap — e.g. by adding a `test:unit:rust:gvproxy` make target — is wider than this PR and deserves its own change.

### DIFF
--- a/src/boxlite/src/net/gvproxy/config.rs
+++ b/src/boxlite/src/net/gvproxy/config.rs
@@ -40,7 +40,11 @@ pub struct PortMapping {
 ///
 /// This structure encapsulates all configuration needed to create a gvproxy
 /// virtual network, replacing the previous approach of hardcoding values in Go.
-#[derive(Debug, Clone, Serialize, Deserialize)]
+///
+/// `Debug` is implemented manually (see below) because the derived form would
+/// print the raw `ca_key_pem` — a PKCS8 CA private key — which must never
+/// land in a log file.
+#[derive(Clone, Serialize, Deserialize)]
 pub struct GvproxyConfig {
     /// Unix socket path for the network tap interface.
     /// Caller-provided to ensure each box gets a unique, collision-free path.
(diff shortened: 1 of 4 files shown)


G4614:
Heads-up for reviewers: #576 has been closed and folded into this PR.

The standalone PR #576 (`fix(net): redact ca_key_pem / ca_cert_pem in Debug impls (defense-in-depth)`) is logically the other half of this change — same threat model (CA private key / user secrets must not leak through log or Debug output), but covering a distinct callsite (`Debug` derive on `GvproxyConfig` / `NetworkBackendConfig`) rather than the `trace!(config = %config_json, ...)` site fixed here. Each half is incomplete on its own:

- Without #575 → the current `RUST_LOG=trace` leak is still live.
- Without #576 → any future `tracing::debug!(?config, ...)` or `format!("{:?}", config)` silently re-introduces the same leak class.

To make the threat→fix→regression-guard loop a single review unit, #576's commit (originally `2ea9ae4`) was cherry-picked onto this branch as `ada6b59`. This PR's HEAD is now `ada6b59`, two commits total, fast-forward push (no force-push). The integrated body above (Summary / Why neither half is complete / Test plan / Out of scope) replaces both original bodies.

Cross-references:
- #576 close comment: https://github.com/boxlite-ai/boxlite/pull/576#issuecomment-4517176552
- Original #575 commit: a4c4172 (unchanged, still the first commit here)
- Original #576 commit: 2ea9ae4 → cherry-picked as ada6b59

DorianZheng (approved):
LGTM

PR #576, opened 2026-05-22
fix(net): redact ca_key_pem / ca_cert_pem in Debug impls (defense-in-depth)

G4614:
## Summary

- `GvproxyConfig` and `NetworkBackendConfig` both held the MITM proxy's PKCS8 CA private key in `ca_key_pem: Option<String>` (and the matching cert PEM in `ca_cert_pem`) and **derived** `Debug`. Any log line that used `{:?}` on one of these — or on a struct that transitively held one, e.g. `InstanceSpec.network_config` — would have printed the full PEM.
- Defense-in-depth: PR #575 (companion) already cuts off the JSON-serialization path that was the one known leak site, but a future `tracing::debug!(?config, ...)` or `format!(\"{:?}\", config)` would silently regress.
- Implement `Debug` manually on both types so `ca_key_pem` / `ca_cert_pem` render as `Some(\"[REDACTED]\")` while every other field renders identically to the previous derived form. `secrets: Vec<…>` is unchanged because the inner `Secret` / `GvproxySecretConfig` already redact their `value` field via their own custom `Debug` impl.

## Why no current behavior changes

I grep'd the codebase: there are **zero** production `{:?}` callers on either type today (one match in `tests/secret_substitution.rs` is on `GvproxySecretConfig`, a different type that is untouched here). `Serialize` / `Deserialize` derives are unchanged, so the wire format is byte-identical. The only observable change is the rendered form of `format!(\"{:?}\", config)` for these two structs, and nothing in production code reads that today.

## Test placement note

The companion assertion for `NetworkBackendConfig::Debug` lives in `src/net/mod.rs` as a `#[test]`, reachable from the standard `make test:unit:rust` target. `GvproxyConfig` lives behind `#[cfg(feature = \"gvproxy\")]`, and `make test:unit:rust` uses `--no-default-features` to avoid the Go runtime link (see `make/test.mk:177`), so an in-crate `#[test]` for it would never compile under that target. Put that assertion in `tests/gvproxy_debug_redaction.rs` gated on `#[cfg(feature = \"gvproxy\")]` — `make test:integration:rust` already passes `--features krun,gvproxy --test '*'`, so it's picked up by the existing target with no make changes.

## Out of scope (flagged for follow-up)

The whole `net::gvproxy` submodule has ~28 in-crate `#[test]` items that are likewise invisible to `make test:unit:rust`. Fixing that gap (e.g. adding a `test:unit:rust:gvproxy` make target that runs `--features gvproxy --lib`) is broader than this PR and deserves its own change.

## Test plan

- [x] `make test:unit:rust` → 705/705 boxlite lib + 36/36 boxlite-shared, no regressions; `NetworkBackendConfig` redaction test passes here.
- [x] `cargo nextest run -p boxlite --features krun,gvproxy --test gvproxy_debug_redaction` → 1 passed; the relocated `GvproxyConfig` test compiles + runs under the gvproxy feature.
- [x] `make clippy` → clean.
- [x] Reverse verification: re-deriving `Debug` on `GvproxyConfig` (regression injection) made the integration test fail with the sentinel CA-key bytes echoed in the captured output. Restoring the manual impl returned it to green. Confirms the assertion catches the leak it claims to.

### DIFF
--- a/src/boxlite/src/net/gvproxy/config.rs
+++ b/src/boxlite/src/net/gvproxy/config.rs
@@ -40,7 +40,11 @@ pub struct PortMapping {
 ///
 /// This structure encapsulates all configuration needed to create a gvproxy
 /// virtual network, replacing the previous approach of hardcoding values in Go.
-#[derive(Debug, Clone, Serialize, Deserialize)]
+///
+/// `Debug` is implemented manually (see below) because the derived form would
+/// print the raw `ca_key_pem` — a PKCS8 CA private key — which must never
+/// land in a log file.
+#[derive(Clone, Serialize, Deserialize)]
 pub struct GvproxyConfig {
     /// Unix socket path for the network tap interface.
     /// Caller-provided to ensure each box gets a unique, collision-free path.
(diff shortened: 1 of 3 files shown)


G4614:
Superseded by #575 — merged into a single PR since both address the same threat model (CA key / secret leaks via log + Debug paths). The original commit (`2ea9ae4`) has been cherry-picked onto #575's branch as `ada6b59` with the bodies/test plans integrated. Closing this PR.

Issue #455, opened 2026-05-01
fix(guest): apply youki PR #3504 to fix InitReady/IntermediateReady race under concurrent exec


DorianZheng:
## Summary

Concurrent exec calls intermittently fail with:
```
received unexpected message: InitReady, expected: IntermediateReady(0)
```

## Root Cause

libcontainer (youki v0.5.7) has a race condition in `container_main_process`. The intermediate and init processes are forked as **siblings** via `clone_sibling(CLONE_PARENT)` and both write to the **same** `main_sender` channel:

```
Intermediate: fork → send(IntermediateReady) → exit
Init (sibling): fork → setup → send(InitReady)
                same socket, two writers, no ordering guarantee
```

Under CPU contention (VM, high concurrency), the scheduler can delay the intermediate process, letting `InitReady` arrive before `IntermediateReady`. The main process's `wait_for_intermediate_ready()` panics on the unexpected message.

runc avoids this by using **separate pipes** for intermediate→main and init→main communication. youki shares one channel.

## Upstream

- **youki PR #3504**: https://github.com/youki-dev/youki/pull/3504 (open, not merged)
- Fix: cache `InitReady` when received out of order during `wait_for_intermediate_ready()`, return it later when `wait_for_init_ready()` is called
- Maintainer acknowledged the fix is acceptable short-term; long-term fix is splitting channels (matching runc)
- Neither v0.5.7 nor v0.6.0 include the fix

## Reproduction

```bash
make test:integration  # on EC2 or any VM environment
```

Failing tests:
- `test_zygote_concurrent_env_isolation`
- `test_zygote_high_concurrency_16`
- `test_zygote_repeated_bursts`
- `test_concurrent_exec_workdir`

## Proposed Fix

Patch the vendored libcontainer's `channel.rs` to apply the caching approach from youki PR #3504. This is a minimal, safe change (adds a `bool` field and a loop) until upstream merges and releases.

## Affected Files

- `~/.cargo/registry/src/.../libcontainer-0.5.7/src/process/channel.rs` (or vendored equivalent)

PR #577, opened 2026-05-22
fix(auth): treat empty BOXLITE_API_KEY as unset in `auth status`

G4614:
## Summary

`auth status` short-circuited on `std::env::var(API_KEY_ENV).is_ok()`. That predicate is true for `BOXLITE_API_KEY=""` — easy to produce in CI / shell scripts via `export BOXLITE_API_KEY=$UNSET_VAR`. The runtime in `cli.rs` and the sibling `auth whoami` both correctly skip the env path when the value is empty (`Ok(key) && !key.is_empty()`), so the three views disagreed: `status` reported `Logged in (env)` while `whoami` and every actual authenticated REST call fell back to the stored profile.

Mirror the canonical `!key.is_empty()` check so `auth status` agrees with `whoami` / runtime, and add a regression test that pins the contract.

## Repro (on `main`)

```
$ BOXLITE_API_KEY="" boxlite auth status
Logged in to:    http://localhost:8100
Credential:      API key (from BOXLITE_API_KEY env var)

$ BOXLITE_API_KEY="" boxlite auth whoami
Not logged in.
```

Same env, contradictory output. After the fix both report `Not logged in.`.

## Cross-reference

| file | check |
|------|-------|
| `src/cli/src/cli.rs:217-218` (runtime) | `Ok(key) && !key.is_empty()` |
| `src/cli/src/commands/auth/whoami.rs:68-69` | `Ok(api_key) && !api_key.is_empty()` |
| `src/cli/src/commands/auth/status.rs:46` (before) | `is_ok()` — missing the empty check |

## Test plan

- [x] `cargo nextest run -p boxlite-cli --test auth` → 7/7 pass (6 existing + 1 new regression).
- [x] `make test:unit:rust` → 704/704 boxlite lib + 36/36 boxlite-shared, no regressions (1 leaky is the pre-existing zygote race #455).
- [x] `cargo fmt --all --check` → clean.
- [x] `make clippy` → clean.
- [x] **Reverse verification**: reverting the production fix to bare `is_ok()` makes `status_treats_empty_env_api_key_as_not_logged_in` fail with `Logged in to: http://localhost:8100` echoed in the captured stdout assertion. Restoring the fix returns it to green.
- [x] **Manual repro**: built the CLI on `main` and on this branch and ran `auth status` / `auth whoami` with `BOXLITE_API_KEY=""` — contradictory on `main`, consistent on the branch.


DorianZheng (dismissed):
LGTM

PR #578, opened 2026-05-22
fix(runtime): rebuild invalid guest rootfs overlay

G4614:
## Summary

- Detect an invalid per-box `guest-rootfs.qcow2` during restart/reuse before it reaches libkrun.
- Discard the bad overlay and recreate it from the shared guest rootfs cache.
- Keep missing backing-file failures as explicit storage errors instead of masking deleted cache data.

## Test plan

- [x] `cargo fmt --check -p boxlite`
- [x] `git diff --check`
- [x] `BOXLITE_DEPS_STUB=1 cargo test -p boxlite litebox::init::tasks::guest_rootfs::tests -- --nocapture`

Covered cases:

- valid guest rootfs overlay is reused unchanged
- zero-filled invalid overlay is rebuilt
- too-short overlay is rebuilt
- invalid UTF-8 backing path overlay is rebuilt
- remove failure while discarding invalid overlay returns a clear error
- missing backing file remains an explicit storage error and does not delete the overlay

### DIFF
--- a/src/boxlite/src/disk/qcow2.rs
+++ b/src/boxlite/src/disk/qcow2.rs
@@ -838,31 +838,62 @@ impl Qcow2Helper {
     }
 }
 
-/// Read the backing file path from a qcow2 disk image header.
+/// Why reading a qcow2 header failed.
 ///
-/// Returns `None` if the qcow2 has no backing file (offset or size is 0).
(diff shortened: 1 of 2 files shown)


DorianZheng:
Duplicated in https://github.com/boxlite-ai/boxlite/pull/525

PR #587, opened 2026-05-25
fix(make): propagate test failures instead of masking via POSIX `;` chaining (#550)

### LINKED ISSUE #550: Should test:unit:rust / dist:python / dist:c chain commands with && (or accumulate $status) instead of ;?
## What I noticed

While running the test matrix on a fresh clone, I noticed `make test:unit:rust` reported `exit 0` even though `cargo` had clearly printed `error: test failed` and 29 lib tests under `-p boxlite` had failed. Tracing it back, it looks like a POSIX shell behaviour in a few recipes — but I wasn't sure if it was intentional, so figured I'd ask before sending a PR.

## The pattern

Three recipes in `make/test.mk` and `make/dist.mk` chain multiple commands inside one shell with `;` instead of `&&`:

**`make/test.mk:175-183` — `test:unit:rust`**
```make
@if command -v cargo-nextest >/dev/null 2>&1; then \
    cargo nextest run -p boxlite --no-default-features --lib $(NEXTEST_FILTER); \
    cargo nextest run -p boxlite-shared --lib $(NEXTEST_FILTER); \
else \
    cargo test -p boxlite ... ; \
    cargo test -p boxlite-shared ... ; \
fi
```
POSIX shell evaluates `if A; then B; C; fi` to the rc of `C`. If `-p boxlite` lib tests fail but `-p boxlite-shared` passes, the recipe exits 0 — and CI sees green.

Reproducer: in an environment where some `-p boxlite` tests fail (e.g. AppArmor `unprivileged_userns` blocks bwrap on Ubuntu 24.04 default), a captured log shows `test result: FAILED. 674 passed; 29 failed` and `error: test failed, to rerun pass -p boxlite --lib`, followed by a final line of `exit=0`.

**`make/dist.mk:5-15` — `dist:python` Linux branch**
```make
source .venv/bin/activate; \
bash $(SCRIPT_DIR)/build/build-guest.sh; \
cibuildwheel --platform linux sdks/python; \
```
`build-guest.sh` failure is masked by `cibuildwheel`'s rc, even though `cibuildwheel` shouldn't really proceed without a fresh guest binary.

**`make/dist.mk:22-28` — `dist:c` per-platform branches**
```make
@if [ "$$(uname)" = "Darwin" ]; then \
    cp target/release/libboxlite.dylib sdks/c/dist/lib/; \
    cp target/release/libboxlite.a sdks/c/dist/lib/; \
elif [ "$$(uname)" = "Linux" ]; then \
    cp target/release/libboxlite.so sdks/c/dist/lib/; \
    cp target/release/libboxlite.a sdks/c/dist/lib/; \
fi
```
If the dynamic-lib `cp` fails but the static-lib `cp` succeeds, dist exits 0 with a half-staged output.

## My questions

1. Is the `;` chaining here intentional — e.g. wanting both crates to attempt regardless of the first crate's outcome, with rc-loss accepted as a tradeoff?
2. If not, would you prefer `&&` (fail-fast, simplest), or `$status` accumulation (both crates always run, all failures surfaced, final exit reflects any failure)? My instinct was `$status` for `test:unit:rust` (where seeing both crates' failures in one run is useful) and `&&` for the two `dist:*` cases (where the downstream command can't proceed without the upstream artifact anyway). Does that match your taste?
3. Side observation: `.pre-commit-config.yaml`'s `lint-fix` and `full-test-matrix` hooks only have `make/quality.mk` in their `files:` regexes. So changes to `make/test.mk` / `make/dist.mk` / other makefiles bypass both hooks even after `pre-commit install`. Is that an intentional scope choice, or would adding `make/.*\.mk` to those filters fit your gating strategy?

Happy to open a PR (~13-line diff across `make/test.mk` and `make/dist.mk`) if any of this resonates — wanted to check intent first.

G4614:
## Summary

Three Makefile recipes chained commands with `;` instead of `&&` / status-accumulation. POSIX evaluates `if A; then B; C; fi` to the rc of `C`, so an earlier failure was silently swallowed — CI sees green even when tests fail. **Meta-bug**: this masks every other bug, so it's the highest-leverage fix to land first.

## What was broken

- **`test:unit:rust`** — a `-p boxlite` lib-test failure was hidden by a passing `-p boxlite-shared` run. Captured log in #550 shows `674 passed; 29 failed` followed by recipe `exit=0`.
- **`dist:python` Linux branch** — `build-guest.sh` failure masked by `cibuildwheel`'s rc; wheel built against stale guest.
- **`dist:c` per-platform branch** — failing dynamic-lib `cp` left the static-lib `cp` to \"succeed\", producing a half-staged `sdks/c/dist/` with exit 0.

## Fix

- `test:unit:rust`: status accumulation (`|| rc=\$?`) so both crates always run and final rc reflects any failure. Added `--no-tests=warn` to keep `make test:unit:rust FILTER=<test-only-in-one-crate>` working (nextest's default `--no-tests=fail` would otherwise exit 4 from the crate with zero matches).
- `dist:python` / `dist:c`: switched intra-branch chains to `&&` so downstream commands can't run against a failed upstream.
- `.pre-commit-config.yaml`: widened `lint-fix` hook's files filter from `make/quality\\.mk` to `make/.*\\.mk` so future `make/test.mk` / `make/dist.mk` edits go through the same autofix gate (the original narrow filter is exactly what let this class of issue land without local validation).

## Test plan

- [x] Happy path: `make test:unit:rust FILTER=container_layout` (matches 1 in boxlite-shared, 0 in boxlite) → exit 0 — no false-fail from the empty crate.
- [x] Failure path: `make test:unit:rust FILTER=can_create_process_in_new_user_ns` on an AppArmor-restricted host (the #544 EACCES panic in `-p boxlite`) → recipe exit 100, make exit 2 — the failure correctly propagates instead of being swallowed by `-p boxlite-shared` finishing clean.
- [x] Shell logic sanity-checked with `bash -c 'rc=0; false || rc=\$?; true || rc=\$?; exit \$rc'` → exit 1.

Closes #550.

### DIFF
--- a/make/dist.mk
+++ b/make/dist.mk
@@ -3,11 +3,11 @@ dist\:python: _ensure-python-deps
 	@. .venv/bin/activate && pip install -q cibuildwheel
 
 	@if [ "$$(uname)" = "Darwin" ]; then \
-		source .venv/bin/activate; \
+		source .venv/bin/activate && \
 		cibuildwheel --only cp314-macosx_arm64 sdks/python; \
 	elif [ "$$(uname)" = "Linux" ]; then \
-		source .venv/bin/activate; \
-		bash $(SCRIPT_DIR)/build/build-guest.sh; \
+		source .venv/bin/activate && \
+		bash $(SCRIPT_DIR)/build/build-guest.sh && \
 		cibuildwheel --platform linux sdks/python; \
 	else \
 		echo "❌ Unsupported platform: $$(uname)"; \
@@ -20,10 +20,10 @@ dist\:c:
 	@mkdir -p sdks/c/dist/lib sdks/c/dist/include
 	@cp sdks/c/include/boxlite.h sdks/c/dist/include/
 	@if [ "$$(uname)" = "Darwin" ]; then \
-		cp target/release/libboxlite.dylib sdks/c/dist/lib/; \
+		cp target/release/libboxlite.dylib sdks/c/dist/lib/ && \
 		cp target/release/libboxlite.a sdks/c/dist/lib/; \
 	elif [ "$$(uname)" = "Linux" ]; then \
-		cp target/release/libboxlite.so sdks/c/dist/lib/; \
+		cp target/release/libboxlite.so sdks/c/dist/lib/ && \
 		cp target/release/libboxlite.a sdks/c/dist/lib/; \
 	fi
 	@echo "✅ C SDK staged in sdks/c/dist/"
--- a/make/test.mk
+++ b/make/test.mk
@@ -175,15 +175,22 @@ test\:integration\:sdk:
 
 # Rust unit tests (parallel via nextest, fallback to serial cargo test).
 # --no-default-features disables gvproxy to avoid Go runtime link issues.
+#
+# Status accumulation: both crates always run (so a `-p boxlite-shared`
+# regression isn't masked by a `-p boxlite` failure aborting first), but
+# the recipe exits non-zero if EITHER crate failed. POSIX shell otherwise
+# evaluates `cmd_a; cmd_b` as the rc of cmd_b, silently swallowing cmd_a.
 test\:unit\:rust:
 	@echo "🧪 Running Rust unit tests..."
-	@if command -v cargo-nextest >/dev/null 2>&1; then \
-		cargo nextest run -p boxlite --no-default-features --lib $(NEXTEST_FILTER); \
-		cargo nextest run -p boxlite-shared --lib $(NEXTEST_FILTER); \
+	@rc=0; \
+	if command -v cargo-nextest >/dev/null 2>&1; then \
+		cargo nextest run --no-tests=warn -p boxlite --no-default-features --lib $(NEXTEST_FILTER) || rc=$$?; \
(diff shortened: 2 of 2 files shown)


G4614:
### Why this slipped — and why it matters more now

Worth calling out the practical detection asymmetry, since it explains both how this lived so long and why the fix matters more in 2026 than it would have in 2020:

- A human running `make test:unit:rust` in a terminal reads the test runner's text output and notices `test result: FAILED. 674 passed; 29 failed` even when the recipe exits 0 — they form their pass/fail judgment from the screen, not from the rc. So the masked rc was effectively invisible to the human-driven workflow that originally wrote and ran these recipes.
- Every non-human consumer — `make`'s `if` evaluation, GitHub Actions, any agent-driven test loop, any wrapping script — has no eyes. They consume the rc and only the rc. The "29 failed" line is just stdout to them.

As more test runs get kicked off by CI and by AI/agent-driven dev loops, the false-green signal lands on exactly the consumer that can't see the "failed" line, which is when it does the most damage. The fix's biggest payoff isn't for the developer who runs `make test:unit:rust` once and reads the output — it's for everything downstream of that recipe in any pipeline.

PR #589, opened 2026-05-25
fix(test:apps): inject boxlite_dev tag via GOFLAGS for apps Go tests (#579)

### LINKED ISSUE #579: make test:apps fails in dev workflow: libboxlite.a not populated at default-bridge path
## Summary

`make test:apps` fails on a fresh dev checkout because every Go-based Nx project under `apps/` (`runner`, `cli`, `proxy`, `ssh-gateway`, `otel-collector`, `boxlite`-containing, etc.) link-fails with:

```
/usr/bin/x86_64-linux-gnu-ld.bfd: cannot find /home/<user>/ws/boxlite/sdks/go/libboxlite.a:
No such file or directory
collect2: error: ld returned 1 exit status
FAIL ./... [setup failed]
```

`make test:unit:go` is unaffected and passes (it uses `-tags boxlite_dev`, which routes to a different libboxlite.a path).

## Reproduction

```bash
# 0. start from clean repo, no prebuilt downloaded, no symlinks
rm -f sdks/go/libboxlite.a sdks/go/include/boxlite.h

# 1. unit:go is fine
make test:unit:go    # builds with -tags boxlite_dev, passes

# 2. apps fails for every Go project
make test:apps       # → cannot find sdks/go/libboxlite.a
```

Verified on `main` @ `93a1f40` (Linux). CI doesn't catch this because `.github/workflows/build-go.yml` line 54 and `.github/workflows/build-runner-binary.yml` line 68 each do an explicit `cp "/tmp/boxlite-c-v${VERSION}-${PLATFORM}/lib/libboxlite.a" sdks/go/` before running tests — populating the prebuilt path that doesn't exist in dev workflow.

## Root cause

The Go SDK has two cgo bridge files selected by Cargo feature tag:

| File | Tag | Lib path | Header path |
|---|---|---|---|
| `bridge_cgo_dev.go` | `boxlite_dev` | `${SRCDIR}/../../target/debug/libboxlite.a` | `${SRCDIR}/../c/include` |
| `bridge_cgo_prebuilt.go` | (default, no tag) | `${SRCDIR}/libboxlite.a` | `${SRCDIR}/include` |

`make dev:go` builds with `-tags boxlite_dev` so its post-build `go build` resolves through the dev bridge. But it never populates the prebuilt-mode paths (`sdks/go/libboxlite.a`, `sdks/go/include/boxlite.h`) that external SDK consumers normally get via `go run github.com/boxlite-ai/boxlite/sdks/go/cmd/setup`.

The Nx Go projects under `apps/` invoke `@nx-go/nx-go:test` → `go test ./...` **without** passing `-tags boxlite_dev`, so they fall into the prebuilt bridge and look for the file at the prebuilt path. In a CI environment the workflows pre-populate that path with the downloaded release artifact; in a dev workspace nothing populates it.

So: **the in-repo dev workflow has no path to produce a libboxlite.a layout that matches what default-tag consumers (i.e. anything not using `-tags boxlite_dev`) expect**. CI works around it with explicit `cp` steps; local devs hit the wall.

## Other failures observed on `make test:apps` (separate issues, flagged here for context)

While verifying, two additional categories of failure were observed on fresh `main` that are independent of the libboxlite.a issue and would benefit from their own issues / PRs:

- **`apps/go.work` is missing entries.** Currently only `./runner`, `./daemon`, `./common-go`, `./api-client-go`, `./libs/computer-use`, `../sdks/go` are listed. So `make test:apps` on `snapshot-manager`, `cli`, `proxy`, `ssh-gateway`, `otel-collector` all fail with:
  ```
  pattern ./...: directory prefix . does not contain modules listed in go.work
  FAIL ./... [setup failed]
  ```
- **`apps/api` Jest preset path resolution** — `Preset ../../jest.preset.js not found relative to rootDir /home/.../apps/api`.

I haven't fixed these; flagging so the apps team knows.

## Mitigation options for the libboxlite.a issue

Three reasonable approaches, ordered by blast radius:

1. **Minimal**: extend `make dev:go` to symlink dev artifacts into the prebuilt-mode paths so default-tag consumers resolve against fresh dev output. ~12 lines (one edit to `make/dev.mk`, add `dev:go` as prereq to `test:apps` in `make/test.mk`, ignore `sdks/go/include/` in `.gitignore`). No code/SDK changes. Doesn't affect CI.

2. **Medium**: collapse the dual workflow — make the CI `cp libboxlite.a sdks/go/` step a call to `make dev:go` (or a new `make sync-go-bridge` target), so CI and local dev funnel through one mechanism. Touches the two workflow YAMLs + a new make target.

3. **Architectural**: drop the dual-bridge pattern entirely. One bridge file picks the lib path dynamically via env / build script. Touches `sdks/go/bridge_cgo_*.go`, requires deciding how external consumers discover the path. Bigger rework.

A working implementation of option 1 sits on a local branch and was test-verified — `runner:test` goes from `[setup failed]` (no file) to successfully linking and reaching actual test code (which then surfaces 2 pre-existing test-code bugs in the runner project, also unrelated). Happy to open a PR with that if option 1 is the preferred direction.

## Environment

- `main` @ `93a1f40`
- Linux x86_64 (Ubuntu 24.04 cloud image, kernel 7.0.0)
- Go 1.25.4, Rust stable, observed without `cmd/setup` having ever run in the workspace

G4614:
Set `GOFLAGS=-tags=boxlite_dev` on the `make test:apps` recipe so the apps' Go tests link the dev-built `target/debug/libboxlite.a` (via `sdks/go/bridge_cgo_dev.go`) instead of the prebuilt bundle absent on dev checkouts — fixes #579.

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -296,9 +296,17 @@ test\:all\:go:
 
 # apps/ workspace test matrix (all Nx projects). FILTER maps to Jest's
 # --testNamePattern (per-runner semantics, like the other suites).
-test\:apps: _ensure-apps-deps
+#
+# Depends on dev:go so target/debug/libboxlite.a is built before any Nx Go
+# project links against it. GOFLAGS=-tags=boxlite_dev (inherited by every
+# `go test` Nx spawns) routes the cgo build through sdks/go/bridge_cgo_dev.go,
+# which links target/debug/libboxlite.a + the in-repo sdks/c/include headers.
+# Without the tag, `go test` uses bridge_cgo_prebuilt.go, which needs the
+# downloaded sdks/go/{libboxlite.a,include} bundle (absent on dev checkouts)
+# and fails #579. Same pattern as test:unit:go above.
+test\:apps: _ensure-apps-deps dev\:go
 	@echo "🧪 Running apps workspace test matrix..."
-	@cd apps && yarn nx run-many --target=test --all --parallel=$$(getconf _NPROCESSORS_ONLN) $(if $(FILTER),-- --testNamePattern '$(FILTER)',)
+	@cd apps && GOFLAGS=-tags=boxlite_dev yarn nx run-many --target=test --all --parallel=$$(getconf _NPROCESSORS_ONLN) $(if $(FILTER),-- --testNamePattern '$(FILTER)',)
 
 # Installer-script smoke test: structural assertions on the rendered
 # install.sh (atomic replace, integrity envelope, pinned-install trust


G4614:
**Caveat — the tag is injected only via `make test:apps`, not bare `yarn nx test <project>`**

Line 309 sets `GOFLAGS=-tags=boxlite_dev` for the `nx run-many` it spawns, so it covers `make test:apps` (and CI, which runs through make). A standalone `yarn nx test <project>` in a shell without `GOFLAGS` does not get the tag. Consequences:

- The 8 pure-Go apps — the tag is a no-op, so they're unaffected.
- `runner` (the only app importing the cgo bridge), run standalone without the tag:
  - Normal dev checkout (no downloaded prebuilt bundle): loud link failure — the #579 symptom. Safe; fix by running via `make`.
  - If a contributor has *also* run the setup downloader and pulled the prebuilt bundle into their checkout: it silently links the released `sdks/go/libboxlite.a` instead of their local `target/debug` build — tests go green but don't exercise local Rust changes.

Neither case affects CI. If standalone `nx test` ever becomes a common workflow, close the gap by switching the Go test targets to `nx:run-commands` with inline `GOFLAGS` (bakes the tag into the target config so it applies however the target is invoked) — not direnv/`options.env` (the latter is dropped by `@nx-go/nx-go:test`, which forwards only `cwd` to the spawned `go`).

---

**Test plan**

- Tag selects the dev bridge (`go list -f '{{.CgoFiles}}'` in `sdks/go`): `GOFLAGS=-tags=boxlite_dev` → `bridge_cgo_dev.go` (in-repo `sdks/c/include` + `target/debug/libboxlite.a`); no tag → `bridge_cgo_prebuilt.go` (needs the absent downloaded bundle, i.e. #579).
- Env reaches the Nx-spawned `go test` (unchanged `@nx-go/nx-go:test` executor): `GOFLAGS=-zzz_bogus yarn nx test api-client-go` → `go: parsing $GOFLAGS: unknown flag -zzz_bogus`; `GOFLAGS=-tags=boxlite_dev yarn nx test api-client-go` → `Successfully ran target test`.

PR #597, opened 2026-05-26
feat(box): --cap option, default all open

G4614:
`--cap NAME=0|1` default all 1

## Test plan

| observed | pre-feat | post-feat |
|---|---|---|
| default container CapEff | `0xa80425fb` (docker 14) | `0x1ffffffffff` (≥38 of 41 OCI caps, incl. SYS_ADMIN) |
| `--cap SYS_ADMIN=0` | silently no-op (per-cap arm added `CAP_` prefix that `Capability::from_str` rejected) | bit 21 cleared in CapEff |
| `--cap ALL=0` | n/a (only `--cap-add`) | every cap bit cleared |
| `--cap ALL=0 --cap SYS_ADMIN=1` | n/a | exactly bit 21 set (later overrides earlier) |
| exec inherits container caps | non-TTY = hardcoded docker 14; TTY = empty default | both replay the container's `cap_overrides` |
| unknown cap name (e.g. `SYS-ADMIN=0`) | guest `warn-and-skip`, box still starts | CLI parse error, box never starts |
| `mount tmpfs` w/ default caps | (worked) | (still works) |
| `mount tmpfs` w/ `--cap SYS_ADMIN=0` | n/a | EPERM (`permission denied (are you root?)`) |
| REST `POST /v1/boxes` carrying `cap_overrides` | field absent on both sides; `deny_unknown_fields` 4xx's the body | field threads client → wire → server → `BoxOptions.cap_overrides` |

28 tests across CLI parse, runtime, guest unit, REST wire, and alpine e2e — two-side verified where the contract is non-obvious (TTY exec inheritance, mount enforcement, cap_overrides round-trip).

DinD/privileged plumbing (`/proc/sys` writable, cgroup mounts) is intentionally out of scope — stacked draft PR.


G4614:
Reverted — branch force-pushed to origin/main (`ahead_by=0 behind_by=3`), all 28 tests + 1293 lines rolled back. Reason: change of approach. The default-all-open + drop-via-`--cap NAME=0` model in this PR was the wrong direction; returning to default-closed + opt-in `--cap-add` (docker-style, which is what main already has). New PR will land that incrementally if needed.

G4614:
Superseded by #1047 (structured capability controls with Docker-compatible defaults).

PR #600, opened 2026-05-27
fix(runner): roll back SignaledHUP/TERM when SDK signal delivery fails

G4614:
while SIGHUP/SIGTERM is not succeed in delivery because the communication between runner and shim is interrupted (also maybe caused by Go SDK/CGo fail, test-based), the info requires fallback

### DIFF
--- a/apps/runner/pkg/boxlite/exec_manager.go
+++ b/apps/runner/pkg/boxlite/exec_manager.go
@@ -704,7 +704,7 @@ func (m *ExecManager) escalate(e *ManagedExec, sig syscall.Signal, name string) 
 	e.handleMu.Lock()
 	if e.closed || e.execution == nil {
 		e.handleMu.Unlock()
-		e.escalationFailedMarkDoomed()
+		e.escalationFailedMarkDoomed(sig)
 		m.killAndEvict(e)
 		return
 	}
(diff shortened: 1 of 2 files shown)


DorianZheng:
Could you add a regression test for this? The pre-fix code is already exercised by TestExecManagerSignalUnsupportedFallsThroughToKill, but the generic-error branch (exec_manager.go:723) and the SIGTERM rollback aren't covered yet.

G4614:
## Test plan

`go test -tags boxlite_dev ./pkg/boxlite/` — each verified FAIL pre-fix / PASS on the fix:

| test | arm | signal | |
|---|---|---|---|
| `TestExecManagerSignalUnsupportedFallsThroughToKill` | `ErrSignalUnsupported` (718) | SIGHUP | pre-existing |
| `TestExecManagerGenericSignalErrorRollsBackSignaledHUP` | `default` (723) | SIGHUP | added |
| `TestExecManagerSigtermFailureRollsBackSignaledTERM` | `default` (723) | SIGTERM | added |

718×SIGTERM is omitted: rollback depends only on `sig`, not the arm, so it's a redundant and near-unreachable cell.

G4614:
> Could you add a regression test for this? The pre-fix code is already exercised by TestExecManagerSignalUnsupportedFallsThroughToKill, but the generic-error branch (exec_manager.go:723) and the SIGTERM rollback aren't covered yet.

fixed in the latest commit, thx

PR #614, opened 2026-05-28
feat(net): auto-remap EXPOSE host port on conflict

G4614:
Auto-remap an image `EXPOSE` port to an OS-allocated ephemeral host port when its desired host port is busy (instead of failing fast),
explicit user `-p` mappings are never remapped.

## Test plan

Unit (`resolve_expose_host_port`): `_uses_desired_when_free` (free → 1:1 `AutoExpose`) and `_falls_back_when_busy` (busy → ephemeral `AutoRemap`).

**Integration `expose_auto_remap`** — pre-bind `0.0.0.0:6379`, then `boxlite run -d redis:alpine` (image EXPOSEs 6379). Verified two-sided (`resolve_expose_host_port` stubbed to never remap vs applied):

| observed (the EXPOSE 6379 mapping) | pre-fix (never remap) | post-fix |
|---|---|---|
| host_port | `6379` (1:1, collides with the pre-bound port) | OS-allocated ephemeral (`≠ 6379`) |
| inspect `Source` | `auto_expose` | `auto_remap` |
| test result | **FAIL** — `assert host_port != 6379` at `expose_auto_remap.rs:155` | **PASS** |

Guest keeps listening on 6379; only the host side moves.

**Integration `user_published_port_keeps_user_source`** (scope guard) — `boxlite run -d -p <free>:<guest> alpine` (no EXPOSE, so the explicit `-p` is the only mapping). Verified two-sided (user `-p` routed through `resolve_expose_host_port` vs not):

| observed (the user `-p` mapping) | broken (user `-p` routed through remap) | correct |
|---|---|---|
| inspect `Source` | `auto_expose` | `user` |
| host_port | chosen (free, so unchanged) | chosen |
| test result | **FAIL** — `assert Source == "user"` at `expose_auto_remap.rs:273` | **PASS** |

### DIFF
--- a/src/boxlite/src/litebox/box_impl.rs
+++ b/src/boxlite/src/litebox/box_impl.rs
@@ -62,6 +62,12 @@ pub(crate) struct LiveState {
     #[allow(dead_code)]
     guest_rootfs_disk: Option<Disk>,
 
+    /// Post-conflict-resolution port mappings handed up from
+    /// `vmm_spawn`/`init`. Consumed once at the `Running` transition
+    /// where they're copied onto `BoxState` so `inspect`/`list` can
+    /// surface them; not used for runtime traffic decisions.
+    port_mappings: Vec<crate::runtime::types::ResolvedPortMapping>,
+
     // Platform-specific
     #[cfg(target_os = "linux")]
     #[allow(dead_code)]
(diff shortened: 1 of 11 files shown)


G4614:

PR #616, opened 2026-05-28
test(c): isolate test_simple_api boxlite home from shared ~/.boxlite

G4614:
set `BOXLITE_HOME` in `test_simple_api`  like `test_null_callback` (dev only)

### DIFF
--- a/sdks/c/tests/CMakeLists.txt
+++ b/sdks/c/tests/CMakeLists.txt
@@ -62,10 +62,21 @@ endforeach()
 # Enable testing
 enable_testing()
 
-# Register tests
+# Register tests. test_simple_api is special-cased below (needs an isolated home).
 foreach(TARGET ${TEST_TARGETS})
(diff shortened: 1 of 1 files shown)


G4614:
Scope note — why this is test-only, and the one user-facing exception:

`Schema version mismatch` (`db/mod.rs:117`) only fires when the on-disk DB is *newer* than the running process; migrations are forward-only, so an upgrade auto-migrates and a normal single-install user never hits it. The common trigger is a dev/CI condition — many branch builds sharing one persistent `~/.boxlite`, where a schema-bumping PR poisons the home for every older-base build after it. That's exactly what this isolates.

The one user-facing path is a **downgrade**: running an older boxlite after a newer one has written `~/.boxlite`. It's uncommon, and the error is already actionable ("Upgrade boxlite to a newer version"), so there's nothing to fix in production here — isolating the test's home is the right and complete scope.

PR #618, opened 2026-05-28
feat-stress(disk_guard): structural reserve via fallocate, replacing the policy walls

G4614:
use fallocate to reserve some space to ensure the box alive and the ability to remove something

### The one test that decides whether #618 works

**`release_unblocks_writes_on_a_full_host`** (GATED integration, `src/cli/tests/reserve_release.rs`)

Three properties have to all be true for the structural reserve to mean anything:

1. The 64 MiB is **physically allocated**, not sparse — other writers can't take it.
2. When the host fs is **truly at f_bavail = 0**, the reserve still owns its blocks.
3. **`unlink(2)` at zero free still works** (metadata-only), and the freed bytes immediately become writable again.

This test exercises all three end-to-end on a real bounded mount:

```
1. bootstrap reserve on $BOXLITE_RESERVE_TEST_HOME
2. progressive-chunk fill (4 MiB → 64 KiB → 4 KiB → 1 byte)
   until even a 1-byte write returns ENOSPC                 ← proves prop 1+2: reserve owns its blocks
3. probe-write 1 byte → expect ENOSPC                       ← confirms host is genuinely full
4. boxlite reserve-release                                  ← proves prop 3: unlink works at 0 free
5. same probe-write → expect success                        ← proves bytes really return to fs
```

Gated behind `BOXLITE_RESERVE_TEST_HOME=/path/to/bounded-mount`. CI can't reach this without a privileged tmpfs setup; the GATED env-var pattern lets reviewers run it on a dev box with two lines:

```sh
sudo mount -t tmpfs -o size=256M tmpfs /tmp/m
BOXLITE_RESERVE_TEST_HOME=/tmp/m \
  make test:integration:cli FILTER=release_unblocks_writes_on_a_full_host
```

Two-side verified in the commit message (efb556ab): the original single-pass version of this test mistakenly passed because the 4 MiB chunk fill left a few-MiB gap, and the 1-byte probe slipped through. The multi-pass cascade was the fix and is documented inline.

### Supporting coverage (21 tests, brief)

The end-to-end above is the load-bearing test. The remaining 21 tests guard properties that have to hold for the e2e to even be reachable, plus orthogonal correctness concerns:

| concern | tests | proves |
|---|---|---|
| reserve is physically allocated, not sparse | `ensure_reserve_allocates_real_blocks` | statvfs `f_bavail` drops by ~64 MiB |
| release returns bytes to fs free pool | `release_returns_bytes_to_host_available` | statvfs `f_bavail` rises by ~64 MiB after unlink |
| bootstrap / idempotency / self-heal | 4 unit + 4 integration | every `boxlite` cmd auto-creates the reserve; double-call no-ops; external `rm`/partial file/4-thread race all top up to exactly 64 MiB |
| **recovery race fix** (my 865ce86f) | 3 unit | hysteresis: `host_free < 2 × RESERVE_BYTES` defers top-up so `boxlite reserve-release` actually leaves headroom for the next `rm` |
| **box-cycle invariance** (my 0a06569e) | 2 integration | release with a running box leaves the box `Running`; 5 release/restore cycles likewise |
| ENOSPC hint plumbing | 4 unit (CLI matcher) | kernel ENOSPC errno 28 / "No space left" substring across anyhow chain → operator hint; ENOENT / 401 don't false-positive |

All seven tests marked under "race fix" / "self-heal" / "box-cycle" are two-side verified — for each, a specific regression was injected and the test was observed to fail on the assertion that names the property. Details in the individual commit messages (865ce86f, 3078b3db, 0a06569e).

### How to run

```sh
make test:unit:rust FILTER=reserve::                                          # 11 unit tests, no VM
make test:integration:cli FILTER='reserve_release|box_survives|looks_like_enospc'  # 11 integration / CLI-binary tests, needs VM + alpine pull
# Plus the GATED e2e above
```

### Known gaps (deliberate, with rationale)

- **Real `f_bavail = 0` for every CLI command** — only the load-bearing e2e reaches this state. Filling shared `/tmp` to verify every other command would destabilize CI; the four ENOSPC matcher tests cover the surface-area indirection that turns a kernel ENOSPC into the operator hint.
- **`fallocate` itself returning ENOSPC at attempt-time** — the hysteresis guard prevents reaching `fallocate` whenever `host_free < 2 × RESERVE_BYTES`; the only entry into that error path is statvfs misreporting, which is production-unreachable.
- **Kernel atomicity of `fallocate` mid-crash** — Linux contract, not boxlite's responsibility. Whatever `meta.len()` the kernel leaves behind, the partial-reserve self-heal test covers recovery on next runtime build.
- **Cross-FS semantics (XFS / btrfs / ZFS)** — CI runs on ext4; XFS / btrfs share the POSIX `fallocate(mode=0)` contract; ZFS is not a supported deployment target (documented in the module header).

### DIFF
--- a/src/boxlite/src/runtime/rt_impl.rs
+++ b/src/boxlite/src/runtime/rt_impl.rs
@@ -193,6 +193,13 @@ impl RuntimeImpl {
             ))
         })?;
 
+        // Lay down (or top up) the recovery reserve. From this moment on
+        // the kernel's free-space accounting hides ~64 MiB from every
+        // writer on the host, including boxlite itself. Replaces the
+        // old per-command `enforce_recovery_budget` policy check: the
+        // floor is now structural, not runtime-policy.
+        crate::util::ensure_reserve(layout.home_dir())?;
+
         let runtime_lock = RuntimeLock::acquire(layout.home_dir()).map_err(|e| {
             BoxliteError::Internal(format!(
                 "Failed to acquire runtime lock at {}: {}",
--- a/src/boxlite/src/util/mod.rs
+++ b/src/boxlite/src/util/mod.rs
@@ -1,9 +1,11 @@
 mod binary_finder;
 mod pid_file;
 pub mod process;
+pub mod reserve;
 
 pub use binary_finder::{RuntimeBinaryFinder, find_binary};
 pub use pid_file::{PidFileReader, PidFileWriter, PidRecord, ProcessIdentity};
+pub use reserve::{RESERVE_BYTES, ReserveReleased, ensure_reserve, release_reserve, reserve_path};
 
 use std::path::PathBuf;
 use std::process::Command;
--- a/src/boxlite/src/util/reserve.rs
+++ b/src/boxlite/src/util/reserve.rs
@@ -376,6 +376,32 @@ mod tests {
         );
     }
 
+    /// The point of the whole reserve is "after release, the host fs really
+    /// gives those bytes back to writers." The earlier
+    /// `allocates_real_blocks` pins the *consumption* direction; this pins
+    /// the *return* direction. Without this, a regression that switched
+    /// `release` to `truncate(0)` without `unlink` would still pass the
+    /// "file is gone" assertion but the blocks could stay reserved until
+    /// the next fsync — and the operator would still hit ENOSPC.
+    #[test]
+    fn release_returns_bytes_to_host_available() {
+        let home = TempDir::new().unwrap();
+        ensure_reserve(home.path()).expect("ensure_reserve");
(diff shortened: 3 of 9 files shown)


G4614:
Closing: this PR's premise no longer holds.

It set out to replace "the policy walls" with a structural `fallocate` reserve, but that mechanism no longer exists on `main` — `git grep -i 'policy wall'` returns nothing. There is nothing left to replace, and the branch is 1195 lines behind on a subsystem that has since changed shape.

If per-box disk reservation is still wanted, it is worth re-deriving against current `main` rather than rebasing this.

PR #619, opened 2026-05-28
fix(jailer): warn when cgroup join fails instead of silently ignoring

G4614:
The cgroup join ran inside `pre_exec` (after fork, before exec) where tracing is unavailable. Errors were discarded with `let _ =`, silently leaving the box running outside its resource limits.

## Fix

Move the join to the parent process via `Jail::post_spawn(pid)`, called immediately after `cmd.spawn()`. The parent has full tracing, so a failed join now emits a warn and the box continues.

```
Before
ShimSpawner::spawn
└─ cmd.spawn()
   └─ pre_exec hook
      └─ let _ = add_self_to_cgroup_raw()   ← join error silently dropped

After
ShimSpawner::spawn
└─ cmd.spawn()
└─ jail.post_spawn(child.id())
   └─ join_cgroup(box_id, pid) → Err → tracing::warn!(...)   ← visible, non-fatal
```

### DIFF
--- a/src/boxlite/src/jailer/builder.rs
+++ b/src/boxlite/src/jailer/builder.rs
@@ -29,6 +29,8 @@ pub struct JailerBuilder {
     layout: Option<BoxFilesystemLayout>,
     preserved_fds: Vec<(RawFd, i32)>,
     detach: bool,
+    /// VM guest memory in MiB, used to derive the host cgroup memory limit.
+    vm_memory_mib: Option<u32>,
 }
 
 impl Default for JailerBuilder {
@@ -47,6 +49,7 @@ impl JailerBuilder {
             layout: None,
             preserved_fds: Vec::new(),
             detach: false,
+            vm_memory_mib: None,
         }
     }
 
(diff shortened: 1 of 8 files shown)


G4614:
**Split into three PRs; this one is now the third.**

What used to be here (1412 lines, 8 files) has been separated so the two independent halves can land without waiting on the rootless work:

- #842 — `enable only the cgroup controllers actually delegated` (+116/−9, 1 file). The atomic `+cpu +memory +pids` write failing as a unit, plus `enable_controllers` running on every `setup_cgroup` rather than only at creation, so a parent left un-delegated by an earlier failure is repaired instead of leaving every later box on that runner unlimited.
- #1441 — `move host cgroup setup into the Jailer and default its limits` (+204/−27, 6 files). The structural move plus `apply_cgroup_defaults`.
- this PR — the rootless systemd-transient-scope enforcement path, the rootless CPU default, and the two stress test files.

Rebased onto current `main`, and the two extracted commits are gone from the series. The remaining eight commits are +1075/−23 across 5 files; the diff against `main` still shows 8 files / +1374 because `place_shim_in_scope` hangs off the `Jailer` #1441 introduces and GitHub cannot use a fork branch as a base. It collapses once #842 and #1441 land.

Filed POL-469 for this, which the PR had been missing since May.

The three `enable_controllers_*` unit tests moved to #842 with the fix they cover — a fix and its test should not land separately.

G4614:
Merged into #1419 — both POL-469 and POL-348 fixes are in one PR since the change is only 14 lines total.

G4614:
Folded into #1419.

PR #621, opened 2026-05-28
test(ffi): de-flake pump_yields_when_queue_full via threshold recalibration

G4614:
De-flake `pump_yields_when_queue_full` by lowering its `progress >= 10` assertion — which sat right on the boundary of the ~8–10 canary ticks the bulk drain produces — to `>= 5`, a threshold recalibration with no production change (`yield_now` doesn't actually starve the canary; a worker-monopolizing producer collapses it to ~2, still caught).

## Test plan

- `make test:unit:rust FILTER=pump_yields` → 15/15 runs pass, non-flaky.
- Two-side verified: inject `std::hint::spin_loop()` in place of the yield (a worker-monopolizing producer) → canary collapses → assertion fails with the named error; restore → passes.

| observed | pre (`>= 10`) | post (`>= 5`) |
|---|---|---|
| cooperative `yield_now` (current prod code) | ~8–10 ticks → flakes on the 9/10 edge (~40% first-try fail) | ~8–10 ticks → passes with margin |
| worker-monopolizing producer (the real regression) | caught (~2 < 10) | caught (~2 < 5) |
| canary max_gap | ~21 ms (no starvation) | ~21 ms (unchanged) |

### DIFF
--- a/sdks/c/src/event_queue.rs
+++ b/sdks/c/src/event_queue.rs
@@ -752,14 +752,17 @@ mod phase2_regression_tests {
         let canary_after = canary.load(Ordering::Relaxed);
         canary_stop.store(1, Ordering::Relaxed);
 
-        // The drain takes >=NUM_EVENTS*10ms = 1000ms. The canary should
-        // have made many ticks during that time IF the producer yielded.
-        // If push_event_with_capacity busy-spinned, the single worker
-        // would have been monopolized and the canary would barely advance.
+        // `boxlite_runtime_drain(rt, 0, …)` empties the queue in bulk, so the
+        // drainer finishes in ~150-250ms — not the ~1000ms a naive
+        // "100 events × 10ms" estimate suggests. In that window a cooperative
+        // producer lets the canary tick ~8-10 times; a producer that busy-spins
+        // or blocks the single worker instead collapses it to ~2 (measured).
+        // `>= 5` separates the two with margin on both sides, without sitting on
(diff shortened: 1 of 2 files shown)


G4614:
The recalibration to `>= 5` cuts the flake rate but doesn't fully close it on a CPU-contended host. Running the suite on an AWS box — where nextest runs the `boxlite-c` crate's tests in parallel, contending for the single tokio worker the canary shares with the producer — `pump_yields_when_queue_full` failed **both** nextest tries, with the canary dipping to **4** ticks:

```
TRY 1 FAIL: canary advanced only 6 ticks; producer likely busy-spinning instead of yielding
TRY 2 FAIL: canary advanced only 4 ticks; producer likely busy-spinning instead of yielding
```

With `>= 5`, the 4-tick run still trips the assert (the 6-tick one would have passed). Suggest lowering to **`>= 3`**: it stays clear of the ~2 busy-spin floor this PR measured, while leaving margin below the cooperative range observed here (4-6; ~8-10 on a quiet host). The discriminator is preserved (cooperative → canary advances; busy-spin → stalls at ~2) — `>= 3` just sits in the gap instead of on the noisy 4-6 boundary.

(Longer term the absolute-tick assertion is inherently timing-fragile under scheduler contention; a relative baseline — e.g. assert cooperative ticks ≫ an in-run busy-spin sample — would be load-independent. But `>= 3` is the minimal change to stop the flake.)

PR #622, opened 2026-05-29
fix(cli): stop box before hard-exit on non-zero run exit code

G4614:
when the box killed badly (non-zero return code), the shim process should be released by RAII 

appeared because of #604 

### Existing exit-code tests (branch ⑥: non-zero exit)

The abnormal-exit `boxlite run` integration tests, each verified two-sided (`run.rs` reverted to `std::process::exit` vs. RAII applied) on this branch. Side B's leak surfaces via `PerTestBoxHome::Drop` panicking with `live shim(s): [pid]` — the `std::process::exit` shortcut bypasses `RuntimeImpl::Drop` → `shutdown_sync`.

| test | side A (RAII) | side B (pre-fix `std::process::exit`) |
|---|---|---|
| `test_run_exit_code_125` | ok (8.14s) | **FAILED** — `live shim(s): [320092]` |
| `test_run_exit_code_custom` | ok (8.14s) | **FAILED** — `live shim(s)` |
| `test_run_signal_exit_code_sigterm` | ok (16.42s) | **FAILED** — `live shim(s): [323543]` |
| `test_run_signal_exit_code_sigkill` | ok (16.42s) | **FAILED** — `live shim(s)` |
| `test_run_signal_exit_code_sigint` | ok (16.42s) | **FAILED** — `live shim(s)` |
| `test_run_python_error_handling` | ok (8.06s) | **FAILED** — `live shim(s): [326872]` |
| `test_run_exit_code_success` (control) | ok (8.14s) | ok (17.85s) — exit 0 never hits the buggy branch |

### Focused reproducer for branch ⑥ (added on review)

`test_run_rm_non_zero_exit_does_not_leak_shim` runs the same scenario (`run --rm alpine:latest sh -c 'exit 7'`) but scans `<home>/boxes/*/shim.pid` for live PIDs **in the test body** — so the no-leak assertion is visible at the call site instead of buried in `PerTestBoxHome::Drop`'s panic.

| step | code state | result |
|---|---|---|
| A | RAII fix applied | `test_run_rm_non_zero_exit_does_not_leak_shim ... ok` (5.79s) |
| B | `run.rs:93` reverted to `std::process::exit(to_shell_exit_code(exit_code))` | **FAILED** — `non-zero boxlite run --rm left live shim PID(s): [286477]` |
| C | RAII fix restored | `test_run_rm_non_zero_exit_does_not_leak_shim ... ok` (6.06s) |

### Other early-return branches in `BoxRunner::run`

Reviewer's request was "execution and litebox dropped automatically in **any** branch". The remaining CLI-reachable early-returns each get their own no-leak test:

| branch | trigger | test | two-sided? |
|---|---|---|---|
| ① `validate_flags?` | `--tty` with non-TTY stdin | `test_run_tty_error_in_pipe` (existing) | n/a — fails before any box is created, structurally cannot leak |
| ② `create_box?` | image pull failure | `test_run_image_pull_failure_does_not_leak_shim` | n/a — pull fails before any shim is spawned, structurally cannot leak |
| ③ `litebox.exec?` | invoking `/etc` (a directory) | `test_run_exec_setup_failure_does_not_leak_shim` | **yes — see below** |
| ④ `detach return Ok(0)` | `-d` flag | `test_run_detach` (existing, manually rm's the box) | n/a — keeping the box alive is the intended behavior |
| ⑤ `streamer.start?` | signal-handler init failure | none — effectively unreachable in any real CLI invocation; covered by `PerTestBoxHome::Drop`'s implicit guard + Rust's stack-unwinding guarantee |
| ⑥ non-zero exit | command exits non-zero | tables above | yes |
| ⑦ `Ok(to_shell_exit_code(0))` | command exits 0 | `test_run_exit_code_success` + 30+ commands as side-effect | implicit |
| panic | internal invariant violated | none — guaranteed by Rust: panic unwinds the stack, Drop runs; tests would exercise the panic, not its cleanup |

Two-sided verification for branch ③ (the only realistic injection point on a path where a shim is actually alive at the failure moment):

| step | code state | result |
|---|---|---|
| A | RAII fix applied | `test_run_exec_setup_failure_does_not_leak_shim ... ok` (4.44s) |
| B | `litebox.exec().await?` replaced with `match { Err => std::process::exit(1) }` to inject the Drop-bypass pattern onto this branch | **FAILED** — `exec-setup failure left live shim PID(s): [395393]` |
| C | injection reverted | `test_run_exec_setup_failure_does_not_leak_shim ... ok` (5.22s) |

### Why this works

Pre-fix, a non-zero command exit took the `std::process::exit` shortcut that bypasses the box teardown the success path runs on return, leaking the microVM's shim (the source of #604's "orphan shims in /tmp"). Post-fix returns the exit code as `Result<i32>` and lets `main` return `ExitCode`; the runtime drops on every return path and `RuntimeImpl::Drop` → `shutdown_sync` SIGTERMs the shim — the same teardown the success path already relied on.

`make fmt:check` + `cargo clippy -- -D warnings` clean; pre-push CLI matrix `277 tests run: 277 passed, 0 skipped` (~83s).

## Summary by CodeRabbit

* **Bug Fixes**
  * Exit codes from executed commands are now properly propagated and returned by the CLI instead of being discarded.
  * Improved shutdown and cleanup logic to ensure orphaned processes are not left behind after command execution.

* **Tests**
  * Added comprehensive test coverage to verify proper process cleanup across various scenarios, including non-zero exits and failure paths.

### DIFF
--- a/src/cli/src/commands/run.rs
+++ b/src/cli/src/commands/run.rs
@@ -33,8 +33,15 @@ pub struct RunArgs {
     pub command: Vec<String>,
 }
 
-/// Entry point
-pub async fn execute(args: RunArgs, global: &GlobalFlags) -> anyhow::Result<()> {
+/// Entry point.
+///
+/// Returns the shell exit code the CLI should exit with (0 on success, the
+/// box's mapped exit code on a non-zero command exit). Returning the code —
+/// instead of calling `std::process::exit` mid-function — lets `BoxRunner`
+/// (and the `BoxliteRuntime` it owns) drop normally, so `RuntimeImpl::Drop`
+/// runs `shutdown_sync()` and stops the box's shim on every return path.
+/// `std::process::exit` would bypass that Drop chain and leak the shim (#622).
+pub async fn execute(args: RunArgs, global: &GlobalFlags) -> anyhow::Result<i32> {
     let mut runner = BoxRunner::new(args, global)?;
     runner.run().await
 }
@@ -53,7 +60,7 @@ impl BoxRunner {
         Ok(Self { args, rt, home })
     }
 
-    async fn run(&mut self) -> anyhow::Result<()> {
+    async fn run(&mut self) -> anyhow::Result<i32> {
         // Validate flags and environment
         self.validate_flags()?;
 
@@ -66,7 +73,7 @@ impl BoxRunner {
         // Detach mode: Print ID and exit
         if self.args.management.detach {
             println!("{}", litebox.id());
-            return Ok(());
+            return Ok(0);
         }
 
         // --tty implies --interactive when stdin is a terminal
@@ -83,18 +90,14 @@ impl BoxRunner {
         );
 
         let exit_code = streamer.start().await?;
-        // Exit with box's exit code
-        if exit_code != 0 {
-            // Tear the box down before hard-exiting. std::process::exit skips
-            // Drop and the async auto-stop/auto-remove, so a non-zero command
-            // exit would otherwise leak the box's shim as a live process (the
-            // success path stops the box via normal teardown when run returns).
-            drop(execution);
-            let _ = litebox.stop().await;
-            std::process::exit(to_shell_exit_code(exit_code));
-        }
-
-        Ok(())
+        // Just return the shell exit code. Returning (vs. calling
+        // `std::process::exit` here) lets `execution`, `litebox`, and the
+        // owning `BoxliteRuntime` drop normally, so `RuntimeImpl::Drop`
+        // runs `shutdown_sync()` and tears the box's shim down — the RAII
+        // teardown the success path already relied on. `std::process::exit`
+        // bypasses Drop entirely and leaked the shim on the non-zero path
+        // (#622).
+        Ok(to_shell_exit_code(exit_code))
     }
 
     async fn create_box(&self) -> anyhow::Result<LiteBox> {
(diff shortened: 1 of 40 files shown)


src/cli/src/commands/run.rs
```diff
@@ -85,6 +85,12 @@ impl BoxRunner {
         let exit_code = streamer.start().await?;
         // Exit with box's exit code
         if exit_code != 0 {
+            // Tear the box down before hard-exiting. std::process::exit skips
```
DorianZheng: try to design RAII around this so we `execution` and `litebox` will be dropped automatically in any branch

src/cli/src/commands/run.rs
```diff
@@ -85,6 +85,12 @@ impl BoxRunner {
...
+            // Drop and the async auto-stop/auto-remove, so a non-zero command
+            // exit would otherwise leak the box's shim as a live process (the
+            // success path stops the box via normal teardown when run returns).
+            drop(execution);
```
DorianZheng: write a test case to cover this issue

G4614: `test_run_rm_non_zero_exit_does_not_leak_shim` added for this, thx

src/cli/src/commands/run.rs
G4614: changed to RAII mode, also wrote test for each branch to confirm it, thx

DorianZheng (approved):
LGTM

PR #626, opened 2026-05-29
fix(cli): read stdin on a dedicated OS thread so TTY exec exits without a stray ENTER

G4614:
Move CLI stdin reads off `tokio::io::stdin()` onto a dedicated `std::thread` (not joined at runtime shutdown) so `exec -ti` returns to the host the moment the in-box shell exits, instead of hanging until a stray ENTER.

## Test plan

`pexpect` PTY against a real box: type `exit`↵ once inside the box and time the host return (`main + #625` baseline; bare main can't reach the hang because TTY exec dies upstream at `#625`):

| observed | pre-fix | post-fix |
|---|---|---|
| host process returns | **HANG ≥8 s** — needs a second ENTER to unblock (dt2 = 8.06 s) | **0.04 s** |
| exit code | 0 (after the stray ENTER) | 0 |

Two automated regressions in `src/cli/src/terminal/mod.rs` (added per @DorianZheng's review):

| test | setup | pre-fix | post-fix |
|---|---|---|---|
| `stream_stdin_does_not_block_runtime_shutdown_after_abort` | `dup2` a writer-less pipe onto fd 0, spawn `stream_stdin`, abort the task (what `StreamManager::start` does at line 191 once `wait()` returns), then time `Runtime::shutdown_timeout(2s)` | hangs ~2.00 s | passes |
| `stream_manager_with_pty_does_not_block_runtime_drop_after_exec_exit` | full select-loop with a real `nix::pty::openpty` slave on fd 0 + stub `Execution` driving a clean shell exit via `result_tx` | hangs ~2.00 s | passes |

Combined wall-clock: <0.3 s post-fix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -492,7 +492,6 @@ dependencies = [
  "futures",
  "gtmpl",
  "gtmpl_value",
- "libc",
  "nix 0.30.1",
  "notify",
  "openidconnect",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -492,6 +492,7 @@ dependencies = [
  "futures",
  "gtmpl",
  "gtmpl_value",
+ "libc",
  "nix 0.30.1",
  "notify",
  "openidconnect",
--- a/src/cli/Cargo.toml
+++ b/src/cli/Cargo.toml
@@ -79,7 +79,6 @@ gtmpl_value = "0.5"
 boxlite = { workspace = true, features = ["rest", "test-support"] }
 boxlite-test-utils = { path = "../test-utils" }
 assert_cmd = "2.1.1"
-libc = "0.2"
 predicates = "3.1.3"
 rstest = "0.21"
 tempfile = "3.24.0"
--- a/src/cli/Cargo.toml
+++ b/src/cli/Cargo.toml
@@ -79,6 +79,7 @@ gtmpl_value = "0.5"
 boxlite = { workspace = true, features = ["rest", "test-support"] }
 boxlite-test-utils = { path = "../test-utils" }
 assert_cmd = "2.1.1"
+libc = "0.2"
 predicates = "3.1.3"
 rstest = "0.21"
 tempfile = "3.24.0"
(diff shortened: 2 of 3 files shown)


G4614:
**The crux — the one-line cause and the change**

The interactive stdin reader used `tokio::io::stdin()`, whose blocking `read(2)` runs on a tokio **blocking-pool** thread that the runtime **joins on shutdown**. When the remote shell exits, the select loop aborts the stdin task (`src/cli/src/terminal/mod.rs:191`):

```rust
if let Some(h) = stdin_handle.as_ref() {
    h.abort();   // cancels the async task — NOT the OS thread already parked in read(2)
}
```

`abort()` only cancels the async task; it cannot interrupt a thread already parked in `read(2)`. So runtime shutdown blocked on that pool thread until the user pressed ENTER to unblock the read — the stray-ENTER hang.

**Fix** — `stream_stdin` (`src/cli/src/terminal/mod.rs:239`) moves the blocking read onto a plain `std::thread` (NOT the tokio blocking pool, so the runtime does **not** join it on shutdown) and forwards bytes over a channel:

```rust
// before
let mut stdin = tokio::io::stdin();
loop { match stdin.read(&mut buf).await { /* … */ } }

// after
let (tx, mut rx) = tokio::sync::mpsc::channel::<Vec<u8>>(16);
std::thread::spawn(move || {
    let mut stdin = std::io::stdin();
    loop {
        match stdin.read(&mut buf) {
            Ok(n) => { if tx.blocking_send(buf[..n].to_vec()).is_err() { break } }
            /* … */
        }
    }
});
while let Some(chunk) = rx.recv().await { /* forward to the box */ }
```

The process now exits promptly on shell exit; the still-parked read is reaped by process exit. Verified on a real box: host prompt returns in ~0.03 s after one `exit`, vs ≥8 s requiring a second ENTER before the fix.

DorianZheng:
@G4614 Please write a test case to reproduce the issue before this diff

G4614:
> @G4614 Please write a test case to reproduce the issue before this diff

unit test and e2e test are both ready, thx

PR #627, opened 2026-05-29
fix(rest): re-attach to drain backlog before emitting terminal result

G4614:
Re-attach once to drain server-buffered stdout when the attach WebSocket is cut before the exit frame, fixing silent stdout loss.
use local `boxlite serve` and REST POST request for E2E test

## Test plan

Two-sided (re-attach disabled vs applied), toggled on the branch since the tests ship with the fix.

### In-process unit reproducer (mock WS)

- `ws_terminal_probe_after_cut_must_not_drop_buffered_stdout` — in-process mock-WS in `rest/litebox.rs`: the first attach is cut before the exit frame, `GET /executions/{id}` reports completed/exit 0, and the backlog (`hello\n`) is served **only** on re-attach.

| observed | pre-fix (re-attach disabled) | post-fix |
|---|---|---|
| consumer stdout (fast cmd, late attach) | **`""`** — clean exit 0, output dropped | `"hello\n"` — backlog drained |
| 50×4 fast-command repro (Go SDK, cloud) | 18–26/50 loss | 0/50 |
| test result | **FAIL** (`left:"" right:"hello\n"`) | **PASS** |

### End-to-end reproducer (real `boxlite serve` + TCP proxy)

`boxlite_rest_attach_drain_after_proxy_cut_serves_stdout` (`src/cli/tests/rest_e2e.rs`) runs the actual production stack: real `boxlite serve` subprocess on an ephemeral port, the production `BoxliteRuntime::rest` client (the path through `attach_ws_pump`), and an inline async TCP proxy in front of the daemon that deterministically cuts the **first** WS upgrade after the 101 response and lets every subsequent connection (HTTP status probes, the second WS for the re-attach) through transparently. Inside the box: `sh -c "echo hello"` in alpine.

| step | code state | result |
|---|---|---|
| A | fix applied | `ok` (6.57 s) — 1st WS cut → status Terminal → re-attach (2nd WS transparent) → runner replays `hello\n` + exit frame |
| B | `terminal_drain_attempted = true` at `rest/litebox.rs:563` (re-attach short-circuited) | **FAILED** — `assertion left == right` `left: "", right: "hello\n"` — the silent stdout loss this PR fixes |
| C | fix restored | `ok` (8.71 s) |

What this end-to-end test catches that the mock unit test doesn't:
- Real WS upgrade handshake between axum (serve) and reqwest-tungstenite (REST client).
- `BoxliteRuntime::rest` end-to-end: box create + exec + attach + status probe through one TCP proxy without any in-process scaffolding.
- The 2nd-WS re-attach actually receiving the runner's backlog over a real WS subprotocol exchange (not a hand-crafted `Message::Binary` reply in the mock).
- `boxlite serve` accepting the proxy-mediated traffic exactly as it would from any external REST consumer.

Orthogonal to #563 (an `sdks/c` FFI drain fix at the SDK layer): this is the REST client attach path, and the loss reproduced on **both `main` and the #563 branch**. Tradeoff: the re-attach replays the runner's full backlog without an offset, so a partial prefix delivered before the cut can be **duplicated** — this matches existing `StillRunning`/`Unavailable` reconnect semantics and is strictly preferable to silent loss (offset-based replay is out of scope).

### DIFF
--- a/src/boxlite/src/rest/litebox.rs
+++ b/src/boxlite/src/rest/litebox.rs
@@ -557,6 +557,10 @@ async fn attach_ws_pump(
     // Sticky across reconnects: once the server has ever sent a frame the
     // exec is real, so a later reconnect uses the steady-state watchdog.
     let mut first_frame_seen = false;
+    // Set once we've already re-attached to drain the backlog after the exec
+    // went terminal. Bounds that drain to a single attempt so a runner that
+    // never sends a closing exit frame can't spin this loop.
+    let mut terminal_drain_attempted = false;
 
     let mut current_stream = Some(initial_stream);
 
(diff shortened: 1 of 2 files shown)


G4614:

PR #628, opened 2026-05-29
feat(secrets): support live box secret updates

G4614:
Adds live box secret updates through the REST API, runner v0/v2, and gvproxy admin socket.

Test plan:
- `go test ./...` in `src/deps/libgvproxy-sys/gvproxy-bridge`
- `BOXLITE_DEPS_STUB=1 cargo build -p boxlite-c`
- `go test -tags boxlite_dev ./pkg/api/dto ./pkg/api/controllers ./pkg/backend ./pkg/boxlite ./pkg/runner/v2/executor` in `apps/runner`
- `BOXLITE_DEPS_STUB=1 cargo test -p boxlite --lib net::gvproxy::config` (compiles; test filter currently reports 0 matched)

Note: `npx nx build api --configuration=development --skip-nx-cache` was not run because `apps/` has no installed Nx modules/node_modules in this checkout.

### DIFF
--- a/src/cli/tests/cp.rs
+++ b/src/cli/tests/cp.rs
@@ -1,0 +1,103 @@
+//! Integration tests for the `boxlite cp` CLI command (host ↔ box file copy).
+//!
+//! Exercises the binary end-to-end: `cp` into a running box, confirm the file
+//! actually landed at the target path with the right content (by `exec cat`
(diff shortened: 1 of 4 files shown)


G4614:
#185 is right, act like docker

PR #633, opened 2026-06-01
fix(rest): cp box->host extracts single file to dest

G4614:
when REST `copy_out` a regular file from box to outside, a dir created in host because of the tar logic

## Test plan

`cargo test -p boxlite --lib --features rest tar_extract` — 4 unit tests, in-memory tar construction:

| scenario | pre-fix | post-fix |
|---|---|---|
| `cp box:/etc/hosts ./hosts` (tar = `./` + `./etc/` + `./etc/hosts`) | `./hosts/etc/hosts` (directory tree) | `./hosts` is a regular file, content = `/etc/hosts` from the box |
| `cp box:/foo ./out/` (trailing slash) | dir | dir (no regression) |
| `cp box:/dir ./dir-copy` (multi-file) | dir | dir (no regression) |
| `--no-overwrite` + dst exists | old unpack overwrites silently | error `exists and overwrite=false`, pre-existing content preserved |

Two-side: temporarily restore the old `archive.unpack` body for `extract_tar_bytes_to_path` → `single_regular_file_lands_at_dest_path` fails with `dest must be a regular file, not a directory tree` and `overwrite_false_refuses_existing_dest` fails with `expected overwrite-refusal error, got: failed to unpack` — those are exactly the two POL-37 symptoms. Restore the fix: 4/4 pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tar archive extraction to correctly handle synthetic directory entries and symlinks.
  * Enhanced overwrite protection during file extraction operations.

* **Tests**
  * Expanded test coverage for archive extraction scenarios, including docker-cp-style archives and symlink handling.
  * Added validation for proper error handling when archives are missing.

### DIFF
--- a/src/boxlite/src/rest/litebox.rs
+++ b/src/boxlite/src/rest/litebox.rs
@@ -246,13 +246,22 @@ impl BoxBackend for RestBox {
         &self,
         container_src: &str,
         host_dst: &Path,
-        _opts: CopyOptions,
+        opts: CopyOptions,
     ) -> BoxliteResult<()> {
         let box_id = self.box_id_str();
 
-        // Download tar from server
+        // Download tar from server. Query params mirror what local copy_out
+        // hands to the guest: a future server that wires these through will
+        // produce a leaner archive (no synthetic parent directory chain when
+        // include_parent=false) so the single-file extraction path below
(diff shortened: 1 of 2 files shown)


G4614:

PR #637, opened 2026-06-01
feat(rest): images pull + list end-to-end (POL-32)

G4614:
REST/SDK runtime support image pull/list

## Test plan

`cargo nextest run -p boxlite-cli --test rest_images` — 4 stub-driven tests:

| scenario | pre-fix | post-fix |
|---|---|---|
| `pull --profile p1 alpine:latest` | `Image operations not supported over REST API`, exit 1 | POST `/v1/images/pull` with `{"reference":"alpine:latest"}`; stdout `Pulled: alpine:latest` + `ID: sha256:...` |
| `pull --profile p1 -q alpine:latest` | same | stdout id only (useful for `IMAGE_ID=$(...)`) |
| `images --profile p1` | same | GET `/v1/images`, renders table with `docker.io/library/alpine` `latest` |
| `pull` against server `422 image_pull_failed` | n/a | exit≠0 + stderr includes server error msg |

Additional server-side coverage:

- `BOXLITE_DEPS_STUB=1 cargo test -p boxlite-cli serve::handlers::images::tests`
  - verifies `POST /v1/images/pull` response correlation matches the pulled manifest digest, not the user's raw reference (`alpine` vs `docker.io/library/alpine:latest`)
  - verifies the handler returns 500 with the requested ref + missing digest when pull succeeds but the cache listing does not contain the pulled digest

Two-side:

- `git checkout HEAD -- pull.rs images.rs` reverts the CLI branching (server + REST client + tests preserved) → `rest_images` 4/4 fail. Restore → 4/4 pass.
- Temporarily changing the server helper back to reference matching (`i.reference == req_reference`) makes `pulled_image_response_matches_by_manifest_digest_not_user_reference` fail with `500` vs expected `200`. Restore digest matching (`i.id == manifest_digest`) → server handler tests pass.

## Known follow-ups

1. SDKs (Node / Python / C / Go) still expose only the local `ImageObject` shape; REST exposure is a separate PR.
2. No streaming progress for image pull (OpenAPI doesn't ask for it either); current behavior is sync-block-until-pulled.
3. `GET /v1/images/{id}` and the HEAD variant from OpenAPI aren't wired here; follow-up.

### DIFF
--- a/sdks/c/src/images.rs
+++ b/sdks/c/src/images.rs
@@ -253,12 +253,12 @@ unsafe fn image_pull(
         let user_data_addr = user_data as usize;
 
         handle_ref.tokio_rt.spawn(async move {
-            let result = core_handle.pull(&image_ref).await.map(|image| {
+            let result = core_handle.pull_info(&image_ref).await.map(|image| {
                 crate::event_queue::OwnedFfiPtr::new_with(
                     Box::new(CImagePullResult::new(
-                        image.reference(),
-                        image.config_digest(),
-                        image.layer_count(),
+                        &image.reference,
+                        &image.config_digest,
+                        image.layer_count,
                     )),
                     free_image_pull_result,
                 )
--- a/sdks/c/src/tests.rs
+++ b/sdks/c/src/tests.rs
@@ -212,7 +212,7 @@ fn test_free_functions_null_safe() {
 }
 
 #[test]
-fn test_runtime_images_unsupported_on_rest_runtime() {
+fn test_runtime_images_supported_on_rest_runtime() {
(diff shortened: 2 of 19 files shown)


G4614:
Closing — this only wires image operations into the self-hosted `boxlite serve` reference server (`src/cli/src/commands/serve/*`); it never touches `apps/api` or `apps/runner`, so it wouldn't actually fix POL-32's reported repro (cloud REST against `api.dev.boxlite.ai`). #703 covers the real gap end-to-end (apps/api controller + Go runner + core SDK), just for pull only (no list yet). Picking that one back up instead — see POL-32 for tracking.

PR #645, opened 2026-06-02
fix(net): allowlist DNS resilience by retrying, fail closed and allow longest-suffix wins

ldarren:
## Issue
When `allow_net` is set, allow-listed hostnames sometimes resolve to `0.0.0.0` inside the guest, even though the host is reachable and `box.create` reports success. The failure is intermittent and silent — the only signal is `allow_zones=N-1 total_zones=N` in the gvproxy log.

  ```
  # inside the guest
  $ getent hosts iapi.example.com
  0.0.0.0   iapi.example.com
  ```

  These are the root causes and fixes:

  1. Host-side resolution had no retry or timeout — one transient hiccup (VPN flap, slow corp DNS) silently dropped the zone.
  2. When resolution definitively failed, `box.create` proceeded with a half-baked sinkhole instead of failing loudly.
  3. Zones were emitted in Go map-iteration order. gvisor-tap-vsock matches first-suffix-wins, so an `subdomain.domain.com.` query could land on a `com.` zone (created for `github.com`) before the `domain.com.` zone and return that zone's `DefaultIP=0.0.0.0`.

  ## Changes

  - [x] Retry host-side `LookupIPAddr` with exponential backoff (4 attempts: 100/300/900/2700 ms) and a 2 s per-attempt deadline
  - [x] Fail-closed: propagate resolver errors through `buildAllowNetDNSZones` → `buildTapConfig` → `gvproxy_create`, surfaced via the new `errOut` channel from #612
  - [x] Sort zones longest-name-first so the most-specific suffix wins; catch-all root zone stays last

## Summary by CodeRabbit

* **New Features**
  * DNS lookups during box creation now use automatic retries with exponential backoff and per-attempt timeouts.
  * DNS resolution is signal-aware and cancels promptly on SIGINT/SIGTERM.

* **Bug Fixes**
  * Deterministic longest-suffix DNS zone ordering to avoid incorrectly sinkholing allow-listed hosts.
  * Box creation fails explicitly if allow-list hostnames cannot be resolved.

* **Documentation**
  * Added troubleshooting and reference guidance for allow_net DNS behavior and recovery.

* **Tests**
  * Expanded tests covering retries, timeouts, cancellation, and ordering.

* **Chores**
  * Log build identity once per process.

### DIFF
--- a/src/deps/libgvproxy-sys/gvproxy-bridge/dns_filter.go
+++ b/src/deps/libgvproxy-sys/gvproxy-bridge/dns_filter.go
@@ -7,25 +7,70 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"regexp"
+	"sort"
 	"strings"
+	"time"
 
 	"github.com/containers/gvisor-tap-vsock/pkg/types"
 	logrus "github.com/sirupsen/logrus"
 )
 
+// Tunables for allowlist DNS resolution. Each allow-listed hostname is
+// resolved against the host OS resolver at box-create time and the
+// resulting IPs are baked into the DNS sinkhole. A transient host-side
+// DNS hiccup (VPN flap, slow corp resolver, mDNSResponder churn) used to
+// silently drop the host's zone, leaving the VM permanently sinkholing
+// it to 0.0.0.0 for the life of the box. We now retry with backoff and
+// fail closed if every attempt fails.
+//
+// Exposed as `var` (not `const`) so tests can override timing without
+// slowing the suite. Production callers MUST treat them as immutable.
+const dnsLookupAttempts = 4
+
+var (
+	dnsLookupInitialBackoffVar = 100 * time.Millisecond
+	dnsLookupBackoffFactor     = 3
+	dnsLookupAttemptTimeoutVar = 2 * time.Second
+)
+
+// hostResolver is the interface buildAllowNetDNSZones uses to look up
+// allow-listed hostnames. The production implementation calls the host
+// OS resolver via net.Resolver. Tests inject a fake to exercise the
+// retry/backoff/fail-closed paths without depending on real DNS.
+type hostResolver interface {
+	LookupIPAddr(ctx context.Context, host string) ([]net.IPAddr, error)
+}
+
+// defaultResolver is the production resolver: PreferGo:false uses the
+// platform's getaddrinfo, which honors VPN/corp DNS configuration.
+var defaultResolver hostResolver = &net.Resolver{PreferGo: false}
+
 // buildAllowNetDNSZones creates DNS zones that implement allowlist filtering.
 //
 // Strategy:
-//   - For each allowed hostname: resolve to IPs, create a zone with A records
-//   - For wildcard patterns (*.example.com): create zone with Regexp records
-//   - Add catch-all root zone "" with DefaultIP 0.0.0.0 (sinkhole)
+//   - For each allowed hostname: resolve to IPs (with retry/backoff), create
+//     a zone with A records.
+//   - For wildcard patterns (*.example.com): create zone with Regexp records.
+//   - Add catch-all root zone "" with DefaultIP 0.0.0.0 (sinkhole).
 //
 // Zone matching is first-match-wins with suffix matching. Specific zones
 // are added before the root zone, so allowed hosts resolve normally while
 // everything else gets sinkholed.
-func buildAllowNetDNSZones(allowNet []string) []types.Zone {
+//
+// Fail-closed: if any allow-listed hostname cannot be resolved after all
+// retry attempts, this function returns an error instead of producing a
+// silently-incomplete sinkhole. The caller is expected to abort box
+// creation rather than ship a misconfigured network.
+func buildAllowNetDNSZones(allowNet []string) ([]types.Zone, error) {
+	return buildAllowNetDNSZonesWith(allowNet, defaultResolver)
+}
+
+// buildAllowNetDNSZonesWith is the testable form: same behavior as
+// buildAllowNetDNSZones, with an injectable resolver.
+func buildAllowNetDNSZonesWith(allowNet []string, resolver hostResolver) ([]types.Zone, error) {
 	zoneRecords := make(map[string][]types.Record)
 
 	for _, rule := range allowNet {
@@ -55,7 +100,9 @@ func buildAllowNetDNSZones(allowNet []string) []types.Zone {
 			zoneRecords[zoneName] = append(zoneRecords[zoneName], types.Record{
 				Regexp: regexp.MustCompile(".*"),
 			})
-			resolveAndAddRecords(domain, domain+".", zoneRecords)
+			if err := resolveAndAddRecords(resolver, domain, domain+".", zoneRecords); err != nil {
+				return nil, err
+			}
 			continue
 		}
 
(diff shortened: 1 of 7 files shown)


DorianZheng:
Small scope nit on `logBridgeBuild` / `bridgeBuildID`: nice debugging aid, but it's unrelated to the allow_net fix and the hand-bumped `"...allownet-dns-resilience-2026-06"` constant will go stale (nobody will remember to bump it). Consider either splitting it into its own PR, or dropping the constant and logging just the auto-derived `vcs.revision`/`vcs.modified`/`vcs.time` fields so it never lies. The retry + fail-closed + longest-suffix-sort changes all look great. 👍

src/deps/libgvproxy-sys/gvproxy-bridge/dns_filter.go
```diff
@@ -93,37 +165,96 @@ func buildAllowNetDNSZones(allowNet []string) []types.Zone {
...
+// treated as a failure so we retry rather than bake an empty zone.
+func lookupWithRetry(resolver hostResolver, hostname string) ([]net.IPAddr, error) {
+	var (
+		ips     []net.IPAddr
+		lastErr error
+	)
+	backoff := dnsLookupInitialBackoffVar
+
+	for attempt := 1; attempt <= dnsLookupAttempts; attempt++ {
+		ctx, cancel := context.WithTimeout(context.Background(), dnsLookupAttemptTimeoutVar)
```
G4614: Maybe it's better to catch the parent cox so that Ctrl-C can end the retry timeout, thx

G4614 (commented):
Also better to add tests to confirm the difference between pre and post fix

Have you meet DNS timeout I/O reported by mistake in this situation?

ldarren:
> Also better to add tests to confirm the difference between pre and post fix
> 
> Have you meet DNS timeout I/O reported by mistake in this situation?

Sure, here are two snippets that demo before/after — easier to grok than test code:
 
Pre-fix (map-iteration order, what gvisor saw):
```go
// gvisor matches first-suffix-wins, no most-specific preference.
// services/dns/dns.go:51 — strings.HasSuffix(q.Name, "."+zone.Name)
 
zones := []Zone{
  {Name: "com.",       DefaultIP: net.IPv4(0,0,0,0),
  Records: []Record{{Name: "github", IP: net.IPv4(20,205,243,166)}}},
  {Name: "sample.com.", DefaultIP: net.IPv4(0,0,0,0),
  Records: []Record{{Name: "iapi",   IP: net.IPv4(100,64,1,1)}}},
}
// query "iapi.sample.com." → matches com. first → no record "iapi.sample"
//                        → returns DefaultIP 0.0.0.0 
```
 
Post-fix (longest-name-first, what we now emit):
```go
zones := []Zone{
  {Name: "sample.com.", DefaultIP: net.IPv4(0,0,0,0),
  Records: []Record{{Name: "iapi",   IP: net.IPv4(100,64,1,1)}}},
  {Name: "com.",       DefaultIP: net.IPv4(0,0,0,0),
  Records: []Record{{Name: "github", IP: net.IPv4(20,205,243,166)}}},
}
// query "iapi.sample.com." → matches sample.com. first → record "iapi"
//                        → returns 100.64.1.1 
```
I'd rather not encode this as a unit test — it would have to copy gvisor's matcher into our tests and would mislead future contributors. TestBuildAllowNetDNSZones_LongestSuffixWinsBeforeRoot already pins the contract on our side.

On DNS timeouts: doesn't happen as misreported — it just happens, on corporate networks with TLS scanners and proxy chains. Slow but legitimate resolution. The original code's context.Background() lookup didn't handle that error path at all; the new retry/fail-closed code does.

src/deps/libgvproxy-sys/gvproxy-bridge/dns_filter.go
ldarren: Good call, in the retry loop now takes a parentCtx parameter threaded down from gvproxy_create to handle interrupt signals

ldarren:
> Small scope nit on `logBridgeBuild` / `bridgeBuildID`: nice debugging aid, but it's unrelated to the allow_net fix and the hand-bumped `"...allownet-dns-resilience-2026-06"` constant will go stale (nobody will remember to bump it). Consider either splitting it into its own PR, or dropping the constant and logging just the auto-derived `vcs.revision`/`vcs.modified`/`vcs.time` fields so it never lies. The retry + fail-closed + longest-suffix-sort changes all look great. 👍

agreed, constant removed

DorianZheng:
solid fix — retry, fail-closed, and longest-suffix ordering each map cleanly onto the three root causes, and the injected-resolver tests are the right shape. two things i'd want resolved before merge, plus some cleanups.

## signal handler inside the FFI export (main.go:985)

`signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)` runs inside `gvproxy_create`, which is a `//export`ed symbol called from the rust runtime embedded in the python/node/go/c SDKs — a host process that owns its own signal disposition. in c-shared mode every `box.create` transiently diverts the host's SIGINT/SIGTERM into go's signal machinery. a host app with its own SIGTERM handler (graceful shutdown, supervisor) could see different behavior during any create. the cancellation goal is right; a library grabbing process signals to achieve it is the part i'd push back on. can the cancellation be driven by a context the rust caller already controls across the FFI boundary instead? and was this exercised with an embedding process that installs its own handlers?

## fail-closed doesn't cover ipv6-only hosts (dns_filter.go:299-329)

`lookupWithRetry` treats `len(ips) > 0` as success, but `resolveAndAddRecords` then drops every non-v4 address (`if ip.IP.To4() == nil { continue }`). an allow-listed host that resolves AAAA-only gets a zone with zero A records, falls through to that zone's `DefaultIP=0.0.0.0`, and silently sinkholes — the exact symptom this PR sets out to kill. it's logged (`v4_count=0`) but not failed. meanwhile docs/reference/README.md:83-90 states create "fails — by design, so a half-baked sinkhole never silently sinkholes an allow-listed host to 0.0.0.0", which is only true for resolution errors, not v4-less results. the code and the doc guarantee disagree; one of them needs to move.

## minor

- config_build comment above `const dnsLookupAttempts` says "exposed as `var` (not `const`) so tests can override timing" — it's describing the `var (...)` block below it but sits on the const. move it so the const isn't covered by an "exposed as var" sentence.
- dns_filter_test.go hand-rolls `contains`/`indexOf` instead of importing `strings` and calling `strings.Contains` — more code, more to get wrong, just import it.
- `flakyResolver.successIPs` and `attemptHook` are read but never set by any test — dead scaffolding, drop until something needs them.
- the zone sort uses `sort.Slice` (unstable) but the comment claims "deterministic" ordering — equal-length zone names get an unspecified order. doesn't affect correctness (equal length can't shadow), and the test only uses distinct-length names so it wouldn't catch a regression; use SliceStable or a lexical tiebreak if you want the comment to hold literally.
- `dnsLookupInitialBackoffVar`/`dnsLookupAttemptTimeoutVar` carry a `Var` suffix that `dnsLookupBackoffFactor` (also a var) doesn't — inconsistent, and it leaks "is a variable" into the name with no const counterpart to disambiguate from.

## worth a note, not a blocker

- worst-case create latency on the failure path is sequential per host: ~4×2s timeouts + ~4s backoff ≈ 12s per unreachable host, so a 10-entry allow_net of dead hosts is ~2min before create fails (interruptible, modulo the signal issue above). a line in docs/reference would save someone the surprise.
- the longest-suffix test asserts slice order as a proxy for gvisor-tap-vsock's first-suffix-wins matching — reasonable since you can't easily drive the real matcher, but the whole fix rests on that upstream contract. a link to the gvisor-tap-vsock matcher code would harden it against an upstream behavior change the test can't see.

ldarren:
Thanks for the review. All addressed in the latest push.

  **Blockers**
  - Signal handler: dropped `signal.NotifyContext` and the `os/signal` / `syscall` imports. `gvproxy_create` runs against `context.Background()` now — no host signal disposition is touched. Trade-off: SIGTERM mid-create can't interrupt the resolver loop (~12s/host worst case on dead allow_net). Considered a 
  sibling `gvproxy_create_with_cancel` + token registry to preserve cancellation, but it's ~110 LOC across go+rust+SDKs for a path that's rarely critical, so I went with the smaller diff. Happy to layer it on as a follow-up if you'd rather have it. Not exercised with an embedder that installs its own SIGTERM
  handler — the change is "we no longer install one", verifiable from the diff + clean `goimports`.
  - AAAA-only fail-closed: `resolveAndAddRecords` now returns an error when the v4 filter strips every record. New test `TestBuildAllowNetDNSZones_AAAAonlyHostFailsClosed` two-side verified (failed against pre-fix code with `v4_count=0` log line; passes now). Doc updated to match.

  **Minor — all done**
  - const-vs-var comment moved onto the `var ( … )` block.
  - `strings.Contains` replaces the hand-rolled `contains`/`indexOf`.
  - `flakyResolver.successIPs` + `attemptHook` removed.
  - Zone sort: `sort.SliceStable` with length-desc + lexical tiebreak.
  - `dnsLookup{InitialBackoff,AttemptTimeout}Var` → drop the `Var` suffix to match `dnsLookupBackoffFactor`.

  **Doc/test polish**
  - `docs/reference/README.md`: latency note (~12s × N dead hosts, sequential) and AAAA-only added to the fail-closed paragraph.
  - Longest-suffix test cites `containers/gvisor-tap-vsock@v0.8.7` and what would/wouldn't trip it on an upstream change.

  `go build ./... && go vet ./... && go test ./... -count=1` clean. Ready when you are.

PR #646, opened 2026-06-02
feat(box): add privileged plumbing for DinD

G4614:
Add Docker-style `advanced.privileged` across the Rust runtime, REST API, runner, CLI, and SDKs. `privileged=true` is normalized to `capabilities.add=["ALL"]` with an empty drop list and removes only `/proc/sys` from the guest readonly paths; capability add/drop remains a separate shape.

Test plan:
- [x] `cargo fmt --all -- --check`
- [x] `cargo check -p boxlite-guest --target aarch64-unknown-linux-musl --tests`
- [x] API Jest tests for REST validation, normalization, and mapping (30 tests)
- [x] API development build and TypeScript check
- [x] Python E2E case added for REST privileged vs `cap_add=ALL` behavior
- [ ] Run `apps/e2e/cases/test_privileged_options.py` on the labeled Linux E2E runner

The Linux E2E workflow remains label-gated because it requires nested KVM. Privileged mode does not add a cgroup mount or clear masked paths.


src/guest/src/container/spec.rs
```diff
@@ -167,6 +167,10 @@ pub fn create_oci_spec(
     devices: &ContainerDevices,
 ) -> BoxliteResult<Spec> {
     let caps = capabilities.to_oci()?;
+    // A container holding CAP_SYS_ADMIN is privileged in the sense DinD needs:
+    // it may write /proc/sys and manage its own cgroups. Anything less keeps
+    // the hardened defaults.
+    let privileged = capabilities.is_privileged();
```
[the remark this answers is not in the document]
G4614: Fixed the stale comment — it still described the earlier SYS_ADMIN-keyed design. It now points at `CapabilitySet::is_privileged`, which requires the full set.

Not wiring privileged into mount construction, though. The writable cgroup2 mount was in an earlier revision and was removed in response to a review finding: with no cgroup namespace in the guest, that hierarchy is the VM-wide one shared with sibling containers and the agent, so a container holding the capability could write `cgroup.procs` for its neighbours. Provisioning an isolated hierarchy instead is a larger change than this PR, and nothing needs it yet.

"DinD unable to manage its cgroups" does not reproduce. On a box booted with this change and no cgroup mount, Docker 29.6.2 starts, pulls images over its own bridge, and runs both single containers and a two-service compose project with working service-name DNS (`getent hosts web` → `172.19.0.2`, HTTP between containers OK). If a workload does turn out to need a writable hierarchy, it should arrive together with the cgroup namespace that makes it safe.

DorianZheng (commented):
the wiring is small, but `is_privileged()` turns an existing capability flag into a sandbox switch, and a few things outside the diff now hang off it.

**guest version gate not bumped** — `MIN_CAPABILITY_GUEST_VERSION` (src/boxlite/src/litebox/init/tasks/guest_init.rs:23) is still `(0, 9, 8)`, but `add=["ALL"]` now means two different things depending on which guest you land on. guest rootfs images are cached per version and reused — the constant's own doc says exactly that — so a host built after this change can boot an old cached 0.9.8 guest, pass the version check, get 41 caps with `/proc/sys` still read-only, and dockerd dies on `ip_forward` with nothing pointing at the cause.

**no privileged flag anywhere** — `ContainerCapabilities` is add/drop only at every layer (cli.rs:473, advanced_options.rs:574, the proto). so you can't ask for a writable `/proc/sys` without also taking all 41 caps, and you can't take all 41 without the wider sandbox. docs/reference/cli/README.md:625 still describes `--cap-add` as capabilities only. and config.json is rebuilt on every start (lifecycle.rs:180 → start.rs:126), so a box created with `--cap-add ALL` before this change gets a writable `/proc/sys` on its next restart after upgrade — no flag touched, no code changed on their side.

**no opt-in gate** — the repo already has one for this shape of change: `ExperimentalFeature::{CustomKernel, NestedVirtualization}` behind `BOXLITE_EXPERIMENTAL` (experimental.rs:16-24). nested virt only hands the workload a `/dev/kvm` node inside its own VM and is gated. this opens the guest's sysctl tree and isn't.

**the capability table is now load-bearing** — `CAPABILITIES_BY_NUMBER` is `[Capability; 41]`, and `supported_capabilities()` does `.take(cap_last_cap + 1)` with no bound check (capabilities.rs:232-236). on a future kernel with `cap_last_cap > 40` it silently truncates: 41 ⊆ 41, so the box is labelled fully privileged while genuinely missing the newest caps — and `ensure_supported` rejects those same caps with "is not supported by the guest kernel". it used to be an input-validation aid; it now decides a sandbox.

**design doc** — docs/architecture/container-capabilities.md:49 (status: accepted) says downstream code "cannot reinterpret the policy", which is what `create_oci_spec` now does. its user-facing advice to prefer `drop=["ALL"]` plus the minimum additions is also now the recipe that turns DinD off.

**description is stale** — it says `--cap-add SYS_ADMIN` counts as privileged (the code deliberately rejects that), that a writable `cgroup2` mount at `/sys/fs/cgroup` is added (still commented out at spec.rs:644), and that `readonlyPaths`/`maskedPaths` are cleared (one entry is dropped, masked paths are untouched). the commit message gets all three right. it also still says "remains draft" but the PR isn't.

**unrelated hunk** — src/deps/libkrun-sys/build.rs:476 is formatting only.

src/guest/src/container/capabilities.rs
```diff
@@ -157,6 +157,18 @@ impl CapabilitySet {
         Ok(capabilities)
     }
 
+    /// Whether the caller asked for the full capability set.
+    ///
+    /// Deliberately not keyed on CAP_SYS_ADMIN alone: `--cap-add SYS_ADMIN` is
+    /// a common narrow request (mounting inside a container) and is not the
+    /// same as asking for privileged behaviour, so widening the sandbox on it
+    /// would surprise callers who wanted only that capability.
```
DorianZheng: the doc says this answers "did the caller ask for the full set", but the code never looks at the `ALL` keyword. it compares the resolved set against the kernel's full set. so `--cap-drop ALL` plus every capability spelled out, or `--cap-add` naming the 27 caps missing from the default 14, also comes out privileged and unmasks `/proc/sys`. you can get the wide sandbox without ever typing `ALL` — and spelling names out instead of using `ALL` is what an audit-minded operator or a policy generator does.

the reverse disagrees too: `check_compatibility` (advanced_options.rs:594) compares the raw add/drop lists, so `add=["ALL"]` and the spelled-out form are rejected as "a different capability policy" for box reuse even though they resolve to the same thing.

src/guest/src/container/capabilities.rs
```diff
@@ -157,6 +157,18 @@ impl CapabilitySet {
...
+    pub(crate) fn is_privileged(&self) -> bool {
+        supported_capabilities()
+            .map(|supported| supported.iter().all(|cap| self.0.contains(cap)))
+            .unwrap_or(false)
```
DorianZheng: four separate things live in these four lines:

- `unwrap_or(false)` throws away a real error. `supported_capabilities()` returns "failed to read guest kernel capability ceiling: {error}", and `resolve()` treats that same error as fatal at line 107. here it just becomes "not privileged", so a `--cap-add ALL` box starts hardened and dockerd fails later with a bare EROFS that names nothing.
- `--cap-add ALL --cap-drop SYS_MODULE` falls off a cliff. drops are applied after `ALL` (lines 114-123), so the set is no longer equal, `/proc/sys` goes back to read-only, and DinD stops working. no error, no warning, no way to tell which drop cost the relaxation.
- `--cap-add ALL --cap-drop ALL` goes the other way and comes out **fully privileged**. `parse_capabilities` (line 202) filters the `ALL` word out of the drop list, so the drop loop at line 120 runs over an empty set. the most locked-down-looking flag pair a user can type gives the widest sandbox.
- this reads `/proc/sys/kernel/cap_last_cap` on every container start. `create_oci_spec` calls it at spec.rs:173 whether or not the branch at spec.rs:552 needs it, and `resolve()` deliberately skips that read when add and drop are both empty. the default box now pays a file read and a 41-element set build it didn't before, on a path whose own comments call ~105ms worth optimising.

and nothing logs which way it went, at any level.

src/guest/src/container/capabilities.rs
```diff
@@ -425,4 +437,19 @@ mod tests {
...
+
+        // A narrow request stays narrow: `--cap-add SYS_ADMIN` is a common
+        // ask (mounting inside a container) and is not `--privileged`, so it
+        // must not widen the sandbox.
+        let narrow =
+            CapabilitySet::resolve(&names(&["SYS_ADMIN"]), &[]).expect("resolve added capability");
+        assert!(!narrow.is_privileged());
+
+        let full = CapabilitySet::resolve(&names(&["ALL"]), &[]).expect("resolve ALL");
+        assert!(full.is_privileged());
```
DorianZheng: this assert can't fail. `resolve(&["ALL"], &[])` returns `supported_capabilities()` as-is (no filtering when drops are empty), and `is_privileged()` calls the same function again and checks it's a subset of itself. corrupt the capability table, break the parse, reorder the entries — it still passes. the two negative asserts above it do carry signal; this one doesn't.

src/guest/src/container/spec.rs
```diff
@@ -167,6 +167,10 @@ pub fn create_oci_spec(
     devices: &ContainerDevices,
 ) -> BoxliteResult<Spec> {
     let caps = capabilities.to_oci()?;
+    // Only a caller asking for every capability gets the relaxed shape; see
+    // CapabilitySet::is_privileged. Narrower requests keep the hardened
+    // defaults.
+    let privileged = capabilities.is_privileged();
```
DorianZheng: nothing tests this line. both new spec tests call `build_linux_spec(..., false)` and `(..., true)` with the bool written by hand, and `only_the_full_capability_set_counts_as_privileged` never builds a spec. change this to `!capabilities.is_privileged()` and the whole suite stays green — every box privileged, or DinD silently never privileged, either way it ships.

no test constructs a real `CapabilitySet::resolve(&["ALL"], &[])` and looks at the resulting spec, `create_oci_spec` is never exercised, and there's no `--cap-add` integration test anywhere under src/cli/tests or src/boxlite/tests.

src/guest/src/container/spec.rs
```diff
@@ -521,7 +526,6 @@ fn build_linux_spec(
     ];
 
     // Readonly paths
-    #[allow(unused)]
     let readonly_paths = [
```
DorianZheng: this list is a hand-written copy of `oci_spec::runtime::get_default_readonly_paths()` — same five entries, same order. the unprivileged path inherits that list live from the crate; the privileged path now uses this frozen copy. Cargo.toml pins `oci-spec = "0.6"` (caret range), so a later 0.6.x that adds an entry reaches unprivileged boxes automatically and silently skips privileged ones — the privileged box quietly loses a protection nobody removed. nothing compares the two lists.

src/guest/src/container/spec.rs
```diff
@@ -545,9 +549,22 @@ fn build_linux_spec(
         builder = builder.devices(devices.to_vec());
     }
 
+    if privileged {
+        // A nested runtime writes /proc/sys/net/ipv4/ip_forward when bringing
+        // up its bridge, so that one entry has to go. The rest of the list
+        // stays: /proc/sysrq-trigger still panics the guest on write, and the
+        // masked set still hides /proc/kcore. Docker's --privileged drops both
+        // lists wholesale; nothing here needs that much.
```
DorianZheng: this promises hardening the spec can't back up, for exactly the containers this branch targets. there's no user namespace (line 394 is commented out) so a `--cap-add ALL` container holds real CAP_SYS_ADMIN in the guest's initial user namespace, and there's no seccomp profile anywhere under src/guest/src/container/ to block `mount(2)`. the readonly entries go on as a plain bind + `MS_REMOUNT|MS_RDONLY` with no `MNT_LOCKED`, so `mount -o remount,bind,rw /proc/sysrq-trigger` and `umount /proc/kcore` should both work from inside the container.

if that holds, the partial relaxation buys nothing over dropping both lists wholesale — it just writes down a guarantee the next reader will believe. worth trying in a real privileged box before the comment claims it.

src/guest/src/container/spec.rs
```diff
@@ -545,9 +549,22 @@ fn build_linux_spec(
...
+        builder = builder.readonly_paths(
+            readonly_paths
+                .iter()
+                .filter(|path| path.as_str() != "/proc/sys")
+                .cloned()
+                .collect::<Vec<_>>(),
+        );
```
DorianZheng: the commit message names one file — `/proc/sys/net/ipv4/ip_forward` — but this drops the whole `/proc/sys` entry. that also opens `/proc/sys/kernel/core_pattern`, `/proc/sys/kernel/modprobe`, and every `vm.*` / `fs.*` knob.

and there's no user namespace and no network namespace here (`build_default_namespaces`, line 383, gives pid/ipc/uts/mount only), so this is the guest's global sysctl tree, not a private one. `ip_forward` flips for the whole box and every sibling container, and `core_pattern` set to `|/tmp/x` runs as guest root outside the container.

the RDONLY remount isn't recursive, so a scope narrower than the whole subtree is reachable.

src/guest/src/container/spec.rs
```diff
@@ -545,9 +549,22 @@ fn build_linux_spec(
...
+    }
+
     builder
```
DorianZheng, line 557: dropping `// .masked_paths(masked_paths)` here removed the last reference to the vec at line 515, so those 12 lines are now dead.

it's also wrong about what actually ships: it lists `/sys/devices/virtual/powercap`, which is **not** masked at runtime, and leaves out `/proc/scsi`, which **is** — the effective set comes from `Linux::default()`. anyone who later "re-enables" it would quietly unmask `/proc/scsi`.

same shape as `readonly_paths`, which lost its `#[allow(unused)]` in this same hunk.

src/guest/src/container/spec.rs
```diff
@@ -1193,4 +1215,48 @@ mod tests {
...
+        assert!(readonly.iter().any(|path| path == "/proc/sysrq-trigger"));
+
+        // Masking is untouched, so kernel memory stays hidden.
+        let masked = linux
+            .masked_paths()
+            .as_ref()
+            .expect("masked paths stay set")
+            .clone();
+        assert!(masked.iter().any(|path| path == "/proc/kcore"));
+    }
```
DorianZheng, line 1381: these assert on values the builder hands out by default, not on anything this PR added. `LinuxBuilder::default().build()` on oci-spec 0.6.7 already returns the five readonly paths and the ten masked paths, the unprivileged branch never calls `.readonly_paths()` or `.masked_paths()`, and no branch anywhere sets masked paths. delete the whole `if privileged` block and the `masked_paths` vec, and lines 1230, 1236 and 1260 stay green.

the second test is named `only_releases_proc_sys` but never checks "only" — it looks for `/proc/sysrq-trigger` and stops. a regression that returned just `["/proc/sysrq-trigger"]` would pass.

G4614:
All nine addressed. The shape changed as a result, so it is easier to describe what the PR does now than to answer point by point.

**Privileged now clears both lists, matching `docker run --privileged`.** Your #8 is the reason. I checked it rather than argued it: in a `--cap-add ALL` box, `/proc/bus` starts `ro` and one `mount -o remount,rw /proc/bus` turns it `rw`. So the subset I was keeping stopped nothing, and the comment claiming otherwise was the worst part of it. A default container has no CAP_SYS_ADMIN, and there the same lists do hold — `mount: permission denied` — so the hardened path is unchanged.

I tried to make them enforceable instead. A user namespace is the mechanism that would do it, and the uid/gid mappings are already built here — only the `LinuxNamespaceType::User` line is commented out. Adding it fails in the guest:

```
build failed: received unexpected message: OtherError("cgroup error: systemd error:
systemd cgroup feature is required, but was not enabled during compile time"),
expected: WriteMapping
```

libcontainer routes a user namespace through the systemd cgroup manager, and `src/guest/Cargo.toml` builds it with `default-features = false, features = ["v2", "libseccomp"]`. That is a guest dependency question, not something this PR can settle, so it is out of scope here and worth its own issue.

**#9 stands and I could not fix it.** It is the whole `/proc/sys` tree, and it is the guest's, not the container's. I tried a network namespace for the privileged branch — it is created, and `/proc/sys/net/*` does become per-container — but the namespace comes up empty, with no veth and no route, so dockerd cannot reach a registry (`failed to resolve reference`). Wiring it into gvproxy is its own piece of work. Reverted; the commit message now says the guest-wide tree opens, without claiming otherwise.

**#1, #2** — `is_privileged` returns `BoxliteResult<bool>` and propagates the ceiling-read failure, since `resolve` treats it as fatal at line 107. On the keyword: I kept the resolved-set comparison and fixed the doc instead. Naming 27 capabilities and typing `ALL` leave the container equally powerful, and the sandbox should follow what it can do, not how the request was spelled — but you are right that the doc claimed something else.

**#3, #4** — both local lists are gone. Nothing consults them now, and as you noted the copy had already drifted from what `Linux::default()` actually ships.

**#5, #6, #7** — the tautological assert is gone. The gap you named in #6 is the one that mattered: `capability_set_decides_which_shape_the_spec_gets` now drives `create_oci_spec` from a real `CapabilitySet` and asserts on the spec it produces, so inverting the predicate at the call site fails the suite instead of passing it.

Verified end to end after the rework: Docker 29.6.2 in a box on this build pulls images and runs both single containers and a two-service compose project with working service DNS.

G4614:
Correction to my previous comment: I wrote that nothing filters `mount(2)`, and implied this repository has no seccomp. The second half was wrong.

There is a seccomp filter — `jailer::seccomp::apply_vmm_filter`, built from `resources/seccomp/*.json` and applied in `src/shim/src/main.rs` — but it wraps the **VMM process on a Linux host** and guards the host from the VMM. It says nothing about what a container may call inside the guest.

The container spec is the part with no seccomp: `create_oci_spec` never sets `linux.seccomp`, so `mount(2)` reaches the guest kernel unfiltered and the remount I demonstrated succeeds. (The `libseccomp` feature on libcontainer compiles the capability in; nothing in this tree emits a profile that would use it.)

So the conclusion holds, but the reasoning I gave for it was sloppy about which layer is protected. The commit message now states it precisely.

src/guest/src/container/spec.rs
G4614: Restored in d1f93962 — both vecs, their `#[allow(unused)]`, and the two commented-out builder lines are back to what main had. This PR should not have touched them.

The drift you note is real (`/sys/devices/virtual/powercap` is in the local copy but not masked at runtime; `/proc/scsi` is masked but absent from the copy — the effective set comes from `Linux::default()`), but it predates this change, so I have left it for a separate commit rather than widening this diff a second time.

src/guest/src/container/spec.rs
```diff
@@ -545,16 +550,42 @@ fn build_linux_spec(
...
+                .source("cgroup2")
+                .options(vec![
+                    "nosuid".to_string(),
+                    "noexec".to_string(),
+                    "nodev".to_string(),
+                ])
+                .build()
+                .map_err(|e| BoxliteError::Internal(format!("Failed to build cgroup2 mount: {e}")))
+        })
+        .transpose()?;
```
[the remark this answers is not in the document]
G4614: This PR does not mount cgroup2. `build_standard_mounts` has the `/sys/fs/cgroup` mount commented out (spec.rs:645-664, disabled for the ~105ms it costs at startup), and `build_linux_spec` leaves `cgroups_path` commented out with it. The privileged branch clears `maskedPaths` and `readonlyPaths` and touches nothing else.

The premise about the namespace is right — `build_default_namespaces` does give only pid/ipc/uts/mount — but with no cgroup mount there is no fresh writable hierarchy for a container to reach. Closing as not applicable to this change.

src/guest/src/container/spec.rs
G4614: Gone — the privileged branch no longer carries a copy of the list. It passes an empty vec, and the unprivileged branch keeps inheriting `Linux::default()` live from the crate, so there is no second copy to drift.

The drift you predicted was already there, in the `#[allow(unused)]` vec at the top of the function: it lists `/sys/devices/virtual/powercap`, which is not masked at runtime, and omits `/proc/scsi`, which is. That vec predates this PR and I have left it alone rather than widen the diff; worth a separate change that deletes the copy in favour of `get_default_maskedpaths()`.

src/guest/src/container/capabilities.rs
G4614: All four checked. Two fixed, one intended, one I want to push back on.

**`unwrap_or(false)` swallowing the error** — fixed. `is_privileged` returns `BoxliteResult<bool>` and the call site propagates.

**`--cap-add ALL --cap-drop ALL` coming out fully privileged** — confirmed and fixed in b3c3f45c. You had the mechanism exactly right: `parse_capabilities` filters the keyword out of both lists, so the additions branch ran its removal loop over an empty set. Reproduced on Linux before the fix:

```
---- container::capabilities::tests::all_on_both_sides_clears_the_set stdout ----
assertion `left == right` failed
  left: 41
 right: 0
```

The fix answers `ALL` in removals before `ALL` in additions. `--cap-drop ALL --cap-add NET_ADMIN` keeps working — same guard, same return — and `all_dropped_then_named_capabilities_added_back` holds that down.

**`--cap-add ALL --cap-drop SYS_MODULE` losing the relaxation** — intended. Privileged is defined as holding the whole ceiling, so one drop leaves it, the same way `docker run --privileged --cap-drop SYS_MODULE` is not privileged. What was wrong is that it happened silently: there is now an `info` at spec build naming the container and which shape it got.

**The `cap_last_cap` read on every container start** — I would rather not optimise this blind. It is one small procfs read plus a 41-element set build; the ~105ms the surrounding comments are about is the cgroup2 mount, a different operation. `resolve` skips the read only when add and drop are both empty, so the case that regresses is the default box. Happy to gate it behind that same condition if you want it gone, but I have no measurement showing it matters and did not want to add a branch on a guess.

src/guest/src/container/capabilities.rs
G4614: Right — that assert compared `supported_capabilities()` against itself. It is gone. `a_narrow_capability_request_is_not_privileged` now carries only the two negative cases you said had signal: the default set, and SYS_ADMIN on top of it.

The positive direction is asserted where it can actually fail: `capability_set_decides_which_shape_the_spec_gets` in spec.rs drives `create_oci_spec` from a real `CapabilitySet` and checks the resulting spec, and `all_on_both_sides_clears_the_set` pins a case that returned the full ceiling until b3c3f45c.

src/guest/src/container/spec.rs
G4614: Fixed. `capability_set_decides_which_shape_the_spec_gets` builds a real `CapabilitySet::resolve(&["ALL"], &[])`, runs it through `create_oci_spec`, and asserts on the spec that comes out; it does the same for `CapabilitySet::default()`. Inverting the predicate at the call site fails it.

Still true that there is no `--cap-add` integration test under src/cli/tests or src/boxlite/tests — that needs a live guest, so it is not something this unit suite can reach. I verified the end-to-end path by hand instead: dockerd starts, pulls over its own bridge, and a two-service compose project resolves service-name DNS inside a `--cap-add ALL` box.

src/guest/src/container/spec.rs
G4614: Tried it in a real privileged box, and you were right — the partial relaxation bought nothing. In a `--cap-add ALL` box:

```
/ # grep /proc/bus /proc/mounts
proc /proc/bus proc ro,nosuid,nodev,noexec,relatime
/ # mount -o remount,rw /proc/bus && grep /proc/bus /proc/mounts
proc /proc/bus proc rw,nosuid,nodev,noexec,relatime
```

The same call in a default container gives `mount: permission denied` — CapEff `a80425fb`, no CAP_SYS_ADMIN. So the entries hold for exactly the containers that were never the concern and fall over for the ones this branch targets.

Both lists are dropped wholesale now, and the comment says what actually holds — the microVM — instead of claiming the spec does. One correction to my own earlier reply on this: I said the repository has no seccomp, which was wrong. It ships a VMM filter (`jailer::seccomp::apply_vmm_filter`) that guards the host from the VMM on a Linux host. What is absent is a container-level profile: `create_oci_spec` never sets `linux.seccomp`, so nothing filters `mount(2)` inside the guest.

src/guest/src/container/spec.rs
G4614: Fair — the commit message named the one knob and the code comment inherited that framing. Both now describe the actual scope.

The comment says the container writes the guest's global sysctl tree rather than a private one, names IP forwarding flipping for every sibling container, and names core_pattern pointing at a program that runs as guest root outside the container. Your reading of `build_default_namespaces` is right: pid, ipc, uts, mount, and no network namespace.

I did try adding one. The namespace is created and the sysctls do become private, but it comes up empty — no veth, no route — so image pulls fail with `failed to resolve reference`. Wiring it needs a veth pair and NAT that this PR does not have, so I reverted it. Same for the user namespace, which is the thing that would make these protections enforceable rather than advisory: libcontainer 0.6 routes userns containers through the systemd cgroup manager, and the guest builds it without the `systemd` feature, so it fails with `systemd cgroup feature is required, but was not enabled during compile time`. Worth its own issue.

src/guest/src/container/spec.rs
G4614: The first half holds and is fixed; the second half is stale.

You were right that `unprivileged_spec_keeps_runtime_path_defaults` asserted on what the builder hands out by default — delete the `if privileged` block and it stayed green. It now pins both lists to their exact contents. Written out as literals rather than compared against `get_default_*`, which is where the value under test comes from and so would hold for any value including the empty one the test exists to catch. Its doc comment says what it guards: the premise, not the branch. Clearing a list only means something while leaving it unset yields a non-empty one, and oci-spec is a caret dependency, so an upstream release that emptied either list would turn the branch into a no-op that no assertion on the privileged side can see.

Spelling them out also caught the drift you flagged in the other thread from the other direction: the real masked set contains `/proc/scsi` and not `/sys/devices/virtual/powercap`, the reverse of the `#[allow(unused)]` copy above.

`only_releases_proc_sys` no longer exists — that was the earlier revision. The current tests are `unprivileged_spec_keeps_runtime_path_defaults`, `privileged_spec_clears_both_path_lists`, and `capability_set_decides_which_shape_the_spec_gets`.

src/guest/src/container/capabilities.rs
G4614: The doc was describing the wrong thing and now says what the code does: keyed on the resolved set, not on the `ALL` keyword.

The spelled-out case coming out privileged is deliberate, and I think it is the right way round. The spec shape should describe what a container can do, not how the request was phrased — a policy generator that emits 41 names has handed over the same power as one that typed `ALL`, and hardening one but not the other would be decided by spelling.

The `check_compatibility` asymmetry is real and I have left it alone: it compares raw add/drop names (advanced_options.rs:607-609) so the two forms are rejected as different policies for box reuse. That is in the host runtime, predates this change, and fixing it means resolving both sides before comparing — a separate change rather than a widening of this one.

src/guest/src/container/capabilities.rs
G4614: On the per-start procfs read — measured rather than argued. Release build, aarch64 guest, page cache warm (the steady state a container start meets; only the first container on a box reads cold), three runs:

| | per call |
|---|---|
| `ceiling_gap()` | 7.3 – 8.3 µs |
| of which the procfs read | 1.4 µs |
| of which the 41-element set build | ~6 µs |

Spread across runs under 15%.

Against the 105 ms cgroup mount that this file's own comments name as the thing worth optimising, that is about 1/14000th. Driving it to zero would save 0.007% of a cost already documented as the reason cgroups are off. A short-circuit would add a branch deciding when the read can be skipped, which is more to read and reason about than the 7 µs buys back — so I am leaving it, and `ceiling_gap` now answers both the privileged verdict and the shortfall from that one read rather than two.

The rest of your four: the discarded error and the `add=ALL drop=ALL` inversion are fixed (the latter in its own commit, with a reproducer verified to fail on the parent — `left: 41, right: 0`). The `drop=SYS_MODULE` cliff is real and stays by design, but no longer silent: a set that can mount and falls short of the ceiling now warns with the missing names, since the symptom is otherwise an EROFS that names nothing.

G4614: Correction to the comparison I just made, not the measurement. I anchored the 7.3 µs against the 105 ms cgroup mount, but that mount is commented out in `build_standard_mounts` — the 105 ms is a cost this code path *avoids*, not one it pays. Anchoring against it was meaningless.

A real denominator, same setup: end-to-end `boxlite run --rm alpine:3.20 true`, warm image, three runs — 2.96 s / 2.83 s / 2.62 s. So `ceiling_gap()` is roughly 1/370000 of what a caller waits for.

That figure covers microVM boot too, so container start proper is some slice of it rather than the whole. It holds as an upper bound either way: 7.3 µs cannot be a meaningful share of a slice of 2.7 s unless container start were itself microseconds, and it involves mounts, pivot_root and fork/exec.

Conclusion unchanged — the per-start read stays, and `ceiling_gap` serves both the verdict and the shortfall from one read. But the number I first reached for was the wrong one.

G4614: One more correction to my earlier measurement reply, found on a self-review pass.

"`ceiling_gap` answers both the verdict and the shortfall from that one read, so explaining a near miss costs no extra procfs work on a path that already runs per container start" — that's true of `ceiling_gap()` in isolation, but it undersells what's actually happening upstream.

`resolve()` (capabilities.rs:106-110) already reads `supported_capabilities()` whenever add/drop is non-empty — which per this file's own docs (`docs/architecture/container-capabilities.md:37-39`) is the recommended usage (`drop=["ALL"]` plus minimal additions). It uses that value for `ensure_supported` and then discards it. `service/container.rs:195` calls `resolve()`, and the same request's `create_oci_spec` (via `lifecycle.rs:105`) calls `ceiling_gap()` moments later, which reads the identical fact again from scratch.

So for any caller who customizes capabilities at all, this branch reads `/proc/sys/kernel/cap_last_cap` twice per container start, not once. The measured 7.3 µs stands as the cost of one `ceiling_gap()` call; it does not account for the pre-existing read in `resolve()` that this PR now duplicates rather than reuses.

I'm not restructuring `CapabilitySet` to thread the already-resolved ceiling through and avoid the second read — doing that means storing it on the struct or widening the return type of `resolve()`, for a savings still on the order of single-digit microseconds against a multi-second container start. That's added surface for a cost this small; happy to revisit if it's ever measured to matter.

src/boxlite/src/runtime/options.rs
```diff
@@ -542,6 +542,12 @@ impl Default for BoxOptions {
 }
 
 impl BoxOptions {
+    /// Expand high-level privileged mode into the explicit guest capability
+    /// policy before the options cross a runtime or SDK boundary.
+    pub fn normalize_privileged(&mut self) {
```
DorianZheng: no need to expose this function, make the user interface as simple as possible

src/boxlite/src/portal/interfaces/container.rs
```diff
@@ -87,12 +87,7 @@ pub struct ContainerInitConfig {
     pub tty: bool,
     /// Guest device nodes to reproduce inside the OCI workload.
     pub devices: Vec<ContainerDevice>,
-    pub advanced: ContainerAdvancedConfig,
```
DorianZheng, line 90: no need to change this

src/boxlite/src/runtime/rt_impl.rs
```diff
@@ -1158,11 +1169,14 @@ impl RuntimeImpl {
         self: &Arc<Self>,
         staging_dir: std::path::PathBuf,
         name: Option<String>,
-        options: BoxOptions,
+        mut options: BoxOptions,
         initial_status: BoxStatus,
     ) -> BoxliteResult<LiteBox> {
         use crate::litebox::config::ContainerRuntimeConfig;
 
+        options.advanced.validate_privileged_capability_conflict()?;
```
DorianZheng: move to options.sanitize

src/shared/proto/boxlite/v1/service.proto
```diff
@@ -385,6 +385,11 @@ message ContainerConfig {
 
 message ContainerAdvancedOptions {
   ContainerCapabilities capabilities = 1;
+
+  // Host-resolved OCI privileged spec shape for DinD. When true, the host
+  // sends capabilities.add=["ALL"] and an empty drop list; the guest must
+  // consume that canonical shape rather than reinterpret user overrides.
+  bool privileged = 2;
```
DorianZheng: Translate this on the host side, so that the protocol only includes atomic options for better backward compatibility.

src/boxlite/src/portal/interfaces/container.rs
G4614: 已恢复 ContainerInitConfig 原有的公开 ContainerAdvancedConfig 边界；高层 privileged 只在 host options 中解析，跨 guest 协议传递的是四个原子 OCI 选项。

src/shared/proto/boxlite/v1/service.proto
```diff
@@ -385,6 +385,10 @@ message ContainerConfig {
 
 message ContainerAdvancedOptions {
   ContainerCapabilities capabilities = 1;
+  bool cgroup_namespace = 2;
```
DorianZheng: what does this mean?

src/shared/proto/boxlite/v1/service.proto
```diff
@@ -385,6 +385,10 @@ message ContainerConfig {
...
+  bool writable_sysfs = 3;
```
DorianZheng: ditto

src/shared/proto/boxlite/v1/service.proto
```diff
@@ -385,6 +385,10 @@ message ContainerConfig {
...
+  bool allow_all_devices = 4;
```
DorianZheng: ditto

src/shared/proto/boxlite/v1/service.proto
```diff
@@ -385,6 +385,10 @@ message ContainerConfig {
...
+  bool unconfined_paths = 5;
```
DorianZheng: ditto

src/shared/proto/boxlite/v1/service.proto
G4614: 表示在 OCI namespaces 中加入独立的 cgroup namespace；已补协议注释。

src/shared/proto/boxlite/v1/service.proto
G4614: 表示让容器内的 /sys 递归 bind 保持可写，包括其中的 cgroup2 子挂载；已补协议注释。

src/shared/proto/boxlite/v1/service.proto
G4614: 表示添加允许所有设备 rwm 的 OCI device-cgroup 规则；已补协议注释。

src/shared/proto/boxlite/v1/service.proto
G4614: 表示移除 OCI 默认的 masked_paths 和 readonly_paths 限制；已补协议注释。

DorianZheng (approved):

PR #652, opened 2026-06-04
feat(security): default sandbox on + expose SecurityOptions to CLI/Go/C (REST stays server-policy-only)

G4614:
Default the sandbox **on** for Linux/macOS and expose a fine-grained `SecurityOptions` through the CLI, Go, and C surfaces. The REST wire stays deliberately closed — the **server** owns the sandbox policy; clients can't set it or relax it.

## What changed

**Core — `SecurityOptions` (`runtime/advanced_options.rs`)**
- `Default` flipped to the fully-enabled profile (secure by default). On Linux: jailer + seccomp + new PID ns + chroot + unprivileged uid/gid; on macOS: sandbox-exec. Previously Linux callers hitting the default ran an **unsandboxed** shim.
- Two named profiles: `SecurityOptions::enabled()` (== `default()`) and `SecurityOptions::disabled()` (explicit opt-out). Finer control = override individual fields (builder / per-field FFI setters) on top of a profile.
- `SecurityOptions::from_preset(name)` resolves the two profiles case-insensitively (`enable`/`enabled`/`on` vs `disable`/`disabled`/`off`); anything else is `InvalidArgument` echoing the offending value.

**CLI**
- `boxlite run|create --security <enable|disable>` (default `enable`). Parsed via `from_preset`; a typo aborts with an `anyhow` error carrying the bad value.

**Go SDK (`sdks/go/security.go`, `options.go`)**
- `NewSecurityOptions()` / `NewSecurityOptionsDisabled()` build a caller-owned `*SecurityOptions` (finalized + `Close()`), with ~28 `Set*/Clear*` field methods; attach via `WithSecurityOptions(spec)`.
- The old `WithSecurity(bool)` shortcut was **removed**.

**C SDK (`sdks/c/src/security.rs`, `options.rs`)**
- `CSecurityOptions` opaque handle: `boxlite_security_options_new` / `_new_disabled` / `_free`, ~28 per-field setters/clearers, attached via `boxlite_options_set_security(opts, handle)`.
- The `boxlite_options_set_security_{enabled,disabled}` two-state shortcuts were **removed** — superseded by the handle (per review).

**REST — server-policy-only (`rest/types.rs`, `commands/serve/types.rs`)**
- Client-side `CreateBoxRequest` is `Serialize`-only with **no** security field, so the wire form structurally can't carry `security` / `security_settings`.
- Server-side request type is `#[serde(deny_unknown_fields)]`: a client smuggling a security knob gets a **400**. An empty body resolves to the server's own `BoxOptions.advanced.security`.

## Structural cleanups (`SecurityOptions`)

- **Single source of truth for the default profile.** Per-field `#[serde(default = "...")]` had silently diverged from `impl Default`, so a partial/empty security block deserialized into a *weaker* sandbox than `default()` (uid unset, no resource limits, no new PID ns on Linux). Moved to **struct-level `#[serde(default)]`** — `{}` now equals `default()`, pinned by `deserializing_empty_equals_default` (two-side verified). Removed the now-dead `default_*()` helpers.
- **Platform-inert fields surfaced.** The struct is flat and mixes Linux-only and macOS-only knobs; `SecurityOptions::warn_inert_fields()` (called at the jailer apply boundary) now logs when a set field is ignored on the current platform instead of failing silently.
- **`network_enabled` → `sandbox_network_enabled`** (Rust/C/Go + regenerated header). It's cross-platform (Linux landlock + macOS seatbelt), not macOS-only as the old doc claimed, and the bare name collided with box-level `BoxOptions.network`. `#[serde(alias = "network_enabled")]` keeps existing JSON working, pinned by `sandbox_network_enabled_accepts_legacy_alias`.
- Doc fixes: dropped the "exactly two states" framing and the stale "jailer enabled on macOS, disabled on Linux" builder doc.

## Test pins

| area | test |
|---|---|
| default flip + serde single-source | `runtime::options::tests::test_box_options_defaults`, `..::deserializing_empty_equals_default` |
| legacy JSON alias | `runtime::options::tests::sandbox_network_enabled_accepts_legacy_alias` |
| CLI flag apply / absent / typo | `management_security_preset_applies_to_box_options`, `..._absent_leaves_default`, `..._typo_surfaces_anyhow_error` |
| REST wire never carries security | `rest::types::tests::test_create_box_request_never_carries_security_on_the_wire` |
| REST server-default + client-reject (400) | `build_box_options_empty_body_lands_on_server_default_security`, `create_box_request_rejects_client_supplied_security_preset`, `..._settings` |
| C handle apply | `set_security_applies_profile_handle_to_options` |
| Go | `TestBuildCOptions_SecurityEnabledDisabled`, `TestBuildCOptions_SecurityUnsetKeepsDefault`, `TestSecurityOptions_FieldSetters` |

## Notes

- There is **no** multi-tier preset (`development`/`standard`/`maximum`) and **no** `set_security_preset` string API on any surface — the only named profiles are `enable`/`disable`; everything finer is set field-by-field on the handle.


sdks/c/include/boxlite.h
```diff
@@ -519,6 +519,21 @@ void boxlite_options_set_auto_remove(CBoxliteOptions *opts, int val);
 
 void boxlite_options_set_detach(CBoxliteOptions *opts, int val);
 
+// Pick a sandbox security preset by name.
```
DorianZheng: Should be consistent with `boxlite_options_new`, `boxlite_options_set_rootfs_path`, etc.

G4614: Addressed in commit `7d3a61d7` (review: strip the security-preset doc block so the header matches sibling setters). The 14-line doc comment on the FFI wrapper was removed so the generated header line `void boxlite_options_set_security_preset(CBoxliteOptions *opts, const char *preset);` is now bare, matching sibling `boxlite_options_set_detach`, `boxlite_options_set_rootfs_path`, etc. Full preset description stays in the Rust source on the internal `options_set_security_preset` helper.

sdks/c/include/boxlite.h
```diff
@@ -537,6 +544,22 @@ void boxlite_options_set_auto_remove(CBoxliteOptions *opts, int val);
 
 void boxlite_options_set_detach(CBoxliteOptions *opts, int val);
 
+void boxlite_options_set_security_enabled(CBoxliteOptions *opts);
+
+void boxlite_options_set_security_disabled(CBoxliteOptions *opts);
```
DorianZheng: these two API can be removed then

G4614: Done — `boxlite_options_set_security_enabled` / `_disabled` were removed in `7008de74`. Security is now configured only via the `CSecurityOptions` profile handle attached through the advanced layer (`boxlite_advanced_options_set_security` → `boxlite_options_set_advanced`), matching core/Python `BoxOptions.advanced.security`.

PR #665, opened 2026-06-06
fix(exec): preserve UTF-8 codepoints split across attach chunks

ldarren:
`route_output` decoded each `ExecOutput` chunk independently with `String::from_utf8_lossy`, so a multi-byte UTF-8 char straddling a chunk boundary became two `U+FFFD` replacements (one per side  of the cut). For TUI workloads (pi-coding-agent, htop, ncdu) this desyncs the renderer's column math vs. what the client terminal sees, producing visible double-render / ghost-line artifacts during streaming output.

  ## Fix

  Hold the trailing partial codepoint across chunks via a per-stream `Utf8StreamDecoder` constructed inside `spawn_attach`. Flush any held bytes as `U+FFFD` on EOF to match prior `from_utf8_lossy` semantics for truncated tails.

  ## Tests

  7 new unit tests in `exec.rs` covering:
  - 3-byte and 4-byte codepoints split at every cut position
  - One-byte-at-a-time delivery
  - ASCII prefix + multi-byte tail
  - Definitively invalid bytes (still emit `U+FFFD`)
  - EOF flush
  - `route_output` integration shape (regression reproducer)

  Two-side verified: 5/7 fail without the fix, 7/7 pass with it. Full `make test:unit:rust` (725+38) green. `make test:unit:node` (69) green against the rebuilt native module.

## Summary by CodeRabbit

* **Bug Fixes**
  * Output streaming preserves multi-byte Unicode across message boundaries, avoids spurious or duplicate replacement characters, and reliably drains held bytes on shutdown, error, or normal end for both stdout and stderr.

* **Tests**
  * Expanded tests for boundary splits, invalid-byte replacement, EOF flushing, idempotent draining, and integration cases to prevent regressions.

### DIFF
--- a/src/boxlite/src/portal/interfaces/exec.rs
+++ b/src/boxlite/src/portal/interfaces/exec.rs
@@ -298,6 +298,12 @@ impl ExecProtocol {
                                     message_count,
                                     "Attach stream cancelled during shutdown"
                                 );
+                                Self::flush_decoders(
+                                    &stdout_tx,
+                                    &stderr_tx,
+                                    &mut stdout_decoder,
+                                    &mut stderr_decoder,
+                                );
                                 break;
                             }
                             msg = stream.message() => msg,
@@ -321,6 +327,16 @@ impl ExecProtocol {
                                     message_count,
                                     "Attach stream error, breaking"
                                 );
+                                // Flush before pushing the error message so
+                                // the held-over partial bytes (as U+FFFD)
+                                // arrive in correct order ahead of the
+                                // synthesized "Attach stream error: …" line.
+                                Self::flush_decoders(
+                                    &stdout_tx,
+                                    &stderr_tx,
+                                    &mut stdout_decoder,
+                                    &mut stderr_decoder,
+                                );
                                 let _ = stderr_tx.send(format!("Attach stream error: {}", e));
                                 break;
                             }
@@ -329,14 +345,12 @@ impl ExecProtocol {
                                 // bytes still in the decoders as U+FFFD,
                                 // matching `from_utf8_lossy` semantics for
                                 // a truncated input at EOF.
-                                let stdout_tail = stdout_decoder.flush();
-                                if !stdout_tail.is_empty() {
-                                    let _ = stdout_tx.send(stdout_tail);
-                                }
-                                let stderr_tail = stderr_decoder.flush();
-                                if !stderr_tail.is_empty() {
-                                    let _ = stderr_tx.send(stderr_tail);
-                                }
+                                Self::flush_decoders(
+                                    &stdout_tx,
+                                    &stderr_tx,
+                                    &mut stdout_decoder,
+                                    &mut stderr_decoder,
+                                );
                                 break;
                             }
                         }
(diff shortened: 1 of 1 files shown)


DorianZheng:
Hi @ldarren. Thanks for file this PR. Please sign the CLA.

DorianZheng:
`decode()` can still emit doubled `U+FFFD` when invalid bytes sit immediately before a codepoint that splits on the chunk boundary — the exact failure this PR targets, just gated behind a non-utf8 prefix.

the `error_len() == Some(n)` arm runs `from_utf8_lossy` over the *whole* remaining tail, so a trailing incomplete-but-valid sequence at the end of that tail gets flattened to `U+FFFD` instead of being held for the next chunk.

repro (replicated `Utf8StreamDecoder::decode` verbatim, `<FFFD>` = U+FFFD):

```
[0xFF, 0xE2] then [0x94, 0x80]      # garbage 0xFF, then "─" (E2 94 80) split across the boundary
  got:  <FFFD><FFFD><FFFD><FFFD>
  want: <FFFD>─

[0xFF,0xFF,0xF0,0x9F] then [0x91,0x8B]   # 2 garbage + "👋" split
  got:  <FFFD><FFFD><FFFD><FFFD><FFFD>
  want: <FFFD><FFFD>👋

[0xE2] then [0x94,0x80]             # clean split, no garbage — fine today
  got:  ─
```

the current tests cover clean splits and isolated invalid bytes, so this slips through. it matters for non-tui exec output (binary dumps, latin-1) where invalid bytes are common and any adjacent split corrupts a real char.

fix is to iterate the error path — emit one `U+FFFD` per definite error, hold only a final `error_len() == None` tail (also drops the now-dead `tail.len() < 4` guard):

```rust
fn decode(&mut self, chunk: &[u8]) -> String {
    let mut buf = std::mem::take(&mut self.holdover);
    buf.extend_from_slice(chunk);
    let mut out = String::new();
    let mut rest: &[u8] = &buf;
    loop {
        match std::str::from_utf8(rest) {
            Ok(s) => { out.push_str(s); break; }
            Err(e) => {
                let v = e.valid_up_to();
                out.push_str(unsafe { std::str::from_utf8_unchecked(&rest[..v]) });
                match e.error_len() {
                    Some(n) => { out.push('\u{FFFD}'); rest = &rest[v + n..]; } // definite error: 1 FFFD, keep scanning
                    None    => { self.holdover = rest[v..].to_vec(); break; }   // incomplete tail: hold for next chunk
                }
            }
        }
    }
    out
}
```

worth adding a test with an invalid byte immediately before a split codepoint to lock it in.

src/boxlite/src/portal/interfaces/exec.rs
```diff
@@ -449,6 +516,91 @@ impl ExecProtocol {
...
+    /// Returns the longest valid UTF-8 prefix as a String; bytes that form
+    /// the start of a possibly-incomplete codepoint at the end are held for
+    /// the next call.
+    fn decode(&mut self, chunk: &[u8]) -> String {
+        // Combine any held-over partial bytes with the new chunk. Allocation
+        // is unavoidable when holdover is non-empty; for the common case
+        // (clean chunk boundaries), holdover is empty and we still need to
+        // own the bytes to splice in any new partial tail.
+        let mut buf = std::mem::take(&mut self.holdover);
+        buf.extend_from_slice(chunk);
```
DorianZheng: the `buf` Vec is allocated and the whole chunk copied on every call — including the common case (empty holdover + fully-valid chunk) where it's pure overhead: that path then does `from_utf8(&buf)` + `.to_string()`, so two allocs and a full memcpy per frame. the comment calls the allocation unavoidable, but when `holdover` is empty you can decode `chunk` in place and only allocate a `Vec` for the ≤3-byte tail you actually hold:

```rust
let buf: Cow<[u8]> = if self.holdover.is_empty() {
    Cow::Borrowed(chunk)
} else {
    let mut b = std::mem::take(&mut self.holdover);
    b.extend_from_slice(chunk);
    Cow::Owned(b)
};
```

on a busy tui stream that's one alloc + one chunk-copy saved per message. (composes with the error-path fix from the other comment.)

src/boxlite/src/portal/interfaces/exec.rs
```diff
@@ -333,22 +374,48 @@ impl ExecProtocol {
         output: ExecOutput,
         stdout_tx: &mpsc::UnboundedSender<String>,
         stderr_tx: &mpsc::UnboundedSender<String>,
+        stdout_decoder: &mut Utf8StreamDecoder,
```
DorianZheng: `route_output` (5 args) and `flush_decoders` (4 args) thread the same four values — `stdout_tx`, `stderr_tx`, and the two decoders — and the attach loop re-stitches them at four call sites (one route + three exit-path flushes). each sender and its decoder are really one unit per stream; grouping the `(tx, decoder)` pairs so the loop holds one value per stream and calls `.route(output)` / `.flush()` would drop the arg threading and keep decode and flush together instead of split across two free fns. minor, but it's coupling that grows with every new exit path — this change already added three.

DorianZheng (commented):
two optional cleanups on the decoder; neither blocks the fix.

PR #595, opened 2026-05-26
feat(infra-local): BoxLite-based local dev stack (L1 + L2 + M5 native runner)

lilongen:
## Summary

Adds **`apps/infra-local/`** — a fully self-hosted BoxLite cloud-MVP
control plane that runs on a single Apple Silicon Mac. One command
(`make stack-up`) brings up the entire stack and lets a developer create
real microVM sandboxes through the dashboard — no AWS, no Docker daemon
for the application boxes; everything runs natively on M5.

Ships as `milestone/infra-local/v0.1.0`. Cold-start → working sandbox +
browser terminal (`root@boxlite:~#`) in ~80 s.

> Dogfood principle: the control-plane services run as **BoxLite boxes**
> (not docker-compose), so any BoxLite weakness is felt by the team
> immediately.

## Architecture

| Layer | What | Where |
|---|---|---|
| **L1 — infra-local** | 10 BoxLite microVM boxes: PostgreSQL, Redis, MinIO, OCI registry, Dex (OIDC), Caddy, Jaeger, pgAdmin, registry-ui, OpenTelemetry collector (+ one-shot minio-init) | libkrun microVMs on macOS Hypervisor.framework, orchestrated by the `boxlite_local` Python package |
| **L2 — control plane** | 4 native processes: NestJS API (`:3001`), Go Runner (`:3003`), Go Proxy (`:4000`), Vite Dashboard (`:3000`) | Native macOS arm64, driven by `make stack-*` |
| **L3 — user sandboxes** | N user-created Ubuntu/Alpine sandboxes | libkrun microVMs spawned by the L2 Runner in `~/.boxlite-runner/` |

The Runner runs **natively on M5** (HVF + libkrun) — single runner host.
Multi-host autoscaler testing (the old Lima route) is intentionally out
of scope here and parked in a separate worktree.

## What works end-to-end (verified)

- **Cold start**: `make stack-nuke && make stack-up` boots all services + auto-seeds + waits for the default snapshot — ~80 s.
- **OIDC login** via Dex (`admin@boxlite.dev` / `password`, plus a normal `test01@boxlite.dev`); API auto-creates user + Personal org + owner row on first login.
- **Sandbox lifecycle**: create → start → stop → destroy via dashboard or `POST /api/sandbox`.
- **Live terminal**: dashboard Terminal → Connect → real interactive shell inside the microVM (browser → Caddy → Proxy → Runner → microVM).
[...]

PR #668, opened 2026-06-07
fix(api): anchor OIDC-created user's Personal org to the default region

lilongen:
## Summary

On first OIDC login, `JwtStrategy.validate` auto-creates the user. It passed `personalOrganizationQuota` but **not** `personalOrganizationDefaultRegionId`, so the downstream chain:

`UserService.create` → `UserCreatedEvent` → `OrganizationService.handleUserCreatedEvent`

created the user's **Personal** organization with `defaultRegionId: undefined`. Any downstream caller that reads `organization.defaultRegionId` then fails for every OIDC-created user.

`AppService`'s admin-seed path already sets `personalOrganizationDefaultRegionId` (`app.service.ts`); this just brings the normal login path in line — a one-field change mirroring existing, proven code.

## Why a separate PR

This was originally bundled into #595 (infra-local). Per @DorianZheng's review ("why change this in this PR?") it's been pulled out, since it's an app-behavior fix unrelated to the local-dev stack.

## Test

Adds `apps/api/src/auth/jwt.strategy.spec.ts` asserting the create DTO carries the configured `defaultRegion.id`.

Two-sided verification (CLAUDE.md reproduce-before-fix):
- **Without the fix:** ❌ fails — `create()` is called with the DTO missing `personalOrganizationDefaultRegionId`.
- **With the fix:** ✅ passes.

```
nx test api --testPathPatterns="jwt.strategy.spec"
```

## Release Notes

* **Bug Fixes**
  * Fixed auto-created user initialization to ensure proper default region assignment during OIDC authentication, preventing potential downstream errors.

* **Tests**
  * Added test coverage for user auto-creation during authentication validation.

### DIFF
--- a/apps/api/src/auth/jwt.strategy.spec.ts
+++ b/apps/api/src/auth/jwt.strategy.spec.ts
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2025 Daytona Platforms Inc.
+ * Modified by BoxLite AI, 2025-2026
+ * SPDX-License-Identifier: AGPL-3.0
+ */
+
+import { Request } from 'express'
+import { JwtStrategy } from './jwt.strategy'
+import { UserService } from '../user/user.service'
+import { TypedConfigService } from '../config/typed-config.service'
+
+const DEFAULT_REGION_ID = 'region-default-id'
+const DEFAULT_ORG_QUOTA = { totalCpuQuota: 8 }
+
+function buildStrategy() {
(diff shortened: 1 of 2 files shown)


law-chain-hot:
Looks good to me.

For MVP, covering the OIDC first-login/new-user path is enough. We are not live yet, so we do not need a backfill for existing org rows.

Only remaining blocker is the CLA/email check before merge. @lilongen  Please try this link: http://cla-assistant.io/boxlite-ai/boxlite

law-chain-hot (approved):

PR #672, opened 2026-06-08
fix(api): un-quote orderBy column refs in sandbox lifecycle crons

lilongen:
## Summary

The sandbox auto-stop / auto-archive / auto-delete cron queries passed pre-quoted column refs (`sandbox."lastBackupAt"`, `activity."lastActivityAt"`) to `QueryBuilder.orderBy(...)`. TypeORM interprets the `alias.property` form and applies its **own** identifier quoting, so the already-quoted column gets double-escaped into invalid SQL and the cron crashes.

Fix: pass the plain `alias.property` and let TypeORM map + quote it. `CustomNamingStrategy` (extends `DefaultNamingStrategy`) preserves the camelCase column name, so the generated SQL is correct.

## Why a separate PR

Extracted from the infra-local stack PR (#595) per review — it's a general API fix to the lifecycle crons, unrelated to the local-dev stack.

## Verification

Verified against dev RDS (the in-VPC API box): the lifecycle crons run without the TypeORM crash. (No unit test added — a meaningful test requires a live Postgres + full TypeORM entity metadata; this codebase has no integration-test harness for the cron query builders.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed ordering behavior in sandbox lifecycle management operations, including auto-stop checks, auto-archive checks, auto-delete checks, and state synchronization to ensure proper sequencing of sandbox operations.

### DIFF
--- a/apps/api/src/sandbox/managers/sandbox.manager.ts
+++ b/apps/api/src/sandbox/managers/sandbox.manager.ts
@@ -117,7 +117,7 @@ export class SandboxManager implements TrackableJobExecutions, OnApplicationShut
             .andWhere('sandbox.pending != true')
             .andWhere('sandbox."autoStopInterval" != 0')
             .andWhere('activity."lastActivityAt" < NOW() - INTERVAL \'1 minute\' * sandbox."autoStopInterval"')
-            .orderBy('sandbox."lastBackupAt"', 'ASC')
+            .orderBy('sandbox.lastBackupAt', 'ASC')
             .limit(100)
             .getMany()
 
@@ -189,7 +189,7 @@ export class SandboxManager implements TrackableJobExecutions, OnApplicationShut
         .andWhere('sandbox.pending != true')
         .andWhere('sandbox."autoArchiveInterval" != 0')
         .andWhere('activity."lastActivityAt" < NOW() - INTERVAL \'1 minute\' * sandbox."autoArchiveInterval"')
-        .orderBy('sandbox."lastBackupAt"', 'ASC')
+        .orderBy('sandbox.lastBackupAt', 'ASC')
         .limit(100)
         .getMany()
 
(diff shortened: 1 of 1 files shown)


DorianZheng:
i don't think the double-escaping happens, so i'm not sure this fix is needed.

with the installed typeorm (0.3.27), order by isn't quoted at build time — the whole query gets one final pass through `replacePropertyNamesForTheWholeQuery`. that pass looks up the bare column name (`lastBackupAt`) in its map. for the already-quoted form `sandbox."lastBackupAt"`, the lookup key becomes `"lastBackupAt"` (quotes included), misses the map, and the token is left untouched. so it stays valid `sandbox."lastBackupAt"`, not double-quoted.

i replayed typeorm's exact regex (using its own `escapeRegExp`) on both forms:

- `sandbox."lastBackupAt"` -> `sandbox."lastBackupAt"` (verbatim, valid)
- `sandbox.lastBackupAt`   -> `"sandbox"."lastBackupAt"` (mapped, valid)

both are valid sql. same story for the `activity."lastActivityAt"` ones.

corroborating: the `where`/`andWhere` clauses in these same queries already use quoted refs (`sandbox."autoStopInterval" != 0`, `activity."lastActivityAt" < ...`) and go through the identical final pass — if quoting crashed, those would crash first, and the pr doesn't touch them.

the pr verified the new code runs but didn't show the old code crashing. could the dev RDS crash have come from something else (a join / the stream query) rather than order by? if you still have the failing sql or stack trace, drop it here — that'd settle it. otherwise this looks like a cosmetic no-op.

lilongen:
Closing this — resolving it via dependency pinning instead of an app-source change.

**Root cause (thanks @DorianZheng — your skepticism was right):** the quoted `orderBy` is *not* a no-op-vs-crash issue on the version we actually run. dev/prod build typeorm **0.3.27/0.3.28**, where the DISTINCT-subquery pagination trigger is `(skip || take) && joinAttributes > 0`. The lifecycle crons use `.limit(100)` (not `.take()`), so they take the plain path and `ORDER BY sandbox."lastBackupAt"` is valid SQL — exactly as you predicted. The crash I originally hit was on **0.3.29**, which widened that trigger to also include `offset || limit`, newly routing these queries through `createOrderByCombinedWithSelectExpression` (which can't resolve the quoted property name → `undefined.databaseName`).

Why the version differed: `apps/yarn.lock` is gitignored, so the floating `^0.3.20` range re-resolved per environment — my local picked 0.3.29, CI built 0.3.28.

**Fix instead:** pin `apps/package.json` to `typeorm@0.3.28`, aligning local with prod and stopping the caret from drifting to the regressed 0.3.29. No apps/api change needed.

Note: the `column ... does not exist` errors currently firing on dev are a separate schema-drift issue, unrelated to orderBy.

The un-quote is still correct hygiene, so if/when we intentionally move to 0.3.29+ we should reapply it first — but that's not needed now.

PR #674, opened 2026-06-08
fix(exec): type-driven command-not-found → 422, classified at the source

lilongen:
## Problem

`box.exec("nonexistent")` returned a nested HTTP 500. A missing or non-executable command is a **user error, not a platform fault** — it must not be a 5xx that false-alarms SRE.

## Why it was non-trivial

youki's workload validator already rejects these with a typed `ExecutorValidationError::ArgValidationError`, but youki **stringifies the error at its own init→main process boundary** (`exec_failed(err.to_string())` → `Message::ExecFailed(String)`). By the time `ContainerBuilder::build()` returns, a user-command error is indistinguishable from a genuine platform failure except by message text — so the host was string-matching youki's volatile wording.

## Fix — classify at the source, no string matching

- **New `ValidatingExecutor`** (a youki `Executor`) wraps `DefaultExecutor`, intercepts the typed `ArgValidationError` from the *real* validator — which runs in the correct mount namespace / FS view, so **zero false-reject risk** — and writes one byte to an `O_CLOEXEC|O_NONBLOCK` pipe created in the zygote.
- **`do_build`** injects it via `.with_executor()`, reads the pipe after a failed `build()`, and tags the outcome with a typed `BuildFailureKind` (`UserCommandError` | `Platform`). A lost signal **fails safe toward `Platform`/500** — it can never downgrade a real platform fault.
- **`Zygote::build`** maps the kind to `BoxliteError::Execution` (422) vs `Internal` (500); the guest carries it across gRPC as a stable `REASON_USER_COMMAND_ERROR` token (shared constant); the host keys 422 off that token, never off youki internals.

Net: host and SDK are fully typed end-to-end; youki coupling is confined to a single guest-side interception.

## Verification

- host + guest (linux-musl cross) compile and clippy clean.
- **Two-side reproducer on a real microVM**: with all production reverted the new tests **FAIL** with `Internal("...not found in $PATH")` (500) pointing at the bug; with the fix they **PASS** with `Execution`/422 — for both command-not-found and not-executable.

7 files, +288/−16.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command execution error handling so invalid or non-executable commands are reported as execution errors instead of generic internal failures.
  * Preserved internal error classification for platform and container startup failures.
  * Added consistent error details for commands that cannot be found or run.
  * Improved error classification for command validation failures without relying on error-message text.

* **Tests**
  * Added coverage for missing and non-executable command scenarios.

### DIFF
--- a/src/guest/src/container/validating_executor.rs
+++ b/src/guest/src/container/validating_executor.rs
@@ -1,0 +1,87 @@
+//! A youki `Executor` that classifies user-command errors at the source.
+//!
+//! ## Why this exists
+//!
+//! When `box.exec("nonexistent")` runs, youki's `DefaultExecutor::validate()`
+//! rejects it with `ExecutorValidationError::ArgValidationError` (program not
+//! found / not executable / no PATH). That error is *typed* — but it dies at
+//! youki's own init→main process boundary: `container_intermediate_process.rs`
+//! does `exec_failed(err.to_string())`, flattening every init failure into a
+//! single `Message::ExecFailed(String)`. By the time `ContainerBuilder::build()`
+//! returns to us, the validation error is indistinguishable (same
+//! `LibcontainerError` variant) from a genuine platform failure (cgroup, mount,
+//! seccomp). The only surviving signal is the message *text*.
+//!
+//! Rather than string-match youki's volatile wording on the host, we intercept
+//! the error *where it is still typed* — inside the container init process,
+//! right after `DefaultExecutor::validate()` returns it — and smuggle a single
+//! byte out through an fd side-channel that bypasses youki's lossy channel. The
+//! zygote reads that byte and tags the build outcome with a typed
+//! [`BuildFailureKind`](super::zygote::BuildFailureKind), which then rides our
+//! own IPC and proto `reason` token up to the host as `BoxliteError::Execution`.
+//!
+//! ## fd lifetime
+//!
+//! `validate()` runs *after* pivot_root but *before* execvp (see the trait doc
+//! and youki `init/process.rs`). The write fd is created in the zygote, set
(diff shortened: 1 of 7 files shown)


src/guest/src/container/validating_executor.rs
```diff
@@ -0,0 +1,87 @@
...
+use libcontainer::workload::default::DefaultExecutor;
+use libcontainer::workload::{Executor, ExecutorError, ExecutorValidationError};
+
+/// Wraps youki's `DefaultExecutor`, intercepting `validate()` to flag
+/// user-command errors over an out-of-band fd. `exec()` / `setup_envs()`
+/// delegate unchanged. `Clone` is required because youki clones the executor
+/// across the fork boundary (`builder_impl.rs`); `RawFd` is `Copy`, so a derive
+/// suffices and the `CloneBoxExecutor` blanket impl handles the trait object.
+#[derive(Clone)]
+pub(crate) struct ValidatingExecutor {
```
DorianZheng, line 47: why name it as `ValidatingExecutor`?

lilongen: It's named after the youki hook it overrides: `Executor::validate()`. `exec()` and `setup_envs()` delegate to `DefaultExecutor` unchanged; only `validate()` is wrapped, so that an `ArgValidationError` (program not found / not executable) writes one byte to the zygote's signal pipe before youki flattens the error to a string.

Agreed the name reads as "adds validation", which it doesn't — it reports the result of youki's existing validation. Happy to rename to something that says that, e.g. `ArgErrorSignalingExecutor` (or `ClassifyingExecutor`, matching the module doc). Which do you prefer?

(Also rebased onto current main; the conflict with the new `build_with_default_sigpipe` / `do_build_init` path is resolved so init builds keep youki's default executor and stay classified as platform failures.)

PR #676, opened 2026-06-08
fix: reject unsupported parameters when creating a box

lilongen:
## What

Creating a box with an unsupported parameter — a typo, or a field a layer
doesn't accept — used to drop it silently. Now it fails fast with a clear
error that names the offending field, so mistakes surface instead of being
ignored.

## Changes

- **Node SDK (`SimpleBox`)** — the constructor validates option keys against
  the known set before touching the runtime; unknown keys throw a `TypeError`
  listing them.
- **REST API** — the box/sandbox creation endpoints reject unknown body fields
  with `400`. Scoped to those handlers; the global validation pipe is left
  unchanged to keep the blast radius small.
- **`boxlite serve`** — already rejected unknown fields; added a regression
  test to lock that behavior in.

Python, Go, and C SDKs already reject unknown parameters (named params /
compile-time checks) and are unchanged.

## Testing

Unit and integration tests on each touched boundary, each confirmed to fail
without the change and pass with it. Also exercised end-to-end over HTTP
(`boxlite serve` plus REST clients) and against the built native module.

## Summary by CodeRabbit

* **New Features**
  * Box and sandbox creation endpoints now enforce stricter validation, rejecting unsupported request fields with detailed error messages.
  * Node SDK constructor validates options at initialization and rejects unsupported settings with clear error feedback.
  * Improved input validation across API and SDK boundaries prevents silent failures and provides better error messaging.

### DIFF
--- a/sdks/node/lib/simplebox.ts
+++ b/sdks/node/lib/simplebox.ts
@@ -252,6 +252,38 @@ export interface SimpleBoxOptions {
 }
 
 /**
+ * Every key the SimpleBox constructor accepts. Used to reject unknown options
+ * at the boundary instead of silently dropping them (napi ignores extra JS
+ * keys at runtime). Keep this in sync with the SimpleBoxOptions interface.
+ *
+ * Subclasses (CodeBox/BrowserBox/ComputerBox/SkillBox/InteractiveBox) strip
+ * their own extra keys before calling super(), so this set never rejects
+ * legitimate subclass usage.
+ */
+const KNOWN_SIMPLE_BOX_OPTION_KEYS: ReadonlySet<string> = new Set([
+  "image",
+  "rootfsPath",
+  "memoryMib",
+  "cpus",
+  "diskSizeGb",
+  "runtime",
+  "name",
+  "autoRemove",
+  "reuseExisting",
+  "detach",
+  "workingDir",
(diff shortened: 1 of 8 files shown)


sdks/node/lib/simplebox.ts
```diff
@@ -251,6 +251,38 @@ export interface SimpleBoxOptions {
...
+/**
+ * Every key the SimpleBox constructor accepts. Used to reject unknown options
+ * at the boundary instead of silently dropping them (napi ignores extra JS
+ * keys at runtime). Keep this in sync with the SimpleBoxOptions interface.
+ *
+ * Subclasses (CodeBox/BrowserBox/ComputerBox/SkillBox/InteractiveBox) strip
+ * their own extra keys before calling super(), so this set never rejects
+ * legitimate subclass usage.
+ */
+const KNOWN_SIMPLE_BOX_OPTION_KEYS: ReadonlySet<string> = new Set([
```
DorianZheng, line 263: this is not a good idea, it's error prone. can we conclude support options automatically?

DorianZheng (changes requested):

DorianZheng:
This PR is error prone. It makes options hard to maintain between known and defined.

PR #680, opened 2026-06-08
fix(runner): make config getters nil-safe before initialization

G4614:
runner cannot panic while getting a nil config

## Tests
- go test ./cmd/runner/config

## Bug Fixes
- Improved stability by ensuring configuration accessor functions safely handle cases where the configuration is not yet initialized, with graceful fallbacks to environment variables for runtime, network, and build engine settings.

## Tests
- Added comprehensive test coverage for configuration fallback behavior when the configuration is uninitialized, including tests for default value handling.

### DIFF
--- a/apps/runner/cmd/runner/config/config.go
+++ b/apps/runner/cmd/runner/config/config.go
@@ -137,17 +137,32 @@ func (c *Config) GetOtelHeaders() map[string]string {
 }
 
 func GetContainerRuntime() string {
+	if config == nil {
+		return os.Getenv("CONTAINER_RUNTIME")
+	}
 	return config.ContainerRuntime
 }
 
 func GetContainerNetwork() string {
+	if config == nil {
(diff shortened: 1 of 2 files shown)


G4614:
Superseded by #724 — same content but as a same-repo PR (not from fork) so the e2e-cloud workflow's OIDC handshake can actually run. GitHub strips `id-token: write` permission from workflow runs triggered by PRs from forks, which made #680's CI structurally unable to pass.

G4614:
Added the runner config nil-safe fix here per request to reuse PR #680.

What changed:
- `GetEnvironment`, `GetContainerRuntime`, `GetContainerNetwork`, and `GetBuildEngine` no longer panic when the package-level runner config has not been initialized.
- Added config tests covering env/default fallback before `GetConfig()`.

Validated locally:
- `go test ./cmd/runner/config`

PR #682, opened 2026-06-09
fix(cloud): reserve executor env for user requests only

G4614:
BOXLITE_EXECUTOR env setting
reject requests from API create/recover, API REST exec proxy, runner HTTP create/recover/exec, and v2 job payloads
accept requests from runner inside / Go SDK

## Tests
- yarn nx test api --testFile=reserved-env.util.spec.ts --testFile=box.service.box-id.spec.ts --testFile=boxlite-proxy.controller.spec.ts --runInBand
- go test ./pkg/boxlite ./pkg/api/controllers ./pkg/runner/v2/executor
- cargo build -p boxlite-c
- bash scripts/build/build-guest.sh --profile debug --dest-dir /tmp/boxlite-pr682-runtime-debug
- SKIP_GUEST_BUILD=1 bash scripts/build/build-runtime.sh --profile debug --dest-dir /tmp/boxlite-pr682-runtime-debug
- BOXLITE_RUNTIME_DIR=/tmp/boxlite-pr682-runtime-debug go test -tags boxlite_dev ./pkg/boxlite -run TestIntegrationReservedExecutorEnvPolicyWithRealVM -count=1 -v

PR #685, opened 2026-06-09
fix(ffi): drain stdout/stderr pumps before Wait terminal event

G4614:
Split-out of #681 — **PR D of 4** (FFI exec drain race fix).

**Stacks on #684** (which stacks on #683 → #682). Diff vs main is large until prereqs land; once they do, this collapses to **1 file, +304/-43**.

## Bug

Go SDK's \`box.Exec\` occasionally returns with stdout chunks still in flight — the caller's callback got \`OnExit\` (or the wait gRPC reply) before the matching \`OnStdout\`/\`OnStderr\` chunks landed. From the caller's POV, the exec had finished but the last few bytes were missing.

Filed as #563. Repros deterministically with the existing \`test_p0_6_exec_stdout_race.py\` regression test (~90% loss against stock 0.9.5).

## Root cause

\`execution_wait\` and \`exit_pump\` in \`sdks/c/src/exec/execution.rs\` pushed their terminal events as soon as the underlying process exited, racing the still-draining stream pumps. The FFI queue is ordered, but wait/exit events were entering ahead of late stdout chunks.

## Fix

Track a \`streams_pending\` count on \`ExecutionHandle\` instead of the previous per-pump receiver list. Both \`exit_pump\` and the \`execution_wait\` task now await \`streams_pending → 0\` (via \`tokio::sync::Notify\` + standard register-then-check pattern to avoid post-notification miss) before pushing terminal events.

Deterministic ordering tests added inline in execution.rs:

- stdout-only: all chunks land before Exit
- mixed stdout/stderr: every chunk lands before Wait
- process exit before pump completion: terminal event waits
- pump completion before process exit: no spurious wait

## xfails

**No PR-B xfail flips** in this PR. The cases this fix unblocks (\`test_p0_6_exec_stdout_race\`) were pre-existing tests, not part of PR-B's 10 xfails. Those 10 (DTO validation, quota, spawn-failed mapping) have separate root causes.

## Stack

| | what |
|---|---|
| A (#682) | reorganize (rename + wiring) |
| B (#683) | 21 new e2e cases (xfails pin bugs) |
| C (#684) | runner exec error mapping fix + flip 1 xfail |
| **D (this)** | **FFI exec drain race fix** |

### DIFF
--- a/.github/workflows/e2e-stack.yml
+++ b/.github/workflows/e2e-stack.yml
@@ -18,9 +18,16 @@ on:
     branches: [main]
     paths:
       - 'sdks/c/src/exec/**'
+      - 'sdks/c/tests/e2e/**'
       - 'sdks/python/src/**'
+      - 'sdks/python/tests/e2e/**'
+      - 'sdks/go/src/**'
+      - 'sdks/go/tests/e2e/**'
+      - 'sdks/node/src/**'
+      - 'sdks/node/tests/e2e/**'
       - 'src/boxlite/src/rest/**'
       - 'src/boxlite/src/cli/**'
+      - 'src/cli/tests/e2e/**'
       - 'apps/runner/**'
       - 'apps/api/src/**'
       - 'scripts/test/e2e/**'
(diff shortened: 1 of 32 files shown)


G4614:
Superseded by #563 — same bug, Go-only fix (folds drain into Execution.Wait, mirroring os/exec's awaitGoroutines step inside Wait()). Keeps the C FFI Wait task decoupled from stream drain instead of coupling them at the FFI layer.

PR #687, opened 2026-06-09
fix(rest): wire `BoxOptions.network` from CreateBoxDto to lower-layer sandbox

G4614:
`box = rt.create(BoxOptions(network=NetworkSpec(mode='disabled')))` over REST silently turned into "no network restriction" — the SDK serialised `network` into the request body, but the NestJS CreateBoxDto didn't include the field, so class-validator stripped it before the mapper saw anything. Adds the field + maps `mode/allow_net` to the lower-layer `networkBlockAll/networkAllowList` that SandboxService and the runner already understand.

Includes a fixup (commit 2) that filters non-CIDR entries out of `allow_net` before passing to the sandbox-side validator, because `validateNetworkAllowList` rejects hostnames as "Invalid IP address" — without this, callers who pass hostnames (per SDK docs) saw their box create call 400 instead of the previous silent no-op.

## Test plan — two-sided verified

Pin: `scripts/test/e2e/cases/test_network_allow_net.py`

Stack: local e2e. Only API restart needed (mapper + DTO change).

| Case | Pre-fix (main) | Post-fix (this PR) |
|---|---|---|
| `test_network_mode_disabled_blocks_all` (`wget https://example.com` with `mode='disabled'`) | exit 0 — filter wasn't applied at all | **nonzero** — `network=disabled` enforced |
| `test_allow_net_lets_listed_host_through` (`wget example.com` with `allow_net=['example.com']`) | exit 0 (by accident — no filter at all) | exit 0 (hostname dropped at mapper, no filter — semantically same outcome, no regression) |
| `test_allow_net_blocks_unlisted_host` (`wget captive.apple.com` with `allow_net=['example.com']`) | exit 0 — nothing blocked | exit 0 — still unblocked because hostname `allow_net` has no path through the sandbox model |

The third case is out of scope: BoxOptions accepts hostnames in `allow_net` (per SDK docs); the sandbox-side `networkAllowList` only takes CIDRs. Wiring hostname → CIDR (DNS resolve, or hostname filter at the runner egress) is an architectural extension, not a DTO fix. Documented as known gap; this PR closes the `mode='disabled'` and CIDR cases.

Commit 1: DTO + mapper.
Commit 2 (fixup): filter non-CIDR before forwarding, restores pre-fix behaviour for hostname inputs (no regression).

### DIFF
--- a/apps/api/src/boxlite-rest/dto/create-box.dto.ts
+++ b/apps/api/src/boxlite-rest/dto/create-box.dto.ts
@@ -4,7 +4,24 @@
  * SPDX-License-Identifier: AGPL-3.0
  */
 
-import { IsOptional, IsString, IsNumber, IsBoolean, IsObject, IsArray, Min } from 'class-validator'
+import { IsOptional, IsString, IsNumber, IsBoolean, IsObject, IsArray, IsIn, ValidateNested, Min } from 'class-validator'
+import { Type } from 'class-transformer'
+
+// Mirrors `BoxOptions.network` on the SDK side. Two simple knobs:
+//   mode        — "enabled" lets allow_net through, "disabled" blocks all
+//   allow_net   — host (and CIDR) allow-list, semantics enforced by gvproxy
+// The API translates this pair to the lower-layer sandbox.networkBlockAll
+// + sandbox.networkAllowList (CSV) in createBoxToCreateSandbox.
+export class CreateBoxNetworkDto {
+  @IsString()
+  @IsIn(['enabled', 'disabled'])
+  mode: 'enabled' | 'disabled'
+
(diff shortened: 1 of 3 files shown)


G4614:
Closing: the gap this PR targeted is already closed on `main`.

`apps/api/src/boxlite-rest/mappers/box-to-box.mapper.ts:70-73` now reads `dto.network.outbound` and derives `networkBlockAll` / `networkAllowList` from it, so `BoxOptions.network` does reach the lower-layer sandbox. That landed via the outbound/inbound split (#996, #1199, #1206) rather than this branch.

Reopen if there is a path through `CreateBoxDto` that still drops the field.

PR #697, opened 2026-06-09
fix(disk): build ext4 rootfs from trees with unreadable (0000) files

DorianZheng:
## Problem

`create_ext4_from_dir` runs `mke2fs -t ext4 -d <src>` as the current user. e2fsprogs opens every source file with `open(O_RDONLY)`, so an **unreadable file (mode `0000`)** — e.g. `/etc/gshadow` in RHEL UBI images like `minio/minio`, `minio/mc` — aborts the unprivileged build:

```
do_write_internal: Permission denied while opening "gshadow" to copy
mke2fs: Permission denied while populating file system
```

Under POSIX DAC the permission classes are mutually exclusive: once the caller matches the **owner** class, only the owner bits are consulted — so the owner of a `0000` file **cannot read it**. Running as root bypasses the check, which is why this only bites unprivileged builds. The image must still record the original `0000` mode.

## Fix (unprivileged path only; root path untouched)

1. **Widen** owner read/search on unreadable entries before `mke2fs`, recording each original full `st_mode`. `chmod` is authorized by *ownership* (not the read bit), and unprivileged OCI extraction leaves every file owned by the current user, so this always succeeds. Walk is **top-down** — a `0000` directory must be widened before it can be listed.
2. **Restore the real mode in the image inode** by extending the existing debugfs pass with `sif <path> mode <orig>` (mke2fs records the *widened* mode, so the image needs correcting). Reuses the `sif … mode 0100555` form already used by `inject_file_into_ext4`.
3. **Restore source-tree modes** afterward (hygiene; after debugfs re-walks the source).

`fix_ownership_with_debugfs` → `normalize_inodes_with_debugfs` since it now normalizes ownership **and** mode.

## Why this approach

Surveyed real tools:

| Tool | Unprivileged FS-image strategy |
|---|---|
| rootless docker / podman / containerd | extract as mapped-root in a userns, keep layers as overlay mounts — **never flatten-read**, so never hit this |
| umoci | flattens to a plain rootfs, keeps the **real mode on disk** (virtualizes only uid/gid via xattr) |
| mkosi / buildah | run `mkfs` inside `unshare(CLONE_NEWUSER)` — hard-fails `EPERM` where unprivileged userns is disabled (Ubuntu 24.04 default, hardened/containerized hosts) |

boxlite flattens to a **real ext4 with no runtime FUSE layer**, so the real `0000` belongs in the inode — the umoci model. This fix needs **zero kernel-feature dependency**, unlike the userns route, which matters for boxlite's container/SSH audience.

## Tests

- `create_ext4_preserves_unreadable_file_mode` — end-to-end: builds the image from a tree with a `0000` `/etc/gshadow`, asserts via `debugfs stat`/`cat` that the image keeps mode `0000` and full content (data crosses the mke2fs+debugfs boundary). Binary-gated; skips cleanly without the runtime or when run as root.
- `widen_unreadable_owner_handles_zero_mode_dir_and_file` — pure unit test for the top-down `0000`-directory ordering (no binaries).

**Two-side verification:** with the production change reverted (test only), the e2e test FAILS with the exact `mke2fs ... while opening "gshadow" to copy` signal; with the fix it PASSES. `disk::` suite green (57 passed); `cargo fmt` + `clippy` clean.

> Note: the local `full-test-matrix` pre-push hook was skipped — it runs `make test` → `make runtime:debug`, which needs the full runtime (recursive submodules + libkrunfw + guest cross-compile) not assembled in this worktree. CI runs the full matrix here.

## Summary by CodeRabbit

* **Bug Fixes**
  * ext4 image creation now handles unreadable-owner files/dirs when run unprivileged: images preserve original modes and contents; inode ownership inside images is normalized to root; normalization failures during image finalization now abort the build, while restoring host permissions is attempted bottom-up and failures are logged.

* **Tests**
  * Added end-to-end and regression tests for unreadable files/dirs and nested-restoration ordering.
  * Added unit test for temporary permission widening and recorded restorations.

### DIFF
--- a/src/boxlite/src/disk/ext4.rs
+++ b/src/boxlite/src/disk/ext4.rs
@@ -85,6 +85,134 @@ fn calculate_disk_size(source: &Path) -> u64 {
     final_size
 }
 
+/// A source entry whose owner-permission bits were temporarily widened so
+/// `mke2fs -d` could read it, with everything needed to restore the original.
+struct WidenedPerm {
+    /// Absolute path inside the ext4 image, e.g. `/etc/gshadow`.
+    ext4_path: String,
+    /// Path on the host source tree (to restore the source mode afterward).
+    source_path: PathBuf,
+    /// Original full `st_mode` (incl. the `S_IFMT` type bits) for `sif … mode`.
+    mode: u32,
+}
+
+/// Temporarily grant the owner read (and search, for directories) on entries the
+/// unprivileged owner cannot otherwise read, so `mke2fs -d` can copy them.
+///
+/// e2fsprogs opens every source file as the calling user; a `0000` file (e.g.
+/// `/etc/gshadow` in RHEL UBI images) is denied because POSIX consults only the
+/// owner-class bits, which have no read bit. `chmod` is authorized by *ownership*
+/// (not the read bit), and unprivileged OCI extraction leaves every file owned by
+/// the current user, so the widen always succeeds. The original modes are returned
+/// so the caller can restore them — both on the source tree and, authoritatively,
+/// inside the image via debugfs: `mke2fs` records the *widened* mode, so the image
+/// must be corrected afterward.
+///
+/// Walks top-down: a `0000` directory cannot be listed until its own owner
+/// read+search bits are restored, so each directory is widened before descent.
+fn widen_unreadable_owner(source: &Path) -> BoxliteResult<Vec<WidenedPerm>> {
+    let mut widened = Vec::new();
+    widen_dir_recursive(source, source, &mut widened)?;
+    Ok(widened)
+}
+
+fn widen_dir_recursive(
+    source_root: &Path,
+    dir: &Path,
+    widened: &mut Vec<WidenedPerm>,
+) -> BoxliteResult<()> {
+    use std::os::unix::fs::MetadataExt;
+
+    // A directory needs owner read+search (0o500) before we can list it.
+    let dir_mode = std::fs::symlink_metadata(dir)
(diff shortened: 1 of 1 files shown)


DorianZheng:
Addressed all 4 review findings in `2b45a565` (single file, `disk/ext4.rs`):

1. **Unconditional source-mode restore** — replaced the free `restore_source_modes` with an RAII `SourceModeGuard` whose `Drop` restores the widened entries. It's constructed before the fallible `mke2fs`/`debugfs` calls, so cleanup now runs on every early-return path, not just success. `widen_unreadable_owner` appends into the guard-owned `Vec`, so even a partial widen failure still restores.
2. **Bottom-up restore** — the guard restores `self.widened.iter().rev()` (children before parents), fixing the EACCES-on-child case. Added reproducer `create_ext4_restores_nested_unreadable_source_modes`: it **fails on the old top-down order** (child left at `0o400`) and passes now — two-side verified (pre-fix `left: 256`, post-fix ok).
3. **Fail-fast debugfs** — `normalize_inodes_with_debugfs` now returns `Err` (operation + exit code + image path + stderr) on a non-zero exit instead of `warn` + `Ok`, since it's the only pass that writes the original `0000` modes into the image.
4. **Test asserts subprocess success** — the e2e test now checks `debugfs stat`/`cat` `.status.success()` (surfacing stderr) before parsing stdout.

`disk::ext4` 8 passed, `disk::` 58 passed; fmt + clippy clean.

law-chain-hot (approved):
LGTM

PR #705, opened 2026-06-09
fix(ffi): drain stdout/stderr pumps before Wait terminal event

G4614:
Re-opened replacement for the previously-closed PR (force-push made the old head commit unreachable; new commit on the same branch).

## Bug

Go SDK's box.Exec occasionally returned with stdout chunks still in flight — the user's callback got OnExit (or the wait gRPC reply) before the matching OnStdout/OnStderr chunks landed on the queue. From the caller's perspective the exec had finished but its stdout was silently truncated.

## Fix

Make exit_pump the sole owner of terminal-event dispatch. Both execution_wait and register_exit fan into it; exit_pump awaits all stream pump receivers before pushing the terminal event. Queue order becomes Stdout/Stderr* -> Exit -> Wait* — all from the same task.

## Test plan

Pin: test_p0_6_exec_stdout_race (scripts/test/e2e/cases/, from #678).

| | Pre-fix | Post-fix |
|---|---|---|
| test_p0_6_exec_stdout_race (stdout loss rate over 10 short execs) | ~70% loss | **0% loss** |

Replaces an earlier split-out attempt that stacked on #682's reorg (now abandoned). Branch rebuilt against current main with a single commit.

### DIFF
--- a/sdks/c/src/exec/execution.rs
+++ b/sdks/c/src/exec/execution.rs
@@ -13,11 +13,11 @@
 use futures::StreamExt;
 use std::os::raw::{c_int, c_void};
 use std::ptr;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicI32, Ordering};
 use std::sync::{Arc, Mutex};
 
 use tokio::runtime::Runtime as TokioRuntime;
-use tokio::sync::oneshot;
+use tokio::sync::Notify;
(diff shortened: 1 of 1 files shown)


G4614:
Closed — this was opened as a placeholder after #685 got auto-closed by a force-push. But the FFI drain-race fix is already tracked at #563 (same content, older PR). Continuing this work on #705 would have been a duplicate. The fix/ffi-exec-drain-race branch will be deleted.

PR #706, opened 2026-06-09
refactor: rename Sandbox -> Box (Part 1: apps/api epicenter)

law-chain-hot:
## What

**Part 1 of N** of the repo-wide product-entity rename `Sandbox` → `Box`.
This PR renames the **`apps/api` epicenter** (the ~851-ref `apps/api/src/sandbox/`
mega-module and its consumers within `apps/api`), and is intentionally a **WIP
draft**: CI will be red until the remaining subsystems + generated-client regen
land (see *Remaining*).

> Context: BoxLite is one feature; internally we no longer want `sandbox` to name
> the product entity — it should be `box`. The OS/security isolation primitive
> `sandbox` (jailer/seccomp/Seatbelt) is a different concept and is **kept**.

## Done in this PR (verified)

- `apps/api` product-entity symbols renamed `Sandbox*`→`Box*`, `sandboxId`→`boxId`,
  files + module dir `sandbox/` → `box/` (231 files).
- **Zero-DDL by construction** (audited): `@Entity('sandbox')` pinned; FK columns
  pinned to `name: 'sandboxId'`; `sandbox_state_enum` + all `sandbox_*`
  index/table names unchanged; **no migration added**.
- Internal `CreateSandboxDto` → `CreateBoxInternalDto` (avoids colliding with the
  existing `boxlite-rest` `CreateBoxDto`).
- `runner-api-client` references temporarily kept as generated `Sandbox*` names
  (flip when the runner is renamed + its client regenerated, Part 2).
- **Verification:** `tsc --build apps/api` == pre-change baseline (0 new type errors).
  No local DB/KVM here, so the remote E2E suite was not run.

## Frozen `sandbox` allowlist (deliberately NOT renamed)

| Category | Examples |
|---|---|
| DB table/column/index/enum | table `sandbox`, col `sandboxId`, `sandbox_state_enum`, `sandbox_*_idx`, `sandbox_usage_periods*` |
| OS isolation primitive (KEEP) | `src/boxlite/jailer/*`, seccomp/Seatbelt/`sandbox-exec`, `--no-sandbox`, `IS_SANDBOX`, Cargo keywords |
| Wire / env | `BOXLITE_SANDBOX_ID` |
| Webhook events | `sandbox.created`, `sandbox.state.updated`, … |
| Telemetry meters | `boxlite.sandbox.*` |
| QueryBuilder aliases | raw-SQL `'sandbox.<col>'` join/where strings |

## Reproducible codemod recipe (for the overnight/a2pr deploy line to rebase)

Identifier-level, quote/snake/dot-protected (so frozen literals survive):

```perl
s/Sandbox/Box/g;
s/(?<!")\bsandboxId\b(?!")/boxId/g;       # protect "sandboxId" SQL columns
s/(?<!")\bsandboxIds\b(?!")/boxIds/g;
s/(?<!['"`])sandbox(?=[A-Z])/box/g;        # camelCase ids; protect quotes/snake
s/(?<!['"`])sandbox(?![\w'"`])/box/g;      # bare ids
s/sandboxes/boxes/g; s/sandboxOtel/boxOtel/g; s/sandboxActivity/boxActivity/g;
s/boxlite\.box/boxlite.sandbox/g;          # restore frozen telemetry
```
then `git mv` paths `sandbox→box`, pin entity `@Entity`/column names, regen clients.

## Remaining (separate installments — NOT in this PR)

1. Codemod `runner` / `cli` / `dashboard` / glue (`proxy`/`ssh-gateway`/`otel`/`daemon`) / `sdk-typescript`.
2. Regen the 9 generated OpenAPI clients (`api-client`, `runner/analytics/toolbox-api-client`, `api-client-go`, swaggo docs, cli docs).
3. Build all + remote E2E (`e4bbd0ea`).
4. **PR B** — `UUID → BoxId` id model (public box-id lookup, dashboard shows box id), stacked on this.

⚠️ **Do not merge yet** — WIP. CI red is expected until 1–3 land.


DorianZheng:
### 🐞 Runtime bug survives into Part 4: `@Param('sandboxIdOrName')` no longer matches the renamed route token

`apps/api/src/box/controllers/box.controller.ts` (head `b20fc6cb`): the route path tokens were renamed to `:boxIdOrName`, but **20 `@Param('sandboxIdOrName')` decorators + their `@ApiParam({ name: 'sandboxIdOrName' })` were left frozen**. NestJS resolves `@Param('sandboxIdOrName')` against a route token that no longer exists, so the handler argument is `undefined` at runtime.

Concrete (the `@Audit` reader in the *same* handler proves the mismatch):

```ts
@Delete(':boxIdOrName')                                  // route token = boxIdOrName
@ApiParam({ name: 'sandboxIdOrName', ... })              // stale → OpenAPI param name wrong
targetIdFromRequest: (req) => req.params.boxIdOrName,    // ✓ reads the real token
async deleteBox(
  @Param('sandboxIdOrName') boxIdOrName: string,         // ✗ reads dead token → undefined
) {
  const box = await this.boxService.destroy(boxIdOrName, ...)   // destroy(undefined, ...)
}
```

**Why "compile-verified" (`tsc == baseline`) doesn't catch it:** the argument to `@Param()` is just a `string` literal, so it's type-valid — the break is purely runtime binding + the generated OpenAPI param name. It would only surface in E2E / a live request.

**Why only this token slipped:** the codemod's quote-protection. `:sandboxIdOrName` (colon-prefixed) gets renamed; `'sandboxIdOrName'` (quote-prefixed) is protected. And the `\bsandboxId\b` rule doesn't match the compound `sandboxIdOrName` (no word boundary), so the plain `sandboxId` params renamed cleanly while `sandboxIdOrName` did not.

**Affected endpoints (20):** `GET/DELETE :boxIdOrName`, `recover`, `start`, `stop`, `resize`, `labels`, `backup`, `public/:isPublic`, `autostop`, `autoarchive`, `autodelete`, `archive`, `ports/:port/preview-url` (+ signed / expire), `build-logs`, `build-logs-url`, `ssh-access` (POST + DELETE).

Stale `@Param` lines: 368, 404, 438, 478, 529, 573, 611, 669, 710, 771, 813, 856, 930, 959, 994, 1028, 1063, 1111, 1155, 1198. Stale `@ApiParam name:` lines: 350, 385, 419, 459, 499, 546, 587, 650, 681, 742, ….

**Fix:** rename `sandboxIdOrName` → `boxIdOrName` in every `@Param(...)` and `@ApiParam({ name: ... })` in this file (the `@Param('boxId')` / `:boxId` routes are already consistent). A general guard for the remaining parts: after the codemod, assert every `@Param('x')` / `@ApiParam({name:'x'})` has a matching `:x` route token.

DorianZheng:
### Request: complete the rename — the frozen `sandbox` literals should become `box` too

The rename currently stops at internal identifiers and keeps the externally-/persistently-observable `sandbox` names frozen for zero-DDL + backward-compat. For the product entity to actually *be* `box`, these should be renamed as well (accepting that each is a breaking/coordinated change rather than a pure refactor). Could we extend the PR (or stack a follow-up) to cover them?

**1. DB schema** — `@Entity('sandbox')`, `@Column/@PrimaryColumn/@JoinColumn({ name: 'sandboxId' })`, `sandbox_state_enum`, `sandbox_usage_periods` / `sandbox_usage_periods_archive` / `sandbox_last_activity`, `organization` columns (`max_*_per_sandbox`, `sandbox_*_rate_limit*`), and all `sandbox_*` indexes.
- Needs a **DDL migration**: `ALTER TABLE … RENAME TO`, `RENAME COLUMN`, `ALTER TYPE sandbox_state_enum RENAME TO box_state_enum`, `ALTER INDEX … RENAME TO`, with a reversible `down()`. Flip the entity `@Entity`/`name:` pins to `box*` so code and schema agree. Run as a post-deploy/contract migration.

**2. Wire — webhook events** — `box-events.constants.ts` / `webhook-events.constants.ts` still emit `'sandbox.created'`, `'sandbox.state.updated'`, … These are delivered to customer endpoints, so renaming to `box.*` breaks consumers. Options: version the webhook payload, or dual-emit (`sandbox.*` + `box.*`) for a deprecation window, then drop the old names.

**3. RBAC permission enum** — `OrganizationResourcePermission.*_SANDBOXES` was reverted because the values are persisted/migration-referenced. Renaming the **string values** to `*_BOXES` needs a data migration of stored permission rows (and any policy docs); the TS member names can rename independently.

**4. Env vars** — `RATE_LIMIT_SANDBOX_*`, `DEFAULT_ORG_QUOTA_MAX_*_PER_SANDBOX`, `ADMIN_MAX_*_PER_SANDBOX`, `ORGANIZATION_SANDBOX_DEFAULT_*`, `RUNNER_*_STARTED_SANDBOXES*`, `BOXLITE_SANDBOX_ID`, `SANDBOX_OTEL_*`, etc. Rename to `*_BOX*` and update all deployment manifests / `.env.example` / infra configs in lockstep.

Understood that the zero-DDL freeze was deliberate to keep the PR non-breaking — the ask is to treat these as the breaking-rename tail (migrations + coordinated wire/env changes) so no `sandbox` remains as the product entity anywhere, not just in source identifiers. Happy to take any of these (e.g. the DDL migration) as a stacked PR if that's easier to review.

law-chain-hot:
Fixed in `cfeb019c` — thanks for the precise catch (and the guard suggestion).

Root cause exactly as you diagnosed: the codemod's quote-protection renamed `:sandboxIdOrName` (colon-prefixed) but froze `'sandboxIdOrName'` (quote-prefixed), and the `\bsandboxId\b` rule didn't match the compound token — so the route token moved to `:boxIdOrName` while `@Param('sandboxIdOrName')` bound a dead token → `undefined` at runtime.

**Fix:**
- `sandboxIdOrName` → `boxIdOrName` in every `@Param(...)` / `@ApiParam({ name })` (pure route param, no DB column).
- controller-level `'sandboxId'` → `'boxId'` in `@Param`/`@ApiParam` **only** — entities keep `@Column/@JoinColumn({ name: 'sandboxId' })` pinned, so the physical table/columns stay unchanged (zero-DDL preserved, deferred to the UUID→boxId PR).
- Implemented your guard across **all** api controllers: every `@Param('x')` / `@ApiParam({name:'x'})` now has a matching `:x` route token — **0 sandbox mismatches**. (The only 3 hits are `organizationId`, which lives on the class-level `@Controller(':organizationId/...')` prefix, not a method token — false positive in a method-only scan.)

`tsc --build api` stays at baseline (the break was runtime binding only, type-invisible as you noted); the E2E stack run will confirm the live binding once the bootstrap is unblocked.

DorianZheng:
Pushed `47526870` (**part 6**) — un-freezes the remaining `sandbox` literals to `box`, completing the rename per the request above. Source-only (163 files, +1014/−831); generated clients left at part-5 state for CI regen.

**Done:**
- **DB**: entity pins flipped `sandbox*`→`box*` (`@Entity`/`@Column name`/`@Index`/`::sandbox_state_enum`) + QueryBuilder aliases & raw-SQL refs. New **reversible** contract migration `post-deploy/1781016743403` renames the physical schema (tables `sandbox`/`_last_activity`/`_usage_periods`/`_archive`, `sandboxId` + `organization.sandbox_*`/`max_*_per_sandbox` columns, enum `sandbox_state_enum`, all `sandbox_*_idx` + `idx_sandbox_*` indexes; `down()` reverses).
- **Wire**: webhook events `sandbox.*`→`box.*`, redis channel, analytics meters `boxlite.sandbox.*`→`boxlite.box.*` (producers + dashboard/ws consumers in lockstep).
- **RBAC**: enum member names `WRITE_SANDBOXES`/`DELETE_SANDBOXES`→`*_BOXES` (values stay `write:boxes`/`delete:boxes` — **no data migration**).
- **Env**: `SANDBOX`→`BOX` env vars across api/runner/proxy/daemon/sdk + infra.

**Verified:** `apps/api` `tsc` == this PR's prior-head baseline (identical error set, **0 new**); Go apps `go build`/`vet` clean.

**⚠️ Follow-ups (env-bound, NOT done here):**
1. **DDL migration is not DB-verified** (no DB in my env) — needs a real run + review before merge.
2. **Regenerate** the 5 clients + cli cobra docs + daemon swaggo from the renamed specs (I reverted my local-env regen to keep this source-only).
3. **PostHog flag** `DASHBOARD_CREATE_SANDBOX`→`DASHBOARD_CREATE_BOX` must also be renamed in the PostHog dashboard, or the flag goes dead.
4. **Webhook rename is breaking** for external consumers (no dual-emit window) — version/announce as needed.

Pushed directly at the maintainer's request; revert with `git revert 47526870` if you'd rather stage it differently.

DorianZheng:
Addressed the review-bot comments in `0b03aaec`:

**✅ Fixed — github-code-quality "Missing await"** (`apps/dashboard/src/hooks/useBoxSession.ts:128`)
Made `boxQuery.queryFn` `async` and `await client.get(boxId)`, keeping the `Client not initialized` guard. (Behavior was already correct — React Query awaited the returned promise — but the async form is clearer and clears the analyzer.)

**ℹ️ CodeQL "Clear-text logging of sensitive information" (7 alerts) — pre-existing, not from this PR**
These flag `log.*("... %s: %v", id, err)` where `err` comes from the generated API client. They are **identical on `main`** (e.g. `apps/ssh-gateway/main.go:217`, `apps/proxy/pkg/proxy/get_sandbox_target.go:313/326/333`) and only surfaced here because the rename put these files in the PR diff (CodeQL default-setup re-scans changed files). `apps/ssh-gateway/main.go` isn't even modified by the rename commits.

I deliberately did **not** bundle a fix into this rename PR — sanitizing credential-tainted API-client errors across cli/proxy/ssh-gateway is a security change that deserves its own reviewed PR (shared redaction helper), not a drive-by in a rename. Happy to open that separately. (They also look like conservative taint flow — the generated client's `GenericOpenAPIError.Error()` returns status+body, not the request `ApiKey`/`Password` — so likely false positives, but worth a dedicated look.)

DorianZheng:
### ✅ Rename-only verification (94-agent sweep over all 937 changed files)

Ran an automated check across every changed file: normalize `sandbox`↔`box` on the pre-rename (`main`) and current versions, diff the residual, classify. Result:

| Verdict | Files |
|---|---|
| Pure rename | 427 |
| Generated churn | 376 |
| Added (rename-driven, incl. files + DDL migration) | 65 |
| Formatter reflow (shorter identifiers) | 49 |
| Flagged for review | 20 |

All 20 review flags triaged to **non-issues for the rename**: the intended DB-name pins (`@Entity('box')`/`@Column({name:'boxId'})`, paired with the contract migration), the documented `CreateSandboxDto`→`CreateBoxInternalDto` collision rename, false-positive git rename-pairings (the dashboard mutation hooks are `R100` renames calling the correct APIs — verified), the CodeQL clear-text-logging autofix commits, and generated-client churn. **No hidden logic change is attributable to the rename.**

Two things surfaced that are **not** rename-related and worth your eyes:

**1. `runnerAdapter` dropped its `ToolboxApi` integration (part 4, `b20fc6cb`).** Both v0 and v2 no longer reference `ToolboxApi`. On `main` the field was declared + initialized but **never called** (`this.toolboxApiClient.` → 0 hits), so this looks like dead-code cleanup — just confirming it was intentional and nothing relied on it.

**2. Generated clients were regenerated with a newer openapi-generator than `main` uses** — `main` is on **7.12.0**, the regen produced **7.23.0**. That version jump (not the rename) is what added unrelated scaffolding to the clients: AWS SigV4 config, `Set` serialization, `GIT_USER_ID`/`GIT_REPO_ID` placeholders in `git_push.sh`, `Accept` headers. There's no generator version pinned in-repo (`openapitools.json` absent on `main`). Recommend **pinning the generator to 7.12.0** (or deciding to bump repo-wide) before the final client regen, so the generated diff stays rename-only instead of carrying a generator upgrade.

PR #728, opened 2026-06-11
chore(apps): remove the Go CLI (apps/cli)

DorianZheng:
The Rust CLI (src/cli) is the product CLI: it covers box create/list/exec/start/stop/info against the same REST API and ported the Go CLI's OIDC login flow verbatim. No CI or release surface builds, ships, tests, or deploys apps/cli — no workflow, Dockerfile, installer, or e2e reference names it; locally it was only swept up by the workspace-wide nx run-many --all targets (make build:apps / test:apps), which simply discover one fewer project after this removal. Since the Box rename it only changed by riding along with refactors (the API-client regen consumer fixes being the last).

Features that existed only here — mcp server, ssh, preview-url, organization and volume management — leave the tree until src/cli grows Rust equivalents; git history remains the porting reference, the same pattern the auth port already used.

Reference cleanup: apps/go.work drops the ./cli use directive (go.work.sum refreshed); api-client-go's set-version target no longer edits apps/cli/go.mod; .gitignore drops the built-binary entry; apps/NOTICE drops apps/cli from the AGPL list.

## Summary by CodeRabbit

* **Chores**
  * Removed the BoxLite CLI application, including all command-line tools for box management, organization operations, volume management, authentication, and MCP server integration.
  * Updated repository configuration to remove CLI-related references and dependencies.


law-chain-hot (approved):
LGTM

PR #732, opened 2026-06-11
fix(apps): vend box S3 credentials from the ECS task role, not a static IAM user

DorianZheng:
## Problem

An IAM audit of `apps/infra` (live account cross-checked) found the box-storage credential path both over-privileged and broken:

- The Api signed STS calls with a **static IAM user** (`S3User`) holding `s3:*` on `*` **and** `sts:AssumeRole` on `*`, delivered as long-lived keys in ECS env vars — the exact anti-pattern AWS's IAM best practices call out ("require workloads to use temporary credentials with IAM roles").
- The role it assumes (`S3_ROLE_NAME=BoxliteS3Role`) **was never provisioned anywhere** — live calls fail with `NoSuchEntity`, so box storage push access has never worked on AWS. Inherited from upstream Daytona, whose public deployment path is MinIO-only (`S3_ROLE_NAME=/` placeholder).
- In the dev stage the S3User's only access key is deactivated, so the static-key path was doubly dead.

## Fix

Keep the architecture — it's AWS's own token-vending-machine pattern (AssumeRole + per-org inline session policy) — and swap the identities under it, then narrow every grant to its verified call surface:

```
before:  S3User (static keys, s3:* + sts:* on *) ──AssumeRole──▶ BoxliteS3Role ✗ (missing)
after:   Api ECS task role (auto-rotated)        ──AssumeRole──▶ S3AccessRole (created here)
            sts:AssumeRole scoped to that one ARN                  trust: exactly the Api task role
            7 enumerated s3 actions on boxlite-volume-*            perms: Get/Put/List on storage bucket
            s3:ListBucket on the storage bucket                    ∩ session policy = per-org prefix
            logs scoped to this region/account
```

**infra (`sst.config.ts`)**
- Drop `S3User`/`S3UserPolicy`/`S3AccessKey`. Add stage-scoped `S3AccessRole` (`boxlite-<stage>-s3-access`; deterministic name declared before the Api so the env var needs no resource dependency; the role is created after the Api because its trust policy references `api.nodes.taskRole.arn`).
- Api task role: `sts:AssumeRole` on that single ARN; **enumerated** volume-bucket actions (`CreateBucket`, `PutBucketTagging`, `ListBucket`, `ListBucketVersions`, `DeleteObject`, `DeleteObjectVersion`, `DeleteBucket` — deliberately not `s3:*`, whose tail is `PutBucketPolicy`/`PutBucketAcl`); `storage` is **de-linked** (link grants `s3:*`) and replaced with the verified need, `s3:ListBucket`; `logs:DescribeLogGroups` scoped from `*` to this region/account.
- Runner `RunnerVolumeS3Policy`: `s3:*` → exactly Mountpoint for Amazon S3's documented set (`ListBucket` + `GetObject`/`PutObject`/`DeleteObject`/`AbortMultipartUpload`).

**api**
- `ObjectStorageService` and `VolumeManager` fall back to the SDK default credential chain (ECS task role) when `S3_ACCESS_KEY`/`S3_SECRET_KEY` are unset; static-key behavior unchanged when both are set (MinIO/local).
- Guards (from review): **both-or-neither** static keys — a lone key now errors instead of silently switching credential source (mirrors `observability-s3.reader.ts:42`); MinIO without keys **fails closed** in both services; `VolumeManager.testConnection` probes the known default bucket (`ListObjectsV2`, `MaxKeys: 1`) instead of `ListBuckets`, removing the account-wide `s3:ListAllMyBuckets` need.

Role chaining caps chained sessions at 1h; the vend already requests exactly 3600s, so no functional change.

## Verification

- `object-storage.service.spec.ts` (4) + `volume.manager.spec.ts` (5): default-chain construction, RoleArn + per-org session policy, static-key passthrough, lone-key rejection, MinIO fail-closed (both services), scoped probe shape, probe skip. **Two-side verified**: with all production changes reverted to HEAD the new-behavior tests fail (5/9, e.g. `Missing config: s3.accessKey`, `ListBucketsCommand is not a constructor`); restored, 9/9 pass.
- Full api suite: green except 2 `default-organization-membership` failures that reproduce identically on `main` (pre-existing, untouched).
- `tsc --noEmit` clean; `nx lint api` clean.
- Not run: `sst diff` (no stage credentials in this environment). The one SST surface this leans on (`Service.nodes.taskRole`) was verified against `sst/sst` `platform/src/components/aws/service.ts`.

## Deploy notes

- First deploy per stage: the new task definition may briefly roll out before `S3AccessRole` exists; vending errors in that window are no worse than today (the target role has never existed). Steady state is correct once the deploy completes.
- The runner policy change is an in-place `RolePolicy` update — it does not touch the protected Runner EC2 (`ignoreChanges` only covers `ami`/`userDataBase64`).
- Pulumi will delete each stage's `S3User` + key on deploy (dev's key is already inactive).
- Out of scope, recommended ops follow-ups from the audit: delete orphaned `RunnerRole-e887e17`/`RunnerProfile-b86f07b` (carries `AmazonS3FullAccess`, unused — verified the live runner uses `RunnerRole-1115ba6`), remove the duplicate `RunnerVolumeS3Policy` inline copy, move deploys off root credentials to the existing `AWSReservedSSO_*` permission sets.
- Known residual (follow-up issue): the `boxlite-volume-*` bucket prefix is account-global, not stage-scoped — stages can reach each other's volume buckets. Fixing it requires a bucket-naming/migration decision (persist `bucketName` on the volume row), tracked separately.

### DIFF
--- a/apps/api/src/admin/services/observability.service.ts
+++ b/apps/api/src/admin/services/observability.service.ts
@@ -8,11 +8,7 @@ import { Inject, Injectable, ServiceUnavailableException } from '@nestjs/common'
 import { ClickHouseService } from '../../clickhouse/clickhouse.service'
 import { TypedConfigService } from '../../config/typed-config.service'
 import { LogEntryDto } from '../../box-telemetry/dto/log-entry.dto'
-import {
-  MetricsResponseDto,
-  MetricDataPointDto,
-  MetricSeriesDto,
-} from '../../box-telemetry/dto/metrics-response.dto'
+import { MetricsResponseDto, MetricDataPointDto, MetricSeriesDto } from '../../box-telemetry/dto/metrics-response.dto'
 import { PaginatedLogsDto } from '../../box-telemetry/dto/paginated-logs.dto'
 import { PaginatedTracesDto } from '../../box-telemetry/dto/paginated-traces.dto'
 import { TraceSpanDto } from '../../box-telemetry/dto/trace-span.dto'
--- a/apps/api/src/app.service.ts
+++ b/apps/api/src/app.service.ts
@@ -191,5 +191,4 @@ Admin API key ensured: ${this.maskApiKeyForLog(value)}
     }
     return `${value.slice(0, 4)}...${value.slice(-4)}`
   }
-
 }
--- a/apps/api/src/box/box.module.ts
+++ b/apps/api/src/box/box.module.ts
@@ -58,16 +58,7 @@ import { BoxStateWaiterService } from './services/box-state-waiter.service'
     UserModule,
     OrganizationModule,
     RegionModule,
-    TypeOrmModule.forFeature([
-      Box,
-      Runner,
-      WarmPool,
-      Volume,
-      SshAccess,
-      Region,
-      Job,
-      BoxLastActivity,
-    ]),
+    TypeOrmModule.forFeature([Box, Runner, WarmPool, Volume, SshAccess, Region, Job, BoxLastActivity]),
   ],
   controllers: [
     BoxController,
--- a/apps/api/src/box/controllers/runner.controller.ts
+++ b/apps/api/src/box/controllers/runner.controller.ts
@@ -20,15 +20,7 @@ import {
 } from '@nestjs/common'
 import { CreateRunnerDto } from '../dto/create-runner.dto'
 import { RunnerService } from '../services/runner.service'
-import {
-  ApiOAuth2,
-  ApiTags,
-  ApiOperation,
-  ApiBearerAuth,
-  ApiResponse,
-  ApiParam,
-  ApiHeader,
-} from '@nestjs/swagger'
+import { ApiOAuth2, ApiTags, ApiOperation, ApiBearerAuth, ApiResponse, ApiParam, ApiHeader } from '@nestjs/swagger'
 import { SystemActionGuard } from '../../auth/system-action.guard'
 import { RequiredApiRole } from '../../common/decorators/required-role.decorator'
 import { SystemRole } from '../../user/enums/system-role.enum'
--- a/apps/api/src/box/managers/volume.manager.spec.ts
+++ b/apps/api/src/box/managers/volume.manager.spec.ts
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2025 Daytona Platforms Inc.
+ * Modified by BoxLite AI, 2025-2026
+ * SPDX-License-Identifier: AGPL-3.0
+ */
+
+import { ListObjectsV2Command, S3Client } from '@aws-sdk/client-s3'
+import { VolumeManager } from './volume.manager'
+
+const mockSend = jest.fn()
+
+jest.mock('@aws-sdk/client-s3', () => ({
+  S3Client: jest.fn().mockImplementation(() => ({ send: mockSend })),
+  CreateBucketCommand: jest.fn().mockImplementation((input) => ({ input })),
+  ListObjectsV2Command: jest.fn().mockImplementation((input) => ({ input })),
+  PutBucketTaggingCommand: jest.fn().mockImplementation((input) => ({ input })),
+}))
+
+describe('VolumeManager S3 client setup', () => {
+  afterEach(() => {
+    jest.clearAllMocks()
+  })
+
+  function buildManager(values: Record<string, unknown>) {
+    const configService = {
+      get: jest.fn((key: string) => values[key]),
+      getOrThrow: jest.fn((key: string) => {
+        const value = values[key]
+        if (value === undefined) {
+          throw new Error(`Missing config: ${key}`)
+        }
+        return value
+      }),
+    }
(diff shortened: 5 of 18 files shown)


law-chain-hot:
Solid change — task-role vending is the right pattern, and the resource-cycle break (declare role *name* up front as a string, create the role resource after the Api so its trust policy can reference `api.nodes.taskRole.arn`) is handled cleanly. Fail-closed on the MinIO path is the right call too. One non-blocking nit:

**Partial static-key config silently falls through to the default chain.** The PR cites `observability-s3.reader.ts` as the same idiom, but that file adds an XOR guard the new code drops:

```ts
// observability-s3.reader.ts:42
if ((accessKey && !secretKey) || (!accessKey && secretKey)) {
  // explicit "require both" error
}

// this PR (object-storage.service.ts / volume.manager.ts)
...(accessKey && secretKey ? { credentials: {...} } : {})  // only one set → silently uses default chain
```

So if someone sets `S3_ACCESS_KEY` but forgets `S3_SECRET_KEY` (a typical misconfig), it won't error — it quietly drops both and falls back to the task-role chain, which makes the failure harder to diagnose. Since the PR already treats that reader as the template, mirroring its both-or-neither guard would close the gap. Low severity (AWS sets neither, MinIO sets both), purely a robustness/diagnosability improvement.

Everything else (MinIO detection via `endpoint.includes('minio')`, the brief first-deploy vending window) is pre-existing or already acknowledged in the description.

law-chain-hot (commented):
Commented

DorianZheng:
@law-chain-hot
Verified and fixed — you're right, and the citation was incomplete on my side: `observability-s3.reader.ts:42-52` does carry the both-or-neither guard, and the new code dropped it. Both `ObjectStorageService.getPushAccess` and the `VolumeManager` constructor now reject a lone key explicitly (`S3_ACCESS_KEY and S3_SECRET_KEY must be set together`) instead of silently switching to the default chain, with spec coverage for the lone-key case in both services (`object-storage.service.spec.ts`, `volume.manager.spec.ts`).

Same commit also narrows the remaining grants to their verified call surfaces (volume-bucket `s3:*` → 7 enumerated actions, `ListAllMyBuckets` dropped in favor of a known-bucket probe, storage de-linked → `s3:ListBucket` only, `DescribeLogGroups` region-scoped, runner policy → Mountpoint's documented set).

DorianZheng:
@coderabbitai
Adjudication of the two nitpicks:

1. **VolumeManager MinIO fail-fast — fixed.** The constructor now rejects MinIO endpoints without static keys (`MinIO requires S3_ACCESS_KEY and S3_SECRET_KEY to be configured`) before building the client, so the failure is a clear boot error rather than a generic auth error from the connection probe. Covered by a new test in `volume.manager.spec.ts`. A both-or-neither guard for partially-set keys (from the human review) landed alongside it.

2. **`endpoint.includes('minio')` heuristic — acknowledged, not changed here.** The heuristic predates this PR (inherited from upstream Daytona's `object-storage.service.ts`) and replacing it with an explicit provider knob (`S3_PROVIDER`) is a deployment-contract change beyond this PR's permissions scope. Reasonable follow-up if MinIO-behind-custom-domain becomes a real deployment shape.

On the docstring-coverage warning: this repo doesn't use docstring conventions in these services (comments are reserved for non-obvious constraints per CLAUDE.md), so no docstrings were added.

DorianZheng:
⚠️ **Correction to the deploy notes**: the recommendation to delete `RunnerRole-e887e17`/`RunnerProfile-b86f07b` as "unused" is **wrong — do not delete it**. A deeper check (instances filtered by instance-profile ARN rather than by `Name=boxlite-runner*` tag) found a **running production instance** depending on it:

- `i-0a2516cdf06c221df`, tagged `boxlite-prod`, c8i.large, launched 2026-05-08, public IP, **default VPC** (i.e. launched manually outside the SST stack) — using `RunnerProfile-b86f07b` → `RunnerRole-e887e17`, role last-used today.

So a manually-launched prod box is riding on the leaked legacy dev role carrying `AmazonS3FullAccess` + ECR PowerUser. That's a worse posture than the audit assumed, but the remediation is migration, not deletion:

1. create a dedicated, scoped prod role + instance profile (SSM core + the Mountpoint volume set, mirroring this PR's `RunnerVolumeS3Policy`),
2. swap it onto `i-0a2516cdf06c221df` with `aws ec2 replace-iam-instance-profile-association` (no downtime; instance-profile swaps apply live),
3. then delete `RunnerRole-e887e17`/`RunnerProfile-b86f07b`.

Completed cleanup so far: the manual duplicate inline policy `RunnerVolumeS3Policy` (unsuffixed) on dev's `RunnerRole-1115ba6` is deleted — the Pulumi-managed `RunnerVolumeS3Policy-865446e` remains and is the one this PR narrows, so the narrowing now actually takes effect on deploy. The per-stage `S3User`s and stale service roles are Pulumi-managed and will be cleaned by deploying this PR per stage, not by hand.

law-chain-hot (approved):
LGTM

PR #715, opened 2026-06-10
Converge A2 + MVP box journey into main (super PR — review & split)

law-chain-hot:
## What this is

A **convergence "super PR"**: it merges everything sitting on the `codex/overnight-20260608` line into `main`, so that `main` becomes the single source of truth again. Intended to be **reviewed then split** — opened as one piece for visibility, not to be squash-merged blind.

`main` ← merge of (52 commits ahead): MVP box journey + agent-ready image catalog + A2 snapshot-manager deletion + Task12B observability wiring + default-org membership + W3C traceId, reconciled against `main`'s `Sandbox→Box` rename (#706) and recent REST/CI fixes.

**Net vs main:** 771 files, +85k/−23.7k.

## Suggested split for review (5 logical blocks)

| # | Block | Rough surface |
|---|---|---|
| 1 | **A2 — delete snapshot-manager machinery** | drops self-hosted registry / backup / build / docker-registry; collapses to `box_template` + `runner_artifact_cache`; direct ghcr pull; 68 files deleted; migrations `17809…`–`17812…` |
| 2 | **MVP box journey** | simplified onboarding, templates-as-images, public Box IDs, archive-lifecycle retirement |
| 3 | **Agent-ready image catalog** | `boxlite/base|python|node` pinned-digest runtime catalog |
| 4 | **Observability (Task12B)** | admin diagnose UX + runner/daemon traceId propagation |
| 5 | **Org membership** | default-org state moved into memberships (backward-compatible) |

## Conflict-resolution policy (527 conflicts)

- **A2-deleted machinery** (snapshot/backup/build/docker-registry): **deletion wins** over main's edits to those files.
- **Generated clients** (`api-client`, `runner-api-client`, `api-client-go`): took **main's regenerated side**; a fresh regen against the merged API surface is still pending (see Caveats).
- **Source files**: kept the A2/MVP side and re-applied **#706's published codemod recipe** (`Sandbox→Box`, with the frozen-literal allowlist preserved — DB table/column/enum, OS-isolation `sandbox`, telemetry `boxlite.sandbox.*`, webhook event names).
- **Renamed-path ports**: 3-way merged onto the post-#706 `box/` paths.
- **Go consumers** aligned to the regenerated client enum names (`BOXSTATE_*`).

## Caveats — read before merging

[...]

PR #734, opened 2026-06-11
fix(build): repair #715 runner build break and org test mock

law-chain-hot:
## Summary
- add the Go SDK `WithPort` option required by the runner build path after #715
- update the default organization membership spec mock argument order to match the current service call

## Verification
- pre-push `make test:changed` was triggered by the local hook and failed in the existing macOS Go/native link test path before push
- pushed with `--no-verify` per handoff flow; CI is the source of truth for this PR

## Summary by CodeRabbit

* **New Features**
  * Added configurable host↔guest port mappings (host port, guest port, protocol, host IP).

* **Bug Fixes**
  * Updated how the toolbox port is exposed between host and guest (mapping direction corrected).

* **Tests**
  * Updated tests and test helpers to reflect the new port-mapping configuration and the adjusted toolbox mapping.

### DIFF
--- a/apps/api/src/organization/services/default-organization-membership.spec.ts
+++ b/apps/api/src/organization/services/default-organization-membership.spec.ts
@@ -42,12 +42,15 @@ function createOrganizationService() {
     }),
   }
 
+  // Argument positions must match OrganizationService's constructor signature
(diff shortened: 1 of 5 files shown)


DorianZheng (approved):
LGTM

PR #735, opened 2026-06-11
feat(box): create boxes from curated images

law-chain-hot:
## Summary

- Box creation boots from one of three **supported pinned OCI images** (base / python / node). The request carries the full OCI ref — there is no key→ref translation layer. An unsupported ref is rejected with a 400 that lists the supported refs.
- The allowlist is a deliberately thin gate at the request boundary ([curated-images.constant.ts](apps/api/src/box/constants/curated-images.constant.ts)): when per-org custom images land, delete that file and its two call sites — no other layer knows the curated set exists.
- A box now has **exactly one identity**: a 12-char base62 id, minted in memory at construction, used as the PG primary key, in the CREATE_BOX payload, as the engine VM name, in runner logs, in the in-box `BOXLITE_BOX_ID` env, and in every API response. The internal uuid (Daytona leftover) is removed.
- Runner contract: `CreateBoxDTO` field renamed to `image` (full OCI ref, aligned with the embedded SDKs' `image` parameter); `BoxId`, `GpuQuota`, `Registry`, and the `authToken` payload field are dropped; `userId`/`osUser` are no longer `validate:"required"`.

## How a box is created (lifecycle)

```text
client
  │ POST /boxes { image: "ghcr.io/boxlite-ai/boxlite-agent-python@sha256:…" }
  ▼
1. boundary gate (box.service)
  │ assertSupportedImage(image)  → 400 + supported list if not allowlisted
  │ quota / org / runner checks  → read-only, no DB trace on failure
  ▼
2. in-memory mint (new Box())
  │ id = "a7Kf93mQpX2z"  ← the single identity, born here, never changes
  │ box.image = full OCI ref, stored verbatim
  ▼
3. tx#1: INSERT boxes (state=UNKNOWN, desired=STARTED)
  │ HTTP thread starts polling for STARTED (≤30s) … event wakes the sync loop
  ▼
4. state machine (same process, background)
  │ redis lock(id) → box-start.action → adapter.createBox(box)
  │ tx#2: INSERT job { resourceId: id, payload: { id, image, cpu/mem/disk, env, network } }
  │ tx#3: UPDATE box → CREATING
  ▼
══ process boundary — below this line only the 12-char id exists ══
5. runner (Go)
  │ poller picks the job → WithName(id) → runtime.Create(image)
  │ logs / mount records / BOXLITE_BOX_ID all use the same id
  ▼
6. engine (Rust)  — pulls the image, boots the microVM, completes the job
  ▼
tx#4: UPDATE box → STARTED → step-3 poller wakes up
  ▼
201 { box_id: "a7Kf93mQpX2z", image: "ghcr.io/…", status: "running" }
```

Notes for reviewers:

- Not one big transaction: a chain of single-row transactions linked by the box id, reconciled by the desired-state sync loop (10s cron + redis locks + conditional updates). A crash mid-chain resumes from the last persisted state.
- The DB stores the resolved ref at create time, so an env-var digest rotation does not affect existing boxes.
- `image` env vars (`BOXLITE_SYSTEM_{BASE,PYTHON,NODE}_IMAGE`) now define the allowlist itself; bootstrap.sh points the local stack at public refs and fixture_setup records the stack's base image in the e2e profile.

## Breaking / deployment

- The baseline migration was edited in place (box PK is now `varchar(12)`, `image` column added, `boxId` column removed). **Requires a DB rebuild** — agreed as the deployment plan since nothing is launched.
- Stacked on #743 (TS cloud SDK removal); merge #743 first, GitHub will retarget this PR to main.

## Verification

- API jest: 38 suites / 112 tests pass
- Go: `go build ./pkg/... && go vet && gofmt -l` clean (final cgo link needs the prebuilt `libboxlite.a`, absent in the dev worktree — environment artifact)
- Dashboard: `tsc --noEmit` clean
- e2e: fixtures adapted (full refs, stack-aware default image via credentials profile); the suite itself runs on the integration branch per plan

## Follow-up (stacked next)

- Drop the legacy `boxId` DTO field (`BoxDto` / admin overview DTO) which currently mirrors `id` for API-contract compatibility, regenerate the four API clients, and switch the dashboard reads from `.boxId` to `.id`. Kept out of this PR so it stays free of generated files.

## Release Notes

* **New Features**
  * Added image selection during box creation from a curated list of supported OCI container images.
  * Introduced external reference tracking for boxes.

* **Improvements**
  * Unified box identifier model; single ID now replaces separate internal/public identifiers.
  * Updated image validation to enforce allowlist acceptance.

* **Documentation**
  * Updated SDKs and API documentation to reflect image handling changes.


apps/api/src/box/runner-adapter/runnerAdapter.ts
```diff
@@ -46,6 +46,7 @@ export interface RunnerAdapter {
   runnerInfo(signal?: AbortSignal): Promise<RunnerInfo>
 
   boxInfo(boxId: string): Promise<RunnerBoxInfo>
+  createBox(box: Box, artifactRef: string): Promise<void>
```
DorianZheng: what is `arti

apps/api/src/box/runner-adapter/runnerAdapter.v2.ts
```diff
@@ -114,6 +114,33 @@ export class RunnerAdapterV2 implements RunnerAdapter {
...
 
+  async createBox(box: Box, artifactRef: string): Promise<void> {
+    // Hand-built payload: keys MUST match the Go dto.CreateBoxDTO json tags
+    // (apps/runner/pkg/api/dto/box.go). The image ref is passed as `artifactRef` —
+    // the runner uses it directly in runtime.Create (NOT `snapshot`).
+    const payload = {
+      id: box.id,
+      boxId: box.boxId,
+      userId: box.organizationId,
+      artifactRef,
```
DorianZheng: image: box.Image,

law-chain-hot:
## Identity model note (pre-empting a likely review question)

**Q: the engine mints its own BoxID and won't accept an external id — so how can the cloud-minted 12-char id be "the" box identity?**

A: the cloud id is never injected as the engine's id. It rides as the engine box **name** (unique, indexed, queryable), exactly the way kubelet attaches the pod UID to containers it doesn't own. Side by side:

```text
        Docker/K8s                            BoxLite
─────────────────────────────────┬─────────────────────────────────
L1 control plane mints the       │ API mints the 12-char id
   public identity (pod UID)     │   (box.entity.ts: generateBoxId)
─────────────────────────────────┼─────────────────────────────────
L2 agent mints nothing; it       │ runner:
   attaches + maps:              │ · attach:  WithName(dto.Id)
 · attach pod UID as a CRI label │             (client.go:229)
 · in-memory pod↔container map   │ · map:     c.boxes[dto.Id]
 · relist by label after restart │ · recover: runtime.Get(dto.Id)
 · backfill runtime id into      │             — name lookup (client.go:322)
   pod status.containerID        │ · backfill: not done (optional follow-up)
─────────────────────────────────┼─────────────────────────────────
L3 runtime mints its own         │ engine mints its own BoxID
   container id; external ids    │   (rt_impl.rs: BoxIDMint::mint)
   cannot be injected            │ · name is unique + persisted in the
 · docker name table = the       │   engine DB = the durable mapping
   durable mapping               │   (rt_impl.rs lookup_box)
 · `docker inspect` exposes the  │ · GetInfo/bx.ID() exposes the engine
   raw Id (display only)         │   id (display only, never referenced
                                 │   as an external identity)
─────────────────────────────────┴─────────────────────────────────
```

Consequences:

- The control plane (API/PG/runner/logs/users) speaks exactly one id. The engine-internal BoxID is a runtime implementation detail, like a containerd id under kubelet.
- The mapping needs no extra table: it lives in the engine DB's unique `name` index (durable) and the runner's box map (hot), with `runtime.Get(name)` as the crash-recovery path.
- Local/embedded mode degrades gracefully: with no control plane, the engine's own 12-char base62 id IS the user-facing id — same shape, single mental model.
- Optional follow-up (parity with `pod status.containerID`): report the engine id back in the CREATE_BOX job result and store it on the box row for forensics. Not needed for this stack.

G4614 (approved):
CAN

law-chain-hot:
Superseded by #753 — same image + single-identity + drop-boxId work consolidated into one PR against main, with external_ref stripped.

PR #741, opened 2026-06-11
fix(sdk): make boxlite_options_add_port carry the full PortSpec

DorianZheng:
The C function took (guest_port, host_port) as two ints — reversed relative to core PortSpec, the Python/Node/Go SDKs, the CLI -p flag, Docker's host:guest convention, and add_volume in the same header — and could not express protocol or host_ip, capping every binding built on the C ABI below what the core model carries.

C SDK:
- New signature: (opts, host_port, guest_port, protocol, host_ip), host-first, field-for-field the core PortSpec order. Old 3-arg form removed outright; the arity change turns stale callers into compile errors instead of silently flipped forwarding.
- Ports are uint16_t; returns BoxliteErrorCode instead of void (InvalidArgument on NULL opts, guest_port 0, or non-UTF-8 host_ip).
- New BoxlitePortProtocol enum, following the BoxliteRegistryTransport pattern. host_ip NULL or "" = all interfaces.
- The C layer stores protocol/host_ip verbatim (same as Node/Python); capability enforcement for values the runtime cannot honor yet stays a core concern.

Go SDK:
- WithPortSpec merged into WithPort(spec PortSpec); the two-int form is gone (Go has no overloading), matching the WithNetwork/WithSecret struct-option style. Named fields replace positional host/guest.
- Zero-value Protocol now normalizes to TCP in toCSpec, matching the core default; UDP and HostIP are still rejected until the runtime honors them, with errors that blame the runtime, not the bridge.
- Bridge passes protocol and host_ip through the new FFI signature and checks the returned error code.

Also corrects the stale "None/0 => dynamically assigned" comment on core PortSpec.host_port (actual behavior mirrors guest_port) and the same false claim in the Go docs, and updates both apps/runner call sites.

## Summary by CodeRabbit

* **Breaking Changes**
  * C SDK port API updated: host-first parameter order, ports are 16-bit, protocol parameter added, host IP allowed, and calls return error codes.

* **New Features**
  * Port forwarding accepts explicit protocol and optional host bind address.
  * Go SDK options now accept a full port spec object when configuring ports.

* **Documentation**
  * Migration guide with before/after C examples; port host-port semantics clarified.

* **Tests**
  * New and revised tests cover port validation and error handling.

### DIFF
--- a/apps/runner/pkg/boxlite/client.go
+++ b/apps/runner/pkg/boxlite/client.go
@@ -265,7 +265,7 @@ func (c *Client) Create(ctx context.Context, boxDto dto.CreateBoxDTO) (string, s
 	if err != nil {
 		return "", "", err
 	}
-	opts = append(opts, boxlite.WithPort(toolboxHostPort, ToolboxGuestPort))
+	opts = append(opts, boxlite.WithPort(boxlite.PortSpec{Host: toolboxHostPort, Guest: ToolboxGuestPort}))
 
 	opts = append(opts, boxlite.WithNetwork(networkSpec(boxDto.NetworkBlockAll, boxDto.NetworkAllowList)))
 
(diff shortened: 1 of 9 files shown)


G4614 (dismissed):
LGTM

law-chain-hot (dismissed):
LGTM

PR #744, opened 2026-06-12
refactor(api): drop the legacy boxId DTO field and regenerate API clients

law-chain-hot:
## Summary

Stacked on #735. The Box entity's identity collapsed there into a single 12-char id; the `boxId` fields on `BoxDto` / the admin overview DTO were kept only as value mirrors of `id` to avoid touching the API contract and generated clients in that PR. This PR removes the mirror everywhere:

```text
before (after #735)                 after (this PR)
───────────────────                 ───────────────
entity/DB:   id  ✅ single          unchanged
BoxDto:      id + boxId (same       id only
             value, legacy shape)
admin DTO:   id + boxId mirror      id only
clients:     Box/AdminBoxItem/      boxId field gone (TS + Go)
             Workspace carry boxId
dashboard:   reads .boxId ?? .id    reads .id
```

- API: `BoxDto.boxId`, admin overview `boxId`, dual-id correlation matching in admin observability — all removed.
- Dashboard: `box-identity` helpers read `.id`; dual-id cache updates and the uuid-shaped-name special case removed (after the DB rebuild a box name can never be a uuid).
- Generated clients: regenerated with the repo's own codegen **inside a disposable BoxLite box** (CI-faithful: Linux + JDK + GNU sed, no host gofmt drift). Drift is exactly the `boxId` removal on the `Box` / `AdminBoxItem` / `Workspace` models in the TS and Go clients.
- Untouched on purpose: route/query parameter names like `:boxId` (they carry the single id), other tables' `boxId` FK column names, and the analytics `ModelsBoxUsage.boxId` (a real data field, not the mirror).

## Verification

- API jest: 38 suites / 112 tests pass
- Dashboard: `tsc --noEmit` clean against the regenerated client
- Regen ran in a BoxLite box (`boxlite run node:22-bookworm`); output is byte-for-byte what CI's api-client-drift check regenerates

## Merge order

#743 (remove TS cloud SDK) → #735 (image pipeline + single box identity) → this PR.

### DIFF
--- a/apps/api-client-go/api/openapi.yaml
+++ b/apps/api-client-go/api/openapi.yaml
@@ -9937,7 +9937,6 @@ components:
     Box:
       example:
         id: fd955d93-e74a-48e7-9f2d-fcbe6dd9e920
-        boxId: aB3cD4eF5gH6
         organizationId: organization123
         name: MyBox
         user: boxlite
@@ -9977,10 +9976,6 @@ components:
         id:
           description: The internal UUID of the box
           example: fd955d93-e74a-48e7-9f2d-fcbe6dd9e920
-          type: string
-        boxId:
-          description: The public Box ID shown to users and SDK clients
-          example: aB3cD4eF5gH6
           type: string
         organizationId:
           description: The organization ID of the box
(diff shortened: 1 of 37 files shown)


G4614 (approved):
zggg

PR #747, opened 2026-06-12
feat(engine): orchestrator external_ref on boxes (CRI uid analog)

law-chain-hot:
## Summary

Stacked on #744. Implements the identity follow-up discussed on #735 (see the "Identity model note" comment there): the engine gains a CRI-style orchestrator reference, and the runner stops squatting the engine's `name` slot to carry the control-plane box id.

```text
before                                after
──────                                ─────
runner: WithName(cloudId)             runner: WithExternalRef(cloudId)
engine: name = cloudId (squatted)     engine: external_ref = cloudId
        name unavailable for users            name free for a future
                                              user-facing box name
```

Industry precedent (verified sources):
- CRI `PodSandboxMetadata.uid` — the orchestrator's id travels as a dedicated metadata field; the runtime mints its own id ([cri-api api.proto](https://github.com/kubernetes/cri-api/blob/master/pkg/apis/runtime/v1/api.proto), `message PodSandboxMetadata`)
- ECS agent attaches `com.amazonaws.ecs.task-arn` as a docker label ([docker_task_engine.go](https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/docker_task_engine.go))
- The caller-supplied-id pattern (OCI runtime spec `create <container-id>`) applies to stateless executors (runc/firecracker), not to stateful daemons like this engine — hence a reference field, not id injection.

## What changed

- **engine**: `BoxOptions.external_ref` (opaque, json-persisted via the existing config blob — no DB migration; `lookup_box` already scans configs). Duplicate refs are rejected at create. Lookup order: exact id → name → external_ref → id prefix. `BoxInfo` exposes it, omitted when unset.
- **C SDK**: `boxlite_options_set_external_ref`, `CBoxInfo.external_ref` (NULL when unset, freed on drop — the leak-count test now expects 5 inner CStrings), header regenerated by the crate build.
- **Go SDK**: `WithExternalRef`, `BoxInfo.ExternalRef`.
- **Node SDK**: field filled with `None` (orchestrator-facing knob; not exposed in JS options yet).
- **CLI**: `inspect` shows `ExternalRef` only when present; `list` columns unchanged (nothing empty is rendered).
- **runner**: `WithName(dto.Id)` → `WithExternalRef(dto.Id)`; restart recovery (`runtime.Get(dto.Id)`) resolves through the external_ref match.

## Verification

- engine lib tests: 735/735; the new `lookup_box_by_external_ref` test was verified two-sided (fails with the lookup clause reverted, passes restored)
- boxlite-c: 55/55
- CLI / Node / Python crates compile; `go vet` clean on `sdks/go` and `apps/runner/pkg/boxlite` (final cgo link needs the prebuilt `libboxlite.a`, absent in this worktree)

## Merge order

#743 → #735 → #744 → this PR. Self-contained: reverting is closing this PR.

### DIFF
--- a/apps/runner/pkg/boxlite/client.go
+++ b/apps/runner/pkg/boxlite/client.go
@@ -221,7 +221,10 @@ func (c *Client) Create(ctx context.Context, boxDto dto.CreateBoxDTO) (string, s
 		memoryMiB = 128
 	}
 	opts := []boxlite.BoxOption{
-		boxlite.WithName(boxDto.Id),
+		// Attach the control-plane box id as the orchestrator reference (CRI
+		// PodSandboxMetadata.uid analog); the engine mints its own internal id
+		// and the name slot stays free for a future user-facing name.
+		boxlite.WithExternalRef(boxDto.Id),
 		boxlite.WithCPUs(cpus),
 		boxlite.WithMemory(memoryMiB),
 		boxlite.WithAutoRemove(false),
--- a/sdks/c/include/boxlite.h
+++ b/sdks/c/include/boxlite.h
@@ -237,6 +237,8 @@ typedef void (*CBoxImageListCb)(struct CImageInfoList*, CBoxliteError*, void*);
 typedef struct CBoxInfo {
   char *id;
   char *name;
+  // Orchestrator-issued external reference; NULL when unset.
+  char *external_ref;
   char *image;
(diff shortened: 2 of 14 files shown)


G4614 (approved):
can

PR #748, opened 2026-06-12
refactor(contracts): single-owner API contracts — cloud API is cloud-only, Box API owns box operations

law-chain-hot:
## Architecture: single owner per contract

```text
BEFORE — box operations & dead surface scattered across contracts
┌─────────────────────────────────────────────────────────────────────┐
│ Cloud API (NestJS openapi.json)   orgs/billing/keys ... + 4 box verbs │ ← box ops leak in
│ toolbox.deprecated  (68 endpoints)                only consumer: TS SDK│ ← dead
│ workspace.deprecated (Daytona alias)              0 consumers          │ ← dead
│ Box API (openapi/box.openapi.yaml)               box create/exec/...   │
└─────────────────────────────────────────────────────────────────────┘

RULE:  an endpoint lives where its capability lives
       engine can fulfil it      → Box API
       needs cloud ingredients   → Cloud API
       (signing keys, gateways, billing clock)

AFTER — one owner each
┌──────────────────────────────┐        ┌──────────────────────────────┐
│ Cloud API  (cloud-only)       │        │ Box API  (the only box-ops    │
│  orgs, billing, api keys,     │        │           contract)           │
│  audit, webhooks,             │        │  create / start / stop /      │
│  signed preview URLs, ssh,    │        │  delete, exec, files,         │
│  archive/recover, org reads   │        │  snapshots                    │
└──────────────────────────────┘        └──────────────────────────────┘
        served by NestJS                  served by apps/api/src/boxlite-rest
        + a drift-workflow guard: cloud spec must never re-grow
          /toolbox//exec//files//workspace; Box API spec stays tenancy-free

DELETED (the bulk of the line count):
  toolbox.deprecated (68 eps)  ·  workspace.deprecated  ·  4 box verbs in box.controller
  apps/libs/toolbox-api-client (generated, 0 callers)  ·  sdk-typescript (retired)
  dashboard Playground/VNC suite (flag default-off)
```

## Why ~90k deletions is not as large as it looks

```text
~95%  deleting GENERATED clients (api-client-go, toolbox-api-client) +
      the retired TS SDK + the dead Playground/VNC suite
~1–2k actual hand-written change: contract rewiring (verb ownership,
      metrics route, handwritten cloudBox Box-API client, drift guard)
```

## Impact

- **Breaking**: removes `toolbox.deprecated` (68 endpoints), `workspace.deprecated`, and 4 box verbs from `box.controller`. Only known consumer was the TS SDK (removed in this PR) — confirm no external callers before merge.
[...]

PR #750, opened 2026-06-12
fix(e2e): skip path_verification tests under BOXLITE_E2E_SKIP_PATH_VERIFY

G4614:
Both checks in `scripts/test/e2e/cases/test_path_verification.py` are local-stack-only — `:3000 in url` and `journalctl -u boxlite-runner` on the pytest host. The cloud e2e workflow already sets `BOXLITE_E2E_SKIP_PATH_VERIFY: '1'` to opt out of the runner-journal autouse fixture; this PR extends the same gate to skip the two top-level tests so they don't fail against the Tokyo ALB.

## Test plan

| | Before | After |
|---|---|---|
| `BOXLITE_E2E_SKIP_PATH_VERIFY=1 pytest .../test_path_verification.py` | 2 FAILED (`:3000` assert, journalctl shell-out) | 2 SKIPPED (reason: "path-verification tests are local-only ...") |
| Unset env (local run) | 2 PASS / FAIL based on local stack | Unchanged — `pytest.mark.skipif` only fires when the env var is truthy |
| Truthy spellings `1/true/yes/on` (case-insensitive) | n/a | Skip |
| Other values incl. `0/false`/unset | Skip (legacy `if os.environ.get(...)` was truthy on any non-empty string — see #748 for the conftest counterpart) | Do not skip |

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed unused dependency from application package.

* **Chores**
  * Updated API schema defaults for role assignment to use empty arrays instead of placeholders.
  * Enhanced CI/CD infrastructure with improved deployment workflows and test automation for cloud-based testing.
  * Added environment variable option to control E2E test behavior for flexible testing scenarios.


G4614:
Closing as stale/superseded during PR queue cleanup. If this still applies, it should be reopened as a small PR against current main.

PR #753, opened 2026-06-13
feat(box): image pipeline + single box identity (drop legacy boxId)

law-chain-hot:
## Summary

Consolidated **box image + identity refactor** in one PR (base `main`). Three coherent, jointly-tested changes:

1. **Image pipeline** — boxes boot from supported pinned OCI refs; `image` is a first-class `Box` column (no key→ref translation); runner contract field is `image`; dashboard image picker; e2e adapted.
2. **Single box identity** — collapse the dual uuid/`boxId` into one 12-char base62 id (the PK). The internal uuid (Daytona leftover) is gone; one id threads PG → job → runner → engine.
3. **Drop legacy `boxId` DTO field + regenerate clients** — remove the `boxId` mirror from `BoxDto`/admin DTO, regenerate the 4 API clients (CI-faithful, run inside a BoxLite box), dashboard reads `.id`.

### Notes
- **`external_ref` is intentionally NOT included.** The cloud box id rides in the engine's `name` slot (the engine mints its own internal id); user-facing names live only in PG. This keeps the runtime contract minimal — no dedicated orchestrator-reference field.
- **The TS cloud SDK removal is NOT part of this PR.** The dashboard image-picker / identity code uses the `cloudBox` adapter (added here) instead of `@boxlite-ai/sdk`, but the `sdk-typescript` package itself is not deleted in this diff. (The branch's merge-base is #743, which was later reverted from `main` by #751 — so the deletion does not appear here and `sdk-typescript` still exists on `main`.) The SDK package removal will be handled as its own follow-up; this PR only migrates the dashboard to `cloudBox`.

## Verification
- Engine lib: 734/734 · boxlite-c: 55/55 · Go `vet` clean (sdks/go + runner pkg; final cgo link needs prebuilt libboxlite.a, absent in worktree)
- API jest: 112/112 (from the consolidated branch; unchanged by the external_ref strip)
- Dashboard `tsc --noEmit` clean · API clients regenerated inside a BoxLite box (CI-faithful)
- Net diff vs `main` contains zero `external_ref` references.

## Deployment
Baseline migration edited in place (box PK = `varchar(12)`, `image` column added, `boxId` dropped) → **requires a DB rebuild** (agreed; nothing launched).

## Release Notes

* **New Features**
  * Box creation now requires selecting from a curated list of supported OCI container images, ensuring consistency and security
  * Streamlined box identification with a unified ID format

* **Changes**
  * Updated API contracts: `image` field is now required for box creation; separate box identifier field removed

### 1. Image: full OCI ref, end-to-end, no translation layer

```text
create request                 PG (box row)            CREATE_BOX payload          runner / engine
  image: "ghcr.io/...@sha256"  ──►  box.image  ───────►  image (verbatim)  ────────►  runtime.Create(image)
        │                            (first-class column)
        ▼
  assertSupportedImage(image)   ← thin allowlist gate at the request boundary only.
   not in allowlist → 400 listing the supported refs.
   Deletable when custom images land; no other layer knows the curated set exists.
```

The pinned OCI ref is the single representation of "image" at every layer — there is no
curated-key ⇄ ref translation step.

### 2. Identity: one 12-char base62 id, no second id

```text
            id = "a7Kf93mQpX2z"   (minted in-memory at create; THE identity everywhere)
                  │
   ┌──────────────┼───────────────────────────────────────────────────────────┐
   ▼              ▼                         ▼                                   ▼
 PG boxes.id    jobs.resourceId        runner box map key                engine VM name
 (primary key)  (= id)                 (= id)                            (WithName(id))
   │              │                       │                                   │
   └─ findOneByIdOrName(id | name)        └─ proxy /boxes/{id}/...             └─ lookup by name slot

cloud box id rides in the engine's name slot (the engine still mints its own internal id,
which never surfaces). No uuid, no separate boxId column, no external_ref field.
```

`uuid` (a Daytona leftover) and the duplicate `boxId` column/DTO field are removed: it offered
neither a smaller-ordered-key benefit nor ABA protection, and kept leaking into runner logs.

### Two ways the CLI/API resolve a box

```text
metadata read (inspect / ls)   →  PG findOneByIdOrName(id | name)         [stops at PG]
operation     (exec / stats)   →  PG resolve → id → runner → engine VM    [reaches the box]
```

### What this PR does NOT add

```text
external_ref (CRI-uid-style field)  →  intentionally excluded.
  The cloud id rides in the engine name slot; user-facing names live only in PG.
  Keeps the runtime contract minimal (no dedicated orchestrator-reference field).
```


law-chain-hot:
Thanks @coderabbitai — all five addressed in 127a7a7c. Quick notes:

- **create-box.dto `image` `@ApiPropertyOptional`** ✅ added (it was the only field on this DTO missing it).
- **adminHelpers `findBoxById`** ✅ now exact, case-sensitive — box ids are 12-char base62, so lowercasing could resolve the wrong box. Good catch.
- **openapi `image` default/example** ✅ switched to pinned ghcr refs so they no longer contradict the allowlist wording.
- **e2e `UUID_RE`** ✅ replaced with an anchored 12-char base62 matcher (`BOX_ID_RE`) across all CLI/SDK entry tests; assertion text updated uuid → box id.

- **warm-pool.spec digest** ✅ fixed, but **not** the suggested way. Asserting `box.image` against `assertSupportedImage(undefined)` would make the test tautological (production sets `box.image = assertSupportedImage(undefined)`, and the assertion would call the same function) — nothing would cross a boundary. Instead I isolate `BOXLITE_SYSTEM_*` env in `beforeEach`/`afterEach` so the pinned fallback is deterministic, and keep the literal assertion. This removes the env-driven flakiness while still proving "default → base ref".

law-chain-hot:
Superseded by #756 — same image + single-identity + drop-boxId work, rebuilt cleanly against main: no cloudBox migration, no SDK package removal (those stay in #748). Only touches sdk-typescript/src/Box.ts as a field mirror so it compiles after the regen.

PR #756, opened 2026-06-13
feat(box): image pipeline + single box identity (drop dual boxId)

law-chain-hot:
## Summary

Backend image + identity refactor against `main`. **No dashboard cloudBox migration, no TS SDK package removal** — those stay in #748.

1. **Image pipeline** — `image` is a first-class `Box` column; the full OCI ref flows end-to-end (request → PG → CREATE_BOX payload → `runtime.Create`) with no curated-key⇄ref translation. A thin allowlist gate validates at the request boundary only. Runner contract field is `image`. e2e adapted.
2. **Single box identity** — collapse the dual uuid/`boxId` into one 12-char base62 id (the PK). The internal engine id is never exposed; `boxId` (the redundant duplicate of the cloud id) is dropped everywhere.
3. **Regenerate the 4 API clients** (CI-faithful, run in a BoxLite box): `Box`/`AdminBoxItem`/`Workspace` lose `boxId`, gain `image`.
4. **Consumers of the dropped `boxId` fixed in place** (this is NOT the SDK removal): dashboard reads switch `boxId`→`id` (6 files); `sdk-typescript`'s `Box` mirrors the new api-client shape (drop `boxId` field, add `image`). The SDK **package** is untouched.

```text
id model:   uuid + boxId   ──►   one 12-char base62 id (cloud id)
                                  engine mints its own internal id, never exposed
image:      key/translation ──►   full OCI ref, single representation, no translation
```

## Scope boundary (what's deferred to #748)

- Removing the `sdk-typescript` package, the dashboard `cloudBox` migration, the Playground/VNC retirement, and the image-picker UI.
- This PR only touches `sdk-typescript/src/Box.ts` (field mirror) so it keeps compiling after the regen — the package deletion is #748's job.

## Verification

- `sdk-typescript` build OK · dashboard `tsc --noEmit` clean · API jest 112/112 · `go vet` clean (runner pkg) · engine unchanged vs `main`.
- API clients regenerated inside a BoxLite box (CI-faithful).

## Deployment

Baseline migration edited in place (box PK = `varchar(12)`, `image` column added, `boxId` dropped) → **requires a DB rebuild** (agreed; nothing launched).

## Release Notes

* **New Features**
  * Added OCI image reference support for box creation with a curated allowlist of supported images.
  * Simplified box identity to use a single unified identifier.

* **Breaking Changes**
  * Updated `Box` schema: removed `boxId` field, added required `image` field.
  * Updated `Workspace` schema: removed `boxId`, made `image` required.
  * Updated `AdminBoxItem` schema: `organizationId` is now optional.

### DIFF
--- a/apps/api-client-go/api/openapi.yaml
+++ b/apps/api-client-go/api/openapi.yaml
@@ -9565,8 +9565,7 @@ components:
           - member
           type: string
         assignedRoleIds:
-          default:
-          - 00000000-0000-0000-0000-000000000001
+          default: []
           description: Array of assigned role IDs
           items:
             type: string
@@ -9596,8 +9595,7 @@ components:
           - member
           type: string
         assignedRoleIds:
-          default:
-          - 00000000-0000-0000-0000-000000000001
+          default: []
           description: Array of assigned role IDs for the invitee
           items:
             type: string
@@ -9939,12 +9937,12 @@ components:
     Box:
       example:
         id: fd955d93-e74a-48e7-9f2d-fcbe6dd9e920
-        boxId: aB3cD4eF5gH6
         organizationId: organization123
         name: MyBox
         user: boxlite
         env:
           NODE_ENV: production
+        image: ghcr.io/boxlite-ai/boxlite-agent-python@sha256:80d562a57f4bc12def4e54dbdb9e7d26d3268fe0767a2955ab5ad718041145d6
         labels:
           boxlite.io/public: "true"
         public: false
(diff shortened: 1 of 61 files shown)


law-chain-hot:
Split up. Superseded by:
- #758 — the curated image allowlist (the unique, security-relevant piece; main's #755 only added a free-form image field).
- single-id collapse (drop uuid/boxId) → separate PR after the SDK/dashboard removal lands, since boxId's only consumers are the TS SDK and the legacy dashboard, which #748 removes.
The image pipeline itself is already on main via #755, so it's dropped here.

PR #768, opened 2026-06-15
fix(ssh-gateway): pin runner host key for the gateway→runner hop

G4614:
## Summary

Source-level security audit fix. The SSH gateway dialed the runner with
`ssh.InsecureIgnoreHostKey()`, so a MITM on the gateway↔runner path could
impersonate the runner and intercept the proxied interactive shell.

## Fix

Add `buildRunnerHostKeyCallback`, which pins the runner host key via
`ssh.FixedHostKey` when `RUNNER_HOST_KEY` (base64 public key) is configured —
mirroring how the gateway already loads its own `SSH_HOST_KEY`. When unset, the
previous behavior is preserved (no deployment breakage) but the downgrade is
logged loudly so operators can opt into verification.

## Test (two-sided)

`TestBuildRunnerHostKeyCallback_PinsConfiguredKey` accepts the matching host
key and rejects a different one; reverting the callback to the pre-fix
`ssh.InsecureIgnoreHostKey()` makes it accept the mismatched key and the test
fails ("MITM not prevented").

Audit finding #8 (medium).

### DIFF
--- a/apps/ssh-gateway/main.go
+++ b/apps/ssh-gateway/main.go
@@ -31,11 +31,42 @@ const (
 )
 
 type SSHGateway struct {
-	port       int
-	apiClient  *apiclient.APIClient
-	hostKey    ssh.Signer
-	privateKey ssh.Signer
-	publicKey  ssh.PublicKey
+	port                  int
+	apiClient             *apiclient.APIClient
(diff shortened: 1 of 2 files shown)


G4614:

PR #778, opened 2026-06-15
[WIP] feat(net): optional allow_net filtering for UDP

G4614:
> ⚠️ **[WIP — decision logic tested; live forwarder NOT integration-verified]**
> Part of a source-level security audit. The UDP policy (`udpAllowed`) is unit
> tested; the live forwarder is gated behind `BOXLITE_UDP_FILTER=true` (OFF by
> default) because its interaction with gvproxy's embedded DNS resolver and the
> datagram relay path could not be validated without a network-stack harness.

## Problem

`allow_net` only filtered TCP. UDP used gvproxy's default NAT forwarder, so a
restricted box could exfiltrate to **any** host over UDP/QUIC (DNS-over-UDP to an
attacker on :53, HTTP/3 on :443, custom UDP channels). The DNS sinkhole governs
only name resolution, not datagram destinations.

## Change
- `udpAllowed(destIP, destPort, filter)` — drops UDP to destinations not permitted
  by the AllowNet filter; reuses `TCPFilter.MatchesIP`, so gateway/internal IPs
  (incl. the DNS resolver) and allowlisted hosts pass. No filter → all UDP allowed.
- `OverrideUDPHandler` installs a filtered `udp.NewForwarder` (mirrors the TCP
  override). Wired in `main.go` behind `BOXLITE_UDP_FILTER=true`, **OFF by
  default** so the live path is unchanged.

## Test (two-sided) — policy, runs locally
`TestUDPAllowed`: attacker `:53`/`:443` and unlisted hosts blocked; allowlisted +
gateway DNS pass. Reverting `udpAllowed` to "always allow" (pre-fix) makes the
blocked cases pass and the test fails. Full bridge suite + `go vet` green.

## What needs a resource to verify (before enabling by default)
- A **gvisor netstack integration harness** (the repo has none — `forked_network_test`
  only reflect-checks the `stack` field) to drive real UDP/DNS packets and confirm:
  (a) non-allowlisted UDP is dropped end to end, (b) **DNS still resolves** (i.e.
  overriding the UDP handler does not bypass the embedded resolver endpoint), and
  (c) the datagram relay/idle-timeout works for allowed flows.

Audit finding #2 (high).

### DIFF
--- a/src/deps/libgvproxy-sys/gvproxy-bridge/forked_udp.go
+++ b/src/deps/libgvproxy-sys/gvproxy-bridge/forked_udp.go
@@ -1,0 +1,149 @@
+package main
+
+// forked_udp.go — Optional AllowNet filtering for UDP.
+//
+// allow_net only constrained TCP: UDP went through gvproxy's default NAT
+// forwarder, so a restricted box could exfiltrate to any host over UDP/QUIC
+// (audit finding #2). This installs a UDP forwarder that drops datagrams to
+// destinations not permitted by the AllowNet filter.
+//
+// WIP / NOT INTEGRATION-VERIFIED: the policy decision (udpAllowed) is unit
+// tested, but the live forwarder path — in particular whether overriding the UDP
+// protocol handler bypasses gvproxy's embedded DNS resolver endpoint, and the
+// datagram relay/idle-timeout behavior — has NOT been validated against a running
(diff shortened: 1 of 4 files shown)


G4614:
Superseded by #1090, which applies allow_net to UDP by default, reuses the upstream UDP forwarder, fails closed on handler installation, and includes end-to-end virtual-network coverage.

PR #781, opened 2026-06-15
fix(seatbelt): narrow macOS sandbox write grants under built-in profile

G4614:
Drops the static seatbelt `(subpath "/private/var/folders")` write grant under the built-in macOS profile.

That grant was vestigial: `vmm/controller/spawn.rs::configure_env` already redirects the shim's `TMPDIR` to a box-scoped path whenever `jailer_enabled && sandbox_profile.is_none()`, and since #742 every box socket binds through `/tmp/bl-{uid}/{box_id}/` (added dynamically via `BoxSockets::policy_paths()`). With both already in place, no per-user temp tree write was actually needed.

Updated TODO in `seatbelt_network_policy.sbpl` documents why narrowing `(allow network-inbound)` is non-trivial (port list is runtime data, doesn't fit `sandbox-exec -D`; `network-bind` vs `network-inbound` distinction matters for the gvproxy NAT return path; needs real EXPOSE e2e). Out of scope for this PR.

## Test plan (macOS arm64)
- [x] Two-side verification of new regression test `test_file_write_policy_excludes_per_user_temp_tree`:
  - Re-added `(subpath "/private/var/folders")` → test **FAILS** at the grant-form check
  - Restored → **27/27** seatbelt unit tests pass, including 4 `test_seatbelt_runtime_*` that spawn real `sandbox-exec`
- [x] **Live VM happy path**: `boxlite run --rm alpine:latest sh -c '...'` → guest boots (Linux 6.12.76 aarch64), command runs, output returned (`uid=0(root)`), auto-stopped. No EPERM / sandbox denial.
- [x] **Live VM EXPOSE port forwarding**: `boxlite run -d -p HOST:8080 alpine:latest sh -c 'nc -l -p 8080'` → host `curl 127.0.0.1:HOST/` returned guest payload via gvproxy NAT; `boxlite stop` cleaned up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restricted temporary file write access to system-only directories.

* **Tests**
  * Added sandbox permission validation tests.

### DIFF
--- a/src/boxlite/resources/seatbelt/seatbelt_file_write_policy.sbpl
+++ b/src/boxlite/resources/seatbelt/seatbelt_file_write_policy.sbpl
@@ -5,15 +5,28 @@
 ; Static write paths for temporary files.
 ; These are needed by many tools and libraries for temporary storage.
 ;
-; Box shared directory and user volumes are added dynamically by seatbelt.rs.
+; Box shared directory, box-scoped tmp, and user volumes are added dynamically
+; by seatbelt.rs.
 ;
 ; Why these paths are needed:
-;   - /private/tmp: Standard temporary directory (macOS canonical path)
-;   - /private/var/tmp: Alternative temporary directory
+;   - /private/tmp: Standard temporary directory (macOS canonical path).
+;     Required for the short binding-symlink parent `/tmp/bl-{uid}/{box_id}`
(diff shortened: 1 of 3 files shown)


DorianZheng (approved):
LGTM

PR #800, opened 2026-06-16
Use cloud Box ID as primary identity

law-chain-hot:
## Summary
- Make the 12-character Cloud Box ID the primary `box.id` and remove the legacy `Box.boxId` entity/DTO/API-client alias.
- Update lookup, cache invalidation, REST mapping, admin/dashboard display, and runner create DTO paths to use `id` directly.
- Regenerate the Cloud API clients so `Box` / admin box item models and list sorting no longer expose `boxId`.

## Notes
- This PR intentionally takes the rebuild/reset-DB path. Fresh databases created from the baseline migration have `box.id` as `varchar(12)` and no `box.boxId` column.
- Existing deployed DBs that already ran the baseline migration will not be converted in place by this PR. Rebuild/drop the existing DB state before deploying this branch if old UUID `box.id` rows or `box.boxId` columns exist.
- `box_last_activity.boxId` and `ssh_access.boxId` remain as relation column names; both foreign-key to `box(id)` and are not the removed `Box.boxId` field.
- Runner HTTP route parameters named `{boxId}` are path variable names for the public box identity and remain valid.

## Validation
- Remote dev machine: `make lint:apps` passed with existing warnings only.
- Remote dev machine: `make test:apps` passed (`36` API suites / `130` tests; app Go test targets passed or matched cache).
- Remote dev machine: `VERSION=0.0.0-dev make build:apps` passed.
- `git diff --check` passed locally.

### DIFF
--- a/apps/api/src/admin/services/observability.service.spec.ts
+++ b/apps/api/src/admin/services/observability.service.spec.ts
@@ -398,7 +398,6 @@ describe('AdminObservabilityService', () => {
     overviewService.listBoxes.mockResolvedValue([
       {
         id: 'box-1',
-        boxId: 'public-box-1',
         organizationId: 'org-1',
         state: 'started',
         runnerId: 'runner-1',
@@ -408,7 +407,6 @@ describe('AdminObservabilityService', () => {
       },
       {
         id: 'box-2',
-        boxId: 'public-box-2',
         organizationId: 'org-1',
         state: 'started',
         runnerId: 'runner-2',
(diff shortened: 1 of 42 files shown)


apps/api/src/admin/services/observability.service.spec.ts
```diff
@@ -429,24 +427,23 @@ describe('AdminObservabilityService', () => {
     expect(result.correlation).toMatchObject({
       traceIds: ['trace-box-1'],
       orgIds: ['org-1'],
-      boxIds: ['box-1', 'public-box-1'],
+      boxIds: ['box-1'],
```
DorianZheng, line 430: no `boxIds`?

DorianZheng (approved):

PR #803, opened 2026-06-16
feat(dashboard): box table cleanup — drop region, tighten density, full image refs in onboarding

law-chain-hot:
## What
**Box table**
- Remove the **Region** column (desktop table + compact mobile meta).
- Drop the stale `min-w-[1360px]` (sized for now-removed columns) to `min-w-[1120px]` so the table fills the viewport instead of horizontally scrolling / stretching columns.
- Tighten row vertical padding (`py-2`→`py-1`, scoped to this table) so rows are ~41px instead of ~46px.
- Remove the now-unused `getRegionName` plumbing from the box-table path and the stale `region` entry in persisted column visibility. `useRegions`/`getRegionName` stay for other consumers (Runners, Org Settings, Box details).

**Responsive**
- Lower `COMPACT_BREAKPOINT` 1200→1024 (`use-mobile`) so the dashboard keeps the full table down to 1024px instead of switching to the compact card layout at 1200px. Shared by Sidebar / Pagination / BoxTableHeader / BoxTable.

**Onboarding**
- Use the fully-qualified image reference `ghcr.io/boxlite-ai/boxlite-agent-base:20260605-p0-r3` across all four SDK snippets (JS/TS, Python, Go, Rust) instead of the short `boxlite/base`.

## Scope
Frontend only. Two unrelated API errors seen during testing (`/api/regions` 500 dev DB drift; `/api/admin/overview` 403 admin-probe) tracked separately.

## Release Notes

* **UI Updates**
  * The Box table now focuses on box images instead of region information, with consistent presentation in both desktop and mobile views.
  * The “Region” column was removed.
  * Mobile/compact layout now activates at a lower screen width, and table sizing/padding were adjusted for improved readability.
  * Default column visibility was updated to align with the new image-focused layout.

* **Chores**
  * Updated the Docker image reference used by onboarding code examples to a newer build.

### DIFF
--- a/apps/dashboard/src/components/BoxTable/columns.tsx
+++ b/apps/dashboard/src/components/BoxTable/columns.tsx
@@ -52,7 +52,6 @@ interface GetColumnsProps {
   handleCreateSshAccess: (id: string) => void
   handleRevokeSshAccess: (id: string) => void
   handleRecover: (id: string) => void
-  getRegionName: (regionId: string) => string | undefined
   handleScreenRecordings: (id: string) => void
 }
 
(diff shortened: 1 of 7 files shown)


DorianZheng (approved):
LGTM

PR #804, opened 2026-06-16
fix(cli): surface cloud-exec stream failures instead of silent success

DorianZheng:
## What
Cloud `exec` streams stdout/stderr over a WebSocket. When that attach fails
(e.g. a 401 on the upgrade), the client recovered the exit code via a status
probe and reported a silent `exit 0` with no output — indistinguishable from a
command that genuinely produced none.

## Changes
- `emit_or_fallback` now carries the disconnect cause in `ExecResult.error_message`
  (was `None`); it is only reached when the output stream failed.
- `StreamManager::start()` surfaces a non-empty `error_message` as an error
  (exit 1 + message) instead of dropping it.
- REST `401`/`403` mapping reads `unauthorized (HTTP 401)` / `forbidden (HTTP 403)`
  instead of a bare `auth: <body>`.
- Adds `debug!`/`trace!` to the previously-uninstrumented WS attach path
  (connect, per-frame, stdout/stderr, exit-frame, status-probe fallback).

## Verification
- Real cloud `exec`: before = silent `exit 0`, empty; after = `exit 1` + clear cause.
- Unit reproducer `terminal::tests::stream_manager_surfaces_exec_error_message_as_error` verified RED→GREEN.
- 4 WS attach unit tests still pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Distinguished error messages for authentication (HTTP 401) versus authorization (HTTP 403) failures
  * Improved error reporting when process execution streaming is interrupted

* **Chores**
  * Enhanced diagnostic logging for WebSocket connection handling

### DIFF
--- a/src/boxlite/src/rest/error.rs
+++ b/src/boxlite/src/rest/error.rs
@@ -56,7 +56,13 @@ pub(crate) fn map_http_error(status: StatusCode, body: &ErrorModel) -> BoxliteEr
 pub(crate) fn map_http_status(status: StatusCode, text: &str) -> BoxliteError {
     match status.as_u16() {
         404 => BoxliteError::NotFound(text.to_string()),
-        401 | 403 => BoxliteError::Config(format!("auth: {}", text)),
+        // Keep the `auth:` prefix (callers key on it) but state the actual
+        // failure: 401 = credentials rejected (expired, or wrong credential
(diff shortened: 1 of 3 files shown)


G4614 (approved):
can

PR #806, opened 2026-06-16
fix(dashboard): recover from a 401 by bouncing to a fresh login

DorianZheng:
## What
The dashboard showed a dead-end "Unauthorized" screen when the access token
became invalid — most commonly when the local Dex box was recreated, rotating
its signing keys, so an unexpired-but-invalid token kept being replayed
(oidc-client-ts only tracks local expiry and can't detect this).

## Change
On a 401 the API client drops the OIDC session (`removeUser`), which the existing
ApiProvider effect turns into a `signinRedirect` (preserving `returnTo`). Recovery
is bounded to one attempt per session via a `sessionStorage` marker (a token still
rejected after re-login surfaces an error instead of bouncing forever); concurrent
401s de-dupe to a single redirect (guard checked before the marker); a failed
redirect start surfaces an error rather than hanging.

Note: this is the minimal scope — the separate billing/analytics clients still
hold their own token and don't share this recovery (pre-existing; tracked as a
follow-up).

## Verification
- Reproducer `apps/dashboard/src/api/apiClient.test.ts` verified RED→GREEN on the 401 path.
- Full dashboard suite green; typecheck clean on touched files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced session recovery handling to prevent infinite redirect loops and duplicate re-authentication attempts when tokens expire.

### DIFF
--- a/apps/dashboard/src/api/apiClient.test.ts
+++ b/apps/dashboard/src/api/apiClient.test.ts
@@ -1,0 +1,65 @@
+// @vitest-environment jsdom
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+const config = { apiUrl: 'http://api.test/api' } as never
+
+// Fresh module per test so the module-level `isHandlingUnauthorized` guard
+// doesn't leak across cases. A custom axios adapter makes every request resolve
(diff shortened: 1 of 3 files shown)


law-chain-hot (approved):
LGTM

PR #809, opened 2026-06-16
feat(api): WS attach accepts OIDC JWTs, not just API keys (closes #807)

### LINKED ISSUE #807: Cloud exec output silent for OIDC/browser logins: WS attach proxy rejects JWTs (API-key only)
## Problem
Cloud `exec` (and any `/attach`) streams stdout/stderr over a WebSocket. The API
WS proxy authenticates the upgrade **only as an API key** —
`BoxliteWsProxyService.authenticate()`
(`apps/api/src/boxlite-rest/boxlite-ws-proxy.service.ts`) calls
`apiKeyService.getApiKeyByValue(token)` with no JWT path (the code comment is an
explicit TODO).

So a **browser/OIDC login** (Auth0 JWT) authenticates fine for REST (exec create,
status, list all 200) but the **WS `/attach` upgrade is 401**, so output never
streams. Symptom: correct exit code, empty stdout/stderr.

Repro: `boxlite --profile cloud auth login` (Browser/OIDC), then
`boxlite --profile cloud exec <box> -- echo hi` → no output. (As of #804 this
now fails loudly instead of a silent `exit 0`.)

## Workaround
`boxlite auth login` with an **API key** instead of browser login.

## Fix (design ready)
In `authenticate()`, after the API-key lookup fails, fall through to JWT:
- reuse `JwtStrategy.verifyToken` (`apps/api/src/auth/jwt.strategy.ts:105` — same
  JWKS/issuer/audience; **do not hand-roll** verification),
- map `sub`→user (mirror the `cid`/`uid` handling),
- take the URL tenant (ATTACH_PATH's optional, currently non-captured
  `(?:[^/]+/)?` group) as the org,
- **membership-check** the user against that org via
  `organizationUserService.findOne` (the cross-org boundary — exactly what the
  API-key path does),
- reject when the URL has no tenant (org is ambiguous for a multi-org JWT user),
- DI-wire `JwtStrategy` into the boxlite-rest module.

⚠️ Security boundary (cross-org box access) — needs review.

DorianZheng:
## What
The API's WS `/attach` proxy authenticated upgrades **only as an API key**, so a
browser/OIDC login (Auth0 JWT) — which works for REST — was rejected 401 on the
attach upgrade and cloud `exec` output never streamed. Closes #807. (PR #804
already made the CLI surface this loudly instead of a silent `exit 0`.)

## Change
`BoxliteWsProxyService.authenticate` now falls through to JWT when the bearer is
not an API key:
- verify via the existing `JwtStrategy.verifyToken` (same JWKS/issuer/audience as
  the HTTP path — no hand-rolled verification),
- map `sub`/`uid` like `JwtStrategy.validate`,
- take the org from the URL `{prefix}` tenant (`ATTACH_PATH` now captures it),
- require org membership — identical to the API-key path.

**Mirrors current behavior exactly** (per scope discussion): membership-only — no
permission gate (there is no `box:exec` permission and the HTTP exec/attach routes
enforce none), no rate-limiting, and a JWT with no/`default` tenant is rejected
(org ambiguous for a multi-org user). `JwtStrategy` is nullable under
`skipConnections`; `authenticate` never throws, so the fire-and-forget upgrade
caller can't leak a hung socket.

## Verification
- `boxlite-ws-proxy.service.spec.ts`: API-key regression + JWT
  member / non-member / cross-org / no-tenant / invalid-token / skipConnections /
  `uid`-claim, plus `matchAttachPath` tenant capture. Verified RED→GREEN (the JWT
  acceptance + membership-check tests fail when the JWT path is neutered).
- Full `nx test api` green (140 tests).

## ⚠️ Security review
Cross-org access boundary. The membership check
`organizationUserService.findOne(urlTenant, userId)` is what stops a JWT user
attaching to boxes in an org they don't belong to — please scrutinize that and
the URL-tenant extraction.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket `/attach` upgrade authentication by adding tenant-aware routing and validating both API keys and JWTs with correct precedence.
  * Clarified legacy vs tenant-prefixed attach paths and ensured unauthorized connections are rejected cleanly.

* **Tests**
  * Expanded WebSocket authentication tests to cover tenant extraction, matching/non-matching attach paths, API-key and JWT authorization rules, claim mapping, and failure cases (including missing/invalid tokens and missing tenant context).

### DIFF
--- a/apps/api/src/boxlite-rest/boxlite-rest-routing.spec.ts
+++ b/apps/api/src/boxlite-rest/boxlite-rest-routing.spec.ts
@@ -87,11 +87,15 @@ describe('BoxLite REST routing', () => {
   })
 
   it('matches websocket attach upgrades with or without a routing prefix', () => {
-    const service = new BoxliteWsProxyService({} as any, {} as any, {} as any, {} as any)
+    const service = new BoxliteWsProxyService({} as any, {} as any, {} as any, {} as any, {} as any)
 
-    expect(service.matchAttachPath('/api/v1/boxes/box-1/executions/exec-1/attach')).toEqual({ boxId: 'box-1' })
+    expect(service.matchAttachPath('/api/v1/boxes/box-1/executions/exec-1/attach')).toEqual({
+      boxId: 'box-1',
+      tenant: undefined,
+    })
     expect(service.matchAttachPath('/api/v1/default/boxes/box-1/executions/exec-1/attach')).toEqual({
       boxId: 'box-1',
+      tenant: 'default',
     })
   })
 })
--- a/apps/api/src/boxlite-rest/boxlite-ws-proxy.service.spec.ts
+++ b/apps/api/src/boxlite-rest/boxlite-ws-proxy.service.spec.ts
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: AGPL-3.0
  */
 
+import type { IncomingMessage } from 'http'
 import { createProxyMiddleware } from 'http-proxy-middleware'
 import { BoxliteWsProxyService } from './boxlite-ws-proxy.service'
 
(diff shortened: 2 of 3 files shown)


G4614 (approved):
can

PR #812, opened 2026-06-16
fix(ffi): isolate exec stream delivery and backpressure

G4614:
Move the Go/C SDK and FFI stream-drain/backpressure work out of #841.

Test plan:
- BOXLITE_DEPS_STUB=1 cargo build -p boxlite-c
- cp target/debug/libboxlite.a sdks/go/libboxlite.a
- cd sdks/go && go test . -run 'TestWaitForStreamDrain|TestStreamStall|TestExecutionStream|TestWait|TestBackpressure' -count=1

### DIFF
--- a/sdks/c/include/boxlite.h
+++ b/sdks/c/include/boxlite.h
@@ -417,6 +417,16 @@ enum BoxliteErrorCode boxlite_execution_stdin_write(CExecutionHandle *execution,
 enum BoxliteErrorCode boxlite_execution_stdin_close(CExecutionHandle *execution,
                                                     CBoxliteError *out_error);
 
+// Pause (paused != 0) or resume (paused == 0) delivery of this execution's
+// stdout/stderr. While paused, the stream pumps stop reading their bounded
+// upstream channel, which fills and back-pressures the guest process's
+// write() — letting a slow consumer throttle the producer instead of buffering
+// unboundedly. The Go SDK drives this from its delivery queue's high/low-water
+// marks. Idempotent; safe to call repeatedly.
+enum BoxliteErrorCode boxlite_execution_set_stream_paused(CExecutionHandle *execution,
+                                                          int paused,
+                                                          CBoxliteError *out_error);
+
 enum BoxliteErrorCode boxlite_execution_wait(CExecutionHandle *execution,
                                              CExecutionWaitCb cb,
                                              void *user_data,
--- a/sdks/c/include/boxlite.h
+++ b/sdks/c/include/boxlite.h
@@ -417,16 +417,6 @@ enum BoxliteErrorCode boxlite_execution_stdin_write(CExecutionHandle *execution,
 enum BoxliteErrorCode boxlite_execution_stdin_close(CExecutionHandle *execution,
                                                     CBoxliteError *out_error);
 
-// Pause (paused != 0) or resume (paused == 0) delivery of this execution's
-// stdout/stderr. While paused, the stream pumps stop reading their bounded
-// upstream channel, which fills and back-pressures the guest process's
-// write() — letting a slow consumer throttle the producer instead of buffering
-// unboundedly. The Go SDK drives this from its delivery queue's high/low-water
-// marks. Idempotent; safe to call repeatedly.
-enum BoxliteErrorCode boxlite_execution_set_stream_paused(CExecutionHandle *execution,
-                                                          int paused,
-                                                          CBoxliteError *out_error);
-
 enum BoxliteErrorCode boxlite_execution_wait(CExecutionHandle *execution,
                                              CExecutionWaitCb cb,
                                              void *user_data,
--- a/src/boxlite/src/rest/litebox.rs
+++ b/src/boxlite/src/rest/litebox.rs
@@ -93,8 +93,8 @@ impl BoxBackend for RestBox {
         let execution_id = resp.execution_id;
 
         // 2. Set up channels for stdout, stderr, stdin, and result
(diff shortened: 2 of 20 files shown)


G4614:
Addressed the CodeRabbit review in c977701. Disposition of each thread:

**Fixed (code changes)**
- **rest/litebox WS pump** — `stdout_tx`/`stderr_tx` now use `try_send` (drop-on-overflow) instead of an awaited `send`. The same loop pumps the Exit/Error control frames, so a back-pressured stdout must never block exit propagation. ✔
- **`stallWatchdog` (sdks/go/exec.go)** — the stall timer is now anchored at *stall-start* (first interval where output is buffered **and** progress hasn't advanced), not from the last progress. A quiet exec that buffers nothing is never counted as stalled, so an idle-but-healthy exec is no longer falsely killed. ✔
- **`await_unpaused` comment (sdks/c)** — corrected the stale "sleep-poll (2ms)" note; the impl blocks on the `watch` channel (zero CPU, version-tracked resume). ✔
- **stall-timeout integration test** — now asserts `Wait` returns *cleanly* (nil error) with the SIGKILL surfaced via a non-zero exit code, instead of discarding the error. ✔
- **idle-window release test** — added a lower-bound assertion (`>= 80ms`) so it proves the drain waits ~one idle window before giving up, not just the upper bound. ✔
- Added **box-free `stallWatchdog` unit tests**: a permanently buffered-but-undelivered stream is killed (after ≥ one timeout), and a quiet exec is never killed. Two-side verified — the quiet-exec guard fails against the pre-fix logic and passes after.

**Intentionally unchanged (with rationale)**
- **`writeChunk` swallows sink `Write` errors** — deliberate, now documented in-code: this runs on the async delivery goroutine with no caller to return to, and a transient sink hiccup shouldn't tear down a healthy exec. A *permanently* failing sink stops accepting bytes → output stays buffered → that's exactly what `StreamStallTimeout` watches for and kills. So a wedged sink is still handled without making every transient write error fatal.
- **drain-matrix shape helper ignores `Wait` errors** — these cases assert on *captured output bytes*; exit/`Wait` semantics are covered by the dedicated drain and stall tests. Coupling the shape assertions to exit semantics would only make them more brittle without adding coverage.

CI: clippy `--workspace --all-targets --all-features -D warnings` clean, rustfmt clean, gofmt clean, runner attach/drain/lag stub tests + new watchdog units green locally.

src/boxlite/src/rest/litebox.rs
```diff
@@ -672,13 +672,17 @@ async fn attach_ws_pump(
...
+                                    // slow consumer would stall exit propagation and hang Wait.
                                     0x01 => {
                                         tracing::trace!(len = text.len(), "WS attach: stdout frame");
-                                        let _ = stdout_tx.send(text);
+                                        let _ = stdout_tx.try_send(text);
                                     }
                                     0x02 => {
                                         tracing::trace!(len = text.len(), "WS attach: stderr frame");
-                                        let _ = stderr_tx.send(text);
+                                        let _ = stderr_tx.try_send(text);
```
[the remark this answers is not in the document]
G4614: Addressed in 2e1981d7. Kept the non-blocking `try_send` but routed both stdout/stderr drops through a `note_dropped_frame` helper:
- `Full` → counted, with a single `warn!` on the first drop (so backpressure-induced output loss is visible without flooding logs at one line per dropped frame), and the running total is reported via `warn!` on the clean Exit path.
- `Closed` → `debug!` only; the consumer being gone is the normal teardown race, not a backpressure drop, so it doesn't inflate the dropped count.

Added a box-free unit test (`note_dropped_frame_counts_full_not_closed`) asserting `Full` increments and `Closed` does not. Verified under `--features rest` (the module is feature-gated): the new test plus the other `rest::litebox` pump tests pass; clippy `--all-targets --all-features` and rustfmt are clean.

sdks/c/include/boxlite.h
```diff
@@ -417,6 +417,16 @@ enum BoxliteErrorCode boxlite_execution_stdin_write(CExecutionHandle *execution,
 enum BoxliteErrorCode boxlite_execution_stdin_close(CExecutionHandle *execution,
                                                     CBoxliteError *out_error);
 
+// Pause (paused != 0) or resume (paused == 0) delivery of this execution's
+// stdout/stderr. While paused, the stream pumps stop reading their bounded
+// upstream channel, which fills and back-pressures the guest process's
+// write() — letting a slow consumer throttle the producer instead of buffering
+// unboundedly. The Go SDK drives this from its delivery queue's high/low-water
+// marks. Idempotent; safe to call repeatedly.
+enum BoxliteErrorCode boxlite_execution_set_stream_paused(CExecutionHandle *execution,
```
DorianZheng: should we do this back-pressures automatically?

G4614: sure, in fact the back-pressure is certainly automatically done.
however  the C API boxlite_execution_set_stream_paused should not be exposed to users, so fixed that

G4614: also code comment fixed

G4614:
Replaced by cross-repo PR #841 from `G4614/boxlite:fix/exec-drain-hang-on-sigkill-cross`, which carries the #812 stack plus the binding-only C hook cleanup.

G4614:
Closing this reopened PR because its head branch is in boxlite-ai/boxlite and cannot be updated by G4614. The split FFI/SDK work is being moved to a new draft PR from G4614's fork.

PR #824, opened 2026-06-17
fix(api): transfer non-JSON runner response to standard JSON format

G4614:
normalize runner API failures into structured Nest `HttpException` JSON responses with error msgs

## Why

`POST /v1/:prefix/boxes` should not leak runner/ALB/proxy HTML or plain-text failures to callers. Create failures now stay JSON-shaped, with non-JSON upstream runner responses represented as `runner_non_json_error`.

## Validation

- `git diff --check`
- `./node_modules/.bin/jest --config api/jest.config.ts api/src/box/errors/runner-api-error.spec.ts api/src/filters/all-exceptions.filter.spec.ts --runInBand`
- `BOXLITE_DEPS_STUB=1 cargo test -p boxlite --features rest flat_nest_error_response_maps_by_code`

Note: `yarn nx test api --testFile=...` was attempted first but Nx/Jest ran the broader API suite and hit the existing `nanoid` ESM transform issue in unrelated specs, so the targeted API specs were rerun directly with the API Jest config.

## Non-JSON runner message handling

Non-JSON runner responses now include a sanitized excerpt in the API `message`. HTML tags, script/style blocks, excess whitespace, and obvious token-like query/header fragments are stripped/redacted before the excerpt is exposed. This keeps useful details such as `502 Bad Gateway` or `upstream connect error` visible without returning raw HTML.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runner error translation: map runner 5xx to API **502 Bad Gateway**, and preserve runner 4xx status; include consistent error details.
  * Added safer handling for non-JSON runner responses by sanitizing messages to redact sensitive/token-like content and remove raw markup.
  * Ensure legacy status handling continues to work for manager cleanup paths.
* **Improvements**
  * Error payloads now reliably expose an optional **`code`** field for better diagnostics.
  * Enhanced client-side REST error parsing to support a flattened error schema and map the new non-JSON runner error to **Network**.

### DIFF
--- a/apps/api/src/box/errors/runner-api-error.spec.ts
+++ b/apps/api/src/box/errors/runner-api-error.spec.ts
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2025 Daytona Platforms Inc.
+ * Modified by BoxLite AI, 2025-2026
+ * SPDX-License-Identifier: AGPL-3.0
+ */
+
+import { HttpStatus } from '@nestjs/common'
+import { RunnerApiError } from './runner-api-error'
+import { sanitizedNonJsonRunnerMessage } from '../runner-adapter/runnerAdapter.v0'
+
+describe('RunnerApiError', () => {
+  it('maps runner 5xx errors to API 502 JSON errors', () => {
+    const error = new RunnerApiError('Runner API returned a non-JSON error response', 503, 'runner_non_json_error')
+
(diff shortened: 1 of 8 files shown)


G4614:
@coderabbitai review

DorianZheng (approved):

PR #837, opened 2026-06-23
feat(secrets): support create and live box secrets

G4614:
Adds create-time and live box secret passthrough from the cloud API to runner-managed BoxLite networking.

Test plan:
- `cd apps && npx tsc -p api/tsconfig.app.json --noEmit`
- `cd apps && npx jest --config api/jest.config.ts api/src/box/managers/box-actions/box-start.action.spec.ts api/src/box/services/box.service.secrets.spec.ts api/src/box/services/job.service.secrets.spec.ts api/src/box/runner-adapter/runnerAdapter.v2.secrets.spec.ts`
- `cd apps/runner && go test -tags boxlite_dev ./pkg/api/dto ./pkg/boxlite ./pkg/runner/v2/executor`
- `cd apps && npx nx build api --configuration=development --skip-nx-cache`

## End-to-end verification (Linux jailer)

The feature was verified end-to-end on a Linux host with the production jailer enabled (bwrap + seccomp fully enforced): boot → live admin-socket secret add → real MITM TLS substitution into upstream traffic (placeholder replaced, absent from the wire).

**Two sandbox fixes (now included in commit `1c8f6fa4`)** are what make the substitution path survive under the jailer — without them the box is silently `SIGSYS`-killed on the first substituted request, or the MITM upstream leg fails cert verification (502):
1. Add `getpeername` (x86_64 syscall 52) to the `vmm` seccomp allowlist next to `getsockname` — gvproxy's MITM TLS-termination path calls it, and the filter's `default_action: trap` otherwise kills the shim/VMM with no log line.
2. ro-bind the host CA trust store (`/etc/ssl/certs`, `/etc/pki`) and `/etc/resolv.conf` into the bwrap sandbox — gvproxy re-originates TLS to the real upstream and must verify its cert / resolve allow-listed hosts.

See the review comment for the full repro, dmesg evidence (`audit type=1326 … syscall=52`), and the seccomp×CA evidence matrix.

**WebSocket coverage (commit `ac8c4f5a`):** the WebSocket MITM path (`handleWebSocketUpgrade`) bypasses `secretTransport`; it substitutes secrets into the upgrade request headers but relayed the upstream's 101 response headers verbatim — the same reflection leak. Upgrade response headers are now scrubbed there too. WebSocket message frames stay verbatim in both directions by design (the value only ever enters via headers, so frames never carry it).

**Compression coverage (commit `71000028`):** scrubbing scans response bytes, so a reflected secret inside a gzip/deflate-compressed body would slip past (a guest requesting `Accept-Encoding: gzip` could decompress and recover it — reproduced e2e). `secretTransport` now requests `Accept-Encoding: identity` upstream and, for upstreams that compress anyway, decompresses gzip/deflate → scrubs → serves identity; encodings it cannot decode are refused (fail closed). This closes the compression bypass but is still not a boundary — any other invertible transform an allow-listed upstream applies to the value (or a pass/fail oracle) can still leak it to a determined malicious guest. **Hosts scoping to trusted upstreams remains the actual control; if the guest is untrusted, run secret-using logic host-side rather than substituting into guest traffic.**


G4614:
## End-to-end verification of live secret substitution on the Linux jailer (bwrap + seccomp)

I ran the real `liveAddSecret` path against an actual upstream (not a mock) on a Linux host with the production jailer enabled, to confirm the feature works from boot → live admin-socket update → real MITM TLS substitution.

**Result: it works — but only after two sandbox fixes that this branch is currently missing.** As-is, the box is silently killed on the first substituted request.

### Repro
Bare box created with secret substitution enabled → `Start` → add a secret live via the gvproxy admin socket (`PUT /services/secrets`) → guest runs `GET https://<host>/headers` with `Authorization: Bearer <BOXLITE_SECRET:livekey>`. Assert the upstream echo contains the real value and not the placeholder.

### Findings

**1. VMM seccomp allowlist is missing `getpeername` (x86_64 syscall 52).**
gvproxy runs inside the shim process under the VMM seccomp filter (applied TSYNC — the code comment even says "covers all threads including gvproxy"). Its MITM TLS-termination path calls `getpeername(2)` on the intercepted connection. The `vmm` filter allows `getsockname` but **not** `getpeername`, and `default_action` is `trap`, so the first real substitution raises `SIGSYS` and kills the shim/VMM **silently** — no log line; the box just disappears and subsequent execs get `Connection refused`. Confirmed via `dmesg`:
```
audit: type=1326 … comm="boxlite-shim" … syscall=52 … code=0x7ffc0000
```
(syscall 52 = `getpeername`; it was the *only* syscall tripped in the whole substitution flow.)

Fix — add `getpeername` next to `getsockname`, mirroring wherever this PR already added `getsockname` for the MITM path:
```json
{ "syscall": "getpeername", "comment": "gvproxy MITM: get remote peer address when terminating intercepted TLS for secret substitution." }
```
Per-arch, this is **not** a uniform "all arches" edit — the socket-syscall coverage this PR added differs by target:
- **`x86_64-unknown-linux-gnu`** — gvproxy runs under the `vmm` filter; add `getpeername` there (vmm 128→129). *This is the arch I verified end-to-end.*
- **`x86_64-unknown-linux-musl`** — has a **separate `gvproxy` filter** in addition to `vmm`, and both carry `getsockname`; add `getpeername` to **both** (vmm 128→129, gvproxy 125→126).
- **`aarch64-unknown-linux-{gnu,musl}`** — the `vmm` filter (59 rules) has **none** of the MITM socket set: it's missing `getsockname`, `getpeername`, `bind`, `listen`, `getsockopt`, `setsockopt`, `sendto`, `shutdown`, `socketpair`, `accept`. Secret substitution would `SIGSYS` on the very first socket op here, well before `getpeername`. **The feature is effectively x86_64-only in this PR.** Before enabling it on ARM, the aarch64 `vmm` filter needs the same socket set the x86_64 filters got — or the feature should be explicitly gated to x86_64. I did **not** blind-port 10 untestable syscalls to aarch64; flagging for a decision.

Verified: with `getpeername` added to x86_64 and seccomp **fully enforced** (`default_action: trap`, no bypass), the substitution passes with no SIGSYS.

**2. The bwrap sandbox does not expose the host CA trust store.**
gvproxy's MITM *upstream* leg re-originates TLS to the real upstream and must verify its certificate. The sandbox binds `/usr /lib /lib64 /bin /sbin` but not `/etc/ssl/certs`, so Go's x509 loader finds no roots → `tls: failed to verify certificate: x509: certificate signed by unknown authority` → every substituted request returns **502**.

Fix — ro-bind the trust store (and, for allow-list DNS to work at all, `/etc/resolv.conf`) into the sandbox:
```rust
// src/boxlite/src/jailer/sandbox/bwrap.rs, after the /usr…/sbin ro-binds
bwrap_cmd.ro_bind_if_exists("/etc/resolv.conf", "/etc/resolv.conf");
bwrap_cmd.ro_bind_if_exists("/etc/ssl/certs", "/etc/ssl/certs");
bwrap_cmd.ro_bind_if_exists("/etc/pki", "/etc/pki");
```
(Without `/etc/resolv.conf`, gvproxy's allow-list resolver falls back to `[::1]:53` (refused) and every allowed host maps to the `0.0.0.0` sinkhole → ECONNREFUSED, independent of secrets.)

### Evidence matrix (Linux, x86_64-gnu, jailer on)
| `getpeername` allowed | `/etc/ssl/certs` bound | result |
|---|---|---|
| no | no | box SIGSYS-killed at `MITM: intercepting` |
| no | yes | still killed (CA bind alone doesn't help) |
| yes (log mode) | yes | PASS |
| **yes (trap enforced)** | **yes** | **PASS — real value substituted into upstream traffic, placeholder absent** |

### Note on persistence
Live-added secrets live in gvproxy's in-memory matcher. The per-box MITM CA is pre-provisioned at boot (which is exactly what this PR's `enable_secret_substitution` guards), but a live-added secret set is not persisted across a box restart. Probably worth a line in the docs.

PR #839, opened 2026-06-23
fix(runtime): embedded runtime survives in-place rollout (split from #681)

G4614:
Make the embedded runtime survive an in-place binary rollout: invalidate the extracted-runtime cache when the embedded manifest changes at the same crate version, and enforce `+x` on extracted executables. Split from #681.

## Test plan
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` + `cargo fmt --check` clean.
- Two-side verified (revert the fix → test fails with the exact symptom → restore → passes):
  - `stamp_matches_current_manifest_requires_hash` — legacy/no-hash and different-hash stamps are refreshed.
  - `runtime_executable_permissions_are_enforced` — an extracted runtime exec gets `0o755`.
  - `copy_shim_to_box_marks_existing_copy_executable`.

### DIFF
--- a/src/boxlite/src/jailer/shim_copy.rs
+++ b/src/boxlite/src/jailer/shim_copy.rs
@@ -90,6 +90,7 @@ pub fn copy_shim_to_box(shim_path: &Path, box_dir: &Path) -> BoxliteResult<PathB
             e
         ))
     })?;
+    ensure_executable(&dest_shim)?;
 
     if copied {
         tracing::debug!(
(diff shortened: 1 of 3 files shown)


src/boxlite/src/jailer/shim_copy.rs
```diff
@@ -107,6 +108,34 @@ pub fn copy_shim_to_box(shim_path: &Path, box_dir: &Path) -> BoxliteResult<PathB
...
+#[cfg(unix)]
+fn ensure_executable(path: &Path) -> BoxliteResult<()> {
+    use std::os::unix::fs::PermissionsExt;
+
+    let mode = std::fs::metadata(path)
+        .map(|metadata| metadata.permissions().mode() & 0o777)
+        .map_err(|e| BoxliteError::Storage(format!("stat {}: {}", path.display(), e)))?;
+    let executable_mode = mode | 0o755;
+    if executable_mode != mode {
+        std::fs::set_permissions(path, std::fs::Permissions::from_mode(executable_mode)).map_err(
```
DorianZheng, line 120: should not have this issue by design, must be a bug

G4614:
Closing: the change this PR set out to make is already on `main`, by a different route.

It invalidated the extracted-runtime cache by checking the embedded manifest hash against the `.complete` stamp. `main` instead puts the hash in the cache directory name — `src/boxlite/src/runtime/embedded.rs:217`, `v{version}-{commit}-{manifest_hash}`, whose own comment says the hash is "what makes the name *correct*". Same property, so there is nothing left here to land.

**One thing in this PR was not resolved and should not be lost with it.** @DorianZheng asked on `src/boxlite/src/jailer/shim_copy.rs:120`, about the `ensure_executable` helper that forced `+x` on the extracted shim:

> should not have this issue by design, must be a bug

That question was never answered, and `shim_copy.rs` is still on `main`. If an extracted shim can really land without its execute bit, the bug is in whatever packs or extracts it — forcing `+x` at the copy site would only have hidden it. Worth chasing separately; it is not carried by this branch either way.

PR #841, opened 2026-06-24
feat(exec): back-pressure and drop observability for C SDK exec stream

G4614:
Add back-pressure and drop observability to the C SDK exec stream path.

- Bound the FFI exec stdout/stderr channels (`mpsc::channel(64)`) so a slow FFI consumer applies back-pressure through the C SDK instead of accumulating unbounded memory
- Implement `boxlite_execution_set_stream_paused` C FFI hook so the Go/C SDK can signal pause/resume to the Rust producer
- Replace silent `let _ = try_send(...)` with `note_dropped_frame` — distinguishes `Full` (backpressure drop, warn + count) from `Closed` (normal teardown, debug)

Does not fix the 256 KB REST-path truncation (POL-454); that requires fixing the runner `streamBus` fan-out.

Test plan:
- [x] `BOXLITE_DEPS_STUB=1 cargo check -p boxlite` — CI pass
- [x] `cargo test -p boxlite --lib note_dropped_frame_counts_full_not_closed` — passes


G4614:
Closing after dev validation: the kill/attach behavior is not a hard stuck case for the bounded sleep scenario, and this PR should not carry a speculative runner change.

G4614:
Updated the head branch G4614:fix/exec-drain-hang-on-sigkill-cross to 8d01491b with the REST/Rust attach stream buffer split only. The previous runner attach changes were cleared from the branch.

Tried to reopen this PR, but GitHub returned: `GraphQL: Could not open the pull request. (reopenPullRequest)`. The fork ref resolves to 8d01491b, while the closed PR API still reports the old head snapshot 36b2276c.

G4614:
Update: rebuilt the branch as a descendant of the old closed head, reverted the old runner attach commit, merged latest main, then cherry-picked the REST/Rust attach buffer commit. GitHub now allows the PR to reopen, and the final PR diff contains only the REST/Rust attach buffering files.

PR #843, opened 2026-06-24
perf(api): scale GET /boxes from O(org) to O(page)

G4614:
## Why

Load-testing showed `GET /v1/{prefix}/boxes` was the sole CPU sink: at 200 rps on 4x0.5-vCPU tasks its p95 was ~414ms while `/health`, `/v1/config`, `/v1/me` all stayed <11ms. A single dev org has ~1.8k boxes, and the endpoint did O(org-size) work on **every** request, so a `min1/max4` service couldn't absorb a 4x surge no matter how fast it scaled (a flat-200 test with 4 pinned warm tasks still saturated CPU at 95% and dropped requests).

Two reasons the query was O(org):
1. `ORDER BY createdAt DESC` had no matching index, so Postgres fetched all the org's rows and sorted them in memory every request.
2. `findAndCount` ran a `COUNT(*)` over all matching rows every request, just to compute `hasMore`.

## What

- **Composite index `(organizationId, createdAt)`** on `box` (pre-deploy migration): the page becomes an index range scan instead of a full fetch + sort.
- **Drop the per-request `COUNT(*)`**: `listBoxesPageDeprecated` fetches `limit+1` rows and derives `hasMore` from the overflow row; the controller uses `hasMore` directly. The public response never exposed a total.

Also bundles the page-size/page-token pagination work (server + Rust SDK list loop) and the dashboard cold-path list cache this built on.

## Test plan

- [x] `box.service.list-paged.spec.ts` rewritten: asserts `findAndCount` is NOT called, `take === limit+1`, hasMore/slice. Two-side verified.
- [x] 50 jest specs across `src/box/services` + `src/boxlite-rest` pass; `tsc --noEmit` clean.
- [ ] `EXPLAIN ANALYZE` the list query on migrated dev + re-run k6 surge.

## Notes

- Migration uses `CREATE INDEX IF NOT EXISTS`; for a large prod table consider `CREATE INDEX CONCURRENTLY`.

Generated with Claude Code

### DIFF
--- a/src/boxlite/src/rest/runtime.rs
+++ b/src/boxlite/src/rest/runtime.rs
@@ -88,8 +88,27 @@ impl RuntimeBackend for RestRuntime {
     }
 
     async fn list_info(&self) -> BoxliteResult<Vec<BoxInfo>> {
-        let resp: ListBoxesResponse = self.client.get("/boxes").await?;
-        resp.boxes.iter().map(|b| b.to_box_info()).collect()
+        // The server pages this endpoint (openapi `pageSize`/`pageToken`), so a
+        // single request only returns one page. Follow `next_page_token` to
+        // preserve the "list every box" contract callers depend on, requesting
+        // the max page size to keep round-trips down.
+        let mut all = Vec::new();
+        let mut page_token: Option<String> = None;
+        loop {
+            let mut query: Vec<(&str, &str)> = vec![("pageSize", "1000")];
+            if let Some(token) = page_token.as_deref() {
+                query.push(("pageToken", token));
+            }
+            let resp: ListBoxesResponse = self.client.get_with_query("/boxes", &query).await?;
+            for box_response in &resp.boxes {
+                all.push(box_response.to_box_info()?);
+            }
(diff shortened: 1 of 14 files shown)


src/boxlite/src/rest/runtime.rs
```diff
@@ -88,8 +88,27 @@ impl RuntimeBackend for RestRuntime {
...
-        let resp: ListBoxesResponse = self.client.get("/boxes").await?;
-        resp.boxes.iter().map(|b| b.to_box_info()).collect()
+        // The server pages this endpoint (openapi `pageSize`/`pageToken`), so a
+        // single request only returns one page. Follow `next_page_token` to
+        // preserve the "list every box" contract callers depend on, requesting
+        // the max page size to keep round-trips down.
+        let mut all = Vec::new();
+        let mut page_token: Option<String> = None;
+        loop {
+            let mut query: Vec<(&str, &str)> = vec![("pageSize", "1000")];
```
DorianZheng, line 98: seems we need a new API other than `list_info` to support next page

PR #851, opened 2026-06-25
fix(jailer): skip --die-with-parent for detached boxes

G4614:
Gate bwrap's `--die-with-parent` on `!detached` so a detached box (`run -d`) isn't SIGKILL'd the instant its launcher exits — a regression from #652 making the jailer default-on that left every detached box born-`Stopped` (history + #602/#275 context in the comment below).

## Test plan
- [x] Guard unit test `apply_sets_die_with_parent_only_for_foreground` (foreground keeps the flag, detached drops it)
- [x] On a real box: `run -d … sleep 300` now stays `Running`; foreground `run` still works
- [ ] e2e-local: the 9 detached failures go green (`lifecycle_journey`, `rm::test_rm_force_running`, 6× `stress_disk`)

## Summary by CodeRabbit

* **New Features**
  * Added support for detached sandbox sessions, allowing boxes to keep running independently of the launching process.

* **Bug Fixes**
  * Fixed sandbox startup behavior so foreground sessions still terminate with the parent process, while detached sessions remain alive.
  * Improved consistency across sandbox setup and tests to match the new detached-session behavior.

### DIFF
--- a/src/boxlite/src/jailer/mod.rs
+++ b/src/boxlite/src/jailer/mod.rs
@@ -500,6 +500,7 @@ impl<S: Sandbox> Jailer<S> {
             resource_limits: &self.security.resource_limits,
             network_enabled: self.security.network_enabled,
             sandbox_profile: self.security.sandbox_profile.as_deref(),
+            detached: self.detach,
         }
     }
 
--- a/src/boxlite/src/jailer/sandbox/bwrap.rs
+++ b/src/boxlite/src/jailer/sandbox/bwrap.rs
@@ -69,10 +69,15 @@ impl Sandbox for BwrapSandbox {
         // =====================================================================
         // Namespace and session isolation
         // =====================================================================
-        bwrap_cmd
-            .with_default_namespaces()
-            .with_die_with_parent()
-            .with_new_session();
+        bwrap_cmd.with_default_namespaces();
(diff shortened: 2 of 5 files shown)


G4614:
@DorianZheng would appreciate your eyes on this — it touches your earlier detach work.

Short version: bwrap's `--die-with-parent` is added **unconditionally** in `BwrapSandbox::apply`, so a detached box (`run -d`) gets `PR_SET_PDEATHSIG`'d the instant its launcher exits and comes up `Stopped`. That overrides the `setsid()` daemon semantics you gave detached boxes in #602. It stayed dormant until #652 flipped the jailer to default-on, at which point every `run -d` started dying (the 9 detached e2e-local failures today).

This PR gates the bwrap-layer `--die-with-parent` on `!detached`, so it respects the same `detach` flag your `pre_exec.rs` layer already honors. Foreground keeps it (the #275 "die with the launcher" safety net is correct there); only detached skips it.

You swapped the parent watchdog for `--die-with-parent` in #275 ("Skip parent watchdog when inside PID namespace — bwrap `--die-with-parent` handles it"). Does gating it on `detached` match your intent, or did you expect the watchdog / `--die-with-parent` to never apply to detached in the first place?

PR #887, opened 2026-06-30
fix(box): keep active terminals and long tasks from auto-stopping

law-chain-hot:
## Summary

Active boxes were getting auto-stopped mid-work because activity was only
recorded on inbound HTTP. A long-running task with no HTTP traffic, or an open
exec/attach terminal, looked idle to the reaper.

- **Default `autoStopInterval` 15 → 60 min** via new `DEFAULT_AUTO_STOP_INTERVAL_MINUTES` constant.
- **Warm-pool assignment** now always resets `autoStopInterval` (to the default, or the DTO value); explicit `0` (disable) is preserved.
- **Exec `/attach` WebSocket** now refreshes `lastActivityAt` immediately and then every 30s while the socket stays connected; the heartbeat stops cleanly on socket `close` / `end` / `error`.

## Behavior

| Scenario | Before | After |
|---|---|---|
| HTTP request hits the box | kept alive | kept alive |
| Terminal `/attach` WebSocket open | reaped 15 min after attach-open | kept alive while WS is connected |
| Tab closed mid-task; bash still running on box | reaped after 15 min | reaped after 60 min (out of scope here) |

## Known limitation

A half-open TCP socket (peer vanishes, no FIN/RST) never emits `close`/`end`/`error`, so the heartbeat keeps refreshing a dead session until OS TCP keepalive trips (~2h on default Linux; `http-proxy` already sets `setKeepAlive(true, 0)`). A naive `socket.setTimeout` idle-timeout would false-positive on an idle-but-alive terminal, so TCP keepalive is the correct layer here. Documented; not fixed in this PR.

## No DB migration

The `@Column({ default: 60 })` change has zero runtime effect because `synchronize: false` and the entity field initializer drives every insert. The next regular `migration:generate` will pick up a correct `15 → 60` ALTER.

## Verification

- Unit RED→GREEN (two-side) on all three new/updated specs.
- Multi-agent adversarial review: 9 confirmed findings, all nit/minor, 0 blocker/major.
- Codex adversarial review: no blocker, no new bug; half-open reasoning confirmed sound.
- End-to-end on a local infra-local stack: idle box stays flat → attach advances `lastActivityAt` every 30s → close stops it; new boxes get `autoStopInterval = 60`.

## Test plan

- [ ] `cd apps && yarn nx test api --testPathPatterns 'box\\.entity\\.spec|boxlite-ws-proxy\\.service\\.spec|box\\.service\\.warm-pool\\.spec'`
- [ ] Manual: open a terminal in the dashboard, leave it idle for >15 min, confirm the box is not stopped
- [ ] Manual: close the terminal tab, wait 60 min, confirm the box is auto-stopped
- [ ] Spot-check: create a box with `autoStopInterval: 0`, confirm it is never auto-stopped

## Summary by CodeRabbit

* **New Features**
  * Box auto-stop now consistently defaults to **15 minutes** when no interval is provided.
  * Websocket attach activity now updates box activity using **30-second throttling** based on inbound client data.

* **Bug Fixes**
  * Warm-pool box assignments now always apply the correct `autoStopInterval`, using the default or any explicit override.
  * Box activity refresh behavior now halts when the client goes silent (dead-peer scenario).

* **Tests**
  * Added coverage for the default auto-stop interval and updated warm-pool and websocket attach timing/stop behavior.


DorianZheng (dismissed):

law-chain-hot:
## Ping/pong
We can use Pong from browser to update activity 

## 30s heartbeat
                 THE 30s HEARTBEAT PATH — WHO GETS HIT, HOW HARD
                 ═══════════════════════════════════════════════
```
  M WebSockets (one 30s timer each)
  ┌──────────────────────────────────┐
  │  every 30s per WS:               │      load ∝ M
  │  updateLastActivityAt(boxId)     │      (in-process timers, ~free)
  └───────────────┬──────────────────┘
                  ▼
  ┌─ REDIS ──────────────────────────┐
  │  1. SET lock NX EX 5   (throttle)│      load ∝ M/30s  → tiny ops
  │     already locked? → DROP       │      e.g. 2000 WS → ~67 ops/s
  │  2. ZADD activity ts boxId       │      load ≤ N/5s (per-box dedup)
  │     (in-memory buffer, dedup     │      e.g. 1000 box → ≤200 ops/s
  │      by boxId, keeps newest ts)  │      Redis capacity ~100k ops/s
  └───────────────┬──────────────────┘      → <0.3% utilization
                  │  every 10s, ONE api instance
                  │  (distributed lock) drains the buffer
                  ▼
  ┌─ POSTGRES ───────────────────────┐
  │  bulk UPSERT box_last_activity   │      load ≈ CONSTANT:
  │  (batched: few SQLs / 10s,       │      a few statements per 10s,
  │   ≤ N single-timestamp rows)     │      regardless of M
  └───────────────┬──────────────────┘
                  ▼
  reaper cron (10s): WHERE lastActivityAt < NOW() - autoStopInterval → stop box
```

apps/runner/pkg/api/controllers/box_activity_touch.go
```diff
@@ -0,0 +1,86 @@
...
+
+const (
+	boxActivityTouchInterval = 45 * time.Second
+	boxActivityTouchTimeout  = 5 * time.Second
+)
+
+type updateBoxActivityFunc func(context.Context, string) error
+
+var updateBoxLastActivity updateBoxActivityFunc = func(ctx context.Context, boxID string) error {
+	apiClient, err := runnerapiclient.GetApiClient()
```
DorianZheng: how runner get api client

law-chain-hot: it imports a code-generated Go SDK (libs/api-client-go, generated from dist/apps/api/openapi.json) baked in at compile time

```
NestJS control-plane API
        │  dumps OpenAPI
        ▼
dist/apps/api/openapi.json
        │  openapi-generator
        ▼
libs/api-client-go   (generated SDK · physical: apps/api-client-go)
        │  runner require + replace, wraps with URL + token
        ▼
runnerapiclient "github.com/boxlite-ai/runner/pkg/apiclient"
   GetApiClient()  → returns apiClient
        │  
        ▼
apiClient.BoxAPI.UpdateLastActivity(ctx, boxID).Execute()
        │
        ▼
POST /box/{boxId}/last-activity  →  control-plane API
```

PR #895, opened 2026-07-01
feat(dashboard): org box resource caps

law-chain-hot:
Two dashboard-only changes (no server API changes).

## 1. Per-org box resource caps (feat)

The create-box dialog fed hardcoded ceilings (8 / 32 / 50) to the CPU / Memory / Disk
steppers, which don't match the backend's per-org limits — users could type values the
API then rejects with a 400. The steppers now read the organization's
`maxCpuPerBox` / `maxMemoryPerBox` / `maxDiskPerBox` (falling back to the built-in
ceilings when a limit is unset / `<= 0`), auto-correct an over-limit entry down to the
cap, and show a red "contact support@boxlite.ai" hint on the field when that happens.

## 2. Quickstart API key collision on re-run (fix)

The Quickstart "Create a key" step created the key under a fixed name
(`sdk-quickstart`), so re-running it (the Regenerate button) hit a 409
"API key with this name already exists". It now retries with an incrementing suffix
(`sdk-quickstart-2`, `-3`, …) on a name conflict, so re-running never surfaces the error.

## Testing

- New unit tests: `CreateBoxDialog` (org-max clamp + contact-support hint) and
  `quickstart-api-key` (default/suffix naming, 409 detection, retry-on-conflict).
- Full dashboard test suite green (62 tests); no new type errors.
- Verified the resource-cap UI in a browser: an over-max CPU entry clamps to the org
  maximum and shows the red hint.

## Summary by CodeRabbit

* **Bug Fixes**
  * Box creation advanced resource inputs (CPU, memory, disk) now clamp only when an organization-level per-box limit is actually set; otherwise the fields remain uncapped.
  * Over-limit entries are corrected more consistently, with the max warning appearing only when relevant and clearing immediately when values drop back under the limit.
  * If the organization changes, the dialog keeps the current form state while updating any limits that tighten.

* **New Features**
  * Improved max guidance under advanced resources, including a support contact link.


apps/dashboard/src/components/Box/CreateBoxDialog.tsx
```diff
@@ -26,20 +26,48 @@ const SUPPORTED_BOX_IMAGES = [
 ] as const
 
 const DEFAULTS = { cpu: 1, memory: 1, disk: 10 }
+// Built-in ceilings, used only as a fallback when the organization does not
+// carry a per-box limit (see resolvePerBoxLimits).
 const LIMITS = { cpu: 8, memory: 32, disk: 50 }
```
law-chain-hot: the fallback value in case there is no org's limit

DorianZheng (approved):

PR #907, opened 2026-07-03
fix(volume): scope managed S3 buckets by stage

### LINKED ISSUE #733: Stage-scope the volume bucket prefix (boxlite-volume-* is account-global)
## Problem

Volume bucket names are account-global, not stage-scoped: `Volume.getBucketName()` returns `boxlite-volume-${id}` (`apps/api/src/box/entities/volume.entity.ts:48`), while multiple stages (dev, e2e-ci, personal) share one AWS account.

Because the bucket name is the only stage signal available to IAM, every policy that scopes to `arn:aws:s3:::boxlite-volume-*` — the Api task role's volume statement and the runner's `RunnerVolumeS3Policy` (`apps/infra/sst.config.ts`) — necessarily spans **all stages' volume buckets**. Concretely: dev's API/runner can list, read, and delete e2e-ci's volume buckets, and vice versa. PR #732 narrowed the *actions* on these grants but could not narrow the *resources* further without this rename.

## Proposed fix

1. Make the prefix configurable: `VOLUME_BUCKET_PREFIX` env (default `boxlite-volume-` for compatibility), set from `sst.config.ts` as `${$app.name}-${$app.stage}-volume-`.
2. Persist `bucketName` on the volume row at creation time so existing volumes keep their current buckets while new volumes get the stage-scoped prefix (avoids orphaning `boxlite-volume-<id>` buckets already in use).
3. Tighten both IAM resource patterns to the stage prefix once the fleet of old-prefix volumes drains (or keep a temporary dual-pattern grant during transition).

## Notes

- Bucket names are globally unique across AWS; the UUID suffix already prevents collisions — this is purely an isolation/IAM-scoping concern.
- The runner mounts buckets by name received from the control plane (`apps/runner/pkg/boxlite/volumes.go`), so no runner code change is needed beyond the IAM pattern.

xhebox:
## Summary
Fixes #733 by persisting the real S3 bucket name for new managed volumes and making SST-managed bucket prefixes stage-scoped. Existing volumes remain compatible through the legacy `boxlite-volume-<volumeId>` fallback; this does not rename buckets or migrate S3 objects.

## Changes
- Add a nullable `volume.bucketName` column and set it for newly-created volumes from `s3.volumeBucketPrefix`.
- Resolve box volume references to canonical volume IDs plus internal bucket names while keeping public REST volume DTOs free of `bucketName`.
- Pass `bucketName` through runner v0/v2 create/recover/start paths and the runner DTO/OpenAPI client.
- Update runner mount logic to prefer the wire `bucketName` while keeping local mount directories under `boxlite-volume-<volumeId>`.
- Wire SST `VOLUME_BUCKET_PREFIX` and stage-scoped S3 IAM resources while retaining legacy `boxlite-volume-*` access.

## How to verify
- `corepack yarn jest --config api/jest.config.ts api/src/box/dto/box.dto.spec.ts api/src/box/services/volume.service.spec.ts api/src/box/managers/volume.manager.spec.ts api/src/box/managers/box-actions/box-start.action.spec.ts api/src/box/services/box.service.spec.ts api/src/box/runner-adapter/runnerAdapter.v2.spec.ts --runInBand`
- `corepack yarn nx run api:lint`
- `bash apps/api/scripts/validate-migration-paths.sh apps/api/src/migrations/pre-deploy/1782972000000-migration.ts`
- `cd apps/runner && go test ./pkg/api/dto`

## Risks / rollout
- Deploy the pre-deploy migration before rolling out API code that writes `volume.bucketName`.
- Roll out runner support before relying on stage-scoped buckets from new API deployments.
- Keep legacy `boxlite-volume-*` IAM access during the transition so old volumes and old payloads continue to work.
- Full `apps/runner/pkg/boxlite` tests still require the native `libboxlite.a` archive / compatible Rust target; this PR only verifies the changed runner DTO package locally.

## Summary by CodeRabbit

* **New Features**
  * Added configurable volume bucket naming with a validated prefix, now supported across API and runner components.
  * Volume attachments are now resolved before box creation, preserving mount path and subpath details.
* **Bug Fixes**
  * Improved handling of volume states so only ready volumes are used, with clearer errors for missing or unavailable volumes.
  * Volume bucket creation and deletion now use the configured naming scheme consistently.


DorianZheng:
one observation on the runner side of the wire change: `volumeBucketName()` hands any non-empty `bucketName` string straight to `mount-s3` (and the mount error/log paths), so the runner's contract is now "whatever string arrives is a mountable bucket". upstream v0.190.0 gated the equivalent input with a canonical-uuid check (`isValidVolumeId`, `apps/runner/pkg/docker/volumes_mountpaths.go`) precisely because that value doubles as the bucket name; that gate never made it into `pkg/boxlite`, and this change moves the trust onto a field the runner no longer derives itself. not exploitable through this PR's own path — the api builds the name from config prefix + volume uuid, and the value lands in argv, not a shell — so nothing blocking, just noting where the validation gap now sits.

apps/runner/pkg/boxlite/volumes.go
```diff
@@ -48,13 +48,22 @@ func getVolumeMountBasePath() string {
 	return "/mnt"
 }
 
+// TODO: Delete this compatibility fallback once every API deployment sends bucketName.
```
DorianZheng: no compatibility needed, this feature is forbidden currently

DorianZheng: remove all similar compatibility code

DorianZheng:
scope change from our side — to be clear, not a defect in the pr: this implements #733 as written, and cleanly. but boxlite is pre-GA and we treat existing volumes as disposable, so backward compatibility with the `boxlite-volume-*` generation is a non-goal. that removes the reason for most of the persistence/wire machinery:

- drop the `volume.bucketName` column and the migration
- drop `bucketName` from the runner wire (dto, openapi client, go struct) and the box entity/dto changes
- iam: single stage-scoped pattern only, no legacy `boxlite-volume-*` grants
- instead give the runner the same `VOLUME_BUCKET_PREFIX` env (default `boxlite-volume-`) and derive `prefix + volumeId` there, replacing the `volumeMountPrefix` constant

please do keep the `resolveVolumes` name→uuid resolution and the ordered `resolveReadyVolumes` — that part fixes a real pre-existing bug independent of bucket naming (a volume referenced by name passes validation, the raw name is stored, and the runner derives a bucket that doesn't exist).

one obligation the derive-everywhere design takes on: `VOLUME_BUCKET_PREFIX` becomes immutable-after-first-use per deployment — changing it orphans every existing volume. a warning to that effect next to the env in both `.env.example`s and `sst.config.ts` is required.

we'll delete the existing `boxlite-volume-*` buckets on our side, so no drain path is needed. this also moots my earlier note about validating the wire `bucketName` — the field goes away.

sorry for the pivot after you built the compat path; it's solid work, we just don't need it pre-GA.

xhebox:
> scope change from our side — to be clear, not a defect in the pr: this implements #733 as written, and cleanly. but boxlite is pre-GA and we treat existing volumes as disposable, so backward compatibility with the `boxlite-volume-*` generation is a non-goal. that removes the reason for most of the persistence/wire machinery:
> 
> * drop the `volume.bucketName` column and the migration
> * drop `bucketName` from the runner wire (dto, openapi client, go struct) and the box entity/dto changes
> * iam: single stage-scoped pattern only, no legacy `boxlite-volume-*` grants
> * instead give the runner the same `VOLUME_BUCKET_PREFIX` env (default `boxlite-volume-`) and derive `prefix + volumeId` there, replacing the `volumeMountPrefix` constant
> 
> please do keep the `resolveVolumes` name→uuid resolution and the ordered `resolveReadyVolumes` — that part fixes a real pre-existing bug independent of bucket naming (a volume referenced by name passes validation, the raw name is stored, and the runner derives a bucket that doesn't exist).
> 
> one obligation the derive-everywhere design takes on: `VOLUME_BUCKET_PREFIX` becomes immutable-after-first-use per deployment — changing it orphans every existing volume. a warning to that effect next to the env in both `.env.example`s and `sst.config.ts` is required.
> 
> we'll delete the existing `boxlite-volume-*` buckets on our side, so no drain path is needed. this also moots my earlier note about validating the wire `bucketName` — the field goes away.
> 
> sorry for the pivot after you built the compat path; it's solid work, we just don't need it pre-GA.

I'll remove all compat later.

PR #908, opened 2026-07-03
feat(dashboard): deal a generated name into the create-box dialog

law-chain-hot:
## What

Opening the Create-Box dialog now deals the user a generated name (e.g.
`cozy-otter`): it lands in the **editable** name field via a short
scramble/decode animation, and a refresh button deals another one. The old
static `my-new-box` placeholder only remains as the empty-field hint.

## How

- **New shared lib `@boxlite-ai/box-name`** (`apps/libs/box-name`) — the single
  source of truth for the adjective/animal word lists and `generateBoxName()`.
  - `apps/api` keeps only the persistence-side collision fallback
    (`persistWithGeneratedBoxName`) and imports the generator from the lib.
  - The dashboard generates names locally from the same lib — **no new server
    endpoints**, no word-list duplication.
- `useGeneratedBoxName` hook + `scramble.ts` drive the reveal animation.

## Collision handling

Client-generated names can collide with the per-org unique-name constraint
(the server's `-boxId` fallback only applies to empty-name creates), so the
dialog catches a create `409`, toasts, and automatically deals a fresh name;
the next click succeeds. Clearing the field still submits no name and the
server auto-names as before.

## Test

- `apps/dashboard/src/lib/scramble.test.ts` — reveal-frame contract (3 tests).
- `apps/api/src/box/utils/box-name-generator.spec.ts` (existing, 5 tests) now
  exercises the shared lib through the re-export.

## Summary by CodeRabbit

* **New Features**
  * Added smarter box-name suggestions when creating a box, with a refresh option for new name ideas.
  * Introduced a shared name generator used across the app for consistent box names.
  * Added animated name scrambling/reveal behavior in the dashboard.

* **Bug Fixes**
  * Improved handling of name conflicts by automatically suggesting a new name instead of showing an immediate failure.
  * Prevented box creation while a new name is still being generated.

* **Tests**
  * Added coverage for the name-scrambling behavior.


apps/libs/box-name/src/index.ts
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright 2025 Daytona Platforms Inc.
```
DorianZheng, line 2: wrong license

PR #923, opened 2026-07-05
perf(file-transfer): stream file and archive payloads

zhangrr:
# PR: Stream file and archive payloads to reduce memory pressure

## Outcome

Convert the heaviest BoxLite file/archive transfer paths from whole-payload buffering to chunked streaming. The PR keeps the existing public REST/gRPC surface stable and focuses on removing avoidable `O(payload_size)` allocations on the host and local REST server.

## Problem

Large copy/export/import operations currently allocate full archives in memory in several places:

- `src/boxlite/src/portal/interfaces/files.rs:37-71` reads a whole tar file into `Vec<UploadChunk>` before starting gRPC upload.
- `src/boxlite/src/litebox/box_impl.rs:496-521` creates a temporary tar for `copy_into`; combined with the `Vec<UploadChunk>` above, upload uses temp disk plus same-sized memory.
- `src/cli/src/commands/serve/handlers/files.rs:15-60` accepts upload bodies as `Bytes`, forcing complete HTTP request buffering before unpack/copy.
- `src/cli/src/commands/serve/handlers/files.rs:91-125` builds download responses with `tar::Builder::new(Vec::new())`, holding the complete tar in memory before response write.
- `src/cli/src/commands/serve/handlers/advanced.rs:69-86` uses `std::fs::read` to load a complete `.boxlite` export archive into memory.
- `src/cli/src/commands/serve/handlers/advanced.rs:92-110` accepts import archives as `Bytes`, again requiring whole-payload buffering.

The protobuf surface is already streaming-friendly: `src/shared/proto/boxlite/v1/service.proto:61-66` defines `Upload(stream UploadChunk)` and `Download(...) returns (stream DownloadChunk)`.

## Similar-project references

- Docker/Moby exposes copy upload/download as reader streams rather than whole buffers: `moby/client/container_copy.go:49-63`, `95-127`.
  - https://github.com/moby/moby/blob/master/client/container_copy.go#L49-L63
  - https://github.com/moby/moby/blob/master/client/container_copy.go#L95-L127
- Kubernetes `kubectl cp` pipes tar generation/consumption with `io.Pipe` and reader streams: `staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go:306-323`, `364-426`, `506`.
  - https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go#L306-L323
  - https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go#L364-L426
  - https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go#L506

## Reliability intent check

I checked whether the original eager buffering looked intentional for retry/resume or network-failure tolerance. The evidence points the other way:

- `git blame` traces the host and guest file-transfer implementation back to `dd61c75cf` (`Add cp support across CLI and SDKs (#175)`). The commit message describes feature rollout and clippy cleanup, not retry/resume semantics.
- The original host implementation explicitly said: `Read entire tar file and build chunks` followed by `Note: For very large files, consider streaming with async_stream crate`. That reads as a known simplification/performance TODO, not a deliberate reliability guard.
- The protobuf API was already streaming at introduction time: `Upload(stream UploadChunk)` and `Download(...) returns (stream DownloadChunk)`.
- The guest upload implementation already consumes the stream chunk-by-chunk and writes it to a temp tar before extracting. That preserves the important failure behavior: if the upload stream is interrupted, extraction does not happen because the stream returns an error before the temp tar is unpacked.
- There is no resumable upload state, chunk checksum, offset acknowledgement, idempotency token, or retry loop in the host/guest protocol. Pre-reading into memory therefore cannot make a network interruption resumable; it only fails later, after allocating memory for the whole tar.

Conclusion: this PR keeps the existing all-or-nothing extraction boundary on the guest temp file, while removing host-side all-payload memory buffering. It does not add resumable transfers, but it does not remove any existing retry/resume mechanism because none exists in the current protocol.

## Scope for this PR

1. Stream `FilesInterface::upload_tar` through a bounded channel instead of collecting every chunk into a `Vec`.
2. Change REST upload/import handlers from `Bytes` extraction to streaming request-body-to-temp-file writes.
3. Change REST download/export handlers to stream response files with `Body::from_stream` instead of reading archives into memory.
4. Keep archive format, route paths, and protobuf messages unchanged.

## Non-goals

- Do not redesign guest-side extraction in this PR. The guest still writes upload/download temp tar files before unpack/pack.
- Do not change SDK API shape or route contracts.
- Do not remove all temporary files; this PR targets peak memory first, while preserving current behavior.

## Verification plan

- Add focused unit coverage for the new chunk producer so it proves upload chunks are produced incrementally and metadata is present on the first chunk.
- Add focused unit coverage for REST body/file helpers so request bodies write to files and response bodies stream file content.
- Run the smallest relevant make target first; broaden only if needed.

## Implementation notes

Changed files:

- `src/boxlite/src/portal/interfaces/files.rs`
  - Replaced eager `Vec<UploadChunk>` construction with a bounded `mpsc` producer and `ReceiverStream` upload body.
  - Added a focused test for first-chunk metadata and chunked data emission.
- `src/cli/src/commands/serve/handlers/body_io.rs`
  - Added shared helpers to write an Axum `Body` stream to disk and stream a response body from a file while owning temp directories until the response stream is dropped.
- `src/cli/src/commands/serve/handlers/advanced.rs`
  - Export now streams the generated `.boxlite` file instead of `std::fs::read` into memory.
  - Import now streams request body bytes into the temp archive instead of extracting `Bytes`.
- `src/cli/src/commands/serve/handlers/files.rs`
  - Upload now streams request body bytes into a temp tar file before extraction.
  - Download now writes the response tar to a temp file and streams that file instead of building `tar::Builder<Vec<u8>>`.
- `src/cli/Cargo.toml`
  - Enabled `tokio` `fs` and `io-util` features for async file streaming in `boxlite serve`.
- `make/test.mk`
  - Added `make test:unit:cli` and wired it into changed-component/core unit test dispatch so CLI source unit tests have a Makefile entry point.

## Verification performed in this environment

- `make setup:build` — passed; installed missing build prerequisites and initialized vendored submodules.
- `make fmt:rust` — passed.
- `make fmt:check:rust` — passed.
- `git diff --check` — passed.
- `make test:unit:rust FILTER=send_upload_chunks_streams_file_with_metadata_only_on_first_chunk` — passed.
- `make test:unit:cli FILTER=stream_file_body` — passed.
- `make test:unit:cli FILTER=write_body_to_file_persists_streamed_body` — passed.
- `make test:unit:cli FILTER=pack_dir_to_tar_streams_from_file_without_vec_buffer` — passed.
- Structural grep confirmed the targeted eager-buffering patterns are gone from the touched transfer paths:
  - no `Vec<UploadChunk>` / `chunks.push` / `futures::stream::iter(chunks)` in `src/boxlite/src/portal/interfaces/files.rs`;
  - no `std::fs::read(archive.path())` in `src/cli/src/commands/serve/handlers/advanced.rs`;
  - no `body: Bytes` request extraction in touched REST transfer handlers;
  - no `tar::Builder::new(Vec::new())` in `src/cli/src/commands/serve/handlers/files.rs`.

## Broader verification notes

- `make test:changed` was attempted after setup. It failed outside the scope of this PR because:
  - changed-component detection included `apps`, but this environment lacks `yarn`;
  - full CLI unit tests hit sandbox/TTY/socket permissions in existing tests (`bind ephemeral: Operation not permitted`, `openpty: EACCES`);
  - full Rust unit tests hit multiple environment-sensitive socket/process/runtime tests.
- The new tests added by this PR all passed via focused `make` targets listed above.

## Pre-merge suggestion

In a full dev/CI environment, rerun:

```bash
make test:changed
```

## Summary by CodeRabbit

* **New Features**
  * Large file uploads and downloads now stream instead of loading entire archives into memory.
  * Import/export and file transfer endpoints now support more efficient handling of big payloads.
  * Request size limits were added to help protect upload-related endpoints.

* **Bug Fixes**
  * Improved tar handling for file transfers to preserve expected contents and structure.
  * Uploads now report file read failures more reliably.


DorianZheng:
nice memory win. moving the host buffer from ram to a temp file at each edge is the right direction. a few things to look at.

**the request size cap is gone**
`upload_files` and `import_box` now take `Body`, not `Bytes`. the `Bytes` extractor enforced axum's 2 mib default limit. `Body` has no limit. the router sets none either. so a client can stream an endless body to `PUT /files` or `POST /boxes/import`. it all writes to a host temp file. the disk fills up. auth is off by default, so no key is needed. please set an explicit limit on the router.

**streaming replies send 200 before the file is read**
`stream_file_body` opens the file lazily inside the stream. but the handler already returned `200 OK`. if the open or a read fails, the client gets 200 and a cut-off body. `export_box` used to return a real 500 on a read error. that 500 path is now gone. a client can keep a truncated archive and think it worked.

**upload is no longer all-or-nothing**
in `send_upload_chunks`, a read error mid-file just drops the channel. the grpc stream then ends cleanly, not with an error. the guest unpacks the chunks it already got. so partial files can land in the container. the caller still sees an error, but the guest already changed state. the old code read the whole file first, so a read error sent nothing. the trigger is rare, but the guarantee changed.

**the tar helpers already exist**
`pack_dir_to_tar` and `unpack_tar_to_dir` do the same job as `boxlite_shared::tar::pack` and `unpack`. `copy_into` and `copy_out` already call those. the shared ones skip the `"."` entry and carry overwrite/mkdir rules. these new copies do not. the two paths will drift. please reuse the shared ones.

**download holds the extracted tree too long**
`download_files` keeps both `temp_dir` and `tar_dir` alive until the stream ends. but `temp_dir` is dead once the tar is packed. a slow client then keeps two copies on disk. drop `temp_dir` right after packing.

**dead code in `send_upload_chunks`**
the empty-file branch never runs. the caller always passes a real tar, which is at least 1024 bytes. it also changes behavior: the old path made the guest return 400, the new one sends an empty chunk. the `sent` count is returned but never used in production. only the test reads it. drop both.

**too many args**
`send_upload_chunks` takes 6 args. two bools sit side by side (`mkdir_parents`, `overwrite`). they are easy to swap by mistake. these are `UploadChunk`'s own fields. group them into one struct.

**the test assumes a full read**
the new test asserts the first chunk is exactly `CHUNK_SIZE`. `read` is allowed to return fewer bytes. so the test can fail for a valid reason. loop the read, or assert on the whole payload.

DorianZheng:
a note on the shape of the helpers. all three hand-rolled adapters here have a library form. other projects use those instead.

**request body → file: axum's own example is 3 lines**

axum ships this exact handler in `examples/stream-to-file/src/main.rs`:

```rust
let body_with_io_error = stream.map_err(io::Error::other);
let mut body_reader = pin!(StreamReader::new(body_with_io_error));
let mut file = BufWriter::new(File::create(path).await?);
tokio::io::copy(&mut body_reader, &mut file).await?;
```

so `write_body_to_file` can drop the manual `next()`/`write_all` loop:

```rust
pub(super) async fn write_body_to_file(body: Body, path: &Path) -> io::Result<()> {
    let reader = StreamReader::new(body.into_data_stream().map_err(io::Error::other));
    let mut reader = std::pin::pin!(reader);
    let mut file = BufWriter::new(tokio::fs::File::create(path).await?);
    tokio::io::copy(&mut reader, &mut file).await?;
    file.flush().await
}
```

**file → response: `ReaderStream`, not `async_stream!` + a copy**

`tokio_util::io::ReaderStream` reads into an internal `BytesMut` and yields owned `Bytes` with no extra copy. it has `with_capacity` for the chunk size. that replaces the `try_stream!` block, the `BoxStream`, the reused `vec![0u8; 64 * 1024]`, and the `Bytes::copy_from_slice` per chunk:

```rust
pub(super) async fn stream_file_body(path: &Path, keep: Vec<tempfile::TempDir>) -> io::Result<Body> {
    let file = tokio::fs::File::open(path).await?;
    let reader = ReaderStream::with_capacity(file, RESPONSE_CHUNK_SIZE);
    let guarded = reader.map(move |chunk| { let _ = &keep; chunk });
    Ok(Body::from_stream(guarded))
}
```

note the open is now eager. the handler can return a real 500 when the file is missing. that also fixes the 200-then-truncated-body problem from the earlier comment.

**grpc upload: pass the stream, skip the task and channel**

tonic polls the outbound stream itself while awaiting the response. its own client-streaming examples (`examples/src/routeguide/client.rs`, `record_route` / `route_chat`) pass a stream straight to the client method. no `mpsc::channel`, no `tokio::spawn`, no join handling. `async-stream` is already a boxlite dep:

```rust
let mut file = File::open(tar_path).await.map_err(...)?;
let outbound = async_stream::stream! {
    let mut buf = vec![0u8; CHUNK_SIZE];
    let mut first = true;
    while let Ok(n) = file.read(&mut buf).await {
        if n == 0 { break; }
        yield UploadChunk {
            dest_path: if first { dest.clone() } else { String::new() },
            container_id: cid.clone(),
            data: buf[..n].to_vec(),
            mkdir_parents,
            overwrite,
        };
        first = false;
    }
};
let response = self.client.upload(outbound).await.map_err(map_tonic_err)?.into_inner();
```

that removes `send_upload_chunks`, the channel, the spawn, the join error mapping, the `sent` counter, the `mem::take`, and the empty-file branch in one go. roughly 55 lines down to 15. one caveat: `while let Ok` still ends the stream silently on a read error. the chunk type can't carry an error, so that limit exists in both shapes. it is the same partial-upload point from the earlier comment, and it needs its own decision.

**tar helpers**

covered in the earlier comment: `boxlite_shared::tar::{pack, unpack}` already exist and are what `copy_into` / `copy_out` use.

cost: one feature flag, `tokio-util = { version = "0.7", features = ["io"] }` in `src/cli/Cargo.toml`. tokio-util is already in the workspace via the boxlite crate, so no new external crate.

G4614:
Verified the focused tests on both macOS locally and the Linux debug box After initializing submodules on the debug worktree, these passed:

- `make fmt:check:rust`
- `git diff --check`
- `make test:unit:rust FILTER=upload_chunks_streams_file_with_metadata_only_on_first_chunk`
- `make test:unit:cli FILTER=stream_file_body`
- `make test:unit:cli FILTER=write_body_to_file_persists_streamed_body`
- `make test:unit:cli FILTER=pack_download_dir_uses_shared_tar_without_parent_entry`

Small doc nit: the PR body lists two stale filter names (`send_upload_chunks...` and `pack_dir_to_tar_streams...`); the names above are the ones that actually run tests.

G4614:
I was able to pin down the read-error concern with a tiny local repro test. Opening a directory as `tokio::fs::File` makes `read()` return an error; with the current `while let Ok(n) = file.read(...)` loop, `upload_chunks` just ends the stream as if it hit EOF.

Temporary test I ran locally:

```rust
#[tokio::test]
async fn upload_chunks_silently_ends_when_source_read_errors() {
    let dir = tempfile::tempdir().unwrap();
    let file = File::open(dir.path()).await.unwrap();
    let mut stream = Box::pin(upload_chunks(file, UploadChunkFields {
        dest_path: "/workspace".to_string(),
        container_id: "container-1".to_string(),
        mkdir_parents: true,
        overwrite: false,
    }));

    assert!(stream.next().await.is_none());
}
```

That test passes today, which confirms the error is swallowed. I would treat this as the main blocker: the stream item or the upload path needs a way to surface source read errors so a partial upload cannot look like a clean EOF.

zhangrr:
Thanks, I added that repro as a regression test. upload_chunks now records source read errors through a shared error holder, and upload_tar checks it after the client-streaming RPC completes so a source read failure is returned as
  BoxliteError::Storage instead of being treated as clean EOF.

DorianZheng:
one more nit. `body_io.rs` names the mechanism, not the domain. every sibling in `handlers/` is a plain noun: `boxes`, `files`, `snapshots`, `executions`. other projects do the same for this exact job: meilisearch's streaming request-body extractor is `extractors/payload.rs`, docker's tar transfer helpers are the `archive` package (moby/go-archive), tower-http's file-to-body streaming lives in `services/fs` and `body.rs`. mechanism names like `ioutils` only appear in dedicated util trees (moby `pkg/ioutils`, crates.io `util/io_util.rs`), never among handlers — and go deprecated `io/ioutil` itself in 1.16. your own doc comment already names the domain: "streaming request/response body helpers for archive-sized payloads". suggest `payload.rs`.

DorianZheng:
the rework looks right. the shared tar reuse also quietly fixes a real hazard: the old download path used `tar::Builder` defaults, which follow symlinks. a guest-controlled absolute symlink could pull host files into the returned tar. `pack` with `follow_symlinks: false` preserves links instead. good outcome.

three things in the new code.

**over-limit chunked bodies return 500, not 413**
`RequestBodyLimitLayer` only returns an early 413 when the request has a `Content-Length` header. a chunked request (or a lying client) gets the body wrapped in `Limited` instead, and the limit error surfaces mid-poll inside the handler. `write_body_to_file` flattens it with `io::Error::other`, so the client sees `500 InternalError "failed to write upload archive: length limit exceeded"` for a client-caused condition. the old `Bytes` extractor returned a proper 413. fix: walk the error source chain, downcast `http_body_util::LengthLimitError`, and return 413 with the standard envelope. typed check, no string matching. (the early content-length 413 also skips the json error envelope, but the status is at least right there.)

**the 512 mib cap duplicates the guest cap with no cross-reference**
`MAX_REQUEST_BODY_BYTES` in `payload.rs` and `MAX_UPLOAD_BYTES` in `src/guest/src/service/files.rs` are the same value in two crates. nothing ties them together. if someone raises one, the other silently disagrees. at minimum comment the coupling; better, share one constant via `boxlite-shared`.

**`async-stream` is now an unused cli dep**
nothing under `src/cli` uses it anymore. the only `async_stream::stream!` in this pr lives in the boxlite crate, which has its own dep. one-line removal in `src/cli/Cargo.toml` while you're already touching that block.

one residual to be aware of, no action needed here: on a host read error mid-upload the grpc stream still ends cleanly, so the guest can extract the chunks it already received. the caller gets an error either way. making the guest discard partial uploads needs a protocol-level signal, which is out of scope for this pr.

zhangrr:
Get it

PR #924, opened 2026-07-06
feat(hooks): delta-trigger the verdict gate on turn tree changes

DorianZheng:
## Problem

#915 unlooped the verdict gate by making it self-declared: the Stop hook validates a dossier if present, but a no-dossier turn always passes and nothing prompts the model to produce one. In practice the verdict-auditor is never invoked. Worse, the soft nudge relies on `systemMessage`, which is shown to the human only — the model gets zero signal on every path, so the gate is pure telemetry.

## Change

- New `turn-baseline.sh` (UserPromptSubmit): snapshots the content-addressed working-tree hash at turn start (`.claude/.turn-baseline`, gitignored) and, while the tree is dirty, injects a one-line audit reminder into model context (UserPromptSubmit stdout reaches the model; Stop `systemMessage` does not).
- `preflight-verdict-check.sh`: with no dossier, a tree that changed since turn start now blocks turn-end with the re-audit instruction. No delta — or a missing/corrupt baseline — still allows (self-declared fallback, so other harnesses and mid-adoption sessions never trap). Present-dossier validation is unchanged.
- `VERDICT_GATE_HARD_BLOCK` defaults to `1`; soft mode remains as rollback/telemetry.
- CLAUDE.md Verify bullet updated to match.

## Why it can't loop (the #892 failure mode)

Default-deny demanded a dossier on every turn, so an audit's completion re-invoked a turn that demanded another audit. The delta keys the demand to evidence of work: all gate artifacts are gitignored, so audits themselves never create a delta, and acknowledgment/completion turns end freely. A residual re-block (stale baseline when a background completion re-opens a turn) converges in one audit, and the harness's 8-consecutive-block cap backstops.

## Tests

- `bash .claude/hooks/preflight-verdict-check.test.sh` — 23/23; the three new delta behaviors were demonstrated red against the #915 script (`got=allow expected=block`) before the implementation.
- `bash .claude/hooks/turn-baseline.test.sh` — 9/9 (baseline correctness, dirty-tree reminder, never-blocks-the-prompt contract).
- Regression: commit-push 21/21, pr-review 30/30.
- Live smoke in this repo: baseline + reminder written; Stop hook allows at no-delta, blocks with the hash pair on a scratch-file delta, allows again after revert.


DorianZheng:
superseded: trigger will be verdict-prose detection, not tree delta

PR #925, opened 2026-07-06
feat(hooks): trigger the verdict gate on detected verdict prose

DorianZheng:
## Problem

Under #915 the turn-end verdict gate is self-declared: the Stop hook validates a dossier if one exists, but nothing ever triggers one, so the verdict-auditor is never invoked in practice. File-less verdicts — "root cause is X", "prod looks healthy" — have no artifact at all, and the soft nudge relies on `systemMessage`, which the model never sees.

Requirements for this rework:
1. Prose verdicts must trigger the check, whether or not files changed.
2. The gate may only loop when the audit finds real issues — never on bookkeeping.

## Change

- `preflight-verdict-check.sh` now DETECTS: it extracts the turn's final assistant message from the transcript (`jq`, no model call), strips code spans/fences, and matches a curated pattern list ("root cause is", "tests pass", "looks healthy", "done", …). A verdict-shaped ending without a fresh dossier blocks with the audit instruction; chat and questions end freely.
- Loops are finding-driven only: a fresh+matching FAIL dossier persists and keeps blocking until a re-audit passes — that loop is the requirement. Stale/mismatched bindings (branch, HEAD, tree, age) are **discarded and re-detected** instead of blocking, which removes the bookkeeping-block class (e.g. a commit moving HEAD out from under a dossier).
- Audits are mandated synchronous (`run_in_background: false` in the block instruction): audit and verdict share one turn, so no completion event exists to re-open anything — the #892 loop cannot form. Validation runs before detection, so a re-ended turn consumes its dossier and never re-detects.
- `VERDICT_GATE_HARD_BLOCK` defaults to `1`; soft mode remains as rollback/telemetry. CLAUDE.md's Verify bullet and the verdict-auditor description updated in lockstep.

## Calibration

Patterns were tuned against 3,429 real assistant messages from this repo's session transcripts: 389 hits (11%), sampled hits overwhelmingly genuine verdicts; one live false negative found in the sweep ("Confirmed reachable …") is fixed and pinned by a test. A pattern miss degrades to #915's self-declared behavior — never a trap; a spurious hit costs one trivial-PASS audit — a tax, not a loop.

## Tests

`bash .claude/hooks/preflight-verdict-check.test.sh` — 25/25, with the 11 new behaviors demonstrated red against the #915 script first (detection cases `got=allow expected=block`; discard cases `got=block expected=allow`). Regression: commit-push 21/21, pr-review 30/30. Live smoke against this session's own transcript produced `decision:block` matching `"root cause is"`.

Replaces the closed delta-trigger attempt (#924), which could not see file-less verdicts.

## Summary by CodeRabbit

* **New Features**
  * Added a more reliable verdict-check flow that re-evaluates recent assistant messages and records decisions for traceability.
  * Enabled synchronous verdict auditing with support for fresh audit dossiers.

* **Bug Fixes**
  * Improved handling of stale or mismatched verdict dossiers so they are ignored and re-detected instead of incorrectly blocking turns.
  * Reduced duplicate judging of the same message and improved behavior when transcripts are missing or formatted differently.
  * Strengthened blocking behavior for verdict-shaped endings when no valid audit is available.


DorianZheng:
round 2 (c1990d41), per review feedback on the trigger design:

- the static pattern list is now the *fallback*, not the trigger. the hook asks haiku ("does this final message assert a verifiable verdict?") with a 20s cap; any failure degrades to the pattern list (now EN+中文), then to the honor system. `VERDICT_CLASSIFIER_CMD` overrides the classifier per harness.
- fixed a live misfire: Stop can fire before the final message lands in the transcript, so detection judged the previous, already-audited message. the hook now records a content-checksum of the judged message, settle-waits ≤2s for a fresh one, and never judges a message twice.
- extraction is harness-agnostic: one convention-based jq (role/type "assistant" + text under `text` keys) verified against real Claude Code and Codex rollout files; an invented future-agent schema is a pinned acceptance test; `VERDICT_EXTRACTOR_CMD` covers non-JSONL formats. new coding agents need zero code here.

suite is now 40 cases; live haiku triage 4/4 through the real hook (quoting-discussion → allow, ship verdict → block, 中文 verdict → block, question → allow, 4-6s each).

DorianZheng:
round 3 (517da8fe): triage prompt now exempts explanation/recap turns.

walking through architecture or restating already-audited facts was triaging YES and costing a full audit per Q&A turn. the prompt now requires fresh outcome-reporting (news of work just done); questions, teaching, and recaps answer NO. calibrated live: the two real explanation messages that false-triggered this session now allow; fresh ship / diagnosis / 中文 verdicts still block. hermetic suite unchanged (40/40) — the prompt is model data, not hook logic.

DorianZheng:
round 4 (9af7d1de): observability + agnostic audit execution.

- decision log: every Stop decision leaves one gitignored line (timestamp, message identity, deciding rung, outcome) — "why did/didn't the gate fire" is now `tail .claude/.verdict-decisions.log` instead of fixture reconstruction. bounded rotation; logging can never fail the hook.
- run-verdict-audit.sh: harness-neutral audit execution — feeds verdict-auditor.md to a headless model (VERDICT_AUDITOR_CMD override → claude -p fallback), succeeds only when a fresh well-formed dossier lands. the block text now offers Task (Claude Code) or this script (everything else), closing the self-audit degradation on non-Claude harnesses. live E2E: a real headless run correctly FAILed a planted bogus claim with concrete findings.

suites: 43/43 gate, 7/7 runner.

PR #926, opened 2026-07-06
feat(volumes): support single-file volume mounts

### LINKED ISSUE #785: SDK: support single-file volume mounts (host path must currently be a directory)
## Summary

Volume mounts only accept directories — a single-file host path is rejected. Sharing one config file into a box therefore isn't expressible; callers must instead share the parent directory, or (as we do) inline the file's contents into the box command with a heredoc.

## Where (as of `07fa30f9`)

`resolve_user_volumes` rejects any non-directory host path:

```rust
// src/boxlite/src/litebox/init/types.rs:66
if !resolved_path.is_dir() {
    return Err(BoxliteError::Config(format!(
        "Volume host path is not a directory: {}",
        vol.host_path
    )));
}
```

This is consistent with the virtio-fs share model (a share is a directory tree), so the constraint is understandable — but Docker/containerd support single-file bind mounts by sharing the parent and bind-mounting the file at the target inside the guest.

## Repro

```python
opts = BoxOptions(..., volumes=[("/host/path/config.yaml", "/etc/app/config.yaml", False)])
# -> BoxliteError::Config: "Volume host path is not a directory: /host/path/config.yaml"
```

## Expected

Support a single-file volume mount — e.g. share the file's parent directory as a virtio-fs tag and bind-mount just that file to `guest_path` inside the guest, leaving the rest of the parent dir invisible.

## Current workaround (downstream)

`apps/infra-local` inlines each config file into the box command via a shell heredoc instead of mounting it. This affects three services today (dex, caddy, minio-init) — see `apps/infra-local/boxlite_local/services.py` (`_DEX_ENTRYPOINT`, `_caddyfile`/`_CADDY_ENTRYPOINT_TEMPLATE`, `_MINIO_INIT_SCRIPT`). Each is a brittle `cat > … <<'__CFG__'` block that a single-file mount would replace.

BatmanByte:
## Summary
Adds support for single-file volume mounts. Previously a volume host path had to be a directory, so a single config file could not be mounted without sharing its whole parent directory or inlining the content via heredocs. Fixes #785.

## Changes
- `resolve_user_volumes`: a single-file host path now shares the file's **parent directory** via virtio-fs and records the file name as `subpath`; directory mounts are unchanged.
- Guest bind-mounts `{volume_dir}/{subpath}` to `guest_path`, so only that one file appears in the container — the rest of the parent dir stays invisible.
- Host jailer grants the VMM sandbox access to the shared **parent dir** (not the file path) so virtio-fs can serve it.
- `subpath` plumbed through `ContainerMount` and the `BindMount` proto (new field `6`; empty = whole-directory mount).

## How to verify
```sh
mkdir -p /tmp/sf && echo key=value > /tmp/sf/app.conf && echo secret > /tmp/sf/other.txt
boxlite run -v /tmp/sf/app.conf:/etc/app.conf:ro alpine \
  sh -c 'cat /etc/app.conf; find / -name other.txt'
# -> prints "key=value"; the sibling other.txt is NOT present in the container
```
Unit tests: `make test:unit:rust` — adds resolver, container-mount, and jailer path-access cases for single-file mounts.

## Risks / rollout
The file's parent directory is shared with the guest VM over virtio-fs (mounted at an internal path); only the requested file is bind-mounted into the container, so the container never sees siblings. A read-write single-file mount makes the shared parent writable at the virtio-fs layer, consistent with how directory mounts already behave.

## Summary by CodeRabbit

* **New Features**
  * Added single-file mounting within shared volumes via an optional mount subpath, including per-file staging when applicable.
  * Extended the mount/init protocol to target the correct in-guest file location.
* **Bug Fixes**
  * Improved sandbox path-access rules for file-backed volumes to avoid granting access to the file path and its parent directory.
  * Refined volume resolution to reliably distinguish directory mounts from single-file mounts.
* **Tests**
  * Added/updated unit tests for single-file classification, mount initialization, staging behavior, and path-access expectations.

### DIFF
--- a/src/boxlite/src/jailer/mod.rs
+++ b/src/boxlite/src/jailer/mod.rs
@@ -170,6 +170,7 @@ pub trait Jail: Send + Sync {
 
 use crate::disk::read_backing_chain;
 use crate::runtime::layout::BoxFilesystemLayout;
+use crate::volumes::{VolumeShare, classify_volume_share};
 use std::path::PathBuf;
 
 // ============================================================================
@@ -320,12 +321,14 @@ fn build_path_access(layout: &BoxFilesystemLayout, volumes: &[VolumeSpec]) -> Ve
         }
     }
 
-    // User volumes
+    // User volumes. Directories are shared directly, so grant the VMM access.
+    // Single files are staged under shared_dir (granted above), so they need no
+    // grant here — this also keeps the file's host siblings out of the sandbox.
     for vol in volumes {
         let p = PathBuf::from(&vol.host_path);
-        if p.exists() {
+        if let Some(VolumeShare::Dir(dir)) = classify_volume_share(&p) {
             paths.push(PathAccess {
-                path: p,
+                path: dir,
                 writable: !vol.read_only,
             });
         }
@@ -856,6 +859,32 @@ mod tests {
         assert!(
             paths.iter().all(|p| p.path != Path::new("/does/not/exist")),
             "Nonexistent volume should be skipped"
+        );
+    }
+
+    #[test]
+    fn test_build_path_access_single_file_grants_no_host_dir() {
+        let dir = tempdir().unwrap();
+        let layout = test_layout(dir.path().to_path_buf());
+
+        let parent = dir.path().join("cfg");
+        std::fs::create_dir_all(&parent).unwrap();
+        let file = parent.join("app.conf");
+        std::fs::write(&file, "k=v\n").unwrap();
+
+        let volumes = vec![VolumeSpec {
+            host_path: file.to_string_lossy().to_string(),
+            guest_path: "/etc/app.conf".to_string(),
+            read_only: true,
+        }];
+
+        let paths = build_path_access(&layout, &volumes);
+
+        // A single file is staged under shared_dir, so it must not widen path
+        // access to the file or its parent (which would expose host siblings).
+        assert!(
+            paths.iter().all(|p| p.path != file && p.path != parent),
+            "single-file volume must not grant its host file or parent dir"
         );
     }
 
--- a/src/boxlite/src/litebox/init/tasks/vmm_spawn.rs
+++ b/src/boxlite/src/litebox/init/tasks/vmm_spawn.rs
@@ -20,7 +20,9 @@ use crate::runtime::types::ContainerID;
 use crate::util::find_binary;
 use crate::vmm::controller::{ShimController, VmmController, VmmHandler};
 use crate::vmm::{Entrypoint, InstanceSpec, VmmKind};
-use crate::volumes::{ContainerMount, ContainerVolumeManager, GuestVolumeManager};
+use crate::volumes::{
+    ContainerMount, ContainerVolumeManager, GuestVolumeManager, stage_single_file,
+};
 use async_trait::async_trait;
 use boxlite_shared::Transport;
 use boxlite_shared::errors::{BoxliteError, BoxliteResult};
@@ -174,15 +176,27 @@ async fn build_config(
     // Add user volumes via ContainerVolumeManager
     let mut container_mgr = ContainerVolumeManager::new(&mut volume_mgr);
     for vol in &user_volumes {
+        // Single-file volume: stage the file into a dedicated dir under the box's
+        // shared tree (already granted to the VMM sandbox) and share that dir, so
+        // virtio-fs never exposes the file's host siblings. Directories share as-is.
+        let share_dir = match &vol.subpath {
+            None => vol.host_path.clone(),
+            Some(file_name) => {
+                let staging_dir = layout.shared_dir().join("user-volumes").join(&vol.tag);
+                stage_single_file(&staging_dir, &vol.host_path, file_name, vol.read_only)?;
+                staging_dir
+            }
+        };
         container_mgr.add_volume(
             container_id.as_str(),
             &vol.tag,
             &vol.tag,
-            vol.host_path.clone(),
+            share_dir,
             &vol.guest_path,
             vol.read_only,
             vol.owner_uid,
             vol.owner_gid,
+            vol.subpath.clone(),
         );
     }
     let container_mounts = container_mgr.build_container_mounts();
--- a/src/boxlite/src/litebox/init/types.rs
+++ b/src/boxlite/src/litebox/init/types.rs
@@ -12,7 +12,7 @@ use crate::runtime::layout::BoxFilesystemLayout;
 use crate::runtime::options::VolumeSpec;
 use crate::runtime::rt_impl::SharedRuntimeImpl;
 use crate::vmm::controller::VmmHandler;
-use crate::volumes::{ContainerMount, GuestVolumeManager};
+use crate::volumes::{ContainerMount, GuestVolumeManager, VolumeShare, classify_volume_share};
 use boxlite_shared::errors::{BoxliteError, BoxliteResult};
 use std::path::PathBuf;
 use std::sync::atomic::Ordering;
@@ -41,6 +41,9 @@ pub struct ResolvedVolume {
     pub owner_uid: u32,
     /// Owner GID of host directory (for auto-idmap in guest).
     pub owner_gid: u32,
+    /// For a single-file mount, the file's name (staged and bind-mounted on its
+    /// own); `None` for a whole-directory mount.
+    pub subpath: Option<String>,
 }
 
 pub fn resolve_user_volumes(volumes: &[VolumeSpec]) -> BoxliteResult<Vec<ResolvedVolume>> {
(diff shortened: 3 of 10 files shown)


DorianZheng:
virtiofs can only share a folder, never a single file. so every runtime ends up doing the same dance: put the file in some shared folder, then bind-mount just that one file inside the vm. the ways to pick that folder:

| option | how it works | good | bad |
|---|---|---|---|
| share the parent folder (this pr) | share the folder around the file, bind only the file into the container | no copy, live edits, works on macos + linux, no root | the guest vm can read the file's siblings. the container can't, but a container escape can — `-v ~/.aws/credentials:...` shares all of `~/.aws` with the vm |
| staging folder + hardlink | per-box folder, hard-link the file into it, share that | vm sees only the one file; edits flow both ways (same inode) | same-filesystem only (fails across disks/volumes, so it needs a fallback story); save-by-rename (vim, sed -i) stops updates — docker file binds have the exact same limit |
| copy into a staging folder | copy the file once at boot | simple, works everywhere | edits after boot don't show up. not really a mount |
| host bind mount into the shared folder | what kata does | exact docker behavior, vm sees only the file | needs a root linux host; macos has no bind mounts — not available to us |

fwiw docker itself never had this problem: a file bind is the same mount(2) call as a folder bind (runc just creates the destination as an empty file instead of a dir). the folder-only limit comes purely from the virtiofs transport.

BatmanByte:
Thanks for the detailed breakdown — that maps the space out really well.

I agree the parent-dir share is the weak spot: the container only sees the file (verified), but the guest VM sees the siblings, so `-v ~/.aws/credentials:...` hands the whole `~/.aws` to the VM — read-only leaks sibling reads, read-write widens it to sibling writes.

I'd like to move this PR to the **staging dir + hardlink** option, since it's the one that keeps "VM sees only the one file" while staying no-copy / live-edit / no-root / macOS+Linux:

- per-box staging dir under the box home; hardlink the file in; virtio-fs shares only that dir; guest bind-mounts the file (the `subpath` plumbing in this PR already covers the bind step)
- fall back to a copy on `EXDEV` (cross-filesystem), where hardlink isn't possible
- jailer then grants only the staging dir — no siblings

Limitations I'd document: cross-fs falls back to copy (so live edits need same-fs), and save-by-rename editors (vim, `sed -i`) break the link — same as Docker's file binds, as you noted.

Does that match the direction you'd want before I rework it? Or would you prefer copy-only for simplicity and accept "no live edits"?

BatmanByte:
Reworked to the **staging dir + hard link** approach (pushed, force-update).

**Mechanism**
- Single file → hard-linked into a per-box dir under the box's shared tree (`shared_dir/user-volumes/{tag}/`), and virtio-fs shares only that dir.
- Cross-filesystem (`EXDEV`) falls back to a copy (guest then sees a boot-time snapshot rather than live edits).
- Guest bind-mounts the one file to `guest_path` (the existing `subpath` plumbing).
- The jailer already grants the shared tree, so file volumes now widen **no** host path — and the file's host siblings are invisible to the VM itself, not just the container. That closes the sibling-exposure you and CodeRabbit flagged.

**Verified (local, macOS + libkrun build)**
- read-only single file: mounts at `guest_path`; a sibling `other.txt` in the source dir is absent everywhere in the guest.
- read-write single file: guest writes propagate back to the host source file — confirms the hard link (same inode), not a copy.
- unit tests: hard-link same-inode + idempotent re-stage; classify/resolver/jailer updated.

**Documented limitations** (as you noted, same as Docker file binds)
- cross-fs falls back to copy → live edits need same-fs.
- save-by-rename editors (vim, `sed -i`) break the link.

Happy to adjust — e.g. copy-only if you'd rather avoid the hard-link edge cases, or a different staging location.

G4614:
## Security regression: `read_only` single-file volumes aren't actually read-only on the host

`stage_single_file` always hard-links the staged file to the real host source (same inode), regardless of `read_only`. The jailer also grants `writable: true` on the whole `shared_dir` the staging dir lives under, with no per-file deny for single-file volumes (directory volumes *do* get `writable: !read_only`).

So for a `read_only: true` mount, any write that reaches the staged path — buggy guest-side `ro` enforcement, or a compromised guest VM — mutates the user's **real host file**, not a throwaway copy.

Reproduced with this test (builds clean against `8c6046e2`, fails as expected):

```
thread '...' panicked at src/boxlite/src/volumes/staging.rs:128:9:
assertion `left == right` failed: a write through a read-only single-file mount reached the real host file (staging hard-links unconditionally, so read_only isn't actually enforced)
  left: "TAMPERED-BY-GUEST\n"
 right: "sensitive=true\n"
```

Patch (test only, no production code changed):

```diff
diff --git a/src/boxlite/src/volumes/staging.rs b/src/boxlite/src/volumes/staging.rs
index 5dac9ac5..66f266cf 100644
--- a/src/boxlite/src/volumes/staging.rs
+++ b/src/boxlite/src/volumes/staging.rs
@@ -93,4 +93,43 @@ mod tests {
 
         assert_eq!(std::fs::read_to_string(&staged).unwrap(), "two");
     }
+
+    /// Security regression: `read_only: true` on a single-file volume should
+    /// mean read-only at the host filesystem level, but it doesn't.
+    ///
+    /// `stage_single_file` always hard-links the staged copy to the real
+    /// host file (same inode), regardless of whether the mount is read-only.
+    /// The jailer also grants `writable: true` on the whole shared_dir the
+    /// staging directory lives under, with no explicit per-file deny for
+    /// single-file volumes (unlike directory volumes, which get
+    /// `writable: !read_only` -- see `jailer::build_path_access`). So any
+    /// write that reaches the staged path -- a bug in guest-side `ro`
+    /// enforcement, or a compromised guest VM -- mutates the user's real
+    /// host file even though they asked for read-only.
+    ///
+    /// This test currently FAILS: a write through the staged file reaches
+    /// the real source, because staging always hard-links. The fix needs to
+    /// make read-only mounts not share an inode with the source (e.g. copy
+    /// instead of hard-link when the volume is read-only) -- which likely
+    /// means threading a `read_only` flag into `stage_single_file` and
+    /// updating its call site in `litebox/init/tasks/vmm_spawn.rs`.
+    #[test]
+    fn read_only_single_file_mount_write_must_not_reach_real_host_file() {
+        let tmp = tempfile::tempdir().unwrap();
+        let source = tmp.path().join("real_host_secret.conf");
+        std::fs::write(&source, "sensitive=true\n").unwrap();
+        let staging = tmp.path().join("staging");
+
+        // Mirrors how vmm_spawn.rs stages a volume today -- nothing here
+        // says this mount is read-only, which is exactly the bug.
+        let staged = stage_single_file(&staging, &source, "app.conf").unwrap();
+
+        let _ = std::fs::write(&staged, "TAMPERED-BY-GUEST\n");
+        assert_eq!(
+            std::fs::read_to_string(&source).unwrap(),
+            "sensitive=true\n",
+            "a write through a read-only single-file mount reached the real host file \
+             (staging hard-links unconditionally, so read_only isn't actually enforced)"
+        );
+    }
 }
```

Suggested direction: copy instead of hard-link when `read_only` is true (needs threading that flag into `stage_single_file` and updating its call site in `litebox/init/tasks/vmm_spawn.rs`).

@BatmanByte — want to take this yourself, or would it help if we pushed a fix on top of your branch?

DorianZheng:
on the `EXDEV => fs::copy` fallback in `stage_single_file`: the copy itself is fine, but two things about it need fixing — it's silent, and it treats read-only and read-write the same.

first, the mechanism can't really be improved for the cross-filesystem case. i checked the alternatives:

- **hardlink** (the primary) can never cross filesystems — inodes are per-fs, `EXDEV` is a hard posix rule, not something we can code around. that's the whole reason this fallback exists.
- **bind mount** is what kata does for single files, and it's strictly better (crosses filesystems *and* stays live). but it needs root/`CAP_SYS_ADMIN` on linux and doesn't exist at all on macos, so it doesn't fit "runs unprivileged, runs on mac".
- **symlink** doesn't work: the guest resolves the link target in its own namespace (libkrun's passthrough opens every lookup `O_NOFOLLOW`), so a symlink to a host path just dangles inside the vm.
- **reflink / clonefile** (copy-on-write) is same-filesystem only too, so it `EXDEV`s in the exact case we're handling — and it's a snapshot anyway, edits don't propagate.

so `fs::copy` is genuinely the only portable fallback. the problem is only how it's used.

what to change:

- **don't do it silently.** a copy is a one-time snapshot — host edits after boot never reach the guest, guest edits never reach the host. right now nothing tells the user their "mount" is actually a frozen copy. at minimum `tracing::warn!` it.
- **split on read-only.** for `:ro` a snapshot is fine (the guest never writes; that's what read-only config at boot means) — copy + warn. for `:rw` a silent copy is data loss: the user expects writes to persist to the host file and they vanish at teardown. that case should hard-error instead, pointing at the two real options — mount the parent directory, or put the file on the same filesystem as the box home.

minor: if you keep the copy, write to a temp name and rename into place so a box that reads mid-copy can't see a half-written file.

BatmanByte:
Fixed in `a39849ff`: `stage_single_file` now takes `read_only` and **copies** for read-only mounts (separate inode → the host source can never be modified through the mount), while read-write still hard-links for live write-back. Call site in `vmm_spawn.rs` passes `vol.read_only`.

One correction on the framing: read-only *is* enforced at the hypervisor level today — `add_virtiofs(..., read_only)` maps to `krun_add_virtiofs3` with the ro flag ([`context.rs:441`](https://github.com/boxlite-ai/boxlite/blob/main/src/boxlite/src/vmm/krun/context.rs#L441), "read-only at the hypervisor level"), so a compromised guest can't write the share. But you're right that single-file ro relied *only* on that, while directory volumes also get sandbox-level `writable: !read_only`. Copying for ro makes the host-immutability structural instead of a dependency on the ro flag, and brings single-file ro to parity with directory ro.

Added a test that fails without the fix (write through a ro staged file reaching the source) — reverting the copy branch reproduces your inode-equality failure. rw hard-link write-back and ro read + sibling isolation both re-verified end-to-end.

BatmanByte:
Done in `d69de80d`. Agreed on all three — and your mechanism survey matches what I found (bind mount needs root/absent on macOS; symlink dangles under libkrun's `O_NOFOLLOW` passthrough; reflink is same-fs too), so `fs::copy` is the only portable fallback and the fix is purely in how it's used:

- **rw + `EXDEV` → hard error** (no silent snapshot / write loss), pointing at the two real options: mount the parent directory, or place the file on the same filesystem as the box home.
- **ro → copy + `tracing::warn!`** that it's a boot-time snapshot (host edits afterward don't reach the guest). ro never writes, so the snapshot is fine — the warn just makes it non-silent.
- **copy → temp file + atomic rename**, so a box reading the staged path mid-copy can't see a half-written file.

Re-verified end-to-end: ro reads + sibling isolation + host source untouched (with the warn emitted), rw hard-link write-back intact. Unit tests cover ro-copy (separate inode, write can't reach source) and rw-hardlink (same inode).

DorianZheng (approved):
verified: compiles clean, 26 volume/jailer/resolver tests green. the two substantive concerns are resolved — read-only single-file is copied to a separate inode (host source can't be modified through the mount), and the cross-filesystem read-write case hard-errors instead of silently snapshotting. remaining items (stage_single_file's read_only bool switching two workflows, inline user-volumes path vs a layout method, Config vs Storage error variant, subpath naming) are non-blocking polish, fine as follow-ups.

PR #928, opened 2026-07-06
feat(githooks): agent-agnostic gate layer at the git boundary

DorianZheng:
## Problem

The gate stack (commit/push audits, verdict gate) enforces only through harness hooks — Claude Code's PreToolUse, Codex's mirror. An agent with no hook system, or one we haven't wired, commits unaudited. Per-harness wiring also means every new coding agent is an integration project.

## Change

Native git hooks are the one interception point every agent shares — they fire for any process that runs `git`, forever.

- `.githooks/pre-commit` / `pre-push`: thin adapters over the existing `preflight-commit-push.sh` — synthesize its PreToolUse payload, translate its `permissionDecision` JSON into hook exit codes, then chain to the framework-managed hooks (prek fmt/lint, test matrix) via `--git-common-dir`. Policy stays single-sourced; the adapters carry zero judgment.
- **Humans stay ungated**: enforcement binds to named harness markers only — `CLAUDECODE`, `CODEX_SANDBOX`, or `AGENT_GATED=1` for any future agent (one env line in its wiring, zero code here). Named markers, not prefix wildcards: plugin artifacts like `CODEX_COMPANION_*` must not gate a session that isn't Codex-driven.
- **Single consumer**: with the layer installed, the PreToolUse gate defers — but only when the delegate hook actually exists at the checkout (a missing delegate would stand both layers down; closed over open). `GITHOOK_DELEGATED` marks the git layer's own call so the layers can't defer to each other.
- Install once per clone: `bash .githooks/install.sh` (sets `core.hooksPath`). The native `--no-verify` flag remains the documented human bypass.

Two defects were found and pinned during development: `--git-path hooks/<name>` honors `core.hooksPath` and resolved the "chained" hook to the adapter itself (self-exec recursion — infinite for humans, double-consume for agents; chaining now uses `--git-common-dir` + a realpath guard), and the original delegation guard failed open when the delegate was absent.

## Tests

`bash .githooks/githooks.test.sh` — 13/13 in hermetic (`env -i`) fixture repos driving real commit/push flows: agent gated / human exempt / audit consumed / framework chaining / both delegation directions / delegate-missing fail-closed. Regression: `preflight-commit-push.test.sh` 21/21 (guard inert without hooksPath), `preflight-pr-review` 30/30. shellcheck clean.

Also fixes a stale repo name in `commit-push-auditor.md`.

## Summary by CodeRabbit

* **New Features**
  * Added Git hook-based checks for commits and pushes, helping prevent actions that don’t meet the required audit flow.
  * Added an installer to enable the new Git hooks path with a confirmation message.
  * Updated the auditor label to match the current repository name.

* **Bug Fixes**
  * Improved hook chaining so existing Git hook behavior still runs after the new checks.
  * Added safeguards to avoid repeated hook execution and to keep behavior consistent across commit and push flows.

### DIFF
--- a/.claude/agents/commit-push-auditor.md
+++ b/.claude/agents/commit-push-auditor.md
@@ -4,7 +4,7 @@ description: Independent auditor that judges a pending git commit or push agains
 tools: Read, Bash, Write
 ---
 
-You are the CLAUDE.md compliance auditor for the boxlite3 repository.
+You are the CLAUDE.md compliance auditor for the boxlite repository.
 
 The parent agent must tell you the exact git command they are about to retry
 (e.g. `git commit -m "..."` or `git push origin <branch>`). Treat that as the
--- a/.claude/hooks/preflight-commit-push.sh
+++ b/.claude/hooks/preflight-commit-push.sh
@@ -64,6 +64,27 @@ else
 fi
 
 repo_root="$(git rev-parse --show-toplevel 2>/dev/null || pwd)"
+
+# Delegate to the git-level gate when installed: with core.hooksPath pointing at
+# .githooks, the same contract is enforced by .githooks/pre-commit|pre-push for
+# EVERY process (any agent, any harness — and humans stay exempt there), so this
+# PreToolUse layer steps aside to keep the audit artifact single-consumer.
+# GITHOOK_DELEGATED marks the call coming FROM that git-level gate — the one
+# caller that must not be deferred, or the two layers would defer to each other
+# and everything would silently pass.
+# Defer ONLY when the delegate hook actually exists at this checkout: git skips
+# missing hooks silently, so hooksPath-configured + delegate-absent (old ref,
+# broken install) would otherwise stand BOTH layers down — a silent bypass.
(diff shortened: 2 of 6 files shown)


DorianZheng:
live-verified on the real repo (this worktree), not just fixtures:

- install → agent commit attempt: the PreToolUse layer deferred (delegation observed live — the deny came from the git-level hook as command output, not from the harness gate) and the commit was rejected rc=1 with the auditor instruction.
- human commit (no harness markers): allowed rc=0, and the real prek chain ran through the adapter's --git-common-dir resolution ("Quality autofix … Skipped") on real worktree topology.
- state restored exactly afterwards (smoke commit dropped, prior hooksPath put back).

one note the live run surfaced: this repo's prek wiring itself uses core.hooksPath (absolute path to the shared .git/hooks), so install.sh overwrites that value. behavior is preserved because the adapters chain to --git-common-dir/hooks — the same directory — but install.sh silently replacing a non-default hooksPath is worth a warning line before this leaves draft.

PR #936, opened 2026-07-07
Feat/boxlite integration skill

mateng0915:
## Summary

Adds a Claude Code / Cursor skill (`boxlite-integration`) that helps developers integrate BoxLite sandboxed execution into their existing AI agent codebases. Install with one command, invoke to get correct integration code without reading SDK docs first.

```bash
npx skills add boxlite-ai/boxlite --skill boxlite-integration
```

## Motivation

Integrating BoxLite into an agent is non-trivial. Developers must understand box lifecycle management, choose the right concurrency model, and apply correct timeout + zombie-process prevention — and the cost of getting any of these wrong is either a resource leak or a security hole. This skill encodes that knowledge and applies it automatically by reading the developer's existing code and injecting the right BoxLite wrapping in the right places.

This mirrors how the Langfuse skill works: install once, invoke when needed, get correct integration code without reading the SDK docs.

## What the skill does

1. **Reads existing agent code** — finds the current execution point (subprocess, `eval()`, `execSync`, etc.) and flags it as a security vulnerability before proceeding.
2. **Determines box lifecycle** — auto-detects from code structure (FastAPI server → long-running; single script → short-lived). Only asks the user when multi-tenancy cannot be inferred from the code.
3. **Applies the integration** — rewrites the execution function using the correct SDK pattern (Python, Node.js, Go, C, or REST client for cloud deployment).
4. **Enforces the critical patterns** — timeout + `execution.kill()` (zombie prevention) and proper box cleanup are always included, with explanations of why they are non-optional.

## Skill structure

```
skills/boxlite-integration/
├── SKILL.md          # Main skill: 5-step integration workflow
└── references/
    ├── python.md     # Python SDK patterns (SimpleBox, CodeBox, REST, timeout)
    ├── node.md       # Node.js SDK patterns (SimpleBox, CodeBox, async disposal)
    ├── go.md         # Go SDK patterns (NewRuntime, lifecycle, timeout)
    ├── c.md          # C SDK patterns (Simple API, Native API, memory management)
    └── patterns.md   # Lifecycle, zombie prevention, concurrency, file transfer
```

## Evaluation

Evaluated against 3 realistic integration scenarios using with-skill vs. without-skill methodology — each run by a fresh agent with no prior context:

**Scenario 1: Python agent using subprocess**
Skill surfaces the security issue upfront and explains why `execution.kill()` is non-optional; baseline omitted the reasoning.

**Scenario 2: FastAPI server with eval() and concurrent users**
Skill correctly identifies multi-tenant context and enforces per-request box isolation; baseline missed timeout handling entirely, leaving zombie processes in the VM. *(source text was garbled here — please confirm this reflects the intended finding)*

**Scenario 3: Node.js agent using execSync**
Skill flags shell injection risk from string interpolation; provides both per-call and singleton lifecycle variants with explicit selection guidance.

Two iterations of refinement were made:
- **Iteration 1→2:** Added explicit dangerous-pattern detection — `eval()`, `shell=True`, and `execSync` with string interpolation are now flagged as security vulnerabilities before integration begins. *(source text was garbled here — please confirm this reflects the intended wording)*
- Promoted zombie-prevention code (`execution.kill()` on timeout) from a reference file into the main skill body so it cannot be missed.

## How to verify

```bash
# Install
npx skills add boxlite-ai/boxlite --skill boxlite-integration

# Preview without installing
npx skills use boxlite-ai/boxlite --skill boxlite-integration
```

Then in Claude Code or Cursor: `/boxlite-integration`

## Changes since review

**Round 1 (CodeRabbit review 1):**
- Fix cleanup teardown: nested try/finally ensures `remove()` runs even if `stop()` raises
- Fix singleton lifecycle bug: replace `async with` anti-pattern with explicit `create → start → shutdown()`
- Add `references/go.md` and `references/c.md` covering all four supported SDKs

**Round 2 (CodeRabbit review 2):**
- Fix Go type mismatch: `Run()` now extracts `result.Stdout` from `ExecResult` (was compile error)
- Fix Go cleanup: defer now calls `Stop()` before `Remove(force=true)` on active box
- Fix Python `safe_exec`: catch `BaseException` instead of `TimeoutError` only — kills guest process on cancellation too
- Fix Node.js quick-start: wrap `box.stop()` in try/finally
- Add input validation step before `box.exec()` (allowlist commands, reject shell metacharacters)
- Pin Go SDK install to `@v0.9.0` instead of floating version

**Round 3 (reviewer — DorianZheng):**
- Go: fix `Exec` calls to variadic form; remove spurious `.Wait()` on `*ExecResult`; replace all `rt.Remove(..., force)` with `rt.ForceRemove()`; rewrite `safeExec` to use `StartExecution` which exposes `.Wait()`/`.Kill()`
- C: rewrite Native API example with correct async callback pattern; `boxlite_box_exec` is sync out-param, `boxlite_create_box`/`start_box`/`execution_wait` are async callbacks
- Node.js: `CodeBox.run()` returns `string` not `{stdout}`; REST requires `JsBoxlite.rest(new BoxliteRestOptions({...}))` not constructor args; exec timeout is 6th positional arg; `SimpleBox.exec()` never throws on non-zero exit
- Python: `SimpleBox.exec` is variadic `(cmd, *args)`; returns `ExecResult` directly with no `.wait()`/`.kill()`; clarify two API levels (SimpleBox vs lower-level Box from `runtime.create()`)
- Python: BoxOptions has no security= param — AdvancedBoxOptions is not re-exported from top-level boxlite; access via from boxlite.boxlite import AdvancedBoxOptions
- Python: execution.wait() returns ExecResult{exit_code, error_message} only — no .stdout; Full Control example updated accordingly
- Python: volume tuple third element must be bool (True), not string "ro"
- Python: except asyncio.TimeoutError misses CancelledError — changed to except BaseException
- patterns.md: volume table "ro" → True; long-lived box example rewritten to use SimpleBox.start() to avoid execution.wait() stdout mismatch; concurrency "safe" claim clarified (same-box execs share filesystem/PIDs/network)
- SKILL.md: Node.js safeExec wrapper scoped to JsBoxlite low-level only; SimpleBox uses { timeoutSecs } option

**Round 4 (CodeRabbit round 4 + reviewer G4614):**
- C: fix on_stdout callback signature to (const uint8_t*, size_t, void*) with fwrite — was const char* with printf("%s") (potential buffer over-read)
- C: check boxlite_options_new return value before passing to boxlite_create_box
- C: add CBoxliteError* to boxlite_execution_wait row in sync/async summary table
- Python: BoxOptions has no security= param — security goes under advanced=AdvancedBoxOptions(security=...)
- Python: execution.wait() returns ExecResult{exit_code, error_message} only; removed result.stdout references; long-lived box example rewritten to use SimpleBox.start() which returns ExecResult with .stdout directly
- Python: volume tuple third element must be bool (True), not the string "ro"
- Python: except asyncio.TimeoutError misses CancelledError — changed to except BaseException
- Python: CodeBox.run() returns stdout unconditionally regardless of exit code (unlike Node.js which raises ExecError)
- patterns.md: clarify same-box concurrent execs share filesystem/PIDs/network — not inter-exec isolated
- Go: update pinned SDK version to v0.9.7


mateng0915:
> [CLA assistant check](https://cla-assistant.io/boxlite-ai/boxlite?pullRequest=936) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/boxlite-ai/boxlite?pullRequest=936) before we can accept your contribution.You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/boxlite-ai/boxlite?pullRequest=936) it.

 I have read the CLA Document and I hereby sign the CLA

DorianZheng (commented):
these are copy-paste examples the skill injects into user code, so they have to match the real sdk. most of the non-trivial ones don't. two recurring root causes:

1. three exec return-shapes are used interchangeably: `CodeBox.run()` returns a plain string (no `.stdout`); high-level `SimpleBox.exec()` / go `Box.Exec()` return an already-waited result (`.stdout`, but no `.wait`/`.kill`); the low-level `Execution`/`JsExecution` handle has `.wait`/`.kill` but no `.stdout`. examples mix them, so `.stdout` and `.wait()/.kill()` land on the wrong object.
2. go.md and the c native-api section are written against signatures that don't exist — they won't compile.

inline notes below, grouped by file. line numbers refer to the new files.

skills/boxlite-integration/references/go.md
```diff
@@ -0,0 +1,217 @@
...
+    defer func() {
+        box.Stop(ctx)
+        rt.Remove(ctx, box.ID(), true) // force=true ensures removal even if stop fails
+    }()
+
+    if err := box.Start(ctx); err != nil {
+        log.Fatal(err)
+    }
+
+    exec, err := box.Exec(ctx, "python", []string{"-c", "print('hello')"})
```
DorianZheng: won't compile. `Box.Exec` is `Exec(ctx, name string, arg ...string) (*ExecResult, error)` (exec.go:181) — variadic, and it already runs+waits internally. two problems here and in every go example:
- passing `[]string{"-c", ...}` to `...string` is a compile error; use `box.Exec(ctx, "python", "-c", "print('hello')")`.
- the return is `*ExecResult{ExitCode, Stdout, Stderr}`, which has no `Wait`/`Kill`. so `exec.Wait(ctx)` (line 54), the `exec.Wait()`/`exec.Kill()` in safeExec (119, 127, 129) and CodeRunner.Run (189) don't exist. for buffered use just read `result.Stdout`; for streaming/kill use `box.StartExecution(...)` which returns `*Execution` (that one has `.Wait(ctx) (int, error)` and `.Kill(ctx)`).

skills/boxlite-integration/references/go.md
```diff
@@ -0,0 +1,217 @@
...
+    defer rt.Close()
+
+    ctx := context.Background()
+    box, err := rt.Create(ctx, "python:slim")
+    if err != nil {
+        log.Fatal(err)
+    }
+    defer func() {
+        box.Stop(ctx)
+        rt.Remove(ctx, box.ID(), true) // force=true ensures removal even if stop fails
```
DorianZheng: `Runtime.Remove` takes `(ctx, idOrName)` — 2 args, no force flag (runtime.go:290). `rt.Remove(ctx, box.ID(), true)` is a compile error (`too many arguments`). force removal is the separate `rt.ForceRemove(ctx, box.ID())` (runtime.go:295). same at lines 87, 101, 179, 198.

skills/boxlite-integration/references/c.md
```diff
@@ -0,0 +1,220 @@
...
+    BoxliteCommand cmd = {
+        .command = "python",
+        .args = args,
+        .argc = 2,
+        .timeout_secs = 30.0,
+    };
+    CExecutionHandle* execution = NULL;
+    int exit_code = 0;
+
+    if (boxlite_execute(box, &cmd, on_output, NULL, &execution, &error) == Ok) {
```
DorianZheng: the native-api exec block doesn't compile/link. there is no `boxlite_execute` symbol — the exec fn is `boxlite_box_exec(CBoxHandle*, const BoxliteCommand*, CExecutionHandle** out, CBoxliteError*)` (boxlite.h:403), with no inline output callback (streaming is registered afterwards via `boxlite_execution_on_stdout/on_stderr/on_exit`). line 123 is also wrong: `boxlite_execution_wait` is async — `boxlite_execution_wait(CExecutionHandle*, CExecutionWaitCb cb, void* user_data, CBoxliteError*)` (boxlite.h:437) — the exit code arrives via the callback's int arg, not a caller `&exit_code`.

skills/boxlite-integration/references/c.md
```diff
@@ -0,0 +1,220 @@
...
+    // Box options
+    if (boxlite_options_new("python:slim", &opts, &error) != Ok) {
+        fprintf(stderr, "options: %s\n", error.message);
+        boxlite_error_free(&error);
+        boxlite_runtime_free(runtime);
+        return 1;
+    }
+
+    // Create + start box
+    if (boxlite_create_box(runtime, opts, &box, &error) != Ok) {
```
DorianZheng: `boxlite_create_box` is async, not a synchronous out-param call: `boxlite_create_box(CBoxliteRuntime*, CBoxliteOptions*, CBoxCreateBoxCb cb, void* user_data, CBoxliteError*)` (boxlite.h:341). the box is delivered to the callback, and callbacks only fire while a drain loop (`boxlite_runtime_drain`) pumps events. `boxlite_create_box(runtime, opts, &box, &error)` (also line 175) passes a `CBoxHandle**` where a callback is expected. the whole native-api section models a synchronous api the c side doesn't have — the simple api section above it is the one that's actually synchronous and correct.

skills/boxlite-integration/references/patterns.md
```diff
@@ -0,0 +1,192 @@
...
+        return "Error: execution timed out"
+```
+
+Note: do **not** use `async with SimpleBox(...) as box` and then assign `box` to a module-level variable — the context manager closes the box when it exits, so the cached reference is already stopped.
+
+For per-request isolation (safer for multi-user):
+```python
+async def execute_code(code: str, timeout: int = 30) -> str:
+    async with boxlite.SimpleBox(image="python:slim") as box:
+        execution = await box.exec("python", ["-c", code])
```
DorianZheng: this is the path recommended for untrusted multi-user code, and it crashes. `SimpleBox.exec` is `exec(cmd, *args, ...)` returning a finished `ExecResult` (simplebox.py:175). `box.exec("python", ["-c", code])` passes the list as a single positional arg (args become `[["-c", code]]`, a list nested in a list) → TypeError; and `execution.wait()` (187) / `execution.kill()` (190) don't exist on `ExecResult`. correct: `result = await box.exec("python", "-c", code, timeout=30); return result.stdout`.

skills/boxlite-integration/references/python.md
```diff
@@ -0,0 +1,166 @@
...
+        volumes=[
+            ("/host/data", "/mnt/data", "ro"),  # read-only input
+        ],
+        security=boxlite.SecurityOptions.maximum(),  # for untrusted AI code
+    ))
+    
+    try:
+        execution = await box.exec("python", ["-c", "print('secure exec')"])
+        result = await execution.wait()
+        print(result.stdout)
```
DorianZheng: `result.stdout` raises AttributeError. the low-level `Execution.wait()` returns only `exit_code` and `error_message` (src/exec.rs:113-119, 168-178); stdout has to be drained from `execution.stdout()`. same in patterns.md:173. for the simple case use `SimpleBox.exec(...)` which returns an `ExecResult` that does carry `.stdout`.

skills/boxlite-integration/references/python.md
```diff
@@ -0,0 +1,166 @@
...
+    
+    box = await runtime.create(boxlite.BoxOptions(
+        image="python:slim",
+        cpus=2,
+        memory_mib=1024,
+        working_dir="/workspace",
+        volumes=[
+            ("/host/data", "/mnt/data", "ro"),  # read-only input
+        ],
+        security=boxlite.SecurityOptions.maximum(),  # for untrusted AI code
```
DorianZheng: `BoxOptions(..., security=...)` raises `TypeError: 'security' is an invalid keyword argument` — the constructor has no `security` param (options.rs:416-434). security lives under the `advanced` field (`AdvancedBoxOptions(security=...)`, advanced_options.rs:274). note `AdvancedBoxOptions` isn't re-exported from the top-level `boxlite` package (only `SecurityOptions` is), so the public way to attach a security preset needs sorting out. same crash in the security-presets example at line 145.

skills/boxlite-integration/references/python.md
```diff
@@ -0,0 +1,166 @@
...
+
+async def main():
+    runtime = boxlite.Boxlite.default()
+    
+    box = await runtime.create(boxlite.BoxOptions(
+        image="python:slim",
+        cpus=2,
+        memory_mib=1024,
+        working_dir="/workspace",
+        volumes=[
```
DorianZheng, line 83: the volume tuple's third element is a bool `read_only`, not `"ro"`. the parser does `read_only = t.get_item(2).extract::<bool>()` (options.rs:606); the string `"ro"` fails to extract → TypeError at create. use `("/host/data", "/mnt/data", True)`. same in the file-transfer table at patterns.md:121. (no sdk uses an `"ro"/"rw"` string — go has `WithVolumeReadOnly`, node uses `readOnly: bool`.)

skills/boxlite-integration/references/python.md
```diff
@@ -0,0 +1,166 @@
...
+```
+
+---
+
+## Timeout + Zombie Prevention
+
+`asyncio.wait_for()` cancels the Python coroutine but does **not** kill the process inside the VM. Always kill explicitly:
+
+```python
+async def exec_with_timeout(box, cmd, args=None, timeout=30):
```
DorianZheng, line 132: altitude: this hand-rolls `asyncio.wait_for()+kill()` around a bare exec, but `SimpleBox.exec` already takes `timeout=` (simplebox.py:181) which flows to `timeout_secs` (box_handle.rs:74) and is enforced inside the guest by a watcher that SIGKILLs the process regardless of host liveness. the wrapper is a weaker client-side version of that — and it's inconsistent with c.md:141 / node.md:93 which do use the built-in. prefer `await box.exec("python", "-c", code, timeout=30)`.

skills/boxlite-integration/references/python.md
```diff
@@ -0,0 +1,166 @@
...
+    execution = await box.exec(cmd, args or [])
+    try:
+        return await asyncio.wait_for(execution.wait(), timeout=timeout)
+    except asyncio.TimeoutError:
```
DorianZheng: this copy catches `asyncio.TimeoutError` (and patterns.md:30/36 catch `TimeoutError`/`Exception`), which miss `asyncio.CancelledError` — a `BaseException` subclass since 3.8. when the task is cancelled (server shutdown, client disconnect) `execution.kill()` is skipped and the guest process leaks — the exact zombie this section says it prevents. the SKILL.md:82 copy uses `except BaseException` and is correct; the three copies have diverged.

skills/boxlite-integration/references/patterns.md
```diff
@@ -0,0 +1,192 @@
...
+
+```python
+results = await asyncio.gather(
+    safe_exec(box, "python", ["-c", "print('task A')"]),
+    safe_exec(box, "python", ["-c", "print('task B')"]),
+    safe_exec(box, "python", ["-c", "print('task C')"]),
+)
+```
+
+This is safe — the VM provides isolation from the host. Concurrent processes inside the VM share the VM's resources, so set appropriate CPU/memory limits.
```
DorianZheng: "this is safe — the vm provides isolation from the host" conflates host-isolation with isolation between the concurrent execs. these three `safe_exec` calls run in the same reused `box`, sharing its filesystem, pids, network and injected secrets — they are not isolated from each other. that contradicts the skill's own multi-tenant warning (SKILL.md:45, "shared box risks data leakage between users"). worth scoping to trusted/cooperative concurrency, with per-request boxes for untrusted multi-tenant code.

skills/boxlite-integration/references/node.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+---
+
+## REST API (cloud / production)
+
+```javascript
+import { JsBoxlite } from '@boxlite-ai/boxlite';
+
+const runtime = new JsBoxlite({
```
DorianZheng: this "cloud / production" example silently runs locally. `new JsBoxlite(options)` takes `JsOptions` = `{ homeDir?, imageRegistries? }` (native-contracts.ts:100-103, 317) — `url`/`apiKey` are ignored (a TS excess-property error, and at runtime a local default runtime). rest is the static factory: `JsBoxlite.rest(new BoxliteRestOptions({ url, credential: new ApiKeyCredential(process.env.BOXLITE_API_KEY) }))` (index.ts:63). as written, every box runs on the developer's machine, not the api.

skills/boxlite-integration/references/node.md
```diff
@@ -0,0 +1,147 @@
...
+const runtime = new JsBoxlite();
+const box = await runtime.create({
+  image: 'python:slim',
+  cpus: 2,
+  memoryMib: 1024,
+  workingDir: '/workspace',
+});
+
+try {
+  const execution = await box.exec('python', ['-c', "print('secure')"], { timeout: 30 });
```
DorianZheng: two issues on the low-level handle. `JsBox.exec(command, args?, env?, tty?, user?, timeoutSecs?, workingDir?)` (native-contracts.ts:265) — `{ timeout: 30 }` binds to the `env` positional (wrong type; the timeout is the 6th positional `timeoutSecs`, not a `{timeout}` bag). and `console.log(execution.stdout)` (line 94) logs the `stdout()` method reference, not output — the low-level `JsExecution` exposes `stdout()` as a stream and `wait()` returns `{exitCode, errorMessage}` with no stdout. drain `await execution.stdout()` + `execution.wait()`.

skills/boxlite-integration/references/node.md
```diff
@@ -0,0 +1,147 @@
...
+});
+```
+
+---
+
+## Timeout Handling
+
+```javascript
+async function execWithTimeout(box, cmd, args = [], timeoutMs = 30000) {
+  const execution = await box.exec(cmd, args);
```
DorianZheng: `box.exec(cmd, args)` on a `SimpleBox`/`CodeBox` returns a resolved `ExecResult` = `{ exitCode, stdout, stderr }` (exec.ts:18, simplebox.ts:489), which has no methods. so `execution.wait()` (line 123) and the `execution.kill()` in the setTimeout throw `TypeError: ... is not a function`. `.wait()/.kill()` exist only on the low-level `JsExecution` from `runtime.create()`→`box.exec()`. same shape in the SKILL.md:90 wrapper and patterns.md:49, which the prose says to use on every exec.

skills/boxlite-integration/references/node.md
```diff
@@ -0,0 +1,147 @@
...
+
+---
+
+## Error Handling
+
+```javascript
+import { ExecError, TimeoutError } from '@boxlite-ai/boxlite';
+
+try {
+  const result = await box.exec('python', ['-c', 'raise ValueError("bad")']);
```
DorianZheng: `SimpleBox.exec()` never throws on a non-zero exit — it resolves to `{ exitCode, stdout, stderr }` (simplebox.ts:489-631). only `CodeBox.run()` throws `ExecError` (codebox.ts:94). so `await box.exec('python', ['-c', 'raise ValueError("bad")'])` resolves normally with `exitCode: 1`, the `catch (err instanceof ExecError)` block is dead, and the failure is swallowed — the opposite of what the example teaches. check `result.exitCode`, or use `CodeBox.run()` if you want the throw.

skills/boxlite-integration/SKILL.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Quick Reference: Minimal Working Example
+
+**Python (simplest path):**
+```python
+import boxlite
+
+async with boxlite.CodeBox() as box:
+    result = await box.run("print('hello from sandbox')")
+    print(result.stdout)
```
DorianZheng: `CodeBox.run()` returns the stdout string itself (`-> str`, codebox.py:49 / `Promise<string>`, codebox.ts:92), so `result.stdout` fails: AttributeError in python, `undefined` in the node quick-ref just below (line 153). use `print(result)` / `console.log(result)`. same at python.md:39 (and the comment at python.md:40 that `result.stderr`/`result.exit_code` are "also available" is false) and node.md:31/44/58/71.

mateng0915: @DorianZheng  The issues you flagged have been addressed,  Could you take another look when you have a chance, Thanks

G4614:
Verified all code examples against current `main` SDK source (pyo3/napi bindings + C header). Found 4 reproducible bugs, 3 confined to Python:

**1. `BoxOptions(security=...)` — wrong kwarg name** (`python.md`, 2 occurrences)
The pyo3 constructor takes `advanced=`, not `security=` (`sdks/python/src/options.rs`). `BoxOptions(image=..., security=security)` raises `TypeError` at call time — right in the "untrusted AI code" security section.

**2. Low-level `execution.wait()` has no `.stdout`** (`python.md` "Full Control", `patterns.md` "Integrating into an LLM Tool Call")
`PyExecution.wait()` returns `PyExecResult{exit_code, error_message}` only (`sdks/python/src/exec.rs`). `result.stdout` → `AttributeError`. This is the skill's flagship before/after example for the most common integration pattern. Note the Node.js equivalent in `node.md` gets this right (reads the stdout stream separately) — Python is the only SDK with this gap.

**3. C `on_stdout` callback signature mismatch** (`c.md` Native API)
`CBoxStdoutCb` is `(const uint8_t* data, size_t len, void* user_data)` — 3 args (`boxlite.h:184`). Example declares `(const char* data, void* user_data)` — 2 args, and does `printf("%s", data)` assuming NUL-termination, which isn't guaranteed for process output. Type-mismatched function pointer + potential buffer over-read.

**4. Python `CodeBox.run()` doesn't raise `ExecError` on failure** (`python.md`)
Comment claims it does; `codebox.py`'s `run()` has no error-raising path — it just returns `result.stdout` regardless of exit code. Node's `CodeBox.run()` *does* raise `ExecError` (`codebox.ts:95`), so this looks like the Node behavior got copied into the Python doc without checking.

Minor: `go.md` pins `@v0.9.0` vs current `v0.9.7` on main (verified the pinned APIs still exist, so not breaking — just stale).

mateng0915:
> Verified all code examples against current `main` SDK source (pyo3/napi bindings + C header). Found 4 reproducible bugs, 3 confined to Python:
> 
> **1. `BoxOptions(security=...)` — wrong kwarg name** (`python.md`, 2 occurrences) The pyo3 constructor takes `advanced=`, not `security=` (`sdks/python/src/options.rs`). `BoxOptions(image=..., security=security)` raises `TypeError` at call time — right in the "untrusted AI code" security section.
> 
> **2. Low-level `execution.wait()` has no `.stdout`** (`python.md` "Full Control", `patterns.md` "Integrating into an LLM Tool Call") `PyExecution.wait()` returns `PyExecResult{exit_code, error_message}` only (`sdks/python/src/exec.rs`). `result.stdout` → `AttributeError`. This is the skill's flagship before/after example for the most common integration pattern. Note the Node.js equivalent in `node.md` gets this right (reads the stdout stream separately) — Python is the only SDK with this gap.
> 
> **3. C `on_stdout` callback signature mismatch** (`c.md` Native API) `CBoxStdoutCb` is `(const uint8_t* data, size_t len, void* user_data)` — 3 args (`boxlite.h:184`). Example declares `(const char* data, void* user_data)` — 2 args, and does `printf("%s", data)` assuming NUL-termination, which isn't guaranteed for process output. Type-mismatched function pointer + potential buffer over-read.
> 
> **4. Python `CodeBox.run()` doesn't raise `ExecError` on failure** (`python.md`) Comment claims it does; `codebox.py`'s `run()` has no error-raising path — it just returns `result.stdout` regardless of exit code. Node's `CodeBox.run()` _does_ raise `ExecError` (`codebox.ts:95`), so this looks like the Node behavior got copied into the Python doc without checking.
> 
> Minor: `go.md` pins `@v0.9.0` vs current `v0.9.7` on main (verified the pinned APIs still exist, so not breaking — just stale).

@G4614 The issues you flagged have been addressed, Could you take another look when you have a chance, Thanks

Issue #927, opened 2026-07-06
Does not really work on native musl host


xhebox:
https://github.com/boxlite-ai/boxlite/blob/51e9d50fcccfce601c42313ca149c3bcac5b1082/.cargo/config.toml#L1-L5

Default `crt-static` makes every `boxlite-***` fails to compile on `cargo build`. It produced errors as in `pyproject.toml`.

https://github.com/boxlite-ai/boxlite/blob/51e9d50fcccfce601c42313ca149c3bcac5b1082/sdks/python/pyproject.toml#L70-L75

It can still be fixed by `cargo build --target x86_64-unknown-linux-musl` as said in `build-shim.sh`. However all popular musl distro had chose to `-crt-static` by default, except the official one: static linking with shared musl is not big deal thanks to the solid ABI, so convenience wins.

Maybe just add these flags on need for CI/CD.

https://github.com/boxlite-ai/boxlite/blob/51e9d50fcccfce601c42313ca149c3bcac5b1082/scripts/build/build-shim.sh#L86-L99

CGO of `gvproxy-bridge` will run into another problem https://github.com/golang/go/issues/13492. Because TLS model is hardcoded into the go runtime, there is no way to workaround it other than a patched GO compiler, which is exactly what openharmony do for their fork.

We'd better just:

1. note musl is unsupported and only used for cross compilation or so somewhere
2. using a patched Go compiler to statically link to musl, if binary size or security matters

DorianZheng:
Could you elaborate a little bit on how to reproduce this issue?

xhebox:
> Could you elaborate a little bit on how to reproduce this issue?

Just run `cargo build` at the root directory using alpine linux or void linux. It is essentially same bug as notes in `pyproject.toml`. It is more of a dev experience problem using musl-based distro. Statically linked binaries should work fine.

```
[xhe@tb14p boxlite]$ cargo build
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /home/xhe/project/boxlite/sdks/node/Cargo.toml
workspace: /home/xhe/project/boxlite/Cargo.toml
error: cannot produce proc-macro for `async-recursion v1.1.1` as the target `x86_64-unknown-linux-musl` does not support these crate types
[xhe@tb14p boxlite]$ cargo build --package boxlite-guest
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /home/xhe/project/boxlite/sdks/node/Cargo.toml
[...]

PR #937, opened 2026-07-07
fix(musl): run static shim without libkrunfw dlopen

xhebox:
## Summary

Fixes the static-musl shim build by linking vendored `libkrun` as a normal Cargo dependency instead of building a nested Rust `staticlib`. This removes duplicate Rust standard-library symbols and the need for `--allow-multiple-definition` while keeping the Linux shim statically linked.

Aligns libkrunfw packaging with the final link mode. Dynamic builds package the shared library for libkrun's `dlopen` path; static-musl builds package the raw `libkrunfw.bin` kernel and configure it through `krun_set_kernel`.

Improves native musl support across the shim, guest, runtime assembly, and local SDK development paths.

Ref https://github.com/boxlite-ai/boxlite/issues/927.

## Changes

- Replace the nested `cargo rustc --crate-type staticlib` libkrun build with a Cargo-managed dependency on the vendored Rust crate.
- Remove duplicate-symbol linker workarounds and obsolete libkrun logging initialization. The vendored crate now shares the shim's Rust runtime and process-wide logger.
- Split libkrunfw artifact behavior by link mode:
  - dynamic Linux and macOS builds package `libkrunfw.so*` or `libkrunfw.dylib`;
  - static-musl builds always build and package `libkrunfw.bin`, including when only the base `krunfw` feature is enabled;
  - `krunfw-source` builds firmware from the vendored source, with shared libraries omitted for static-musl.
- Add runtime external-kernel configuration with `BOXLITE_KRUNFW_EXTERNAL_KERNEL`, `BOXLITE_KRUNFW_KERNEL_PATH`, and `BOXLITE_KRUNFW_KERNEL_FORMAT`.
  - External-kernel loading defaults to off for dynamic builds and on for static-musl.
  - An explicit gate value overrides the platform default.
  - Kernel format defaults to ELF on x86_64 and RAW on other architectures; explicit formats are `raw`, `elf`, `pe-gz`, `image-bz2`, `image-gz`, and `image-zstd`.
- Bundle and copy the selected libkrunfw artifact alongside the shim, and pass the external-kernel controls through the sanitized shim environment.
- Scope `crt-static` flags to the shim and guest Cargo configurations so they do not leak into SDK cdylibs or host-built proc macros.
- Use the native musl Rust target when available, while retaining cross-compilation support for GNU hosts.
- Centralize build-script path, target-directory, and Cargo-profile handling. Build profiles now flow through `BUILD_PROFILE`; `make runtime` and `make cli` default to release, with `BUILD_PROFILE=debug` selecting debug builds.
- Fix related musl portability issues: resource-limit types, GNU-only `libresolv` search paths, libseccomp cross-build handling, target/profile runtime discovery, and Go/cgo unwind linkage.
- Temporarily skip publishing `libkrun-sys` from the release workflow because its vendored path dependency requires a follow-up packaging fix; #962 will address this.
## How to verify

```bash
bash -n scripts/build/build-cli.sh
bash -n scripts/build/build-guest.sh
bash -n scripts/build/build-libseccomp.sh
bash -n scripts/build/build-runtime.sh
bash -n scripts/build/build-shim.sh
make fmt:check:rust
git diff --check

BOXLITE_DEPS_STUB=1 cargo check -p boxlite --tests --features krun
BOXLITE_DEPS_STUB=1 cargo check -p libkrun-sys --features krun,krunfw
BOXLITE_DEPS_STUB=1 cargo test -p boxlite --lib vmm::krun::kernel::tests --features krun
```

For an end-to-end static-musl build after `make setup:build`:

```bash
make runtime
BUILD_PROFILE=debug make runtime
```

The assembled static-musl runtime should contain `libkrunfw.bin` and no `libkrunfw.so*`. Dynamic Linux and macOS runtimes should continue to package the shared libkrunfw artifact.

`libkrun-sys` is intentionally omitted from the crates.io publish step for this PR; the packaging issue is tracked in #962.


xhebox:
Some changes are still needed. It is fixing lots of build problems without compromise.

src/boxlite/src/vmm/controller/spawn.rs
```diff
@@ -148,6 +148,15 @@ impl<'a> ShimSpawner<'a> {
         if let Ok(rust_backtrace) = std::env::var("RUST_BACKTRACE") {
             cmd.env("RUST_BACKTRACE", rust_backtrace);
         }
+        for name in [
```
DorianZheng: duplicate with bwrap.rs, need a class to manage all the env vars

src/boxlite/src/vmm/krun/engine.rs
```diff
@@ -166,6 +172,146 @@ impl Krun {
         guest_args
     }
 
+    fn krunfw_kernel_env() -> BoxliteResult<Option<bool>> {
```
DorianZheng: extract 175-314 as a class

DorianZheng:
`scripts/common.sh` is missing a trailing newline on its last line (the new `boxlite_cargo_feature_args` helper). no functional impact, but `end-of-file-fixer` and most editors will re-add it on the next edit and create a noisy diff — worth appending a `\n`.

src/boxlite/Cargo.toml
```diff
@@ -21,11 +21,14 @@ crate-type = ["rlib"]
...
+libslirp = []                                               # Uses external libslirp-helper binary
+gvproxy = ["dep:libgvproxy-sys"]                             # Uses libgvproxy CGO shared library
 e2fsprogs = ["dep:e2fsprogs-sys"]                            # Bundled mke2fs for ext4 image creation
 bubblewrap = ["dep:bubblewrap-sys"]                          # Bundled bwrap for sandbox isolation (Linux)
-krunfw = ["dep:libkrun-sys", "libkrun-sys/krunfw"]           # Download libkrunfw firmware for runtime bundling
-krun = ["krunfw", "libkrun-sys/krun"]                        # Build + statically link libkrun.a (only for boxlite-shim)
+krunfw = ["dep:libkrun-sys", "libkrun-sys/krunfw"]           # Bundle libkrunfw shared library for libkrun's fallback path
+krun = ["krunfw", "libkrun-sys/krun"]                        # Link vendored libkrun as a Rust dependency (only for boxlite-shim)
+krunfw-kernel = ["krunfw", "libkrun-sys/krunfw-kernel"]      # Build/source libkrunfw kernel artifact for runtime env-enabled krun_set_kernel
+krunfw-kernel-llvm = ["krunfw-kernel", "libkrun-sys/krunfw-kernel-llvm"] # Use LLVM=1 for libkrunfw kernel builds
```
DorianZheng: can we make this conclude automatically instead of adding new flag?

xhebox: Need an agreement on how we treat it. I mean, we could always build libkrunfw.so and libkrunfw.bin, and do llvm detection in build.rs.

But feature flag other than `llvm` may not be eliminated. IIRC, shim/guest all depends on boxlite crate, but only shim needs libkrun.

src/shim/Cargo.toml
```diff
@@ -12,7 +12,7 @@ name = "boxlite-shim"
 path = "src/main.rs"
 
 [dependencies]
-boxlite = { path = "../boxlite", default-features = false, features = ["gvproxy", "krun"] }
+boxlite = { path = "../boxlite", default-features = false, features = ["gvproxy", "krun", "krunfw"] }
```
DorianZheng: why shim need feature `krunfw`

scripts/build/build-cli.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
DorianZheng, line 1: the script style should be consistent with other scripts

src/shim/Cargo.toml
xhebox: I believe that it is a miss modification of agent.

scripts/build/build-cli.sh
xhebox: Modified

DorianZheng (dismissed):
LGTM

xhebox:
> LGTM

@DorianZheng should be able to run lint and test now

PR #951, opened 2026-07-09
feat(serve): accept volumes in REST create box

### LINKED ISSUE #942: boxlite serve REST API does not accept CreateBoxRequest.volumes
## Summary

`boxlite serve` appears to document `CreateBoxRequest.volumes` in the OpenAPI spec, but the Rust REST server implementation does not currently accept or map this field.

This makes it impossible to create a box with host-to-guest mounts through the native REST API, even though BoxLite core/CLI supports volumes via `BoxOptions.volumes`.

## What I observed

In the OpenAPI spec, `CreateBoxRequest` includes `volumes`:

- `openapi/box.openapi.yaml` -> `CreateBoxRequest.volumes`
- `VolumeSpec` has `host_path`, `guest_path`, and `read_only`

But in the Rust serve implementation:

- `src/cli/src/commands/serve/types.rs` -> `CreateBoxRequest` does not include `volumes`
- it also uses `#[serde(deny_unknown_fields)]`, so sending `volumes` to `POST /v1/boxes` is rejected as an unknown field
- `src/cli/src/commands/serve/mod.rs` -> `build_box_options()` does not map volumes into `BoxOptions.volumes`

I checked this against `v0.9.7` and current `main`.

## Why this matters

We are evaluating `boxlite serve` as a production sandbox node for an agent runtime. The desired architecture is:

- run BoxLite as a separate long-lived sandbox daemon/node
- have the agent service call BoxLite over HTTP/WS instead of embedding the sync runtime in-process
- use explicit execution IDs, attach/status/kill, request timeouts, and process isolation

For agent sandboxes we need to mount:

- a writable per-session workspace at `/workspace`
- a read-only shared workspace at `/shared`

Today this works through lower-level BoxLite volume support, but not through `boxlite serve` REST.

## Expected behavior

Either:

1. `boxlite serve` REST should support `CreateBoxRequest.volumes` and map it into `BoxOptions.volumes`, consistent with the OpenAPI spec; or
2. the OpenAPI spec should clarify that REST volume mounts are not currently supported.

## Production/security question

What is the recommended production model for exposing volumes through `boxlite serve`?

Allowing arbitrary REST clients to pass arbitrary host paths seems unsafe in multi-tenant production. A safer production design might be one of:

- `boxlite serve` supports volumes only for trusted/internal clients, with API key + private network isolation
- server-side host path allowlists, e.g. only paths under configured workspace roots can be mounted
- named/server-defined mount profiles, where clients request logical mounts instead of raw host paths
- read-only/read-write policy validation on the server side

For an agent sandbox platform, we would prefer a model where the control plane can request known workspace mounts while the BoxLite node enforces host path boundaries.

Is REST volume support planned, and what production-safe pattern would you recommend?

sycamore792:
## Summary

- accept `volumes` on `boxlite serve` `CreateBoxRequest`, matching the OpenAPI schema
- map REST volume specs into `BoxOptions.volumes` while preserving `read_only: false` as the default
- add a focused unit test covering REST volume mapping

Fixes #942

## Tests

- `cargo fmt --all -- --check`
- `cargo check -p boxlite-cli`
- `cargo test -p boxlite-cli build_box_options_maps_rest_volumes`

## Summary by CodeRabbit

* **New Features**
  * Box creation requests now support mounting host directories into the guest environment.
  * Volume entries can include a source path, destination path, and optional read-only setting.

* **Bug Fixes**
  * Missing read-only settings now default to writable access, preserving expected behavior.
  * Volume mappings are now carried through correctly when creating a box.

### DIFF
--- a/src/cli/src/commands/serve/mod.rs
+++ b/src/cli/src/commands/serve/mod.rs
@@ -23,7 +23,7 @@ use tokio::sync::RwLock;
 use tower_http::catch_panic::CatchPanicLayer;
 use tower_http::request_id::{MakeRequestUuid, PropagateRequestIdLayer, SetRequestIdLayer};
 
-use boxlite::runtime::options::{NetworkConfig, NetworkMode};
+use boxlite::runtime::options::{NetworkConfig, NetworkMode, VolumeSpec as RuntimeVolumeSpec};
 use boxlite::{
     BoxCommand, BoxInfo, BoxOptions, BoxliteRuntime, ExecStdin, Execution, LiteBox, NetworkSpec,
     RootfsSpec,
@@ -655,6 +655,15 @@ fn build_box_options(req: &CreateBoxRequest) -> Result<BoxOptions, boxlite::Boxl
         })?,
         None => NetworkSpec::default(),
     };
+    let volumes = req
+        .volumes
+        .iter()
+        .map(|volume| RuntimeVolumeSpec {
+            host_path: volume.host_path.clone(),
+            guest_path: volume.guest_path.clone(),
+            read_only: volume.read_only,
+        })
+        .collect();
 
     // SecurityOptions is deliberately NOT client-configurable over
     // REST: sandbox security is the operator's policy. The server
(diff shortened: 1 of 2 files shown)


the-mentor:
is there an update on this ?

Issue #848, opened 2026-06-25
Hardening: split libkrun/VMM processes for finer-grained seccomp after xattr allowlist


G4614:
## Context

PR #846 fixes writable virtiofs volumes by allowing the Linux xattr syscall family in the VMM seccomp profile. That is a practical compatibility fix: host-side virtiofs handling can call `fgetxattr`/related syscalls while preserving file metadata such as `security.capability`, and blocking those syscalls SIGSYS-kills the VMM during otherwise normal writes.

However, the current seccomp granularity is VMM-process-wide. The `vmm` filter is applied with TSYNC to all threads in the shim/VMM process, so the xattr allowlist is available not only to the virtiofs path that needs it, but also to other threads sharing the same process boundary, such as libkrun runtime threads and embedded networking/helper threads.

## Problem

This is acceptable as a short-term unblocker, but it is not the ideal long-term security shape. The current architecture forces one broad VMM seccomp profile to cover multiple responsibilities:

- libkrun / core VMM execution
- virtiofs host-side file serving
- networking/helper runtime threads
- other shim-managed runtime work

Because those components share one process-wide filter, any syscall required by one component expands the allowed syscall surface for all of them. PR #846 makes this visible with the xattr family (`getxattr`, `setxattr`, `listxattr`, `removexattr`, including `l*` and `f*` variants).

## Desired direction

Refactor the libkrun/VMM runtime boundary so security policy can be applied at a finer granularity, ideally per process or per narrowly-scoped component. For example:

- Keep the core VMM/libkrun process on the smallest syscall profile it actually needs.
- Run virtiofs/file-serving work behind a separate process boundary with an explicit xattr-capable seccomp profile.
- Keep networking/helper paths on their own profile where possible.
- Avoid one shared process-wide allowlist accumulating every syscall needed by any subsystem.

## Acceptance criteria

- Document the current VMM-process-wide seccomp limitation and why #846 had to allow xattr syscalls there.
- Design a process/component split that supports narrower seccomp profiles.
- Introduce separate seccomp profiles for at least the core VMM and virtiofs/file-serving component.
- Ensure writable virtiofs volumes still work without granting xattr syscalls to unrelated runtime components.
- Add regression coverage showing writable virtiofs metadata operations continue to pass under the narrowed profile.

## Related

- PR #846: allows xattr syscall family in the current VMM seccomp profile to prevent virtiofs writes from SIGSYS-killing the VMM.

PR #962, opened 2026-07-09
refactor(deps): split libkrunfw sidecar package

### LINKED ISSUE #927: Does not really work on native musl host

xhebox:
## Summary

Separates Rust dependencies from runtime assets.

`libkrun` is now a normal Cargo dependency, while e2fsprogs, bubblewrap, and libkrunfw are built explicitly before runtime assembly. This removes the `*-sys` crates that existed only to trigger native builds and substantially simplifies `boxlite/build.rs`.

This also removes all executable rpath injection from the runtime and C consumer paths. The shim resolves its copied sidecar libraries through a sandbox-scoped library search path, while C quickstarts and examples link the prepared static `libboxlite.a` directly. Dynamic C linking remains documented as an explicit alternative.

Ref #848, close #927.

## Changes

- Replace `libkrun-sys` with a direct dependency on `libkrun`.
- Remove `e2fsprogs-sys` and `bubblewrap-sys`.
- Move vendored native sources under `vendor/`.
- Add explicit builders for e2fsprogs, bubblewrap, and libkrunfw.
- Assemble runtime assets through `make runtime-assets`.
- Keep source builds on the assembled local runtime; packaged crates continue downloading release runtime artifacts.
- Remove `BOXLITE_DEPS_STUB` and Cargo JSON/sed-based runtime discovery.
- Preserve runtime `BOXLITE_RUNTIME_DIR` and `BOXLITE_RUNTIME_URL` overrides.
- Verify libkrun's derived `-krun.sock` endpoint in the existing VM network test.
- Keep publishing `libgvproxy-sys` with its existing excluded-source behavior and document its final-link limitation.
- Remove the shim's macOS `@loader_path` and Linux `$ORIGIN` rpaths.
- Remove `-Wl,-rpath`, `BUILD_RPATH`, `INSTALL_RPATH`, and `install_name_tool -add_rpath` usage from C consumer paths.
- Set `DYLD_LIBRARY_PATH` and `LD_LIBRARY_PATH` to the copied shim directory inside the macOS Seatbelt and Linux bubblewrap launch paths, without inheriting host search paths.
- Prepare debug and release C static archives in the local `dev:c` and `dist:c` entrypoints, including the Go symbol fix and bundled unwinder when required; the C workflow now calls `make dist:c` instead of patching the archive separately.
- Link C quickstarts and examples directly against `libboxlite.a`; keep both static and explicit dynamic linking examples in the C reference documentation.

## Verification

- `git diff --check`
- `bash -n scripts/build/fix-go-symbols.sh`
- Verified that applying `fix-go-symbols.sh` twice is idempotent.
- Verified that `simple_api_demo.c` links using only the prepared `libboxlite.a`, without `-lunwind` or other added native dependency flags.
- Native macOS sandbox/dyld behavior remains to be exercised on a macOS runner.

## Summary by CodeRabbit

* **New Features**
  * Runtime is now assembled from pre-built runtime assets, including libkrunfw, e2fsprogs, and bwrap components.
  * Improved network allowlist behavior with DNS sinkholing, HTTPS inspection, and secret substitution (including TCP/MITM and WebSocket upgrade handling).
  * Updated networking identifiers and added public libkrun disk/kernel format constants.
* **Bug Fixes**
  * More reliable runtime asset verification/assembly during release packaging.
  * Sandbox execution now sanitizes environment variables more strictly.
* **Documentation**
  * C SDK and Quickstart guidance updated to prefer static linking.
* **Tests**
  * Expanded coverage for runtime asset packaging, networking/socket paths, DNS sinkholing, allowlist routing, and MITM/secret substitution.


G4614:
why should we change names for so many files

also can we split the prs for different functions

xhebox:
> why should we change names for so many files
> 
> also can we split the prs for different functions

Only gvproxy is renamed, since it is a real dependency. All other `-sys` crates were removed leaving an empty dir.  And I don't think that deep dir is good anymore.

Splitting can be done, but co-existing old/new path will be really troublesome for ai to get it right. It definitely will leave some **compatibility safeguard non-sense**, even if this is only *building*.

The whole point of the refactor is splitting build into scripts, which centralize those misc builds and speed up the building on local machine a little bit by removing them from `cargo xxxx` commands.

G4614:
> > why should we change names for so many files
> > also can we split the prs for different functions
> 
> Only gvproxy is renamed, since it is a real dependency. All other `-sys` crates were removed leaving an empty dir. And I don't think that deep dir is good anymore.
> 
> Splitting can be done, but co-existing old/new path will be really troublesome for ai to get it right. It definitely will leave some **compatibility safeguard non-sense**, even if this is only _building_.
> 
> The whole point of the refactor is splitting build into scripts, which centralize those misc builds and speed up the building on local machine a little bit by removing them from `cargo xxxx` commands.

makes sense, thx

PR #963, opened 2026-07-09
feat(logs): add opt-in host-side container log capture

### LINKED ISSUE #909: Opt-in host-side capture of container stdout/stderr (container logs)
**Title:** Opt-in host-side capture of container stdout/stderr (container logs)

## Feature

An option to have a box's container stdout/stderr captured to a file on the
host — regardless of how the process was started (image-default ENTRYPOINT or a
declared `entrypoint`) and whether the box is detached. Something like:

```typescript
const box = new SimpleBox({
  image: "nginx:alpine",
  captureLogs: true,   // → $BOXLITE_HOME/boxes/<id>/logs/entrypoint.log
  // or: logPath: "/var/log/my-workloads/box-1.log"
});
```

Similar to how `docker logs` works: the daemon always has the container's
output on disk, and any client (or none) can read it later.

## Why

Containers built to the Docker logging convention write their logs to
stdout/stderr — official images enforce it (`nginx` ships
`/var/log/nginx/access.log -> /dev/stdout`). Under boxlite that output is not
obtainable on the host today, so any image following the standard convention
runs silently. `exec()` has streaming support, but that only covers processes
started through `exec()`, and only while the launching client is alive — for
detached boxes, or after a client restart, there's no way to get output at all.

Our use case: boxlite as a workload runtime where a host-side agent (Grafana
Alloy) tails per-box log files and ships them to Loki. The boxes are often
detached and outlive the process that created them, so capture has to be
host-side and independent of any SDK client.

## Reproduction

The script below runs four cases — unmodified `nginx:alpine` (image
ENTRYPOINT), a declared entrypoint on an attached box, the same detached, and
an `exec()`'d process — generating output with a unique marker in each, then
scans the box directory (including `logs/console.log`) for it:

<details>
<summary>log-capture-gaps-test.ts</summary>

```typescript
import { existsSync, readFileSync, readdirSync, statSync } from "node:fs";
import { homedir } from "node:os";
import { join } from "node:path";

// Demonstrates which process output IS and IS NOT obtainable on the host with boxlite.
// Written to accompany a feature request for host-side log capture of container output.
//
// Cases:
//   1. Image-default ENTRYPOINT (nginx:alpine — Docker convention: logs go to stdout/stderr,
//      access.log/error.log are symlinks to /dev/stdout & /dev/stderr)
//   2. Declared entrypoint, attached box (direct-binary awk emitter, no shell involved)
//   3. Declared entrypoint, detached box (same emitter, detach: true)
//   4. exec()'d process on a running box (the one case with an output API)
//
// For each case the script generates output containing a unique marker, then looks for that
// marker everywhere it could plausibly land on the host: the box's logs/console.log and every
// readable file under the box dir. The final table shows the gap.

const RUN = Date.now().toString(36);
const BOXLITE_HOME = process.env.BOXLITE_HOME ?? join(homedir(), ".boxlite");
const NGINX_PORT = 18080;

const SKIP_EXT = [".qcow2", ".img", ".sock", ".db", ".pid"];
const MAX_SCAN_BYTES = 5 * 1024 * 1024;

function sleep(ms: number) {
  return new Promise((resolve) => setTimeout(resolve, ms));
}

function* walk(dir: string): Generator<string> {
  let entries;
  try {
    entries = readdirSync(dir, { withFileTypes: true });
  } catch {
    return;
  }
  for (const e of entries) {
    const p = join(dir, e.name);
    if (e.isDirectory()) yield* walk(p);
    else if (e.isFile()) yield p;
  }
}

/** Returns the files under the box dir containing the marker. */
function findInBoxDir(boxId: string, marker: string): string[] {
  const hits: string[] = [];
  for (const file of walk(join(BOXLITE_HOME, "boxes", boxId))) {
    if (SKIP_EXT.some((ext) => file.endsWith(ext))) continue;
    try {
      if (statSync(file).size > MAX_SCAN_BYTES) continue;
      if (readFileSync(file, "utf-8").includes(marker)) hits.push(file);
    } catch {
      // unreadable — not a log a host agent could tail anyway
    }
  }
  return hits;
}

function consoleLogInfo(boxId: string): string {
  const p = join(BOXLITE_HOME, "boxes", boxId, "logs", "console.log");
  return existsSync(p) ? `exists (${statSync(p).size}B)` : "absent";
}

interface CaseResult {
  name: string;
  outputGenerated: boolean;
  foundAt: string[];
}

const results: CaseResult[] = [];

// One line per second to both streams, exec'd directly (no shell required in the chain).
function awkEmitter(marker: string): string[] {
  return [
    "/usr/bin/awk",
    `BEGIN { for (i = 0; i < 60; i++) { ` +
      `print "${marker} stdout " i; fflush(); ` +
      `print "${marker} stderr " i > "/dev/stderr"; fflush(); ` +
      `system("sleep 1") } }`,
  ];
}

async function case1_imageDefaultEntrypoint() {
  console.log("\n─── Case 1: image-default ENTRYPOINT (nginx:alpine) ───");
  const box = new SimpleBox({
    image: "nginx:alpine",
    name: `gaps-nginx-${RUN}`,
    ports: [{ hostPort: NGINX_PORT, guestPort: 80, protocol: "tcp" }],
    autoRemove: false,
  });
  try {
    await box.exec("true"); // boots the VM; nginx (image ENTRYPOINT) starts as container init
    await sleep(2000);

    // Each request makes nginx write an access-log line — to stdout, per the image's
    // /var/log/nginx/access.log -> /dev/stdout symlink.
    const marker = `/marker-${RUN}`;
    let ok = false;
    for (let i = 0; i < 5 && !ok; i++) {
      try {
        const res = await fetch(`http://localhost:${NGINX_PORT}${marker}`);
        ok = res.status > 0;
      } catch {
        await sleep(1000);
      }
    }
    console.log(`request reached nginx : ${ok ? "YES (access-log line was generated)" : "NO — case inconclusive"}`);
    await sleep(2000);

    const found = findInBoxDir(box.id, marker);
    console.log(`console.log           : ${consoleLogInfo(box.id)}`);
    console.log(`access-log line found : ${found.length ? found.join(", ") : "NOWHERE on host"}`);
    results.push({ name: "1. image-default ENTRYPOINT (nginx)", outputGenerated: ok, foundAt: found });
  } finally {
    await box.stop().catch(() => {});
  }
}

async function entrypointCase(name: string, detach: boolean) {
  console.log(`\n─── ${name} ───`);
  const marker = `MARKER_${RUN}_${detach ? "D" : "A"}`;
  const box = new SimpleBox({
    image: "alpine:latest",
    name: `gaps-entrypoint-${detach ? "detached" : "attached"}-${RUN}`,
    entrypoint: awkEmitter(marker),
    autoRemove: false,
    detach,
  });
  try {
    await box.exec("true");
    await sleep(5000); // emitter has produced ~5 lines on each stream by now

    const found = findInBoxDir(box.id, marker);
    console.log(`console.log           : ${consoleLogInfo(box.id)}`);
    console.log(`entrypoint output found: ${found.length ? found.join(", ") : "NOWHERE on host"}`);
    results.push({ name, outputGenerated: true, foundAt: found });
  } finally {
    await box.stop().catch(() => {});
  }
}

async function case4_execStreaming() {
  console.log("\n─── Case 4: exec()'d process (the contrast) ───");
  const marker = `MARKER_${RUN}_EXEC`;
  const box = new SimpleBox({
    image: "alpine:latest",
    name: `gaps-exec-${RUN}`,
    autoRemove: false,
  });
  try {
    const result = await box.exec("echo", marker);
    const obtained = result.stdout.includes(marker);
    console.log(`exec stdout obtained  : ${obtained ? "YES — exec() returns/streams output" : "NO"}`);
    results.push({
      name: "4. exec()'d process",
      outputGenerated: true,
      foundAt: obtained ? ["<returned by exec() API>"] : [],
    });
  } finally {
    await box.stop().catch(() => {});
  }
}

async function main() {
  const version = (await import("@boxlite-ai/boxlite/package.json")).default.version;
  console.log(`boxlite ${version}  ${process.platform}/${process.arch}  BOXLITE_HOME=${BOXLITE_HOME}`);

  await case1_imageDefaultEntrypoint();
  await entrypointCase("Case 2: declared entrypoint, attached box", false);
  await entrypointCase("Case 3: declared entrypoint, detached box", true);
  await case4_execStreaming();

  console.log("\n================== SUMMARY: is process output obtainable on the host? ==================");
  for (const r of results) {
    const status = !r.outputGenerated ? "INCONCLUSIVE" : r.foundAt.length ? `YES — ${r.foundAt.join(", ")}` : "NO ❌";
    console.log(`  ${r.name.padEnd(45)} ${status}`);
  }
  console.log("=========================================================================================");
  console.log("Cases 1–3 are the gap: container stdout/stderr is not captured anywhere host-visible,");
  console.log("regardless of how the entrypoint is defined or whether the box is detached.");
}

main();

```
</details>

```
─── Case 1: image-default ENTRYPOINT (nginx:alpine) ───
request reached nginx : YES (access-log line was generated)
console.log           : exists (7477B)
access-log line found : NOWHERE on host

─── Case 2: declared entrypoint, attached box ───
console.log           : exists (7477B)
entrypoint output found: NOWHERE on host

─── Case 3: declared entrypoint, detached box ───
console.log           : exists (7477B)
entrypoint output found: NOWHERE on host

─── Case 4: exec()'d process (the contrast) ───
exec stdout obtained  : YES — exec() returns/streams output

================== SUMMARY: is process output obtainable on the host? ==================
  1. image-default ENTRYPOINT (nginx)           NO ❌
  Case 2: declared entrypoint, attached box     NO ❌
  Case 3: declared entrypoint, detached box     NO ❌
  4. exec()'d process                           YES — <returned by exec() API>
=========================================================================================
Cases 1–3 are the gap: container stdout/stderr is not captured anywhere host-visible,
regardless of how the entrypoint is defined or whether the box is detached.
```

Container output is host-obtainable in none of cases 1–3; only the `exec()`
case has an API.

Happy to test pre-releases — and thanks for boxlite!

xhebox:
## Summary

Add opt-in host-side persistence for container init and entrypoint stdout/stderr through `captureLogs` / `capture_logs`. When enabled, output is written to `{box}/logs/{container_id}.log`, including for detached boxes, without changing existing `exec()` streaming.

Captured logs are bounded with size-based rotation: the active file rotates at 10 MiB and four numbered archives are retained (`.log.1` through `.log.4`).

Closes #909.

## Changes

- Expose the default-off capture option across the runtime, SDKs, REST/OpenAPI, cloud API, and runner.
- Prepare the box-managed logs directory on the host and mount it into the guest as a dedicated writable share.
- Start capture before container init so short-lived entrypoint output is preserved and setup failures cannot leave orphaned libcontainer state.
- Drain stdout and stderr with one Tokio task per container instead of dedicated OS threads.
- Rotate logs in the guest writer using rename/reopen when the active file would exceed 10 MiB, retaining five files total.
- Initialize size accounting from existing file metadata so capture resumes correctly after recovery.
- On normal shutdown, wait up to one second for pipe EOF, remaining output, and the final flush; abort only if that bounded drain times out.
- Bypass warm-pool reuse when capture is requested because an existing VM cannot retrofit capture before init starts.
- Parse Cargo build-script JSON with `jq` during runtime assembly, avoiding GNU/BSD `sed` differences on Darwin.

## Log layout

```text
{box}/logs/{container_id}.log    # active, up to 10 MiB
{box}/logs/{container_id}.log.1  # newest archive
{box}/logs/{container_id}.log.2
{box}/logs/{container_id}.log.3
{box}/logs/{container_id}.log.4  # oldest archive
```

Rotation is performed by the guest-side writer that owns the file descriptor. This avoids `copytruncate` races and lets inode-aware host tailers follow the active path after rename/reopen.

## How to verify

- `cargo check -p boxlite-guest`
- `cargo test -p boxlite-guest container::stdio`
- `cargo test -p boxlite-guest`
- `cargo clippy -p boxlite-guest --all-targets -- -D warnings`
- `git diff --check`
- Cargo build-script JSON fixture extraction with a Darwin-style `package_id`

The stdio tests cover capture and drain behavior, rotation and oldest-file pruning, resuming size accounting from an existing log, and continuing to drain when rotation fails.

## Summary by CodeRabbit

* **New Features**
  * Added an optional `captureLogs` (`capture_logs`) setting to capture container init stdout/stderr to host-visible log files when creating or recovering boxes.
  * Enabled end-to-end log-capture support across BoxLite APIs and major SDKs (C, Go, Node.js, Python), including bypassing warm-pool reuse when enabled.
  * Captured logs are written through the init pipeline with size-based rotation and retention.
* **Database Updates**
  * Persisted the preference per box (default: off) via a schema migration.

### DIFF
--- a/src/boxlite/src/litebox/init/tasks/vmm_spawn.rs
+++ b/src/boxlite/src/litebox/init/tasks/vmm_spawn.rs
@@ -11,7 +11,7 @@ use crate::litebox::init::types::resolve_user_volumes;
 use crate::net::{NetworkBackend, NetworkBackendConfig};
 use crate::pipeline::PipelineTask;
 use crate::rootfs::guest::{GuestRootfs, Strategy};
-use crate::runtime::constants::{guest_paths, mount_tags};
+use crate::runtime::constants::{guest_paths, logs, mount_tags};
 use crate::runtime::id::BoxID;
 use crate::runtime::layout::BoxFilesystemLayout;
 use crate::runtime::options::BoxOptions;
@@ -27,7 +27,7 @@ use async_trait::async_trait;
 use boxlite_shared::BoxTransport;
 use boxlite_shared::errors::{BoxliteError, BoxliteResult};
 use std::collections::{HashMap, HashSet};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 pub struct VmmSpawnTask;
 
@@ -77,20 +77,26 @@ impl PipelineTask<InitCtx> for VmmSpawnTask {
         };
 
         // Build config and get outputs
-        let (instance_spec, volume_mgr, rootfs_init, container_mounts, network_backend) =
-            build_config(
-                &box_id,
-                &options,
-                &layout,
-                &container_image_config,
-                &container_disk_path,
-                guest_disk_path.as_deref(),
-                &container_id,
-                &runtime,
-                reuse_rootfs,
-            )
-            .await
-            .inspect_err(|e| log_task_error(&box_id, task_name, e))?;
+        let (
+            instance_spec,
+            volume_mgr,
+            rootfs_init,
+            container_mounts,
+            network_backend,
+            log_capture_path,
+        ) = build_config(
+            &box_id,
+            &options,
+            &layout,
+            &container_image_config,
+            &container_disk_path,
+            guest_disk_path.as_deref(),
+            &container_id,
+            &runtime,
+            reuse_rootfs,
+        )
+        .await
+        .inspect_err(|e| log_task_error(&box_id, task_name, e))?;
 
         // Spawn VM
         let handler = spawn_vm(&box_id, &instance_spec, &options, &layout)
@@ -101,6 +107,7 @@ impl PipelineTask<InitCtx> for VmmSpawnTask {
         ctx.guard.set_handler(handler);
         ctx.volume_mgr = Some(volume_mgr);
         ctx.rootfs_init = Some(rootfs_init);
+        ctx.log_capture_path = log_capture_path;
         ctx.container_mounts = Some(container_mounts);
         // Hand the box's one network backend to LiveState assembly (init/mod.rs).
         ctx.network_backend = network_backend;
@@ -135,6 +142,7 @@ async fn build_config(
     crate::portal::interfaces::ContainerRootfsInitConfig,
     Vec<ContainerMount>,
     Option<Box<dyn NetworkBackend>>,
+    Option<String>,
 )> {
     // BoxTransport setup
     let transport = BoxTransport::unix(layout.socket_path());
@@ -151,6 +159,7 @@ async fn build_config(
 
     // SHARED virtiofs - needed by all strategies
     volume_mgr.add_fs_share(mount_tags::SHARED, layout.shared_dir(), None, false, None);
+    let log_capture_path = configure_log_capture(options, layout, container_id, &mut volume_mgr)?;
 
     // Add container rootfs disk (COW overlay workflow):
     // 1. Base disk: Pre-built ext4 image with container layers merged
@@ -254,13 +263,60 @@ async fn build_config(
         exit_file: layout.exit_file_path(),
         detach: options.detach,
     };
-
     Ok((
         instance_spec,
         volume_mgr,
         rootfs_init,
         container_mounts,
         network_backend,
+        log_capture_path,
+    ))
+}
+
+fn configure_log_capture(
+    options: &BoxOptions,
+    layout: &BoxFilesystemLayout,
+    container_id: &ContainerID,
+    volume_mgr: &mut GuestVolumeManager,
+) -> BoxliteResult<Option<String>> {
+    if !options.capture_logs {
+        return Ok(None);
+    }
+
+    let managed_log_path = layout.container_log_path(container_id.as_str());
+    let managed_log_dir = layout.container_logs_dir(container_id.as_str());
+    std::fs::create_dir_all(&managed_log_dir).map_err(|e| {
+        BoxliteError::Storage(format!(
+            "failed to create container log directory {}: {}",
+            managed_log_dir.display(),
+            e
+        ))
+    })?;
+    std::fs::OpenOptions::new()
+        .create(true)
+        .append(true)
+        .open(&managed_log_path)
+        .map_err(|e| {
+            BoxliteError::Storage(format!(
+                "failed to create container log file {}: {}",
+                managed_log_path.display(),
+                e
+            ))
+        })?;
+
+    volume_mgr.add_fs_share(
+        mount_tags::CONTAINER_LOGS,
+        managed_log_dir,
+        Some(logs::GUEST_DIR),
+        false,
+        None,
+    );
+
+    Ok(Some(
+        PathBuf::from(logs::GUEST_DIR)
+            .join("console.log")
+            .to_string_lossy()
+            .into_owned(),
     ))
 }
 
--- a/src/boxlite/src/runtime/options.rs
+++ b/src/boxlite/src/runtime/options.rs
@@ -415,6 +415,14 @@ pub struct BoxOptions {
     #[serde(default)]
     pub tty: bool,
 
+    /// Capture the container init process stdout/stderr as CRI-formatted records.
+    ///
+    /// This is opt-in because it creates an additional writable guest share.
+    /// When enabled, logs are written to `logs/{container_id}/console.log` under
+    /// the box-managed directory, with stream identity, timestamps, and rotation.
+    #[serde(default)]
+    pub capture_logs: bool,
+
     /// Secrets for MITM proxy injection into outbound HTTP(S) requests.
     ///
     /// Each secret maps a placeholder string to a real value. When the box
@@ -511,6 +519,7 @@ fn default_detach() -> bool {
 impl Default for BoxOptions {
     fn default() -> Self {
         Self {
+            capture_logs: false,
             cpus: None,
             memory_mib: None,
             disk_size_gb: None,
--- a/src/guest/src/container/lifecycle.rs
+++ b/src/guest/src/container/lifecycle.rs
@@ -106,6 +106,7 @@ impl Container {
         user: &str,
         user_mounts: Vec<UserMount>,
         tty: bool,
+        log_capture: Option<PathBuf>,
     ) -> BoxliteResult<Self> {
         let rootfs = rootfs.as_ref();
         let workdir = workdir.as_ref();
@@ -199,7 +200,8 @@ impl Container {
         } else {
             // Pipes exist before the container does. The guest holds stdin's
             // write-end, so init's read() blocks instead of seeing EOF.
-            let (stdio, init_fds) = ContainerStdio::pipes()?;
+            let (mut stdio, init_fds) = ContainerStdio::pipes()?;
+            stdio.start_output_pump(log_capture)?;
             start::create_container_with_stdio(
                 container_id,
                 &state_root,
@@ -369,10 +371,10 @@ impl Container {
         )
     }
 
-    /// Drain init process stdout and stderr.
+    /// Take the bounded diagnostic tails for init stdout and stderr.
     ///
-    /// Reads all available data from the init process pipes using non-blocking I/O.
-    /// Can only be called once — subsequent calls return empty strings.
+    /// The always-on output pump owns the pipes and retains the latest bytes
+    /// per stream. Taking the tails clears them; subsequent calls are empty.
     ///
     /// # Returns
     ///
@@ -404,7 +406,7 @@ impl Container {
     pub fn diagnose_exit(&mut self) -> String {
         let container_state_path = self.container_state_path();
 
-        // Drain init process output before building diagnostics
+        // Take the bounded init output tails before building diagnostics
         let (init_stdout, init_stderr) = self.drain_init_output();
 
         // Try to load container state from libcontainer
@@ -483,7 +485,8 @@ impl Container {
     /// # Returns
     ///
     /// Ok(()) on successful shutdown, or if container was already stopped.
-    pub fn shutdown(&self, timeout_ms: u64) -> BoxliteResult<()> {
+    pub async fn shutdown(&mut self, timeout_ms: u64) -> BoxliteResult<()> {
+        const LOG_DRAIN_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(1);
         self.is_shutdown
             .store(true, std::sync::atomic::Ordering::SeqCst);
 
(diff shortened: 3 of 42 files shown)


src/guest/src/container/stdio.rs
```diff
@@ -139,6 +246,122 @@ impl ContainerStdio {
...
+            if let Err(error) = self.rotate().await {
+                tracing::warn!(
+                    path = %self.path.display(),
+                    %error,
+                    "Failed to rotate container init log; continuing with current file"
+                );
+            }
+        }
+
+        if let Err(error) = self.file.write_all(chunk).await {
```
DorianZheng: `stream` is only read by the failure warning below — the file itself gets `write_all(chunk)` verbatim, so nothing in `{cid}.log` distinguishes stdout from stderr, and there are no timestamps.

that has a second effect: both pipes are read in 8 KiB chunks and written raw, so a chunk ending mid-line can be followed by the other stream's chunk, splicing two partial lines together with no framing that lets a reader undo it.

#909's use case is grafana alloy → loki. alloy's `stage.cri {}` and `stage.docker {}` both key on a per-line prefix, so against raw bytes neither parses and no `stream`/`time` labels come through. the two established shapes both split on `\n` and prefix per line — docker json-file (`{"log","stream","time"}`) and the cri text line (`<RFC3339Nano> <stdout|stderr> <F|P> <msg>`), where `P`/`F` marks a line split at a max length so a tailer can reassemble it.

worth settling what `{cid}.log` is claiming to be before this ships — it's the difference between the bytes being on disk and the reporter's pipeline working.

src/boxlite/src/litebox/init/tasks/vmm_spawn.rs
```diff
@@ -254,13 +263,52 @@ async fn build_config(
...
+            ))
+        })?;
+
+    volume_mgr.add_fs_share(
+        mount_tags::CONTAINER_LOGS,
+        layout.logs_dir(),
+        Some(logs::GUEST_DIR),
+        false,
+        None,
+    );
```
DorianZheng, line 313: 4th arg is `read_only`, so this shares the whole per-box `logs/` dir writable — and that dir also holds `console.log` and `boxlite-shim.log`. the guest ends up with write access to the host-side forensic record, not just to its own container log.

it lands at `/run/boxlite/logs` in the guest and isn't in `user_mounts`, so the container namespace can't reach it directly — this is guest-escape blast radius rather than direct container access. scoping the share to a dedicated subdir (`logs/container/`) would keep the capture file writable and leave console/shim logs out of reach.

for what it's worth, the micro-vm runtimes that ship this keep the guest off the host fs for logs entirely — kata pumps container stdio over vsock and opens the log file on the host side, firecracker-containerd does the same with a vsock port per stream. in both, the serial console stays kernel+agent only, which is exactly what `console.log` already is here.

src/guest/src/container/lifecycle.rs
```diff
@@ -167,9 +172,14 @@ impl Container {
...
-        let (stdio, init_fds) = ContainerStdio::new()?;
+        let (mut stdio, init_fds) = ContainerStdio::new()?;
 
-        // Create and start container with custom stdio
+        // Prepare capture before creating libcontainer state so setup failures
+        // cannot leave an orphaned created-but-unowned container. The readers
+        // are active before init starts, preserving short-lived output.
+        if let Some(log_capture) = log_capture {
+            stdio.start_capture(log_capture)?;
+        }
```
DorianZheng: `start_capture` only runs when a path is passed, so with `capture_logs` off — the default — nothing ever reads the stdout/stderr read-ends. they stay held by `ContainerStdio` and are only touched by the one-shot 4 KiB `drain_output` on failure.

so the ~64 KiB pipe buffer still fills for any container writing more than that to stdout/stderr, and init blocks on `write()`. an nginx image with the stock access-log→`/dev/stdout` symlink reaches it under sustained traffic. turning capture on fixes it as a side effect; leaving it off doesn't.

draining unconditionally and only *writing* when capture is on would cover the default path too — and would remove the reason warm-pool reuse has to be bypassed, since the conditional part becomes the behaviour rather than the mount.

src/guest/src/container/stdio.rs
```diff
@@ -124,10 +138,103 @@ impl ContainerStdio {
...
+                            Err(error) => {
+                                tracing::warn!(%error, "Failed to read container init stderr");
+                                stderr_open = false;
+                            }
+                        }
+                    }
+                }
+            }
+
+            writer.flush().await;
```
DorianZheng: `tokio::fs::File` buffers, and the only flushes are here at task end and inside `rotate()`. for a long-running container nothing forces bytes out in between, so a host tailer can sit on stale data for an unbounded stretch — which is the "alloy tails the file" case from #909.

worth confirming what tokio actually submits per `write_all`; if it doesn't reach the os promptly this needs a flush per chunk, or a periodic one.

src/boxlite/src/runtime/options.rs
```diff
@@ -386,6 +386,13 @@ pub struct BoxOptions {
     #[serde(default)]
     pub user: Option<String>,
 
+    /// Capture the container init process stdout/stderr to a host-visible log file.
+    ///
+    /// This is opt-in because it creates an additional writable guest share.
+    /// When enabled, logs are written under the box-managed logs directory.
+    #[serde(default)]
+    pub capture_logs: bool,
```
DorianZheng, line 424: #909 asked for two knobs — `captureLogs: true` for the managed path and `logPath: "/var/log/my-workloads/box-1.log"` for a caller-chosen one. only the first is here, and the path is fixed at `{cid}.log`.

deliberate scope cut? the reporter points grafana alloy at a specific directory, so the custom path may be the part they actually need.

src/guest/src/container/stdio.rs
```diff
@@ -124,10 +138,103 @@ impl ContainerStdio {
...
+            task.abort();
+            let _ = task.await;
+        }
+    }
+
     /// Drain all available output from init process stdout and stderr.
     ///
     /// Takes ownership of the pipe read-ends and reads with non-blocking I/O.
-    /// Can only be called once — subsequent calls return empty strings.
+    /// Can only be called once. If log capture owns the pipe read-ends, this
```
DorianZheng: once capture owns the read-ends this returns empty, so `diagnose_exit` loses the init stdout/stderr it used to surface on startup failure — precisely when capture is enabled. the bytes are in `{cid}.log`; reading its tail here would keep the diagnostic intact.

DorianZheng (commented):
shape matches what #909 asked for — opt-in and default off, host-visible file, survives detach, rotation by rename/reopen rather than copytruncate, and capture starting before init so short-lived output isn't lost.

three things stand out: the on-disk format is raw bytes today, so stdout/stderr aren't separable and alloy has nothing to parse; the writable share covers the whole `logs/` dir including `console.log` and `boxlite-shim.log`; and the pipe-fill stall remains on the default capture-off path.

PR #970, opened 2026-07-10
fix(tunnel): preserve responses after half-close

G4614:
Upgrade gvproxy so tunnel responses remain readable after the client closes its write side.

Test plan:
- [x] `go test ./... -count=1` in `src/deps/libgvproxy-sys/gvproxy-bridge`
- [x] `make dev:python` on the Linux debug host
- [x] Local Python SDK tunnel received the complete HTTP response after `shutdown_write()`


G4614:
Closing for now: the current change does not resolve cloud TCP half-close across the NLB TLS-termination boundary. The transport design needs to be revisited before continuing.

PR #986, opened 2026-07-14
fix(guest): treat zombie init as stopped

G4614:
Treat a zombie container init process as stopped when loading guest container status.

Test plan:
- [x] `cargo fmt --all --check`
- [x] ` 'source ~/.cargo/env && cd ~/boxlite-pol104-zombie-init && cargo test -p boxlite-guest load_container_status -- --nocapture'`
- [ ] Pre-push hook completed locally; blocked by local mac native runtime integration build, branch pushed with `--no-verify` after targeted Linux guest validation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container status reporting by checking that a container marked as running still has a live, matching init process on Linux.
  * If the init PID is missing, is zombie/dead, or the recorded init start-time no longer matches current `/proc` data, the status is now shown as **stopped**.
* **Tests**
  * Added Linux-only coverage for init PID zombie scenarios and mismatched start-time cases.

### DIFF
--- a/src/guest/src/container/start.rs
+++ b/src/guest/src/container/start.rs
@@ -274,6 +274,8 @@ pub(crate) fn cleanup_bundle_directory(bundle_path: &std::path::Path) {
 pub(crate) fn load_container_status(
     container_state_path: &Path,
 ) -> BoxliteResult<libcontainer::container::ContainerStatus> {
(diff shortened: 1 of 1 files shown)


G4614:
ref #590 the similar method for oci init process

PR #996, opened 2026-07-16
Split NetworkSpec into outbound/inbound directions

G4614:
## Summary

`NetworkSpec` modeled guest egress only; whether a box's exposed services are reachable from outside had no field anywhere. This PR splits the two directions in the core type and adapts every dependent surface to compile — without exposing inbound configuration yet (that's #1206, stacked on this).

## Before/after

```
BoxOptions.network: NetworkSpecV2 {                              (options.rs)
  outbound: NetworkSpec::Enabled{allow_net}|Disabled,             — the pre-split enum, untouched
  inbound:  InboundNetworkSpec::Enabled{allow_net}|Disabled,      — new: Enabled=public (default), Disabled=private
}
impl From<NetworkSpec> for NetworkSpecV2                         — fills inbound with its default

NetworkSpecV2::try_from(NetworkConfig{outbound, inbound})        — one validation point
  ├─ rejects disabled outbound + allow_net (as before)
  ├─ rejects non-empty inbound allow_net ("not supported yet" — no enforcement sink exists)
  └─ legacy egress-only wire shape still deserializes, with a deprecation warning
BoxOptions::sanitize                                             — repeats the inbound allowlist check at create,
                                                                   catching FFI callers that build the spec directly
```

## Rust source compatibility

`NetworkSpec` is untouched — same name, same variants, same wire form — so pre-split source and already-persisted box configs keep working. Verified by `network_spec.rs::pre_split_network_spec_source_shape_still_compiles`, which compiles as an external crate against the public API:

| pre-split code | change needed |
| --- | --- |
| `NetworkSpec::Enabled { allow_net: v }` literal | none |
| `match spec { NetworkSpec::Enabled { .. } => … }` | none |
| persisted `{"Enabled":{...}}` / `"Disabled"` | none |
| `BoxOptions { network: spec, .. }` | add `.into()` |
| `match opts.network { … }` | read `opts.network.outbound` |

The last two are unavoidable: a struct field has exactly one type, and the pre-split enum cannot also be the two-direction container. rustc points at both and suggests the fix.

## Stack

1. **this PR** — core split + compile adaptations
2. #1206 — expose inbound across CLI/SDKs/serve + `NetworkInfo` read side (stacked on this)
3. ~~#1199 — apps/api REST DTO alignment~~ — **merged** (`2a567eb5`); its API accepts both the flat and nested wire shapes, so #1206's client is unblocked

## Test plan

- `boxlite` core 999/999 (incl. the source-compat test), `boxlite-cli` 191/191, C 72/72 (1 ignored), Node 21/21
- Python: `cargo check -p boxlite-python --tests` clean (`cargo test -p boxlite-python` fails to link libpython on current main in this environment — pre-existing, verified against a clean origin/main worktree)
- `make fmt:check:rust` and `make fmt:check:go` clean

## Docs

`docs/reference/rust/README.md` now documents the `NetworkPolicy` container, both direction types, the `OutboundNetworkPolicy` alias, why `inbound.allow_net` must be empty, and the `From<NetworkSpec>` path pre-split callers use.

### DIFF
--- a/src/boxlite/src/runtime/options.rs
+++ b/src/boxlite/src/runtime/options.rs
@@ -329,6 +329,13 @@ pub struct BoxOptions {
     pub rootfs: RootfsSpec,
     pub volumes: Vec<VolumeSpec>,
     pub network: NetworkSpec,
+    /// Inbound reachability of the services this box exposes. A sibling of
+    /// `network` rather than a field inside it: the two directions are
+    /// independent, and `network` keeps its pre-split meaning (egress
+    /// only). Same type as `network` — both directions have identical
+    /// shape.
+    #[serde(default)]
+    pub inbound_network: NetworkSpec,
     /// Explicit host publication for the local runtime.
     ///
     /// Remote runtimes reject port mappings; use a box network tunnel for
@@ -525,6 +532,7 @@ impl Default for BoxOptions {
             rootfs: RootfsSpec::default(),
             volumes: Vec::new(),
             network: NetworkSpec::default(),
+            inbound_network: NetworkSpec::default(),
             ports: Vec::new(),
             auto_remove: default_auto_remove(),
             auto_stop: None,
@@ -582,17 +590,16 @@ impl BoxOptions {
         self.advanced.validate_privileged_capability_conflict()?;
         self.advanced.capabilities.validate()?;
 
-        if matches!(self.network.outbound, OutboundNetworkSpec::Disabled) && !self.ports.is_empty()
-        {
+        if matches!(self.network, NetworkSpec::Disabled) && !self.ports.is_empty() {
             return Err(boxlite_shared::errors::BoxliteError::Config(
                 "ports require network.outbound.mode=\"enabled\"".to_string(),
             ));
         }
 
-        // Wire conversions already reject this (NetworkSpec::try_from), but
-        // FFI callers (C/Go) build NetworkSpec directly — catch them at
-        // create. See try_from for the rationale; lift once enforcement lands.
-        if matches!(&self.network.inbound, InboundNetworkSpec::Enabled { allow_net } if !allow_net.is_empty())
+        // Wire conversions already reject this (the InboundNetworkConfig
+        // TryFrom), but FFI callers (C/Go) set the spec directly — catch
+        // them at create. See try_from for the rationale.
+        if matches!(&self.inbound_network, NetworkSpec::Enabled { allow_net } if !allow_net.is_empty())
         {
             return Err(boxlite_shared::errors::BoxliteError::Config(
                 "inbound.allow_net is not supported yet; remove it \
@@ -693,7 +700,7 @@ pub struct OutboundNetworkConfig {
     pub allow_net: Vec<String>,
 }
 
-/// Wire shape for [`InboundNetworkSpec`], aligned field-for-field with
+/// Wire shape for the inbound direction, aligned field-for-field with
 /// [`OutboundNetworkConfig`]. `Enabled` means services the box exposes are
 /// publicly reachable; `Disabled` means they are private (unreachable from
 /// outside the box). `allow_net` exists for shape symmetry with outbound but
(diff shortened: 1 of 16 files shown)


zombee0:
Does the current implementation establish a new runner-to-box connection for every HTTP request? Does each request require re-authentication? Is the user base characterized by low concurrency? Is the performance acceptable?

apps/api/src/boxlite-rest/boxlite-proxy.controller.ts
```diff
@@ -176,6 +179,38 @@ export class BoxliteProxyController {
...
+        req,
+        res,
+        next,
+        { proxyTimeout: 0 },
+      )
+    }
+
+    const { url } = await this.boxService.getPortPreviewUrl(boxId, authContext.organizationId, port)
+    return res.status(200).json({ url })
+  }
```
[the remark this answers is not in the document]
G4614: Not applying this. The descriptor endpoint only issues a URI/ticket and does not touch the guest runtime. Calling startHint here would mutate desired state and could start billing merely because a client requested endpoint metadata; actual lifecycle changes must remain explicit control-plane operations.

apps/proxy/pkg/proxy/proxy.go
```diff
@@ -193,8 +198,9 @@ func StartProxy(ctx context.Context, config *config.Config) error {
 	})
 
 	httpServer := &http.Server{
-		Addr:    fmt.Sprintf(":%d", config.ProxyPort),
-		Handler: router,
+		Addr:              fmt.Sprintf(":%d", config.ProxyPort),
+		Handler:           connectAwareHandler(http.HandlerFunc(proxy.handleTunnelConnect), router),
+		ReadHeaderTimeout: 10 * time.Second,
 	}
```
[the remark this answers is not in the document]
G4614: Not applying a server-wide ReadTimeout. This listener intentionally carries long-lived CONNECT tunnels and streaming preview traffic, so an overall read deadline would terminate valid idle or long-running sessions. The PR-added header timeout was removed to preserve main behavior, while tunnel establishment now has a dedicated bounded timeout.

apps/proxy/pkg/proxy/tunnel.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	defer clientConn.Close()
+	if _, err := buffered.WriteString("HTTP/1.1 200 Connection Established\r\n\r\n"); err != nil {
+		return
+	}
+	if err := buffered.Flush(); err != nil {
+		return
+	}
+
+	proxyTunnelStreams(clientConn, runnerConn)
```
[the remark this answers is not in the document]
G4614: indeed the def of product, the stream is for public

apps/proxy/pkg/proxy/proxy.go
```diff
@@ -194,7 +199,7 @@ func StartProxy(ctx context.Context, config *config.Config) error {
 
 	httpServer := &http.Server{
 		Addr:    fmt.Sprintf(":%d", config.ProxyPort),
-		Handler: router,
+		Handler: connectAwareHandler(http.HandlerFunc(proxy.handleTunnelConnect), router),
```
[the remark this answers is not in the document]
G4614: Fixed in b9f84c30. CONNECT requests now register with the proxy shutdown wait group for the full handler/tunnel lifetime, and the lifecycle test verifies shutdown blocks until the tunnel closes.

G4614:
Split into #992 (core/OpenAPI), #994 (SDKs and SDK E2E), and #995 (API/proxy/runner); all three branches are rebuilt from the latest main.

src/boxlite/src/runtime/options.rs
```diff
@@ -327,6 +327,10 @@ pub struct BoxOptions {
     pub rootfs: RootfsSpec,
     pub volumes: Vec<VolumeSpec>,
     pub network: NetworkSpec,
+    /// Whether service preview endpoints are public or private. `None` lets the
+    /// REST control plane use its default policy.
+    #[serde(default)]
+    pub service_access: Option<ServiceAccess>,
```
DorianZheng: move inside  NetworkSpec. NetworkSpec {Inbound, Outbound}. support inbound cidr in the future

G4614:
@coderabbitai review

G4614:
@coderabbitai review

src/boxlite/src/runtime/types.rs
```diff
@@ -424,8 +424,8 @@ impl BoxInfo {
             cpus: config.options.cpus.unwrap_or(DEFAULT_CPUS),
             memory_mib: config.options.memory_mib.unwrap_or(DEFAULT_MEMORY_MIB),
             network: Some(NetworkInfo {
-                mode: network_config.mode,
-                allow_net: network_config.allow_net,
+                mode: network_config.outbound.mode,
+                allow_net: network_config.outbound.allow_net,
```
DorianZheng, line 443: How do the user know allow_net means outbound allow_net only?

G4614: Thx, the BoxInfo.network will be refactored as  
```
    outbound: NetworkDirectionInfo {
        mode: network_config.outbound.mode,
        allow_net: network_config.outbound.allow_net,
    },
    inbound: NetworkDirectionInfo {
        mode: network_config.inbound.mode,
        allow_net: network_config.inbound.allow_net,
    },
    published_ports: crate::litebox::ports::resolved_from_state(config, state),
```

G4614: Also this is a break to the publish contract, should be signed

G4614: Splitting this into a separate PR: #1198 adds `outbound`/`inbound` to `NetworkInfo` so `allow_net` (and `mode`) are readable per-direction, not just outbound.

G4614: Update: landed directly in this PR (see the latest push) — `NetworkInfo` now has `outbound`/`inbound`, each `{mode, allow_net}`, mirroring `NetworkSpec`. (Earlier I'd split this into #1198, but core + all 4 SDK bindings are compile-coupled in the same Cargo workspace, so they need to ship together — #1198 is closed, folded back in here. The REST DTO side of this ships separately as #1199, since apps/api isn't part of that workspace.)

src/boxlite/src/runtime/options.rs
```diff
@@ -524,7 +524,7 @@ impl Default for BoxOptions {
             env: Vec::new(),
             rootfs: RootfsSpec::default(),
             volumes: Vec::new(),
-            network: NetworkSpec::default(),
+            network: NetworkPolicy::default(),
```
DorianZheng: why change the name

G4614: because if we want to accept the old `pub enum NetworkSpec: {Enabled{allow_net}, Disabled }` for Rust SDK, we can't just change it to `pub struct NetworkSpec { outbound, inbound }` and let it accept the old non-struct format at the same time

G4614: like this?
```
pub enum NetworkSpec {
    Enabled { allow_net: Vec<String> },   
    Disabled,                              
    Policy { outbound: OutboundNetworkSpec, inbound: InboundNetworkSpec },
}
```

apps/api/src/boxlite-rest/boxlite-box.controller.ts
```diff
@@ -79,6 +79,12 @@ export class BoxliteBoxController {
         auto_pause: req.body?.auto_pause,
         auto_delete: req.body?.auto_delete,
         auto_resume: req.body?.auto_resume,
+        network: req.body?.network
+          ? {
+              outbound: req.body.network.outbound,
+              inbound: req.body.network.inbound,
+            }
+          : undefined,
```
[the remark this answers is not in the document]
G4614: No longer applicable to this PR: the `apps/api` REST layer was split out into #1199, which has since merged. Its DTO normalizes the legacy flat `network.{mode,allow_net}` shape into `outbound` before validation, so the shape is accepted rather than rejected, and the audit log sees the normalized body.

DorianZheng (dismissed):

DorianZheng (approved):

PR #1006, opened 2026-07-17
feat(cli): add env file support

### LINKED ISSUE #1005: Add `--env-file` support to `boxlite run`, `create`, and `exec`
# Description
BoxLite currently supports passing environment variables individually with `--env` / `-e`. For workflows involving many variables, repeating `-e` is inconvenient and makes commands difficult to maintain.

Add repeatable `--env-file <FILE>` support to:
`boxlite run`
`boxlite create`
`boxlite exec`

Environment files should use standard dotenv syntax, including quoted values, comments, export prefixes, escapes, and variable substitution.

# Expected behavior
```
boxlite run --env-file .env alpine:latest
boxlite create --env-file app.env --name app alpine:latest
boxlite exec --env-file test.env app -- pytest
```

Multiple files can be supplied:
```
boxlite run \
  --env-file .env \
  --env-file .env.local \
  -e LOG_LEVEL=debug \
  alpine:latest
```

Precedence should be:
1. Earlier environment files
2. Later environment files
3. Explicit --env / -e values
Later definitions override earlier ones. A bare `-e KEY` should continue to inherit the value from the host environment.

niyue:
## Summary
Adds repeatable `--env-file` support to `boxlite run`, `create`, and `exec`, making it easier to configure guest environments without specifying every variable individually.
This PR closes https://github.com/boxlite-ai/boxlite/issues/1005

## Changes
* Add shared environment handling for `run`, `create`, and `exec`.
* Parse environment files using `dotenvy`, supporting standard dotenv syntax such as quotes, comments, export, escapes, and variable substitution.
* Apply files in command-line order, with later files overriding earlier files and explicit `-e` values taking highest precedence.
* Preserve existing `-e KEY=value` and host lookup behavior for bare `-e KEY`.
* Add unit and integration coverage for parsing, precedence, missing or invalid files, and all three commands.
* Document the new option and its precedence rules.

## How to verify
```
make test:integration:cli FILTER=environment::tests
make test:integration:cli FILTER=env_file
```

## Summary by CodeRabbit

* **New Features**
  * Added repeatable `--env-file FILE` support to `boxlite run`, `create`, and `exec`.
  * Environment files are merged in order; later definitions override earlier ones, and explicit `-e/--env` takes precedence.
  * `-e KEY` can be loaded from the host environment; if missing, earlier values are removed and a warning is shown.
* **Bug Fixes**
  * Clarified/support detached `--tty` behavior when standard input is not a TTY.
* **Documentation**
  * Updated CLI reference and shared flag groups with `--env-file` examples and dotenv precedence/override behavior.
* **Tests**
  * Added/updated integration tests covering precedence, overrides, missing/invalid env-file behavior, and `--tty` + `--detach` scenarios.

### DIFF
--- a/docs/reference/cli/README.md
+++ b/docs/reference/cli/README.md
@@ -256,13 +256,14 @@ takes the command's exit code; `boxlite ps` shows it stopped and
 
 - Default (foreground): streams stdout/stderr to the terminal, exits with the box command's exit code. If the command was killed by signal *N*, exits with `128 + N` (Unix convention, see [Exit Codes](#exit-codes)).
 - `-d`/`--detach`: prints the box ID to stdout and exits `0` immediately; `auto_remove` is force-disabled in this mode so the box outlives the CLI process.
-- `--tty` with non-TTY stdin: fails with `the input device is not a TTY.`
+- Foreground `--tty` with non-TTY stdin: fails with `the input device is not a TTY.` Detached `-d -t` is allowed because the command does not attach to local stdin.
 
 **Examples:**
 
 ```bash
 boxlite run alpine:latest echo "Hello"
 boxlite run -it --rm alpine:latest /bin/sh
+boxlite run --env-file .env -e DEBUG=1 python:slim python app.py
 boxlite run -d --name web -p 8080:80 nginx:alpine
 boxlite run -v $(pwd):/work -w /work alpine:latest ls -la
 boxlite run --cpus 4 --memory 4096 python:slim python -c "print(2+2)"
@@ -291,6 +292,7 @@ Run a command in a *running* box. The `--` separator is required (`src/cli/src/c
 boxlite exec mybox -- echo "hello"
 boxlite exec -it mybox -- /bin/sh
 boxlite exec -e DEBUG=1 -w /app mybox -- pytest tests/
+boxlite exec --env-file test.env mybox -- pytest tests/
 ```
 
 ---
@@ -319,16 +321,20 @@ implicitly, because starting it runs that command. Start it deliberately with
 |------|-------|-------------|
 | `--rootfs PATH` | — | Use a prepared rootfs path instead of pulling/resolving an image |
 | `--env KEY=VALUE` | `-e` | Set environment variables (repeatable) |
+| `--env-file FILE` | — | Read environment variables from a file (repeatable) |
 | `--workdir PATH` | `-w` | Working directory inside the box |
 
 Also uses [`ResourceFlags`](#resourceflags) + [`PublishFlags`](#publishflags) + [`VolumeFlags`](#volumeflags) + [`ManagementFlags`](#managementflags).
 
-> Note: `create` accepts `--env` and `--workdir` directly rather than via `ProcessFlags` (no `-i`/`-t`/`-u` here, since no command is being executed).
+> Note: `create` flattens `EnvironmentFlags` directly and defines `--workdir`
+> itself rather than using `ProcessFlags`; it therefore does not expose
+> `-i`/`-t`/`-u`, since no command is executed.
 
 **Examples:**
 
 ```bash
 boxlite create --name mybox alpine:latest
+boxlite create --env-file .env --name configured alpine:latest
 boxlite create -p 8080:80 -v /data:/app/data --name web nginx:alpine
 boxlite create --rootfs /path/to/rootfs --name local-rootfs
 ```
@@ -599,19 +605,36 @@ boxlite completion fish > ~/.config/fish/completions/boxlite.fish
 
 Several commands flatten shared `clap` `Args` structs. Each is documented here once.
 
+### `EnvironmentFlags`
+
+Used by `run`, `create`, and `exec`.
+
+| Flag | Short | Description |
+|------|-------|-------------|
+| `--env KEY=VALUE` | `-e` | Set an environment variable; a bare key inherits from the host |
+| `--env-file FILE` | — | Read environment variables from a file; repeatable |
+
+Environment files use dotenv syntax, including quoted values, `export`
+prefixes, comments, escapes, and variable substitution. Files are applied in
+command-line order, later files override earlier files, and explicit `-e`
+values have the highest precedence. A bare key is supported by `-e` and
+inherits its value from the host. These values enter the guest; use BoxLite
+secret injection for credentials that must remain outside the VM.
+
 ### `ProcessFlags`
 
-Used by `run` and `exec` (defined at `src/cli/src/cli.rs:208-281`).
+Used by `run` and `exec` and includes [`EnvironmentFlags`](#environmentflags).
 
 | Flag | Short | Description |
 |------|-------|-------------|
 | `--interactive` | `-i` | Keep STDIN open even if not attached |
 | `--tty` | `-t` | Allocate a pseudo-TTY (stdout and stderr are merged in TTY mode) |
-| `--env KEY=VALUE` | `-e` | Set environment variables (repeatable; if value omitted, inherits from host) |
 | `--workdir PATH` | `-w` | Working directory inside the box |
 | `--user NAME[:GROUP]` | `-u` | Run as `name`/`uid`[:`group`/`gid`] |
 
-`--tty` implies `--interactive` when stdin is a TTY. `--tty` without a TTY-attached stdin is a hard error.
+In foreground mode, `--tty` implies `--interactive` when stdin is a TTY and
+fails when stdin is not a TTY. Detached `-d -t` is allowed because the command
+does not attach to local stdin.
 
 ### `ResourceFlags`
 
--- a/docs/reference/cli/README.md
+++ b/docs/reference/cli/README.md
@@ -622,9 +622,10 @@ take precedence over values defined earlier in the same file; if the host
 variable is unset, an earlier same-file value is used. Substitution is evaluated
 independently for each file, so references cannot use values defined by an
 earlier `--env-file`. A UTF-8 BOM at the start of a file is ignored. A bare key
-is supported by `-e` and inherits its value from the host. These values enter the
-guest; use BoxLite secret injection for credentials that must remain outside the
-VM.
+is supported by `-e` and inherits its value from the host. If the key is absent
+from the host, it remains unset and removes any value loaded from an earlier
+`--env-file`. These values enter the guest; use BoxLite secret injection for
+credentials that must remain outside the VM.
 
 ### `ProcessFlags`
 
--- a/docs/reference/cli/README.md
+++ b/docs/reference/cli/README.md
@@ -614,18 +614,24 @@ Used by `run`, `create`, and `exec`.
 | `--env KEY=VALUE` | `-e` | Set an environment variable; a bare key inherits from the host |
 | `--env-file FILE` | — | Read environment variables from a file; repeatable |
 
-Environment files use dotenv syntax, including quoted values, `export`
-prefixes, comments, escapes, and variable substitution. Files are applied in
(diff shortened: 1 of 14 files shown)


src/cli/src/environment.rs
```diff
@@ -0,0 +1,172 @@
...
+        self.resolve_with_lookup(|key| std::env::var(key).ok())
+    }
+
+    fn resolve_with_lookup<F>(&self, lookup: F) -> anyhow::Result<Vec<(String, String)>>
+    where
+        F: Fn(&str) -> Option<String>,
+    {
+        let mut resolved = Vec::new();
+
+        for path in &self.env_files {
```
DorianZheng, line 36: a `${VAR}` in a later env file can't see variables from an earlier file: each file gets its own fresh dotenvy parser, and nothing is fed between files. `base.env` with `PROJECT=demo` plus `svc.env` with `SERVICE=${PROJECT}-api` produces `SERVICE=-api`, silently (unless PROJECT also happens to be in the host env). the docs say later files override earlier ones, which reads like a merged view.

src/cli/src/environment.rs
```diff
@@ -0,0 +1,172 @@
...
+            let entries = dotenvy::from_path_iter(path)
```
DorianZheng: two things about this parse call:

when a file redefines a variable that also exists on the host and then uses it, dotenvy expands `${...}` from the host first. host `HOME=/Users/alice` + file `HOME=/app` then `CONFIG=${HOME}/cfg` gives `HOME=/app` but `CONFIG=/Users/alice/cfg` — the file's own line doesn't win inside `${...}`.

also, `from_path_iter` never strips a utf-8 BOM (only dotenvy's `load()` does), so a `.env` saved by windows notepad or powershell fails on its first key and aborts the whole command.

DorianZheng (commented):
notes from a pass over the env-file feature; parser behavior checked against dotenvy 0.15.7 source. the important ones: parse errors echo secret values, `${...}` substitution has two surprises (host env wins over the file's own lines; files can't see earlier files), and `run -u` quietly becomes active.

src/cli/src/environment.rs
```diff
@@ -0,0 +1,172 @@
...
+                .with_context(|| format!("failed to open environment file '{}'", path.display()))?;
+            for entry in entries {
+                let (key, value) = entry.with_context(|| {
```
DorianZheng: if a line fails to parse, the error prints the whole raw line — value included — to stderr (dotenvy's parse error carries the line text, and main prints the full chain). an env file usually holds secrets, so a broken line like `DB_PASS=p@ss'w0rd` echoes the password. worth masking the line content here.

separately, this context says 'failed to parse' for every mid-iteration error; pointing `--env-file` at a directory opens fine and then fails on read, so the user sees a parse error for what is a path mistake.

src/cli/src/environment.rs
```diff
@@ -0,0 +1,172 @@
...
+) -> anyhow::Result<()>
+where
+    F: Fn(&str) -> Option<String>,
+{
+    let (key, explicit_value) = match entry.split_once('=') {
+        Some((key, value)) => (key, Some(value.to_string())),
+        None => (entry, None),
+    };
+
+    if key.is_empty() || key.chars().any(char::is_whitespace) {
```
DorianZheng: this is stricter than the old `-e` handling: `-e ""` used to warn and continue, and `-e =v` / `-e 'A B=1'` were accepted, but now all three abort the command. probably good hardening, just worth a mention in the docs since scripts that build `-e` args from lists can hit it.

also `-e` and env files now disagree on what a legal key is: `-e 1FOO=bar` passes this check, but the same line inside an env file fails dotenvy's key grammar. moving a var from `-e` into a file can turn a working flag into a hard error.

src/cli/src/environment.rs
```diff
@@ -0,0 +1,172 @@
...
+        anyhow::bail!("invalid environment variable name '{}' in --env", key);
+    }
+
+    let value = explicit_value.or_else(|| lookup(key));
+    let Some(value) = value else {
```
DorianZheng, line 72: a bare `-e TOKEN` with TOKEN unset on the host skips, so a TOKEN from an earlier `--env-file` stays in the box. docker treats this case as unset — the var is removed from the final set. the warning also says 'skipping', but the file value actually survives, which is easy to misread.

src/cli/src/environment.rs
```diff
@@ -0,0 +1,172 @@
...
+        tracing::warn!(
```
DorianZheng, line 74: this skip branch has no test anymore — the deleted cli.rs test was the only one asserting a missing bare key ends up absent. the remaining tests either hit the lookup or pass an empty `-e` list, so a regression here would pass the suite.

src/cli/src/environment.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::fs;
+    use tempfile::TempDir;
+
+    #[test]
+    fn resolve_files_then_explicit_values() {
```
DorianZheng, line 100: file-overrides-file is never actually tested: the only key set in two files (`SHARED`) is also overridden by `-e` both here and in the integration test, so the assertion passes even if the second file's value stopped winning. one key shared by two files and absent from `-e` would pin it.

src/cli/src/commands/run.rs
```diff
@@ -133,20 +140,18 @@ impl BoxRunner {
...
-    }
-
-    fn validate_flags(&self) -> anyhow::Result<()> {
-        // Check TTY availability if requested
-        if self.args.process.tty && !io::stdin().is_terminal() {
-            anyhow::bail!("the input device is not a TTY.");
-        }
-
-        Ok(())
+        self.args.process.configure_command(
```
DorianZheng: routing run through `configure_command` quietly turns on `-u`: on main, `-u` was parsed but never reached the command for `run` (only exec used it), so `boxlite run -u 1000:1000 ...` ran as the image default; now it execs as uid 1000. real behavior change, no test covers it, and the pr doesn't mention it.

it also sends env and workdir twice — once baked into the box config via `apply_to`, once on the exec request here. the guest overlays by key so output is the same, but run's env now has two owners, and a later exec on a kept box only sees the box-config copy.

src/cli/src/commands/run.rs
```diff
@@ -133,20 +140,18 @@ impl BoxRunner {
...
+            BoxCommand::new(program)
+                .args(args)
+                .tty(self.args.process.tty),
```
DorianZheng: tty is set here and then again inside `configure_command`; exec relies on `configure_command` alone.

src/cli/src/cli.rs
```diff
@@ -336,17 +312,13 @@ pub struct ProcessFlags {
...
-    /// Internal helper for dependency injection of environment variables
-    fn apply_to_with_lookup<F>(&self, opts: &mut BoxOptions, lookup: F) -> anyhow::Result<()>
-    where
-        F: Fn(&str) -> Option<String>,
-    {
+    pub fn apply_to(
+        &self,
+        opts: &mut BoxOptions,
+        environment: &[(String, String)],
+    ) -> anyhow::Result<()> {
```
DorianZheng: nothing in this body can fail anymore now that env resolution moved out, so the `Result` return only produces a dead `?` at the call site and dead `.expect`s in tests. `ResourceFlags::apply_to` already returns `()`.

src/cli/src/commands/create.rs
```diff
@@ -61,10 +62,10 @@ impl CreateArgs {
         self.volume.apply_to(&mut options, global.home.as_deref())?;
         self.network.apply_to(&mut options)?;
         options.working_dir = self.workdir.clone();
+        options.env.extend(self.environment.resolve()?);
```
DorianZheng: this block (workdir + env + entrypoint onto `BoxOptions`) is the same rule as `ProcessFlags::apply_to`, and with `apply_env_vars` deleted the env half is now inlined in both places. a change to how these land on `BoxOptions` has to be made twice.

docs/reference/cli/README.md
```diff
@@ -573,19 +579,36 @@ boxlite completion fish > ~/.config/fish/completions/boxlite.fish
...
+
+Used by `run`, `create`, and `exec`.
+
+| Flag | Short | Description |
+|------|-------|-------------|
+| `--env KEY=VALUE` | `-e` | Set an environment variable; a bare key inherits from the host |
+| `--env-file FILE` | — | Read environment variables from a file; repeatable |
+
+Environment files use dotenv syntax, including quoted values, `export`
+prefixes, comments, escapes, and variable substitution. Files are applied in
```
DorianZheng: a `${VAR}` that doesn't resolve becomes an empty string with no warning — a typo like `${DB_HSOT}` silently produces a broken value. worth a caveat here since this sentence advertises substitution.

src/cli/src/environment.rs
niyue: Fixed. Each file is evaluated indepently, doc updated

src/cli/src/environment.rs
niyue: Fixed. 
1) Doc updated
2) The BOM issue seems to be a `dotenvy`'s issue. I logged https://github.com/allan2/dotenvy/issues/168 and submitted https://github.com/allan2/dotenvy/pull/169 to address it. Before a new version of `dotenvy` is released, it is handled in this repo

src/cli/src/environment.rs
niyue: Good catch. Secrect is avoid to be logged here, and error handling is improved as well to avoid misleading error message.

src/cli/src/environment.rs
niyue: A bare -e TOKEN with no host value now removes any value loaded from an earlier --env-file. The warning now says leaving unset

src/cli/src/environment.rs
niyue: Fixed. The old behavior is restored, --env-file intentionally follows dotenvy's stricter key grammar, and the docs now clarify that env files use dotenv syntax.

src/cli/src/environment.rs
niyue: Fixed. A dedicated regression test is added, confirming that a bare -e TOKEN remains absent when the host has no value and no earlier value exists.

src/cli/src/environment.rs
niyue: Fixed. Test `later_env_file_overrides_earlier_file` is added to cover this scenario.

docs/reference/cli/README.md
niyue: Fixed. Doc updated for this behavior

src/cli/src/commands/create.rs
niyue: Fixed.

src/cli/src/cli.rs
niyue: Fixed. Code simplified for this part

src/cli/src/commands/run.rs
niyue: Fixed. `run` no longer routes its initial command through `configure_command`; it now sets only program, args, and TTY.

`exec` still uses `configure_command` for per-command overrides.

src/cli/src/commands/run.rs
niyue: Fixed as part of the `run` command wiring cleanup.

DorianZheng:
@niyue Please fix the conflicts

niyue:
@DorianZheng I've addressed the conflicts, and rebased onto the latest master. Please check it out. Thanks.

PR #1008, opened 2026-07-19
feat: expose configurable auto-stop intervals

zombee0:
## Changes
- 

## How to verify
- 

## Risks / rollout
- 

## Summary by CodeRabbit

* **New Features**
  * Set and retrieve hosted box auto-stop intervals in minutes when creating boxes and via the box-scoped autostop endpoint; `0` disables auto-stop.
  * Dashboard now supports viewing and updating the interval.
  * SDKs updated to expose `autoStopInterval/auto_stop_interval` in options and box details (C, Go, Node.js, Python, plus API clients).

* **Bug Fixes**
  * Enforced non-negative whole-minute intervals within the allowed max value; tightened request/response contracts to integers.

* **Documentation**
  * Updated API/SDK documentation and examples to the correct `setAutoStopInterval` naming and integer constraints.

* **Tests**
  * Added/expanded validation and mapping tests for auto-stop interval behavior.


zombee0:
I will change the interval config with minutes

xhebox:
> I will change the interval config with minutes

Will we change it back to minutes? I don't think that is a good idea, since it is too long for future pause/resume.

zombee0:
> > I will change the interval config with minutes
> 
> Will we change it back to minutes? I don't think that is a good idea, since it is too long for future pause/resume.

When i change it from minute to second, i think minute is too long especially when our billing is base on second.
but it’s auto-stop, A very short auto-stop interval would make our implementation cost more (we'd have to report active status and check for auto-stop more frequently), so I think using minutes is appropriate. If users really want to save as much as possible, they can manage the lifecycle themselves; auto-stop just provides a good enough capability.

xhebox:
> When i change it from minute to second, i think minute is too long especially when our billing is base on second. but it’s auto-stop, A very short auto-stop interval would make our implementation cost more (we'd have to report active status and check for auto-stop more frequently), so I think using minutes is appropriate. If users really want to save as much as possible, they can manage the lifecycle themselves; auto-stop just provides a good enough capability.

OK. Then do my implementation change to minutes, too? I also changed units to seconds as written in the draft spec of 'auto pause/resume'

FYI: I have a draft branch here https://github.com/xhebox/boxlite/tree/auto-pause-resume

src/boxlite/src/litebox/mod.rs
```diff
@@ -122,6 +122,14 @@ impl LiteBox {
         self.box_backend.stop().await
     }
 
+    /// Set the hosted REST auto-stop interval in seconds.
+    ///
+    /// `0` disables auto-stop. This operation is supported by REST-backed
+    /// boxes; local runtimes do not run the hosted control-plane scheduler.
+    pub async fn set_auto_stop_interval(&self, interval: u32) -> BoxliteResult<()> {
```
DorianZheng: This should be a options argument only. Don't support it dynamically yet. We can do it in the future to simplify the PR

PR #1022, opened 2026-07-23
feat(exec): drain guest output before attach

BatmanByte:
## Summary

Drain each guest execution stdout and stderr as soon as it is created. This removes the dependency on a host Attach consumer for the guest process to make progress.

Guest stdio uses the Tokio I/O reactor rather than tokio fs. Idle pipes no longer occupy blocking-pool workers. Stdin uses the same reactor-backed path because PTY stdin and stdout descriptors share the non-blocking setting.

The guest retains at most 1 MiB of output for a late Attach. When that ring overwrites old bytes, Attach emits an OutputDropped control event with separate stdout and stderr byte counts. The host renders the notice and resets only the affected UTF-8 decoder.

## Scope

This PR intentionally contains only guest exec stdio, the shared gRPC protocol, and the local gRPC consumer needed to render it. It does not change REST wait behavior, the runner, API, CLI, or host output queueing.

## Verification

- make test:integration:rust FILTER=main_command_exits_after_large_output_without_attach
- make test:integration:rust FILTER=late_attach_reports_output_dropped
- make test:integration:rust FILTER=test_zygote_concurrent_stdin_pipes
- make clippy

The integration commands ran outside the sandbox against real macOS Hypervisor VMs.

make fmt:check is currently blocked before Rust checking by 186 pre-existing Prettier violations in untouched generated apps client files; make fmt:check:rust passed.

### DIFF
--- a/src/boxlite/src/portal/interfaces/exec.rs
+++ b/src/boxlite/src/portal/interfaces/exec.rs
@@ -409,32 +409,59 @@ impl ExecProtocol {
         match output.event {
             Some(exec_output::Event::Stdout(chunk)) => {
                 tracing::trace!(len = chunk.data.len(), "Received exec stdout");
-                if let Some(lost_bytes) =
-                    stdout.receive(chunk.offset, chunk.data, chunk.total_bytes)
-                {
-                    Self::report_gap(stderr, "stdout", lost_bytes);
-                }
+                Self::forward_output(
+                    stdout,
+                    "stdout",
+                    chunk.offset,
+                    chunk.data,
+                    chunk.total_bytes,
+                );
             }
             Some(exec_output::Event::Stderr(chunk)) => {
                 tracing::trace!(len = chunk.data.len(), "Received exec stderr");
-                if let Some(lost_bytes) =
-                    stderr.receive(chunk.offset, chunk.data, chunk.total_bytes)
-                {
-                    Self::report_gap(stderr, "stderr", lost_bytes);
-                }
+                Self::forward_output(
+                    stderr,
+                    "stderr",
+                    chunk.offset,
+                    chunk.data,
+                    chunk.total_bytes,
+                );
             }
             None => {}
         }
     }
 
-    fn report_gap(stderr: &mut OutputTracker, source: &str, lost_bytes: u64) {
+    fn forward_output(
+        output: &mut OutputTracker,
+        source: &str,
+        offset: Option<u64>,
+        data: Vec<u8>,
+        total_bytes: Option<u64>,
+    ) {
+        match output.receive(offset, data, total_bytes) {
+            ReceivedOutput::Data { data, lost_bytes } => {
+                if let Some(lost_bytes) = lost_bytes {
+                    Self::report_gap(output, source, lost_bytes);
+                }
+                output.stream.send_bytes(data);
+            }
+            ReceivedOutput::End { lost_bytes } => {
+                if let Some(lost_bytes) = lost_bytes {
+                    Self::report_gap(output, source, lost_bytes);
+                }
+            }
+            ReceivedOutput::Ignore => {}
+        }
+    }
+
+    fn report_gap(output: &mut OutputTracker, source: &str, lost_bytes: u64) {
         tracing::warn!(
             source,
             lost_bytes,
             "Guest output buffer dropped older output"
         );
-        let _ = stderr.stream.tx.send(format!(
-            "[boxlite] {source} output dropped {lost_bytes} bytes\n"
+        let _ = output.stream.tx.send(format!(
+            "[boxlite] {source} output dropped {lost_bytes} bytes\r\n"
         ));
     }
 
@@ -744,6 +771,17 @@ struct OutputTracker {
     stream: DecodedStream,
 }
 
+enum ReceivedOutput {
+    Data {
+        data: Vec<u8>,
+        lost_bytes: Option<u64>,
+    },
+    End {
+        lost_bytes: Option<u64>,
+    },
+    Ignore,
+}
+
 impl OutputTracker {
     fn new(tx: mpsc::UnboundedSender<String>) -> Self {
         Self {
@@ -757,15 +795,15 @@ impl OutputTracker {
         offset: Option<u64>,
         data: Vec<u8>,
         total_bytes: Option<u64>,
-    ) -> Option<u64> {
+    ) -> ReceivedOutput {
         if let Some(total_bytes) = total_bytes {
             let Some(offset) = offset else {
                 tracing::warn!(total_bytes, "Exec output end frame has no offset");
-                return None;
+                return ReceivedOutput::Ignore;
             };
             if !data.is_empty() || offset != total_bytes {
                 tracing::warn!(offset, total_bytes, "Invalid exec output end frame");
-                return None;
+                return ReceivedOutput::Ignore;
             }
             if total_bytes < self.expected_offset {
                 tracing::warn!(
@@ -773,19 +811,23 @@ impl OutputTracker {
                     total_bytes,
                     "Exec output end frame moved backwards"
                 );
-                return None;
+                return ReceivedOutput::Ignore;
             }
 
             let lost_bytes = total_bytes - self.expected_offset;
             self.expected_offset = total_bytes;
             self.stream.flush();
-            return (lost_bytes > 0).then_some(lost_bytes);
+            return ReceivedOutput::End {
+                lost_bytes: (lost_bytes > 0).then_some(lost_bytes),
+            };
         }
 
         let Some(offset) = offset else {
             self.expected_offset += data.len() as u64;
-            self.stream.send_bytes(data);
-            return None;
+            return ReceivedOutput::Data {
+                data,
+                lost_bytes: None,
+            };
         };
 
         if offset < self.expected_offset {
@@ -794,7 +836,7 @@ impl OutputTracker {
                 offset,
                 "Exec output chunk moved backwards"
             );
-            return None;
+            return ReceivedOutput::Ignore;
         }
 
         let lost_bytes = offset - self.expected_offset;
@@ -802,8 +844,10 @@ impl OutputTracker {
             self.stream.flush();
         }
         self.expected_offset = offset + data.len() as u64;
-        self.stream.send_bytes(data);
-        (lost_bytes > 0).then_some(lost_bytes)
+        ReceivedOutput::Data {
+            data,
+            lost_bytes: (lost_bytes > 0).then_some(lost_bytes),
+        }
     }
 
     fn flush(&mut self) {
@@ -1301,11 +1345,11 @@ mod tests {
             &mut stderr,
         );
 
-        assert_eq!(stdout_rx.try_recv().ok(), Some("after-gap".to_string()));
         assert_eq!(
-            stderr_rx.try_recv().ok(),
-            Some("[boxlite] stdout output dropped 1 bytes\n".to_string())
+            stdout_rx.try_recv().ok(),
+            Some("[boxlite] stdout output dropped 1 bytes\r\n".to_string())
         );
+        assert_eq!(stdout_rx.try_recv().ok(), Some("after-gap".to_string()));
         assert_eq!(stderr_rx.try_recv().ok(), Some("─".to_string()));
     }
 
@@ -1340,9 +1384,10 @@ mod tests {
 
         assert_eq!(stdout_rx.try_recv().ok(), Some("hello".to_string()));
         assert_eq!(
-            stderr_rx.try_recv().ok(),
-            Some("[boxlite] stdout output dropped 5 bytes\n".to_string())
+            stdout_rx.try_recv().ok(),
+            Some("[boxlite] stdout output dropped 5 bytes\r\n".to_string())
         );
+        assert!(stderr_rx.try_recv().is_err());
     }
 
     #[test]
--- a/src/boxlite/src/portal/interfaces/exec.rs
+++ b/src/boxlite/src/portal/interfaces/exec.rs
@@ -414,6 +414,23 @@ impl ExecProtocol {
                 tracing::trace!(len = chunk.data.len(), "Received exec stderr");
                 stderr.send_bytes(chunk.data);
             }
+            Some(exec_output::Event::Dropped(dropped)) => {
+                tracing::warn!(
+                    stdout_bytes = dropped.stdout_bytes,
+                    stderr_bytes = dropped.stderr_bytes,
+                    "Guest output buffer dropped older output"
+                );
+                if dropped.stdout_bytes > 0 {
+                    stdout.flush();
+                }
+                if dropped.stderr_bytes > 0 {
+                    stderr.flush();
+                }
+                let _ = stderr.tx.send(format!(
+                    "[boxlite] output dropped (stdout: {} bytes, stderr: {} bytes)\n",
+                    dropped.stdout_bytes, dropped.stderr_bytes
+                ));
+            }
             None => {}
         }
     }
@@ -1165,6 +1182,45 @@ mod tests {
         }
         assert_eq!(received, "─");
         assert!(stderr_rx.try_recv().is_err());
+    }
+
+    #[test]
+    fn output_drop_only_flushes_the_affected_utf8_decoder() {
+        use boxlite_shared::{OutputDropped, Stderr as StderrMsg, exec_output};
+
+        let (stdout_tx, mut stdout_rx) = mpsc::unbounded_channel::<String>();
+        let (stderr_tx, mut stderr_rx) = mpsc::unbounded_channel::<String>();
+        let mut stdout = DecodedStream::new(stdout_tx);
+        let mut stderr = DecodedStream::new(stderr_tx);
+
+        let output = |event| ExecOutput { event: Some(event) };
+        ExecProtocol::route_output(
+            output(exec_output::Event::Stderr(StderrMsg { data: vec![0xE2] })),
+            &mut stdout,
+            &mut stderr,
+        );
+        ExecProtocol::route_output(
+            output(exec_output::Event::Dropped(OutputDropped {
+                stdout_bytes: 1,
+                stderr_bytes: 0,
+            })),
+            &mut stdout,
+            &mut stderr,
+        );
+        ExecProtocol::route_output(
+            output(exec_output::Event::Stderr(StderrMsg {
+                data: vec![0x94, 0x80],
+            })),
+            &mut stdout,
+            &mut stderr,
+        );
+
+        assert!(stdout_rx.try_recv().is_err());
+        assert_eq!(
+            stderr_rx.try_recv().ok(),
+            Some("[boxlite] output dropped (stdout: 1 bytes, stderr: 0 bytes)\n".to_string())
+        );
+        assert_eq!(stderr_rx.try_recv().ok(), Some("─".to_string()));
     }
 
     /// Flushing a DecodedStream must drain held-over bytes, leave the
--- a/src/boxlite/src/portal/interfaces/exec.rs
+++ b/src/boxlite/src/portal/interfaces/exec.rs
@@ -337,8 +337,8 @@ impl ExecProtocol {
                     // cut, doubling visible columns and desyncing TUI cursor
                     // math (see https://github.com/.../issues/...). Holding
                     // the trailing partial across chunks fixes this.
-                    let mut stdout = DecodedStream::new(stdout_tx);
-                    let mut stderr = DecodedStream::new(stderr_tx);
+                    let mut stdout = OutputTracker::new(stdout_tx);
+                    let mut stderr = OutputTracker::new(stderr_tx);
 
                     loop {
                         // Use select! to handle cancellation while streaming
@@ -375,7 +375,8 @@ impl ExecProtocol {
                                 // synthesized "Attach stream error: …" line.
                                 stdout.flush();
                                 stderr.flush();
-                                let _ = stderr.tx.send(format!("Attach stream error: {}", e));
+                                let _ =
+                                    stderr.stream.tx.send(format!("Attach stream error: {}", e));
                                 break;
                             }
                             None => {
@@ -404,35 +405,67 @@ impl ExecProtocol {
         });
     }
 
-    fn route_output(output: ExecOutput, stdout: &mut DecodedStream, stderr: &mut DecodedStream) {
+    fn route_output(output: ExecOutput, stdout: &mut OutputTracker, stderr: &mut OutputTracker) {
         match output.event {
             Some(exec_output::Event::Stdout(chunk)) => {
                 tracing::trace!(len = chunk.data.len(), "Received exec stdout");
-                stdout.send_bytes(chunk.data);
+                if let Some(lost_bytes) =
+                    stdout.receive(chunk.offset, chunk.data, chunk.total_bytes)
+                {
+                    Self::report_gap(stderr, "stdout", lost_bytes);
+                }
             }
             Some(exec_output::Event::Stderr(chunk)) => {
                 tracing::trace!(len = chunk.data.len(), "Received exec stderr");
-                stderr.send_bytes(chunk.data);
+                if let Some(lost_bytes) =
+                    stderr.receive(chunk.offset, chunk.data, chunk.total_bytes)
+                {
+                    Self::report_gap(stderr, "stderr", lost_bytes);
+                }
             }
             Some(exec_output::Event::Dropped(dropped)) => {
-                tracing::warn!(
-                    stdout_bytes = dropped.stdout_bytes,
-                    stderr_bytes = dropped.stderr_bytes,
-                    "Guest output buffer dropped older output"
+                Self::report_legacy_drop(
+                    stdout,
+                    stderr,
+                    dropped.stdout_bytes,
+                    dropped.stderr_bytes,
                 );
-                if dropped.stdout_bytes > 0 {
-                    stdout.flush();
-                }
-                if dropped.stderr_bytes > 0 {
-                    stderr.flush();
-                }
-                let _ = stderr.tx.send(format!(
-                    "[boxlite] output dropped (stdout: {} bytes, stderr: {} bytes)\n",
-                    dropped.stdout_bytes, dropped.stderr_bytes
-                ));
             }
             None => {}
         }
+    }
+
+    fn report_gap(stderr: &mut OutputTracker, source: &str, lost_bytes: u64) {
+        tracing::warn!(
+            source,
+            lost_bytes,
+            "Guest output buffer dropped older output"
+        );
+        let _ = stderr.stream.tx.send(format!(
+            "[boxlite] {source} output dropped {lost_bytes} bytes\n"
+        ));
+    }
+
+    fn report_legacy_drop(
+        stdout: &mut OutputTracker,
+        stderr: &mut OutputTracker,
+        stdout_bytes: u64,
+        stderr_bytes: u64,
+    ) {
+        tracing::warn!(
+            stdout_bytes,
+            stderr_bytes,
+            "Legacy guest output buffer dropped older output"
+        );
+        if stdout_bytes > 0 {
+            stdout.flush();
+        }
+        if stderr_bytes > 0 {
+            stderr.flush();
+        }
+        let _ = stderr.stream.tx.send(format!(
+            "[boxlite] output dropped (stdout: {stdout_bytes} bytes, stderr: {stderr_bytes} bytes)\n"
+        ));
     }
 
     fn spawn_wait(
@@ -733,6 +766,78 @@ impl DecodedStream {
         if !tail.is_empty() {
             let _ = self.tx.send(tail);
         }
+    }
+}
+
+struct OutputTracker {
+    expected_offset: u64,
+    stream: DecodedStream,
+}
+
+impl OutputTracker {
+    fn new(tx: mpsc::UnboundedSender<String>) -> Self {
+        Self {
+            expected_offset: 0,
+            stream: DecodedStream::new(tx),
+        }
+    }
+
+    fn receive(
+        &mut self,
+        offset: Option<u64>,
+        data: Vec<u8>,
+        total_bytes: Option<u64>,
+    ) -> Option<u64> {
+        if let Some(total_bytes) = total_bytes {
+            let Some(offset) = offset else {
+                tracing::warn!(total_bytes, "Exec output end frame has no offset");
+                return None;
+            };
+            if !data.is_empty() || offset != total_bytes {
+                tracing::warn!(offset, total_bytes, "Invalid exec output end frame");
+                return None;
+            }
+            if total_bytes < self.expected_offset {
+                tracing::warn!(
+                    expected_offset = self.expected_offset,
+                    total_bytes,
+                    "Exec output end frame moved backwards"
+                );
+                return None;
+            }
+
+            let lost_bytes = total_bytes - self.expected_offset;
+            self.expected_offset = total_bytes;
+            self.stream.flush();
+            return (lost_bytes > 0).then_some(lost_bytes);
+        }
+
+        let Some(offset) = offset else {
+            self.expected_offset += data.len() as u64;
+            self.stream.send_bytes(data);
+            return None;
+        };
+
+        if offset < self.expected_offset {
+            tracing::warn!(
+                expected_offset = self.expected_offset,
+                offset,
+                "Exec output chunk moved backwards"
+            );
+            return None;
+        }
+
+        let lost_bytes = offset - self.expected_offset;
+        if lost_bytes > 0 {
+            self.stream.flush();
+        }
+        self.expected_offset = offset + data.len() as u64;
+        self.stream.send_bytes(data);
+        (lost_bytes > 0).then_some(lost_bytes)
+    }
+
+    fn flush(&mut self) {
+        self.stream.flush();
     }
 }
 
@@ -1163,16 +1268,20 @@ mod tests {
 
         let (stdout_tx, mut stdout_rx) = mpsc::unbounded_channel::<String>();
         let (stderr_tx, mut stderr_rx) = mpsc::unbounded_channel::<String>();
-        let mut stdout = DecodedStream::new(stdout_tx);
-        let mut stderr = DecodedStream::new(stderr_tx);
+        let mut stdout = OutputTracker::new(stdout_tx);
+        let mut stderr = OutputTracker::new(stderr_tx);
 
-        let mk_stdout = |bytes: Vec<u8>| ExecOutput {
-            event: Some(exec_output::Event::Stdout(StdoutMsg { data: bytes })),
+        let mk_stdout = |offset, bytes: Vec<u8>| ExecOutput {
+            event: Some(exec_output::Event::Stdout(StdoutMsg {
+                data: bytes,
+                offset: Some(offset),
+                total_bytes: None,
+            })),
         };
 
         // "─" split into [E2] and [94 80] across two messages.
-        ExecProtocol::route_output(mk_stdout(vec![0xE2]), &mut stdout, &mut stderr);
-        ExecProtocol::route_output(mk_stdout(vec![0x94, 0x80]), &mut stdout, &mut stderr);
+        ExecProtocol::route_output(mk_stdout(0, vec![0xE2]), &mut stdout, &mut stderr);
+        ExecProtocol::route_output(mk_stdout(1, vec![0x94, 0x80]), &mut stdout, &mut stderr);
 
         // First message: holdover only, no emission.
         // Second message: complete "─" emitted.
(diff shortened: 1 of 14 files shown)


BatmanByte:
@codex review

src/boxlite/src/rest/litebox.rs
```diff
@@ -983,49 +971,47 @@ fn parse_control_frame(text: &str) -> ControlFrame {
...
+                info.status
+            )),
+        },
+        Err(error) => ExecResult {
+            exit_code: -1,
+            error_message: Some(format!("wait request failed: {error}")),
+        },
+    };
+    let _ = result_tx.send(result);
 }
```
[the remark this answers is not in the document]
BatmanByte: Fixed in f28b16c6. `/wait` is now a bounded 45-minute long poll: runner and `boxlite serve` return `running` at the boundary, the API proxy allows 50 minutes, and the REST client allows 55 minutes then transparently starts another request. The client also retries `BoxliteError::Network` failures with bounded exponential backoff. Added regression tests for both a `running` response and a dropped first connection.

BatmanByte:
Review follow-up:

- The long-lived REST wait / proxy timeout finding is addressed in f28b16c6. Each server poll is bounded to 45 minutes, the API proxy has a 50-minute limit, and the SDK has a 55-minute request limit. `running` responses renew the poll; transient transport failures retry with bounded exponential backoff.
- The CLI `error_message` parity finding is valid but intentionally left for a separate follow-up. This change remains limited to making REST wait work across the existing one-hour ALB idle timeout.
- I did not abort the REST WebSocket pump when `wait()` resolves. That would cut output for callers that read stdout/stderr after waiting. A separate output-consumer lifecycle/cancellation policy is needed for that resource-retention tradeoff.

Tests include REST wait retry coverage and the runner wait-timeout controller test.

BatmanByte:
@codex review

src/boxlite/src/rest/litebox.rs
```diff
@@ -983,49 +974,79 @@ fn parse_control_frame(text: &str) -> ControlFrame {
...
+                tracing::debug!(execution_id, error, "wait request failed; retrying");
+                let delay = retry_backoff;
+                retry_backoff = retry_backoff.saturating_mul(2).min(WAIT_RETRY_BACKOFF_MAX);
+                delay
+            }
+            Err(error) => {
+                let _ = result_tx.send(ExecResult {
+                    exit_code: -1,
+                    error_message: Some(format!("wait request failed: {error}")),
+                });
```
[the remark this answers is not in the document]
BatmanByte: Fixed in c7b2ce1a. A 404 from the new `/wait` route now switches the SDK to the existing status route and continues using the same `running` retry loop. The fallback is attempted only once: a 404 from the status route still surfaces as the execution genuinely being unavailable. Added a regression test that verifies the route sequence and real exit-code propagation.

openapi/box.openapi.yaml
```diff
@@ -631,6 +632,30 @@ paths:
...
+      - $ref: "#/components/parameters/boxId"
+      - $ref: "#/components/parameters/execId"
+    get:
+      operationId: waitExecution
+      summary: Wait for execution completion
+      description: |
+        Waits for a terminal state for up to 45 minutes, then returns either
+        the terminal result or `running`. This control request is independent
+        from WebSocket output, so callers can wait without consuming stdout
+        or stderr. SDKs retry `running` responses until terminal completion.
```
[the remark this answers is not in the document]
BatmanByte: Fixed in c7b2ce1a. The `200` response now states that it can carry either a terminal result or `status: running` after the 45-minute long-poll boundary.

BatmanByte:
@codex review

src/boxlite/src/portal/interfaces/exec.rs
```diff
@@ -393,26 +430,78 @@ impl ExecProtocol {
...
                     tracing::debug!(execution_id = %execution_id, error = %e, "Attach failed");
-                    let _ = stderr_tx.send(format!("Attach failed: {}", e));
+                    let _ = stderr_tx.try_send(format!("Attach failed: {}", e));
                 }
             }
         });
     }
 
-    fn route_output(output: ExecOutput, stdout: &mut DecodedStream, stderr: &mut DecodedStream) {
+    async fn route_output(
```
DorianZheng: a single task routes both stdout and stderr through the bounded(1024) channels here, and `send_text` awaits the send. a caller that reads only one stream (the pattern shown in the exec_handle docs) fills the other channel; `route_output` then parks on `send().await`, the task stops polling the guest stream, and the stream that *is* being read never advances. meanwhile the guest's 1 MiB ring overflows and drops output. the pre-change unbounded senders couldn't block here.

src/boxlite/src/rest/litebox.rs
```diff
@@ -799,11 +802,11 @@ async fn attach_ws_pump(
                                 match *channel {
                                     0x01 => {
                                         tracing::trace!(len = text.len(), "WS attach: stdout frame");
-                                        let _ = stdout_tx.send(text);
+                                        let _ = stdout_tx.send(text).await;
```
DorianZheng: this send is now bounded and sits inside the single `select!` loop, so a full stdout channel blocks stdin forwarding and the keepalive ping too (same for the stderr send just below). an interactive exec whose consumer stops reading stdout — or an `exec().wait()` that never reads output — fills the 1024 slots, the pump parks on `send().await`, stdin stops flowing, no ping goes out, and the idle watchdog can reap a healthy connection.

src/boxlite/src/rest/litebox.rs
```diff
@@ -983,49 +974,89 @@ fn parse_control_frame(text: &str) -> ControlFrame {
...
+            Err(BoxliteError::NotFound(_)) if !using_status_endpoint => {
+                tracing::debug!(
+                    execution_id,
+                    "wait endpoint unavailable; falling back to status polling"
+                );
+                path = format!("/boxes/{}/executions/{}", box_id, execution_id);
+                using_status_endpoint = true;
+                continue;
+            }
+            Err(error) => {
```
DorianZheng: this terminal arm turns every non-`Network`, non-`NotFound` error into `ExecResult{exit_code:-1}`. an enveloped 503 (Portal), a `timeout` envelope (Internal), a bare 500 (Internal), and 429 (ResourceExhausted) all land here instead of retrying, and `Execution::wait()` caches the result via `get_or_try_init` — so a momentary gateway hiccup is permanently reported to the caller as the command having failed.

apps/api/src/boxlite-rest/boxlite-proxy.controller.ts
```diff
@@ -128,6 +128,27 @@ export class BoxliteProxyController {
     )
   }
 
+  @Get(':boxId/executions/:execId/wait')
+  async proxyExecWait(
```
DorianZheng: this endpoint invites a wait-only client (no attach, no other calls), but `proxyToRunner` writes `lastActivityAt` once at request start. a 25-min command polled only through `/wait` has its activity stamped at t=0, so the auto-pause sweeper (900s) can pause the box mid-wait. with `autoResume` off the exec stays frozen; with it on the box flaps and time is lost.

src/guest/src/service/exec/output.rs
```diff
@@ -0,0 +1,204 @@
...
+use async_stream::stream;
+use boxlite_shared::{exec_output, ExecOutput, OutputDropped, Stderr, Stdout};
+use futures::{Stream, StreamExt};
+use std::collections::VecDeque;
+use std::pin::Pin;
+use std::sync::Arc;
+use tokio::sync::{watch, Mutex};
+use tonic::Status;
+
+const BUFFER_CAPACITY_BYTES: usize = 1024 * 1024;
```
DorianZheng, line 13: every `ExecutionState` now retains an `OutputManager` ring up to this 1 MiB cap, and the guest `ExecutionRegistry` has no remove/evict path, so nothing frees it after the exec exits. a box that runs many output-heavy commands over its lifetime pins ~1 MiB each until it stops — guest RSS grows monotonically toward OOM.

src/boxlite/src/rest/litebox.rs
```diff
@@ -983,49 +974,89 @@ fn parse_control_frame(text: &str) -> ControlFrame {
...
+        {
+            Ok(info) if matches!(info.status.as_str(), "completed" | "killed" | "timed_out") => {
                 let _ = result_tx.send(ExecResult {
                     exit_code: info.exit_code.unwrap_or(-1),
-                    error_message: Some(cause.clone()),
+                    error_message: info.error_message,
+                });
+                return;
+            }
+            Ok(info) if info.status == "running" => {
```
DorianZheng: on this running branch the task loops without ever observing a dropped `Execution` handle (closed `result_tx`) — it only sends on the terminal arms. if the caller drops the handle while the command is still running, the poll task keeps going until `WAIT_TOTAL_TIMEOUT` (25h): ~1 req/45min on `/wait`, or ~1 req/s on the status fallback. the previous bounded attach pump exited when its receivers dropped.

apps/runner/pkg/api/controllers/boxlite_exec.go
```diff
@@ -182,6 +184,25 @@ func BoxliteGetExecution(ctx *gin.Context) {
...
+	exec, err := execManager.GetForBox(execId, boxId)
+	if err != nil {
+		ctx.JSON(classifyExecError(err), gin.H{"error": err.Error()})
+		return
+	}
+
+	select {
+	case <-exec.Done:
+		ctx.JSON(http.StatusOK, executionInfoFromManagedExec(exec))
+	case <-time.After(executionWaitPollTimeout):
```
DorianZheng: `time.After` arms a timer that Go can't collect until it fires, even when `exec.Done` or the client-cancel case wins the select first. every fast-completing or disconnected wait leaves a live timer+channel pinned for the full 45 minutes; under sustained wait traffic they pile up on the runtime timer heap.

src/boxlite/src/rest/litebox.rs
```diff
@@ -622,15 +634,7 @@ async fn attach_ws(
...
-            tracing::debug!(path = %path, error = %e, "WS attach: connect failed; falling back to status probe (output will not stream)");
-            emit_or_fallback(
-                client,
-                box_id,
-                execution_id,
-                &result_tx,
-                format!("WS connect failed: {}", e),
-            )
-            .await;
+            tracing::debug!(path = %path, error = %e, "WS attach: connect failed; output will not stream");
```
DorianZheng: when the ws upgrade fails (e.g. a proxy tunnels http but blocks the upgrade) attach just logs here and returns, and completion then arrives from `wait_execution` with `error_message: None` on a clean exit. the caller sees a clean exit code and empty stdout/stderr with no signal that output was silently lost — the distinction between 'no output' and 'output lost' is gone.

src/boxlite/src/rest/litebox.rs
```diff
@@ -983,49 +974,89 @@ fn parse_control_frame(text: &str) -> ControlFrame {
...
+            Ok(info) if matches!(info.status.as_str(), "completed" | "killed" | "timed_out") => {
                 let _ = result_tx.send(ExecResult {
                     exit_code: info.exit_code.unwrap_or(-1),
-                    error_message: Some(cause.clone()),
+                    error_message: info.error_message,
+                });
+                return;
+            }
+            Ok(info) if info.status == "running" => {
+                retry_backoff = WAIT_RETRY_BACKOFF_INITIAL;
```
DorianZheng: the running arm resets `retry_backoff` to the 1s initial every iteration, so against any server that returns `running` without blocking (the 404→status fallback, or a short-timeout intermediary) the client polls at a fixed ~1 req/s with no backoff for the exec's whole lifetime — tens of thousands of requests for a long exec, which can itself trip rate limiting.

DorianZheng (commented):
most of these trace to two changes in this PR — the unbounded→bounded output channels, and the narrowed wait-retry predicate:

- one task routes both streams through blocking bounded sends, so reading only one of stdout/stderr can stall the other (and on the ws pump a full channel also blocks stdin + the keepalive ping).
- `wait_execution` retries only `Network`, so a transient 503/timeout/429 becomes a cached `exit_code:-1`.

the rest are resource retention (guest ring + leaked wait tasks/timers), the auto-pause vs long-poll interaction, and a few decoder/observability items.

src/boxlite/src/rest/litebox.rs
```diff
@@ -1602,33 +1792,23 @@ mod tests {
...
         let attach = tokio::spawn(async move {
-            attach_ws(
-                &client, "box1", "exec1", stdin_rx, stdout_tx, stderr_tx, result_tx,
-            )
-            .await;
+            attach_ws(&client, "box1", "exec1", stdin_rx, stdout_tx, stderr_tx).await;
         });
 
-        let res = tokio::time::timeout(Duration::from_secs(3), result_rx.recv())
+        tokio::time::timeout(Duration::from_secs(3), attach)
```
DorianZheng: this test holds the server silent for 400ms then only asserts the attach task returns within 3s — it never inspects the `ExecResult`, though the doc comment just above says the emitted result must name the watchdog as cause. break the client watchdog entirely and this still passes: an ordinary server-side close is indistinguishable from a watchdog-driven exit here.

src/guest/src/service/exec/output.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    fn spawn_stderr(&self, stderr: ExecStderr) {
+        let manager = self.clone();
+        tokio::spawn(async move {
+            manager.drain(stderr, OutputSource::Stderr).await;
+        });
+    }
+
+    async fn drain<S>(&self, mut stream: S, source: OutputSource)
```
DorianZheng, line 204: this drain loop (and the spawn helpers) emit no tracing — the previous forwarding tasks logged per-exec (`execution=…, "…forwarding task ended"`). a stuck or prematurely-ended exec can no longer be correlated to an exec_id or a specific stream from guest logs.

src/boxlite/src/portal/interfaces/exec.rs
```diff
@@ -393,26 +430,78 @@ impl ExecProtocol {
...
                 tracing::trace!(len = chunk.data.len(), "Received exec stdout");
-                stdout.send_bytes(chunk.data);
+                stdout.send_bytes(chunk.data).await;
             }
             Some(exec_output::Event::Stderr(chunk)) => {
                 tracing::trace!(len = chunk.data.len(), "Received exec stderr");
-                stderr.send_bytes(chunk.data);
+                stderr.send_bytes(chunk.data).await;
+            }
+            Some(exec_output::Event::Dropped(_)) => {
```
DorianZheng: the `Dropped` event is stream-agnostic and flushes both decoders. a loss on one stream then corrupts an in-flight split codepoint on the other: if stderr is mid-codepoint (holding e.g. 0xE2) when a stdout burst overflows the ring, `stderr.flush()` emits U+FFFD for the held byte and the surviving continuation bytes decode to more U+FFFD — a healthy `─` on stderr becomes replacement chars though stderr lost nothing.

src/boxlite/src/rest/litebox.rs
```diff
@@ -983,49 +974,89 @@ fn parse_control_frame(text: &str) -> ControlFrame {
...
-            _ => {
-                // Server says the exec is still running — surface the
-                // synthesized cause so the caller sees the disconnect.
+            Err(BoxliteError::Network(error)) => {
+                tracing::debug!(execution_id, error, "wait request failed; retrying");
+                let delay = retry_backoff;
+                retry_backoff = retry_backoff.saturating_mul(2).min(WAIT_RETRY_BACKOFF_MAX);
+                delay
             }
+            Err(BoxliteError::NotFound(_)) if !using_status_endpoint => {
```
DorianZheng: `using_status_endpoint` latches true on the first `/wait` 404 and never resets, and a route-unsupported 404 is indistinguishable from a transient/real-exec 404 here. one early 404 (exec not yet visible on the addressed runner, or a proxy-level 404) permanently downgrades the client to immediate status polling for the whole exec, even against a server that supports the long-poll.

src/boxlite/src/portal/interfaces/exec.rs
```diff
@@ -685,31 +774,52 @@ impl Utf8StreamDecoder {
...
         Self {
-            tx,
+            tx: Some(tx),
             decoder: Utf8StreamDecoder::default(),
         }
     }
 
     /// Decode a wire chunk and forward any completed text to the receiver.
-    fn send_bytes(&mut self, data: Vec<u8>) {
+    async fn send_bytes(&mut self, data: Vec<u8>) {
```
DorianZheng: on shutdown cancellation this path can lose output: `send_bytes` decodes and *then* awaits the send, so a cancel mid-send drops the already-decoded chunk; and `try_flush` only handles `Closed`, not `Full`, so a held partial's trailing U+FFFD is dropped when the channel is full. the normal EOF/error path awaits the flush and delivers these, so the shutdown path silently diverges.

src/boxlite/src/rest/litebox.rs
```diff
@@ -983,49 +974,89 @@ fn parse_control_frame(text: &str) -> ControlFrame {
...
+                        "wait endpoint returned non-terminal status: {}",
+                        info.status
+                    )),
                 });
                 return;
             }
-            _ => {
-                // Server says the exec is still running — surface the
-                // synthesized cause so the caller sees the disconnect.
+            Err(BoxliteError::Network(error)) => {
```
DorianZheng: `Network` errors are retried until `WAIT_TOTAL_TIMEOUT` (25h) with no consecutive-failure cap, so a clearly-persistent outage (connection refused / bare 502 / DNS failure all map to `Network`) makes `Execution::wait()` hang for up to a day instead of failing fast.

BatmanByte:
@codex review

src/guest/src/service/exec/output.rs
```diff
@@ -0,0 +1,276 @@
...
+                        state.attachment_next_sequence = Some(next_sequence);
+                        Next::Item(dropped_output(state.pending_dropped.take()))
+                    } else if next_sequence < state.next_sequence {
+                        let index = (next_sequence - state.oldest_sequence) as usize;
+                        let output = state
+                            .entries
+                            .get(index)
+                            .expect("ring sequence must exist")
+                            .output
+                            .clone();
```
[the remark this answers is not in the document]
BatmanByte: Confirmed. This PR adds bounded per-execution retention but does not yet define the terminal lifecycle for that retained output, so repeated completed executions can accumulate memory. I am not applying the suggested local pop-on-send change here: it would only help attached executions and would silently change the late-Attach replay guarantee while leaving unattached completed executions retained.

I will address this in a focused follow-up with a complete lifecycle: retain the ring while execution is active and for a bounded terminal grace window; release entries already handed to the sole Attach stream; clear terminal rings after the window when no attachment is active; and preserve terminal Wait results independently from output retention. The follow-up will also define what a late Attach observes after expiry, rather than reusing OutputDropped byte counts whose current meaning is ring overwrite. This needs tests for attached, never-attached, terminal, and reconnect cases.

src/guest/src/service/exec/exec_handle.rs
```diff
@@ -354,3 +393,39 @@ impl ExecHandle {
...
+            .worker_threads(1)
+            .max_blocking_threads(1)
+            .enable_all()
+            .build()
+            .unwrap();
+        runtime.block_on(async {
+            let (read_fd, _write_fd) = nix::unistd::pipe().unwrap();
+            let mut stdout = ExecStdout::new(read_fd).unwrap();
+            let output_read = tokio::spawn(async move { stdout.next().await });
+            tokio::time::sleep(Duration::from_millis(50)).await;
```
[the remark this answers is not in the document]
BatmanByte: Fixed in 7fff2e0c. The test now waits for the output stream to be polled into Pending before submitting the blocking-pool probe, instead of relying on a 50 ms scheduling delay.

src/shared/proto/boxlite/v1/service.proto
```diff
@@ -354,9 +354,14 @@ message ExecOutput {
   oneof event {
     Stdout stdout = 1;
     Stderr stderr = 2;
+    OutputDropped dropped = 3;
   }
 }
 
+// The guest's bounded output buffer overwrote older bytes before Attach read
+// them. This is a control event, not guest stderr.
+message OutputDropped {}
```
DorianZheng: an alternative that avoids this control event: drop the `dropped` variant and put a per-stream absolute byte `offset` on `Stdout`/`Stderr`. the reader infers loss from an offset jump.

simple e2e:
```
guest: each stream keeps a running byte counter; stamp every chunk
       Stdout{ data, offset }   # offset = bytes emitted on this fd before this chunk
       ring drops oldest on overflow -> the next surviving chunk's offset jumps

host:  keep expected[fd]; per chunk:
         if offset > expected[fd]:  lost = offset - expected[fd]
                                    reset THIS stream's decoder   # only this one
         text = decode(data);  expected[fd] = offset + len(data)
```

why this is nicer than the empty marker:
- carries the byte count this message can't, and it's per-stream, so a stdout overflow no longer flushes the stderr decoder and corrupts an intact `-` (the issue flagged on the `Dropped` handler in exec.rs).
- survives reconnect: a resumed attach compares its last offset with no dependence on having seen a prior control event, so a dropped/late marker can't hide a gap. a data plane over tunnels reconnects often.
- one scalar folds into the existing messages -> no third oneof variant for every consumer (host, runner, sdks) to handle.

tradeoff: readers must track a per-stream offset to notice a gap (vs just seeing an event), and offset base/reset-on-attach semantics need pinning. also orthogonal to the bigger retention issue: the ring is still kept per-exec with no eviction. a bounded per-session on-disk log would reuse the same offset contract (offset = file position) and remove both the marker and the leak.

BatmanByte: Thanks. Per-stream offsets are a useful foundation for a future resumable output protocol, but I do not think they are a complete replacement for `OutputDropped` in #1022.

The current manager has one shared-capacity ring. Consider stdout writing 1 KiB and then reaching EOF, while stderr later fills the ring and evicts that stdout entry. This can happen both before Attach and while an attached response is flow-controlled. If no later stdout chunk survives, there is no stdout offset jump for the client to observe. An initial retention watermark covers only the pre-Attach case; the live case still requires a per-stream gap/watermark event when the retention cursor passes the attachment cursor. That control-plane event serves the same essential purpose as `OutputDropped`.

`OutputDropped` reports the exact lost byte counts per stream, and the host resets only the decoder whose count is non-zero. It therefore handles the no-later-chunk case without requiring clients to maintain offset state. Reconnect is intentionally out of scope for this PR: Attach permits one non-resumable consumer, so offset does not yet unlock that benefit.

Separate stdout/stderr rings would avoid cross-stream eviction, but that is a larger policy change: it needs per-stream capacity rules, bounded/split chunks, and a defined merge order for replay. It also does not by itself define resume semantics or terminal retention.

I will keep the shared ring plus `OutputDropped` for #1022, and treat per-stream offsets as a follow-up protocol design when we add resumable Attach or persisted output. The completed-execution retention issue remains orthogonal and needs the explicit output lifecycle follow-up.

DorianZheng: fair on the no-later-chunk case — a per-chunk offset alone can't see loss on a stream that already hit EOF before its bytes were evicted. but that case is coverable without a control event and without a persisted log:

- put a per-fd byte `offset` on `Stdout`/`Stderr` (bytes emitted on that fd before this chunk).
- carry the fd's final `total_bytes` on its end-of-stream frame.

the host then derives loss per stream, and resets only the stream that lost bytes:
```
mid-stream:  chunk.offset > expected[fd]  -> lost = offset - expected[fd];  reset THIS decoder
tail (EOF):  total_bytes  > received[fd]  -> lost = total - received;       reset THIS decoder
```
your stdout-writes-then-EOF-then-evicted example is caught by the EOF total (received 0, total 1000 -> 1000 lost), so it needs neither `dropped` nor a file — the shared ring stays as-is.

honest tradeoffs: the consumer becomes stateful (tracks a per-fd offset) instead of just reacting to an event; and the final count still rides the end-of-stream frame, so it's a signal folded onto an existing frame rather than a mid-stream oneof variant — not literally zero signaling. the upside is it keeps the per-stream byte counts and selective decoder reset you already added, drops the `dropped` variant, stays in-memory, and the same offset unlocks resumable attach later for free.

retention/eviction lifecycle stays a separate concern either way.

BatmanByte:
@codex review

src/shared/proto/boxlite/v1/service.proto
BatmanByte: Agreed. I updated the Attach contract along these lines: normal stdout/stderr frames carry a per-stream absolute `offset`; each stream emits an empty terminal frame whose `offset` and `total_bytes` equal that stream’s final byte count; and the unpublished `OutputDropped` variant has been removed.

The host now tracks `expected_offset` independently for stdout and stderr. An offset jump flushes only that stream’s decoder, and the terminal `total_bytes` detects the no-later-chunk case after eviction. The tests cover a stream ending before its sibling, data fully evicted before Attach, an offset gap, and keeping an intact sibling’s split UTF-8 sequence intact.

Attach remains a single non-resumable consumer in this PR, so this does not claim reconnect support, but the per-stream offsets leave that path open without another wire change. Thanks for pushing on the EOF case; that was the missing piece in my earlier assessment.

src/guest/src/service/exec/output.rs
```diff
@@ -0,0 +1,392 @@
...
+        let sequence = state.next_sequence;
+        state.next_sequence += 1;
+        let stream = state.stream_mut(source);
+        stream.enabled = true;
+        let offset = stream.total_bytes;
+        stream.total_bytes += byte_len as u64;
+        stream.last_sequence = Some(sequence);
+        let output = data_output(source, data, offset);
+
+        while state.buffered_bytes + byte_len > BUFFER_CAPACITY_BYTES {
```
DorianZheng, line 234: eviction here is unconditional — it never checks whether a consumer is attached. an attached, actively-reading consumer that falls behind a >1 MiB burst loses the evicted bytes: replay skips the hole via `next_sequence = oldest_sequence` (:132). the `mpsc(100)` in `ExecutionState::attach` sits downstream of replay (ring -> grpc) and backpressures only that, so the fd -> ring drain, and thus the producing process, is never slowed.

src/guest/src/service/exec/exec_handle.rs
```diff
@@ -35,44 +34,79 @@ impl ExecStdin {
...
 
-        // Read chunks as they arrive (works for both PTY and pipes)
         let stream = stream! {
             let mut buf = [0u8; 1024];
             loop {
-                match reader.read(&mut buf).await {
+                match reader.async_io(Interest::READABLE, |fd| read_fd(fd, &mut buf)).await {
                     Ok(0) => break,  // EOF
                     Ok(n) => yield buf[..n].to_vec(),
                     Err(_) => break,
```
DorianZheng: any non-`WouldBlock` read error breaks the loop and is indistinguishable from clean eof. the end-frame then reports the truncated length as the total, so the host sees no offset gap and treats a truncated stream as complete. (the stdin `write_all` above turns a transient error into a permanent forward failure the same way.)

src/boxlite/src/portal/interfaces/exec.rs
```diff
@@ -404,20 +405,39 @@ impl ExecProtocol {
...
-    fn route_output(output: ExecOutput, stdout: &mut DecodedStream, stderr: &mut DecodedStream) {
+    fn route_output(output: ExecOutput, stdout: &mut OutputTracker, stderr: &mut OutputTracker) {
         match output.event {
             Some(exec_output::Event::Stdout(chunk)) => {
                 tracing::trace!(len = chunk.data.len(), "Received exec stdout");
-                stdout.send_bytes(chunk.data);
+                if let Some(lost_bytes) =
+                    stdout.receive(chunk.offset, chunk.data, chunk.total_bytes)
+                {
+                    Self::report_gap(stderr, "stdout", lost_bytes);
```
DorianZheng: a stdout gap is reported on the stderr channel here (`report_gap(stderr, "stdout", ...)`). a consumer capturing stdout alone loses data with no signal, and a consumer parsing stderr gets `[boxlite] ...` spliced into real stderr. the stderr-gap branch just below is correct; only the stdout branch is cross-wired.

src/boxlite/src/portal/interfaces/exec.rs
```diff
@@ -404,20 +405,39 @@ impl ExecProtocol {
...
     }
 
+    fn report_gap(stderr: &mut OutputTracker, source: &str, lost_bytes: u64) {
+        tracing::warn!(
+            source,
+            lost_bytes,
+            "Guest output buffer dropped older output"
+        );
+        let _ = stderr.stream.tx.send(format!(
+            "[boxlite] {source} output dropped {lost_bytes} bytes\n"
```
DorianZheng: trailing bare `\n`: in raw/pty mode (opost cleared) this advances a row without a carriage return — staircase + column desync in a tui. `\r\n`, or routing the notice through the opost-aware path, avoids it.

src/boxlite/src/portal/interfaces/exec.rs
```diff
@@ -719,6 +739,78 @@ impl DecodedStream {
...
+            );
+            return None;
+        }
+
+        let lost_bytes = offset - self.expected_offset;
+        if lost_bytes > 0 {
+            self.stream.flush();
+        }
+        self.expected_offset = offset + data.len() as u64;
+        self.stream.send_bytes(data);
```
DorianZheng: `send_bytes(data)` forwards the surviving post-gap bytes before the caller runs `report_gap`, so the "dropped N bytes" marker trails the recovered output and reads as a mid-stream loss rather than one at the gap point.

src/guest/src/service/exec/output.rs
```diff
@@ -0,0 +1,392 @@
...
+        }
+    }
+
+    pub(crate) async fn attach(&self) -> Result<AttachStream, Status> {
+        {
+            let mut state = self.inner.lock().await;
+            if state.attached {
+                return Err(Status::already_exists("Already attached"));
+            }
+            state.attached = true;
```
DorianZheng, line 120: `attached` is set once and never cleared, so if the single attach stream drops (a transient disconnect) it can't reconnect — `attach` returns AlreadyAttached for the rest of the execution — even though the ring still holds recent output and the new `offset`/`total_bytes` fields are exactly what resume-from-offset needs. as written those wire fields aren't used for reconnect, so either they're dead weight or this latch blocks the feature they're for.

src/guest/src/service/exec/state.rs
```diff
@@ -288,88 +270,36 @@ impl ExecutionState {
     }
 
     /// Attach to execution output.
-    ///
-    /// Takes stdout/stderr from handle and starts forwarding tasks.
-    /// Returns stream of output chunks.
     pub async fn attach(
         &self,
-        exec_id: &str,
+        _exec_id: &str,
```
DorianZheng: `_exec_id` is unused and the per-stream `info!(... forwarding task ended)` logs that existed before are gone, so the attach path no longer carries the execution id or logs task lifecycle — harder to correlate a stuck or silently-dropping attach against the wait/exit records.

src/guest/src/service/exec/output.rs
```diff
@@ -0,0 +1,392 @@
...
+                    } else if state.stderr.ready_to_end(next_sequence) && !stderr_end_sent {
+                        stderr_end_sent = true;
+                        Next::Item(end_output(OutputSource::Stderr, state.stderr.total_bytes))
+                    } else if next_sequence < state.next_sequence {
+                        let index = (next_sequence - state.oldest_sequence) as usize;
+                        let output = state
+                            .entries
+                            .get(index)
+                            .expect("ring sequence must exist")
+                            .output
```
DorianZheng, line 162: two things on this replay step:
- it deep-clones the `ExecOutput` (its `Vec<u8>`) while holding the mutex, and `push` keeps buffering into the ring after a consumer attaches, so a live stream is copied twice for its whole life. `Arc<[u8]>`/`Bytes` payloads would make this a refcount bump.
- the `Next::Wait` exit at :165 (`updates.changed().await.is_err()`) is unreachable: this closure holds a `manager` clone, hence a `watch::Sender`, so `changed()` never errors. `Next::Done` is the only way out, and it needs a drain to finish naturally; if drains are aborted (shutdown) without the attach consumer also being torn down, this stream hangs. today the release path aborts the consumer too, so it's latent.

DorianZheng (commented):
the earlier compile break — fallible `ExecHandle::new` against the infallible `#[cfg(test)]` callers — is resolved on this push (`.expect(...)` at the four sites), so that's cleared.

output buffering is the piece worth a deliberate decision. draining at exec-creation into a bounded ring is right for the no-consumer case, but eviction is unconditional (inline at output.rs:221), so a consumer that is attached and reading, yet slower than a >1 MiB burst, silently loses the evicted bytes — the downstream `mpsc(100)` backpressures replay, not the fd drain, so the producer is never slowed. previously a full bounded channel blocked the reader, and an attached consumer couldn't lose data. the call: backpressure while a consumer is attached and caught up, or accept a lossy attach with a reliable gap notice.

the same evicting stream feeds ssh: `spawn_output_pump` (src/guest/src/service/ssh/bridge.rs:597, outside this diff) forwards stdout/stderr raw with no offset/gap handling, so an interactive session faster than its client loses bytes with no notice at all.

remaining points are inline. cleanup only, no behavior change: the `drain_tasks` `StdMutex<Vec<JoinHandle>>` (output.rs:20), the near-duplicate `spawn_stdout`/`spawn_stderr` (:176/:187, `drain` is already generic), and the dead `stream.enabled = true` in `push` (:215, a drain only exists for an already-enabled stream).

src/guest/src/service/exec/exec_handle.rs
```diff
@@ -35,44 +34,79 @@ impl ExecStdin {
...
+    Ok(())
+}
+
+fn read_fd(fd: &OwnedFd, buffer: &mut [u8]) -> io::Result<usize> {
+    nix::unistd::read(fd.as_raw_fd(), buffer).map_err(Into::into)
+}
+
+fn write_fd(fd: &OwnedFd, buffer: &[u8]) -> io::Result<usize> {
+    nix::unistd::write(fd, buffer).map_err(Into::into)
+}
```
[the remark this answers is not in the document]
BatmanByte: Fixed in dc31d495. read_fd/write_fd now retry EINTR. A real output read error is propagated through OutputManager, ExecutionState, and Attach as gRPC INTERNAL instead of being converted into a successful EOF. PTY-master EIO remains the one explicit EOF case.

src/guest/src/service/exec/output.rs
BatmanByte: Fixed in dc31d495. The 1 MiB ring remains a deliberate bounded-retention policy, so loss is allowed but no longer silent: every gRPC stdout/stderr frame carries that stream absolute offset, and its empty terminal frame carries final total_bytes. Consumers detect either a mid-stream offset jump or a tail gap at EOF. SSH tracks the same counters and emits the exact per-stream marker on the affected SSH output stream.

src/boxlite/src/portal/interfaces/exec.rs
BatmanByte: Fixed in dc31d495. Host gap markers now go to the affected decoded stream, so stdout-only consumers observe stdout loss and intact stderr is not modified. The same rule is applied to the in-process SSH bridge.

src/boxlite/src/portal/interfaces/exec.rs
BatmanByte: Fixed in dc31d495. OutputTracker now separates validation from delivery: it flushes the affected decoder, emits the gap marker, then forwards surviving bytes. Markers use CRLF; the SSH bridge follows the same ordering and line ending.

src/guest/src/service/exec/state.rs
BatmanByte: Fixed in dc31d495. ExecutionState::attach now retains exec_id and logs execution output forwarding ended; OutputManager also logs an output-reader failure with its stream. The duplicate stdout/stderr drain spawners were consolidated.

src/guest/src/service/exec/output.rs
BatmanByte: Partially fixed in dc31d495: the unreachable watch::Receiver error branch is removed; the manager owner makes sender closure impossible. I am not changing the Vec clone here: avoiding it needs a Bytes/Arc payload change across the generated protobuf and all Attach consumers, which is a separate wire/API change rather than a local cleanup.

src/guest/src/service/exec/exec_handle.rs
BatmanByte: Fixed in dc31d495. Non-PTY read failures now become gRPC INTERNAL through Attach, so no terminal total_bytes can certify a truncated stream. The PTY EIO-on-slave-close case is handled separately as its documented EOF condition.

src/guest/src/service/exec/output.rs
```diff
@@ -0,0 +1,423 @@
...
+                        } else if next_sequence < state.next_sequence {
+                            let index = (next_sequence - state.oldest_sequence) as usize;
+                            let output = state
+                                .entries
+                                .get(index)
+                                .expect("ring sequence must exist")
+                                .output
+                                .clone();
+                            next_sequence += 1;
+                            Next::Item(output)
```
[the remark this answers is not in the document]
BatmanByte: Fixed in b5d053ee. ReaderFailure now records the next internal sequence at failure time. Attach replays every retained entry before that boundary, then yields INTERNAL; output captured after the failure is intentionally not presented as a complete stream. The regression test asserts buffered stdout is delivered before the error.

src/boxlite/src/portal/interfaces/exec.rs
BatmanByte: Fixed in dc31d495. Gap markers now use CRLF in both the Host and SSH output paths, so raw PTY consumers do not advance without returning to column zero.

src/guest/src/service/exec/output.rs
BatmanByte: Intentionally deferred. #1022 keeps Attach as one non-resumable consumer, so clearing the latch alone would permit a second stream without defining the resume cursor, replay start point, duplicate policy, or terminal retention. The new per-stream offset/total fields make that follow-up possible, but resumable Attach needs a request cursor plus tests for disconnect, replay, expiry, and concurrent reconnect; it should land as a separate protocol design.

DorianZheng:
https://github.com/boxlite-ai/boxlite/actions/runs/30514470192/job/90781243192?pr=1022

failed, please take a look and fix it before merge

PR #1029, opened 2026-07-25
feat(ssh): embed real SSH server in boxlite-guest via direct tunnel

nieyy:
## Summary
- Embeds a russh-based SSH server in boxlite-guest, exposed through the existing direct VM tunnel, with a typed control path across guest, Runtime, C/Go SDKs, and the Runner.
- Adds Hosted API app-scoped access grants and short-lived temporary SSH credentials, with dashboard wiring, OpenAPI/client regen, and pre-deploy migrations.
- `BOXLITE_GUEST_SSH_ENABLED` (guest) and `SSH_ISSUANCE_ENABLED` (API) both default to disabled, requiring explicit opt-in per environment for controlled rollout.
- Adds an E2E lifecycle test, an SSH-specific load/security test script (concurrency, auth-rejection burst, connect/disconnect churn), a rollback runbook, and an observability signal-mapping guide (explicitly flagged NOT_IMPLEMENTED for dashboards/alerts, pending an assigned owner and target system).

Design doc (external source of truth, not checked into this repo):
https://github.com/nieyy/boxlite_integration_test/blob/main/docs/designs/2026-07-23-boxlite-direct-tunnel-real-ssh-design-zh.md

## Test plan
- [x] `cargo test -p boxlite --lib guest_ssh_enablement_tests` (3/3 pass)
- [x] `cargo clippy -p boxlite --all-targets -- -D warnings` (clean)
- [x] `cargo nextest` for `boxlite-guest` `ssh` module in Linux Docker (real behavior + negative-auth tests)
- [x] `npx jest configuration.spec.ts box-access-grant.service.spec.ts temporary-ssh-credential.service.spec.ts` (27/27 pass)
- [x] `tsc --noEmit`, eslint clean on touched TS files
- [x] `make guest`, `make clippy`, `make build:apps`
- [x] `ruff check` / `py_compile` on new Python scripts; load-test decision logic verified via monkeypatched dry run
- [ ] Live E2E (`scripts/test/e2e/cases/test_real_ssh.py`) against a deployed stack — **BLOCKED_EXTERNAL**, needs a built `boxlite` wheel + live API/Runner/VM stack
- [ ] Live SSH load test (`make test:stress:ssh`) — **BLOCKED_EXTERNAL**, needs a reachable box with SSH enabled + `proxytunnel`
- [ ] Staging rollout + rollback drill (`scripts/deploy/rollback-real-ssh.sh`) — **BLOCKED_EXTERNAL**, needs a staging deployment
- [ ] Dashboards/alerts — **NOT_IMPLEMENTED**, no owner or target system assigned yet

## Summary by CodeRabbit

* **New Features**
  * Added box-scoped access grants and one-time temporary SSH credentials (create/list/revoke), with connection details for endpoint, proxy command, host-key fingerprint, and known-hosts.
  * Updated API/client and dashboard UI to use the new access-grant → temporary-credential flow (private-key copy/download included).
  * Added rollout toggles for SSH issuance and enabling guest SSH listeners.
* **Bug Fixes**
  * Improved SSH public-key parsing/validation and strengthened guest process I/O handling.
  * Added more resilient SSH access-set application error mapping and best-effort cleanup during box destroy/start.
* **Tests & CI**
  * Added guest test routing plus real-SSH E2E/stress coverage; expanded Linux-only guest unit testing.


src/boxlite/src/litebox/init/tasks/vmm_spawn.rs
```diff
@@ -309,6 +309,11 @@ fn build_guest_entrypoint(
     builder.with_arg("--notify");
     builder.with_arg(&ready_notify_uri);
 
+    if let Some(ssh_listen_addr) = guest_ssh_listen_addr() {
+        builder.with_arg("--ssh-listen");
```
DorianZheng, line 313: why do we need this?

DorianZheng:
the guest side here is the valuable part — a russh server reached over the existing tunnel, host key persisted at `host_key.rs:32`, staged behind two default-off flags. none of the below touches that. it's about the interface wrapped around it.

line refs to this branch are against `7c0b2358`; refs describing current behaviour — `box.controller.ts:729/:772` and `box.service.ts:1496/:1501/:1512/:1519` — are against `main` at `53c7aea2`, since this pr deletes those lines.

## 1. this replaces a shipped endpoint rather than adding one

`main` already serves the same user story:

```
main    POST   /box/:boxIdOrName/ssh-access?expiresInMinutes=60   box.controller.ts:729 -> box.service.ts:1496
        DELETE /box/:boxIdOrName/ssh-access?token=…               box.controller.ts:772 -> box.service.ts:1519
        -> SshAccessDto { id, boxId, token, expiresAt, createdAt, updatedAt, sshCommand }
```

this pr deletes both (−128 lines across `box.controller.ts` + `box.service.ts`) and does two different things to them:

| path | before | after |
|---|---|---|
| `POST :boxIdOrName/ssh-access` | `createSshAccess` | rebound → `createTemporarySshCredential` (`temporary-ssh-credential.controller.ts:39`) |
| `DELETE :boxIdOrName/ssh-access` | `revokeSshAccess` | **no route** — revoke moved to `DELETE :boxIdOrName/ssh-access/:credentialId` (`:101`) |

`box.controller.ts` on this branch has zero `@Delete` routes, so an existing caller's revoke 404s rather than getting a changed contract. create keeps its url but changes shape: operation renamed, request gains required `grantId` + `publicKey`, response drops `token`. every generated client method for both renames.

the comment left at `box.controller.ts:729` says `DELETE /:boxIdOrName/ssh-access` "now lives in TemporarySshCredentialController" — the route table doesn't back that.

worth stating explicitly whether retiring `createSshAccess`/`revokeSshAccess` is intended. `validateSshAccess` and the `SshAccess` rows are deliberately preserved "for rollback isolation", which reads like the gateway is meant to keep working — if so the two flows now disagree about what `/ssh-access` means.

## 2. the runtime facade is wired to the opposite backend from the one that can serve it

`litebox/ssh.rs:65` documents the handle as "internal control facade … used by the Runner … the public sdk credential-issuance helper is a separate Hosted API client and must not call this facade". but it ships in the published c header and go sdk:

```
sdks/c/include/boxlite.h    boxlite_box_ssh, _ssh_replace_access_set, _ssh_status,
                            _ssh_free, _ssh_status_free  + 4 types
sdks/go/ssh.go              box.SSH().ReplaceAccessSet(ctx, gen, entries) / .Status(ctx)
rest/litebox.rs:478         Err(Unsupported("… not available over REST"))
sdks/python, sdks/node      nothing
```

the operation a user wants — *issue me a credential* — only the hosted api can perform, and that's the rest backend. the operation exposed — *replace the guest's access set* — only the runner should perform, and it's published to everyone. inverting these would let python and node get the feature and take the c/go control surface back out of the abi.

## 3. the push model is what forces the rest of the machinery

pushing per-credential public keys into the guest is what creates the generation counter, `BoxSshReconciliationService`, the `ACTIVE -> REVOKING -> REVOKED` cascade, and the whole `gRPC -> Rust -> C ABI -> Go -> Runner` path. everything downstream of `service Ssh` (`service.proto:84`) and `PUT /:boxId/ssh/access-set` (`runner/pkg/api/server.go:162`) exists to carry that one push.

prior art on the same problem — short-lived ssh into ephemeral compute:

| | user-facing surface | guest-side auth | per-credential push? |
|---|---|---|---|
| daytona (upstream this was ported from) | `POST /sandbox/{id}/ssh-access?expiresInMinutes=60` → `{token}` | gateway terminates ssh | no |
| e2b | `ssh -o 'ProxyCommand=websocat … wss://8081-%h.e2b.app' user@<id>` | template's own sshd | no |
| coder | `coder ssh <ws>` | `agent/agentssh/agentssh.go:303` `NoClientAuth: true` | no |
| vault ssh | `POST /ssh/sign/:name {public_key, ttl, valid_principals}` → `{signed_key}` | `TrustedUserCAKeys` | no — ca |
| ec2 instance connect | `SendSSHPublicKey` | sshd | yes — "the key remains for 60 seconds", no revoke api, no version counter |
| this pr | 6 endpoints / 2 resources / appKey + byo pubkey | russh `auth_publickey` over a pushed set | yes, + generation counter + reconciler |

the only push-model peer has neither a revoke api nor a version counter, because a 60s ttl removes the need for both. the ca model removes the push outright.

## proposal

**guest trusts a ca instead of a pushed set.** the guest already takes its ssh config as a spawn arg (`vmm_spawn.rs:313`, `--ssh-listen`); add one more on the same path:

```
--ssh-listen 0.0.0.0:22  --ssh-ca-key "ssh-ed25519 AAAA…"
```

russh 0.62 already has the hook — `Handler::auth_openssh_certificate(user, cert)`. auth becomes a pure function of the cert: signature key == ca, cert type user, principals contains root, valid_after <= now < valid_before.

that deletes `service Ssh` + its 4 messages + `SshIdentityStatus`, `AccessState`/`AccessSnapshot`/`ReplaceError`, the generation counter, the `box_ssh_access_generation` table, the reconciler, the grant state machine, the runner endpoint, and the whole c/go/rust control surface. no rpc, no reconcile loop, no api↔guest ordering hazard.

**keep `/ssh-access` and extend it** rather than renaming operations:

```
POST   /box/:boxIdOrName/ssh-access?expiresInMinutes=60  { publicKey? }  -> SshAccess   (exists; extend)
DELETE /box/:boxIdOrName/ssh-access?token=…                                             (exists; keep)
GET    /box/:boxIdOrName/ssh-access                                      -> SshAccess[] (new)
```

`SshAccess` keeps `id`/`boxId`/`token`/`expiresAt`/`sshCommand` and gains `certificate`, `privateKey` (only when `publicKey` omitted, returned once), `fingerprint`, `host`, `knownHosts`, `proxyCommand`. `token` stays as the revoke handle; it stops being the authenticator once the guest checks certs.

optional `publicKey` is a deliberate divergence from vault: for a short-lived credential to a disposable microvm, letting the api mint the keypair keeps the existing one-call-one-artifact ergonomics. callers who care supply their own.

**`BoxAccessGrant` dropped.** letting a non-org-credentialed app mint access is a delegation feature, not an ssh feature — a box-scoped scoped token belongs in the auth module, built once, generically. inside this pr it costs 3 endpoints, 2 enums, a status machine, a second auth guard, and a one-time secret users must store. is there a caller for it today, or is it anticipatory?

**sdk becomes user-facing in all four languages**, on the rest backend where it can actually be served:

```
LiteBox::ssh_access() -> { create, list, revoke }
  rest/litebox.rs   implemented
  local backend     Unsupported     (no proxy, no ca in a local runtime)
box.ssh_access.create()  ·  box.sshAccess.create()  ·  box.SSHAccess().Create()
```

net against `main`, excluding `GET /box/ssh-access/validate` which all three keep:

| | main | this pr | proposal |
|---|---|---|---|
| user-facing ssh endpoints | 2 | 6 | 3 |
| …removed from main | — | 2 | 0 |
| renamed client operations | — | 2 | 0 |
| api entities / enums | 1 / 0 | 4 / 4 | 1 / 1 |
| grpc services / messages / enums | 0 | 1 / 4 / 1 | 0 |
| public c functions / types | 0 | 5 / 4 | 0 |
| runner endpoints | 0 | 1 | 0 |
| reconciler | none | 1 | none |
| sdks covered | — | rust, go, c (internal) | rust, python, node, go |

the honest cost: revocation becomes ttl-bounded. `DELETE` stops renewal and blocks new sessions, but a live session on an issued cert survives to `valid_before`. vault accepts exactly this, and the 300s default / 3600s ceiling already chosen here bounds it. if immediate per-credential revoke is a hard requirement, a krl — a list of revoked cert serials — is append-only and O(revoked), so it still needs no generation counter, no full-set replace and no reconciler.

two things to decide either way: per-box ca (blast radius one box, rotation free on restart) vs one org ca with `valid_principals` scoped to the box id; and whether c needs ssh at all.

## smaller, independent of the above

**ssh endpoint drops the region override.** every sibling hostname derivation in `box.service.ts` consults a per-region override before returning — `:697` `getPortPreviewUrl` and `:721` `getNetworkTunnelUrl` and `:756` `getSignedPortPreviewUrl` all read `region.proxyUrl`, and the `createSshAccess` this pr deletes read `region.sshGatewayUrl` (`:1512`). `temporary-ssh-credential.service.ts:207` derives `${SSH_ENDPOINT_PORT}-${encodeDirectPreviewBoxId(box.id)}.${proxyDomain}` from config alone — the file has no `region` reference. `region.proxyUrl` is settable (`region.service.ts:88,273`) though the seeded default region leaves it null (`app.service.ts:76-84`), so this is scoped to multi-region / custom-proxy deployments. no routing check run — flagging it as a divergence from every sibling, not a measured failure. `getNetworkTunnelUrl(box, 22)` already does this correctly.

**`createSshAccess` revokes everything on every call** (`box.service.ts:1501`, `revokeSshAccess(box.id)` with no token → `delete({ boxId })`). that has to go if more than one credential can be live at once, which the list endpoint implies.

**path naming**: `/ssh-access` over `/ssh` — every sibling in that controller is a compound noun (`preview-url`, `signed-preview-url`, `toolbox-proxy-url`, `last-activity`), `DELETE /box/:id/ssh` would read as "disable ssh on this box", and leaving `/ssh` unclaimed keeps it free for box-level ssh config later.

nieyy:
Thanks, this is a solid point. The CA-based approach makes sense but it's a big enough change that I'd rather do it as its own PR instead of reworking this one.

PR #1056, opened 2026-07-27
feat(rest): expose S3 managed volumes

G4614:
~~Expose existing S3 managed volumes through the BoxLite REST API and carry volume mounts into REST-created boxes.~~

**Rebuilt on top of main** now that this PR's backend content has been split out and merged separately:
- #1191 — volume REST CRUD (`/v1/volumes` create/list/get/delete)
- #1192 — REST box-create volume mount wiring (`VolumeSpec.source`)
- (follow-up, still open) #1196 — makes `create` return immediately instead of blocking, adds client-side `wait_until_ready`

What's left here is the remaining, still-unmerged piece: **CLI + Node/Python SDK support for attaching a managed volume to a box**, on top of the server-side contract #1191/#1192 already shipped.

- CLI: `--mount src=volume://<id>,target=<path>` on `create`/`run`
- Node/Python SDKs: accept a scheme-qualified volume source (`volume://<id>`) alongside the existing host-path volume option

Test plan:
- `cargo build -p boxlite-cli` / `cargo test -p boxlite-cli --bin boxlite volume` — 30 tests, all pass
- `cargo check -p boxlite-node` / `cargo check -p boxlite-python` — clean
- `cargo fmt --all -- --check` — clean
- Fixed a pre-existing test bug found during verification: `test_volume_flags_managed_volume_rejects_missing_id`'s fixture never reached the code path it meant to exercise (see commit message for detail)

### DIFF
--- a/sdks/node/src/options.rs
+++ b/sdks/node/src/options.rs
@@ -259,8 +259,11 @@ pub struct JsEnvVar {
 #[napi(object)]
 #[derive(Clone, Debug)]
 pub struct JsVolumeSpec {
-    /// Path on host machine
-    pub host_path: String,
+    /// Scheme-qualified source such as `volume://vol_123`.
+    pub source: Option<String>,
+
+    /// Path on host machine for local host-path mounts.
+    pub host_path: Option<String>,
 
     /// Path inside container
     pub guest_path: String,
@@ -269,13 +272,32 @@ pub struct JsVolumeSpec {
     pub read_only: Option<bool>,
 }
 
-impl From<JsVolumeSpec> for VolumeSpec {
-    fn from(v: JsVolumeSpec) -> Self {
-        VolumeSpec {
-            host_path: v.host_path,
+impl TryFrom<JsVolumeSpec> for VolumeSpec {
+    type Error = boxlite_shared::errors::BoxliteError;
+
+    fn try_from(v: JsVolumeSpec) -> Result<Self, Self::Error> {
+        let host_path = match (v.source, v.host_path) {
+            (Some(source), _) => {
+                if !source.starts_with("volume://") {
+                    return Err(Self::Error::InvalidArgument(
+                        "volume source must use the volume:// scheme".into(),
+                    ));
+                }
+                source
+            }
+            (None, Some(host_path)) => host_path,
+            (None, None) => {
+                return Err(Self::Error::InvalidArgument(
+                    "volume source or hostPath is required".into(),
+                ));
+            }
+        };
+
+        Ok(VolumeSpec {
+            host_path,
             guest_path: v.guest_path,
             read_only: v.read_only.unwrap_or(false),
-        }
+        })
     }
 }
 
@@ -389,8 +411,8 @@ impl TryFrom<JsBoxOptions> for BoxOptions {
             .volumes
             .unwrap_or_default()
             .into_iter()
-            .map(VolumeSpec::from)
-            .collect();
+            .map(VolumeSpec::try_from)
+            .collect::<Result<Vec<_>, _>>()?;
 
         // Convert network spec
         let network = match js_opts.network {
@@ -694,6 +716,41 @@ mod tests {
         for registry in cases {
             assert!(js_image_registry_into_core(registry).is_err());
         }
+    }
+
+    #[test]
+    fn js_volume_source_requires_volume_scheme() {
+        let volume = VolumeSpec::try_from(JsVolumeSpec {
+            source: Some("volume://vol_123".into()),
+            host_path: None,
+            guest_path: "/data".into(),
+            read_only: None,
+        })
+        .unwrap();
+
+        assert_eq!(volume.host_path, "volume://vol_123");
+        assert_eq!(volume.guest_path, "/data");
+        assert!(!volume.read_only);
+
+        let err = VolumeSpec::try_from(JsVolumeSpec {
+            source: Some("vol_123".into()),
+            host_path: None,
+            guest_path: "/data".into(),
+            read_only: None,
+        })
+        .unwrap_err();
+
+        assert!(err.to_string().contains("volume:// scheme"));
+
+        let err = VolumeSpec::try_from(JsVolumeSpec {
+            source: None,
+            host_path: None,
+            guest_path: "/data".into(),
+            read_only: None,
+        })
+        .unwrap_err();
+
+        assert!(err.to_string().contains("source or hostPath is required"));
     }
 
     #[test]
--- a/sdks/node/lib/native-contracts.ts
+++ b/sdks/node/lib/native-contracts.ts
@@ -77,7 +77,29 @@ export interface JsEnvVar {
   value: string;
 }
 
-export interface JsVolumeSpec {
+export type JsVolumeSpec =
+  | {
+      /** Scheme-qualified managed volume source, e.g. volume://vol_123. */
+      source: string;
+      guestPath: string;
+      readOnly?: boolean;
+    }
+  | {
+      /** Path on the local host for host-path mounts. */
+      hostPath: string;
+      guestPath: string;
+      readOnly?: boolean;
+    };
+
+export interface JsVolumeSourceSpec {
+  /** Scheme-qualified managed volume source, e.g. volume://vol_123. */
+  source: string;
+  guestPath: string;
+  readOnly?: boolean;
+}
+
+export interface JsHostPathVolumeSpec {
+  /** Path on the local host for host-path mounts. */
   hostPath: string;
   guestPath: string;
   readOnly?: boolean;
--- a/sdks/node/lib/simplebox.ts
+++ b/sdks/node/lib/simplebox.ts
@@ -15,6 +15,7 @@ import type {
   JsBox,
   JsBoxInfo,
   JsBoxOptions,
+  JsVolumeSpec,
   NativeBoxConnection,
   NativeBoxTunnel,
   JsExecStderr,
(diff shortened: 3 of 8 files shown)


apps/api/src/boxlite-rest/boxlite-volume.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+  @Get()
+  @RequiredOrganizationResourcePermissions([OrganizationResourcePermission.READ_VOLUMES])
+  async list(@AuthContext() authContext: OrganizationAuthContext): Promise<{ volumes: RestVolumeResponse[] }> {
+    const volumes = await this.volumeService.findAll(authContext.organizationId)
+    return { volumes: volumes.map((volume) => this.toResponse(volume)) }
+  }
+
+  @Get(':volumeId')
+  @RequiredOrganizationResourcePermissions([OrganizationResourcePermission.READ_VOLUMES])
+  @UseGuards(VolumeAccessGuard)
```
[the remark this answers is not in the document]
G4614: `READ_VOLUMES` is present in `OrganizationResourcePermission`, and the existing `VolumeController` uses the same permission for list/get routes. Keeping these decorators preserves the established organization-level authorization; `VolumeAccessGuard` remains on the per-volume get route.

apps/api/src/boxlite-rest/boxlite-volume.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+  async remove(@Param('volumeId') volumeId: string, @Query('force') force?: string): Promise<void> {
+    try {
+      await this.volumeService.delete(volumeId)
+    } catch (error) {
+      if (force === 'true' && error instanceof NotFoundException) {
+        return
+      }
+      throw error
+    }
+  }
```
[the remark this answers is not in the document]
G4614: Fixed: `VolumeAccessGuard` now permits a missing volume only for `DELETE ?force=true`, and `VolumeService.delete(..., true)` treats missing/already-deleting volumes idempotently. Covered by guard, controller, service, and local MinIO control-plane tests.

apps/api/src/boxlite-rest/boxlite-volume.controller.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+        {
+          id: volume.id,
+          created_at: createdAt.toISOString(),
+          size_bytes: 0,
+        },
+      ],
+    })
+    expect(volumeService.findAll).toHaveBeenCalledWith('org-1')
+  })
+})
```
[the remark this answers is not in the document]
G4614: Added get/remove/forced-delete controller coverage plus dedicated `VolumeAccessGuard` and `VolumeService.delete` tests. The full API suite passes: 57 suites, 279 tests.

apps/api/src/boxlite-rest/boxlite-volume.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  private toResponse(volume: Awaited<ReturnType<VolumeService['findOne']>>): RestVolumeResponse {
+    return {
+      id: volume.id,
+      created_at: volume.createdAt.toISOString(),
+      size_bytes: 0,
+    }
```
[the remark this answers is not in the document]
G4614: Fixed in `5f34db69`: the REST API now omits unknown `size_bytes` instead of reporting a misleading zero. The Rust REST client already deserializes the field as optional, so SDK consumers receive `None` until real usage accounting exists.

apps/api/src/boxlite-rest/boxlite-volume.controller.ts
```diff
@@ -0,0 +1,58 @@
...
+  @Delete(':volumeId')
+  @HttpCode(204)
+  @RequiredOrganizationResourcePermissions([OrganizationResourcePermission.DELETE_VOLUMES])
+  @UseGuards(VolumeAccessGuard)
+  async remove(@Param('volumeId') volumeId: string, @Query('force') force?: string): Promise<void> {
+    await this.volumeService.delete(volumeId, force === 'true')
+  }
+
+  private toResponse(volume: Awaited<ReturnType<VolumeService['findOne']>>): RestVolumeResponse {
+    return {
```
mamamiyear, line 71: I think the 'state' field should be included in list view, otherwise, it's hard to check state for outer caller.
And, if name is specificed, the volume.name should be included too.

apps/api/src/boxlite-rest/boxlite-volume.controller.ts
```diff
@@ -0,0 +1,58 @@
...
+@UseGuards(CombinedAuthGuard, OrganizationResourceActionGuard)
+export class BoxliteVolumeController {
+  constructor(private readonly volumeService: VolumeService) {}
+
+  @Post()
+  @RequiredOrganizationResourcePermissions([OrganizationResourcePermission.WRITE_VOLUMES])
+  async create(@AuthContext() authContext: OrganizationAuthContext): Promise<RestVolumeResponse> {
+    const volume = await this.volumeService.create(authContext.organization, {})
+    return this.toResponse(await this.volumeService.waitForReady(volume.id, 30))
+  }
```
mamamiyear, line 37: Is it to rollback when create failed? Beacuse the api means that to create volume and it should be success, if not, it should be rollback which try to delete the bucket created just now.

apps/api/src/boxlite-rest/boxlite-volume.controller.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  @Get()
+  @RequiredOrganizationResourcePermissions([OrganizationResourcePermission.READ_VOLUMES])
+  async list(@AuthContext() authContext: OrganizationAuthContext): Promise<{ volumes: RestVolumeResponse[] }> {
+    const volumes = await this.volumeService.findAll(authContext.organizationId)
+    return { volumes: volumes.map((volume) => this.toResponse(volume)) }
+  }
```
mamamiyear, line 44: Consider the api to support filter by 'state', and add pagination.

apps/api/src/boxlite-rest/boxlite-volume.controller.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  @Get(':volumeId')
+  @RequiredOrganizationResourcePermissions([OrganizationResourcePermission.READ_VOLUMES])
+  @UseGuards(VolumeAccessGuard)
+  async get(@Param('volumeId') volumeId: string): Promise<RestVolumeResponse> {
+    return this.toResponse(await this.volumeService.findOne(volumeId))
+  }
```
mamamiyear, line 51: GET response view should include more details of volume, not only same as list view.

sdks/node/src/options.rs
```diff
@@ -269,13 +272,28 @@ pub struct JsVolumeSpec {
...
+            (Some(source), _) => {
+                if !source.starts_with("volume://") {
+                    return Err(Self::Error::InvalidArgument(
+                        "volume source must use the volume:// scheme".into(),
+                    ));
+                }
+                source
+            }
+            (None, Some(host_path)) => host_path,
+            (None, None) => String::new(),
```
mamamiyear: Maybe here should throw error when  both`source` and `hostPath` are empty, like python sdk's behavior.

mamamiyear (commented):
需不需要考虑为已有 box 挂载 volume？

apps/api/src/boxlite-rest/boxlite-volume.controller.ts
G4614: 已补充：list 现在返回 name 和 state，外部可以直接判断 volume 状态。

apps/api/src/boxlite-rest/boxlite-volume.controller.ts
G4614: 这里保留现有异步语义：超时只表示等待窗口结束，后台创建可能仍会成功；此时立即强制删除会与 VolumeManager 的创建状态机竞态。已让响应暴露 state，调用方可继续查询并在终态后删除。

apps/api/src/boxlite-rest/boxlite-volume.controller.ts
G4614: 本次先补上了返回值中的 state，因此调用方可以判断每项状态。分页和服务端过滤会单独扩展，避免在这个打通 REST 链路的 PR 中改变 findAll 契约。

apps/api/src/boxlite-rest/boxlite-volume.controller.ts
G4614: GET 已改为详细视图，除 id/name/state/created_at 外还返回 updated_at、last_used_at 和 error_reason。

sdks/node/src/options.rs
G4614: fixed: when lack of `source` and  `hostPath`  at the same time ,Node SDK returns `InvalidArgument`, also test added

apps/api/src/boxlite-rest/boxlite-volume.controller.ts
mamamiyear: Accoding that design, the api should return when volume has been created (in state PENDING), not waitting for success and blocking request.

G4614: You're right, and I've made this change — fixed in G4614/boxlite@3664efb79, which lives in the split-out #1191 (this PR's volume-CRUD half; see #1191 for the rebase-and-split context). `create` now returns 202 immediately with `state=pending_create` instead of blocking up to 30s in `waitForReady`, matching what the pre-existing classic `VolumeController` already did. Callers can poll GET or subscribe to the existing `volume.state.updated` webhook (`apps/api/src/box/subscribers/volume.subscriber.ts`) instead of the request holding a connection open near typical gateway timeout windows.

DorianZheng (approved):

PR #1065, opened 2026-07-28
fix(hooks): read mtime via GNU stat -c first across preflight hooks

lilongen:
## Problem

The preflight hooks read a file's mtime via a BSD-first chain:
`stat -f '%m' … || stat -c '%Y' …`. On GNU/uutils, `stat -f` means
`--file-system`: the format and the file become two operands, the format
errors to stderr (swallowed) while the file's FS block prints to **stdout**
(not swallowed) and it exits 1 — so the `||` also runs `-c %Y`, and the mtime
variable captures the FS block **plus** the epoch. Under `set -u` that
concatenation crashed the Stop hook with `File: unbound variable` in the
`$((now - mtime))` arithmetic.

## Fix

Reorder every site to GNU `-c %Y` first (real BSD lacks `-c` and still falls
through to `-f`), and guard each mtime to digits so non-numeric `stat` output
can no longer crash the arithmetic. All 7 occurrences across 5 hooks:
`preflight-verdict-check`, `preflight-pr-review`, `preflight-commit-push` (×2),
`run-verdict-audit` (×2), `run-commit-push-audit`.

## Verification

- Red→green: old order → `File: unbound variable`; patched → numeric epoch,
  arithmetic succeeds.
- `bash -n` clean on all 5 hooks; grep confirms no BSD-first `stat -f '%m'` remains.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility across Linux, macOS, and BSD environments when checking file timestamps.
  * Prevented invalid timestamp values from disrupting freshness, expiration, and audit validation checks.
  * Ensured stale or malformed validation markers are handled safely and require renewed verification when necessary.

### DIFF
--- a/.claude/hooks/preflight-commit-push.sh
+++ b/.claude/hooks/preflight-commit-push.sh
@@ -140,7 +140,8 @@ valid_handoff_command_hash() {
   handoff_kind="$(jq -r '.command_kind // ""' "$handoff_file" 2>/dev/null || echo '')"
   handoff_diff_hash="$(jq -r '.diff_hash // ""' "$handoff_file" 2>/dev/null || echo '')"
   handoff_command_hash="$(jq -r '.command_hash // ""' "$handoff_file" 2>/dev/null || echo '')"
-  handoff_mtime="$(stat -f '%m' "$handoff_file" 2>/dev/null || stat -c '%Y' "$handoff_file" 2>/dev/null || echo 0)"
+  handoff_mtime="$(stat -c '%Y' "$handoff_file" 2>/dev/null || stat -f '%m' "$handoff_file" 2>/dev/null || echo 0)"
+  [[ "$handoff_mtime" =~ ^[0-9]+$ ]] || handoff_mtime=0
   now_epoch="$(date +%s)"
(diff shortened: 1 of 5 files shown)


lilongen:
Closing as superseded — #1081 (`687a2c9b`) landed the same fix: GNU `stat -c '%Y'`
probed first, BSD `stat -f '%m'` second, each guarded with `^[0-9]+$`.

#1115 then moved the hook bodies to `.agents/hooks/` and left `.claude/hooks/*.sh`
as symlinks, so all five files here conflict by type (blob vs symlink) — no hunks
to resolve, and the edits would target paths that already have the behavior.

On `main` (`8048500d`) every `stat` under `.agents/hooks/` sits inside a
GNU-first, digit-guarded `file_mtime_epoch()`:
`preflight-commit-push.sh:93,97` · `preflight-pr-review.sh:77,81` ·
`preflight-verdict-check.sh:124,128` · `run-commit-push-audit.sh:31,35`.
`run-verdict-audit.sh` no longer reads mtime — #1125 dropped it.

PR #1061, opened 2026-07-27
feat(billing): add end-to-end prepaid usage billing

ltstriker:
## Summary

Introduce end-to-end prepaid usage billing for Box workloads, including
failure-safe usage metering, versioned rating, wallet settlement, payment
flows, and balance-based access enforcement.

This also adds Billing APIs and dashboard views, Stripe and fake-provider
operations, generated clients, rollout runbooks, and layered verification
through a KVM golden-path smoke test.

Tracking: POL-144

Suggested review order: migrations → metering/runtime → rating/wallet →
payments/enforcement → API/UI/clients → verification/CI.

## Changes

- Meter Box lifecycle using allocation- and time-based `FULL` and
  `DISK_ONLY` usage periods.
- Add runtime leases, runner epochs, runtime generations, stale-job fencing,
  complete Box inventory, orphan cleanup, and process liveness checks so
  compute billing stops when a runtime is no longer authoritative.
- Archive closed periods with stable identities and rate them against
  versioned pricing plans using decimal math and pricing snapshots.
- Settle charges into an immutable organization wallet ledger, spending free
  balance before paid balance and carrying fractional-cent remainder between
  periods.
- Add fake and Stripe payment providers for payment-method setup, idempotent
  manual top-ups, auto-reload, webhooks, reconciliation, refunds, disputes,
  health checks, and operator recovery.
- Enforce prepaid access for Box create, start, and proxy auto-start, and stop
  active Boxes after billing access is lost. Enforcement remains disabled by
  default for staged rollout.
- Expose Billing overview, pricing, per-Box usage, receipts, payment setup,
  top-up, and auto-reload APIs.
- Add dashboard views for balances, usage trends, limits, payment methods,
  top-ups, auto-reload, and receipts.
- Synchronize the TypeScript and Go OpenAPI clients with the Billing and
  runner protocol additions.
- Add pre-deploy migrations, SST configuration, payment/archive runbooks,
  unit and PostgreSQL integration suites, Stripe sandbox coverage, fault
  system tests, and the Billing golden-path E2E workflow.

## Architecture

```mermaid
flowchart LR
  Runner["Runner inventory<br/>epoch + sequence + generation"]
  Runtime["Runtime lease<br/>liveness + fencing"]
  Period["Usage periods<br/>FULL / DISK_ONLY"]
  Archive["Usage archive"]
  Rating["Versioned rating"]
  Settlement["Wallet settlement"]
  Wallet["Immutable wallet ledger"]
[...]

PR #1071, opened 2026-07-28
feat(runner): report runtime generation and fence stale commands

G4614:
Adds runner-side runtime generation tracking and stale-command fencing, split out of #1061.

Test plan:
- [x] `go -C apps/runner build ./pkg/...`
- [x] `go -C apps/runner test ./pkg/runner/v2/runtimegeneration/...`
- [x] `gofmt -l apps/runner/pkg apps/runner/cmd` — clean
- [ ] `go -C apps/runner test ./pkg/runner/v2/{executor,healthcheck,poller}/...` — cgo link needs `libboxlite.a`, not built locally; runs in CI
- [ ] Regenerate the Go client — five files were taken from #1061 by hand and will not match the "Regenerate API clients and diff" check


G4614:
Scope notes for review.

The Rust liveness probe from #1061 (`liveness.rs`, `util/process.rs`, `rt_impl.rs`, `box_impl.rs`) is not here. `RuntimeLivenessProbe` is `pub(crate)` and Go never calls it, and those four files are where #1061 conflicts with `main` — six conflicts, the largest 102 lines. They want their own slice.

The generated Go client is included only for the five files this protocol needs (`model_runner_box_observation.go`, `model_runner_healthcheck.go`, `model_job_type.go`, `api_jobs.go`). `api_billing.go` and `api_admin.go` are not. `client.go` is left at `main` because #1061's version registers `BillingAPI`. As the test plan says, this should be regenerated before merge rather than merged as hand-picked output.

PR #1073, opened 2026-07-28
feat(api): reconcile runtime leases and fence stale runners

G4614:
Adds API-side runtime lease reconciliation and stale-runner fencing, split out of #1061 and stacked on #1071.

Test plan:
- [x] `npx tsc -p api/tsconfig.app.json --noEmit` — clean; the same check on #1071 is also clean, so nothing here is inherited
- [x] `npx jest --config api/jest.config.ts --testPathPatterns runtime-lease` — 9 passed
- [x] Mutation check on those 9: removing the lease-expiry re-check, the `runtimeAuthorized` guard, or the `desiredState` guard each fails exactly one case
- [x] Audited that every non-billing writer of `runtimeAuthorized`, `runtimeGeneration`, `runtimeUnavailable`, `lifecycleJobId` and `executionEpoch` in #1061 is present here, that no `usage/` or `billing/` import remains under `box/` or `config/`, and that every relative import resolves
- [x] Migration up/down/up against PostgreSQL 15 on a dev Box — after `down`, the table list, `box` columns and index list are identical to the pre-migration baseline, and `up` re-applies cleanly; note `box_runtime_cleanup` needs the `uuid-ossp` extension for `uuid_generate_v4()`
- [ ] Lock-ordering audit across `observeRunnerSnapshot` → `acceptRunnerEpoch` → `evaluateStartedJob` — four tables take `pessimistic_write` and unit tests cannot cover ordering


G4614:
@ltstriker two things I changed while splitting, both worth your confirmation.

**A port in place of the direct metering dependency.** `runtime-lease.service.ts` called `UsagePeriodWriter` in three places. Those now go through a `RuntimeUsageSink` port with a no-op default; metering implements it in the next slice. `MeteringMode` moved to `box/enums/` since the reconciler needs it to describe a transition even when nothing is metering, and the lease TTL reads `runtime.leaseSeconds` rather than `billing.runtimeLeaseSeconds`. With no usage periods in existence the no-op is correct rather than merely inert.

**`box.repository.ts` split by content.** 146 added lines, six of them metering. I took the file whole and removed the import, the constructor injection and four `transition()` calls, which seemed more reliable than reassembling eleven hunks. If the repository is meant to meter on create, that belongs in the metering slice — worth confirming the layering is intended.

Also: the migration is rewritten rather than renumbered. #1061 grows this schema across three migrations written weeks apart, in the opposite order from the dependencies, and none has run anywhere — so there is no history to replay. Fifteen files become one.

PR #1075, opened 2026-07-28
fix(export): stop shipping the guest rootfs disk in box archives

G4614:
Box archives no longer carry the guest rootfs disk, which is host-global state the importing host rebuilds from its own version-keyed cache.

It is a thin COW overlay over the shared guest rootfs cache (keyed by bootstrap image + guest binary version) and holds no user state; clone and snapshot-restore already treat it as disposable. Export was the exception, and flattening stripped its backing reference, so an imported box booted from the archived copy instead of the importing host's correctly-versioned cache. No archive version bump — the disk was already optional, so old importers handle its absence and new ones ignore its presence.

Test plan:
- [x] `cargo nextest run -p boxlite --no-default-features --lib -E 'test(~export_omits_the_guest_rootfs_disk)'` — new regression test
- [x] Two-side verified: with all production changes reverted the test still compiles and fails on its assertion, `got ["manifest.json", "disk.qcow2", "guest-rootfs.qcow2"]`; restored, it passes
- [x] `make test:integration:rust FILTER=export` — 15/15 passed (real VM), including `test_export_import_roundtrip`, which imports and then starts the box
- [x] `cargo nextest run -p boxlite --no-default-features --lib` — 952/954; the two failures (`seatbelt` sandbox, vendored-libkrun source read) reproduce on a pristine tree
- [x] `cargo fmt --all -- --check`, `cargo clippy -p boxlite --no-default-features --all-targets -- -D warnings`

### DIFF
--- a/src/boxlite/src/litebox/archive.rs
+++ b/src/boxlite/src/litebox/archive.rs
@@ -77,11 +77,13 @@ pub struct ArchiveManifest {
 // ── Build ───────────────────────────────────────────────────────────────
 
 /// Build a zstd-compressed tar archive.
+///
+/// Carries the manifest and the container disk only. The guest rootfs disk is
+/// not exported — see `do_export_flatten`.
 pub(crate) fn build_zstd_tar_archive(
     output_path: &Path,
     manifest_path: &Path,
     container_disk: &Path,
-    guest_disk: Option<&Path>,
     compression_level: i32,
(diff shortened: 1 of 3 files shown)


G4614:
Folded into #1080 — that PR already contained this commit, so keeping them split cost an extra review without shrinking the diff.

PR #1086, opened 2026-07-29
test(usage): cover box usage period ledger and its crash paths

lntotk:
## Summary

The box usage ledger is written by in-process event handlers with no outbox and no retry, so a box row and its usage period can disagree whenever the API dies between committing one and writing the other. Nothing currently detects that, and most of the resulting drift never heals. This adds verification for the healthy path and for that crash window, plus the reconciliation query the API lacks.

## Changes

- `usage.lifecycle.integration.spec.ts` — 29 scenarios driving every transition through the production `BoxRepository.update()`, so the handlers consume the events a real box produces rather than hand-built ones. Covers start / stop / restart / resize / delete, the daily roll-over, the archive sweep, and the one-open-period invariant.
- `usage.crash-recovery.integration.spec.ts` — 18 scenarios pinning the four lost-transition shapes (START / STOP / DESTROY / resize) through an API restart, the roll-over, and the box's next real transition, plus two handler-level failures: a lost write surfacing only as a log line, and the graceful-shutdown drain returning while a handler is still writing.
- `apps/scripts/usage-period-chaos/` — failure injection that kills a real process holding the real `UsageModule` graph, against a database it creates and drops itself.
- `apps/infra-local/scripts/` — the same faults injected into a running local stack on a real box, and `usage-ledger-audit.sql`: ten finding classes across both billing directions and the period chain, with an exit code so it can run as a cron check or a CI gate.
- `docs/investigations/box-usage-period-ledger.md` — where the window is, what each scenario measured, and the options for closing it.

## How to verify

Both specs need a Postgres and a Redis and create their own databases; they skip when `DB_HOST`, `DB_DATABASE` or `REDIS_HOST` are unset.

```bash
cd apps
DB_HOST=127.0.0.1 DB_PORT=5432 DB_USERNAME=postgres DB_PASSWORD=postgres \
DB_DATABASE=postgres REDIS_HOST=127.0.0.1 REDIS_PORT=6379 \
  npx jest --config api/jest.config.ts --rootDir api --testPathPatterns usage

node scripts/usage-period-chaos/run.mjs            # 8 injection scenarios
node scripts/usage-period-chaos/check-audit-classes.mjs
```

## Risks / rollout

No production code is touched — specs, tooling and docs only.

`usage.crash-recovery.integration.spec.ts` deliberately pins the shutdown drain's current behaviour. Landing the `activeJobs` change that the investigation recommends will turn that assertion red by design; the doc's staleness note flags it.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive integration coverage for usage billing lifecycles, rollovers, archiving, concurrency, and crash-recovery/lost-transition scenarios.
* **New Tools**
  * Added local failure-injection and chaos scripts to reproduce interrupted usage-ledger behavior.
  * Added a read-only SQL audit report to detect under-/over-billing, missing/stale periods, structural inconsistencies, and drift.
* **Documentation**
  * Added an investigations guide and local playbook for reproducing, diagnosing, and auditing usage-ledger drift, with links from the docs index.

### DIFF
--- a/apps/api/src/usage/usage.crash-recovery.integration.spec.ts
+++ b/apps/api/src/usage/usage.crash-recovery.integration.spec.ts
@@ -1,0 +1,556 @@
+/*
+ * Copyright 2025 BoxLite AI
+ * SPDX-License-Identifier: AGPL-3.0
+ */
+
+import { RedisModule, getRedisConnectionToken } from '@nestjs-modules/ioredis'
+import { EventEmitterModule } from '@nestjs/event-emitter'
+import { Test, TestingModule } from '@nestjs/testing'
+import { TypeOrmModule } from '@nestjs/typeorm'
+import type { Redis } from 'ioredis'
+import { setTimeout as sleep } from 'timers/promises'
+import { DataSource, IsNull, Repository } from 'typeorm'
+import { RedisLockProvider } from '../box/common/redis-lock.provider'
+import { BoxLastActivity } from '../box/entities/box-last-activity.entity'
+import { Box } from '../box/entities/box.entity'
+import { BoxDesiredState } from '../box/enums/box-desired-state.enum'
+import { BoxState } from '../box/enums/box-state.enum'
+import { BoxRepository } from '../box/repositories/box.repository'
+import { CustomNamingStrategy } from '../common/utils/naming-strategy.util'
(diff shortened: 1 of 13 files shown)


DorianZheng (approved):
LGTM!

PR #1009, opened 2026-07-20
feat: add auto pause and resume lifecycle options at box creation time

xhebox:
Refs #1003

## Summary
Adds create-time lifecycle options for AutoPause, AutoDelete, and AutoResume, and keeps the legacy autostop/autodelete endpoints working by converting minutes to seconds.

## Changes
- Box creation accepts `auto_pause`, `auto_delete`, and `auto_resume` in both the internal and BoxLite REST APIs.
- Backend sweeper pauses idle boxes and deletes long-stopped boxes based on the new second-based fields.
- Proxy and WebSocket paths auto-resume stopped boxes when `auto_resume` is true.
- Legacy `POST /box/{id}/autostop/{interval}` and `POST /box/{id}/autodelete/{interval}` endpoints are retained; input is converted from minutes to seconds.
- Dashboard, SDKs (Go/Node/Python/C), and Rust runtime expose the new options.
- Added design doc and user guide under `docs/`.

## How to verify
- `make fmt:check:rust`
- `cd apps && ./node_modules/.bin/tsc -p api/tsconfig.app.json --noEmit`
- `cd apps && ./node_modules/.bin/eslint "api/src/box/**/*.ts"`

## Risks / rollout
- Old `autodelete/0` no longer deletes immediately; it now disables auto-delete, consistent with the new `0 = disabled` semantics. Existing clients relying on the old immediate-delete behavior need to migrate.

## Summary by CodeRabbit

- **New Features**
  - Added AutoPause, AutoDelete, and AutoResume controls with second-based intervals.
  - Boxes can automatically resume for supported proxy operations after pausing.
  - Dashboard creation now includes advanced lifecycle settings and validation.
  - Lifecycle settings are available across REST, C, Go, Node.js, and Python SDKs.
- **Bug Fixes**
  - Improved concurrent auto-resume handling and prevented requests from reaching boxes that are not ready.
  - Activity tracking now distinguishes user operations from observation-only requests.
- **Documentation**
  - Added lifecycle guides and updated API references, defaults, and migration behavior.

PR #1091, opened 2026-07-30
feat: repair a box startup whose job completion was lost

ltstriker:
## Summary

Repair boxes left in `CREATING` or `STARTING` when the runner starts the container successfully but its job-completion callback never reaches the API.

Persist lifecycle-scoped `Container.Start` evidence in BoxLite, use it to reconcile only the matching stalled startup job through the normal completion path, and prevent concurrent workers from claiming or finalizing the same job row from stale state.

### Startup recovery

Before (`origin/main` at `48b49977`)

```text
Executor.Execute
  (apps/runner/pkg/runner/v2/executor/executor.go:58)
  ├─ executeJob
  │    (executor.go:108)
  │    └─ createBox / startBox
  │         (apps/runner/pkg/runner/v2/executor/box.go:18,38)
  │         └─ Client.Create / Client.Start
  │              (apps/runner/pkg/boxlite/client.go:219,312)
  │              └─ Box.Start
  │                   (sdks/go/box_handle.go:43)
  │                   └─ … C ABI …
  │                        └─ BoxImpl::start
  │                             (src/boxlite/src/litebox/box_impl.rs:269)
  │                             └─ ensure_container_started
  │                                  (box_impl.rs:979)
  │                                  └─ Container.Start succeeds
  │                                       — no durable success evidence is recorded
  └─ updateJobStatus
       (executor.go:161)
       ← BUG: exhausted completion-callback retries can leave the API job IN_PROGRESS

BoxSyncService.PerformSync
  (apps/runner/pkg/services/box_sync.go:101)
  ├─ GetRemoteBoxStates
  │    (box_sync.go:65)
  │    └─ fetch only STARTED boxes
  │         ← BUG: excludes the stuck CREATING/STARTING box
  └─ SyncBoxState
       (box_sync.go:90)
       — never reached for an excluded box

BoxService.updateState
  (apps/api/src/box/services/box.service.ts:1259)
  ← BUG: rejects runner reports while the API box is CREATING or STARTING
```

After (`035a13b2`)

```text
Executor.Execute
  (apps/runner/pkg/runner/v2/executor/executor.go:58)
  ├─ executeJob
  │    (executor.go:108)
  │    └─ createBox / startBox
  │         (apps/runner/pkg/runner/v2/executor/box.go:18,38)
  │         └─ Client.Create / Client.Start
  │              (apps/runner/pkg/boxlite/client.go:219,320)
  │              └─ Box.Start
  │                   (sdks/go/box_handle.go:43)
  │                   └─ … C ABI …
  │                        └─ BoxImpl::start
  │                             (src/boxlite/src/litebox/box_impl.rs:269)
  │                             └─ ensure_container_started
  │                                  (box_impl.rs:985)
  │                                  └─ record_started
  │                                       (box_impl.rs:1023)
  │                                       └─ mark_started + save_box
  │                                            (state.rs:362; manager.rs:157)
  │                                            — persists state, PID, and started_at together
  └─ updateJobStatus
       (executor.go:161)
       — remains the normal completion path

BoxSyncService.PerformSync
  (apps/runner/pkg/services/box_sync.go:170)
  ├─ GetLocalContainerStates
  │    (box_sync.go:57)
  │    └─ Client.ListInfo
  │         (apps/runner/pkg/boxlite/client.go:490)
  │         └─ Go Runtime.ListInfo
  │              (sdks/go/info.go:94)
  │              └─ … C ABI …
  │                   └─ RuntimeImpl::list_info
  │                        (src/boxlite/src/runtime/rt_impl.rs:636)
  │                        └─ BoxInfo::new
  │                             (src/boxlite/src/runtime/types.rs:421)
  │                             ↩ state and started_at return from one snapshot
  ├─ GetRemoteBoxStates
  │    (box_sync.go:95)
  │    └─ fetchRunnerBoxes
  │         (box_sync.go:129)
  │         — fetches STARTED plus CREATING/STARTING candidates
  ├─ canReport
  │    (box_sync.go:226)
  │    — a transitional box requires local STARTED and non-nil started_at
  └─ SyncBoxState
       (box_sync.go:159; only when canReport returns true)
       └─ BoxController.updateBoxState
            (apps/api/src/box/controllers/box.controller.ts:381)
            └─ BoxService.updateState
                 (apps/api/src/box/services/box.service.ts:1313)
                 ├─ findStalledStartupJob
                 │    (box.service.ts:582)
                 │    — matches the claimed CREATE_BOX or START_BOX job
                 └─ JobService.updateJobStatus
                      (apps/api/src/box/services/job.service.ts:232)
                      └─ JobStateHandlerService.handleJobCompletion
                           (job-state-handler.service.ts:37; after transaction commit)
                           └─ handleCreateBoxJobCompletion /
                              handleStartBoxJobCompletion
                                (job-state-handler.service.ts:84,126)
                                — publishes STARTED and releases the existing lock
```

### Job-row concurrency

Before

```text
JobService.pollJobs
  (apps/api/src/box/services/job.service.ts:118)
  └─ claimPendingJobs
       (job.service.ts:455)
       └─ Repository.save
            (job.service.ts:483)
            ← BUG: concurrent pollers can both claim the same PENDING row

JobService.updateJobStatus
  (job.service.ts:232)
  └─ findOne → validate → save
       (job.service.ts:238–264)
       ← BUG: concurrent terminal writers can validate stale state and overwrite it
```

After

```text
JobService.pollJobs
  (apps/api/src/box/services/job.service.ts:118)
  └─ claimPendingJobs
       (job.service.ts:460)
       └─ UPDATE … WHERE status = PENDING RETURNING *
            (job.service.ts:493)
            — only the poller with affected = 1 receives the job

JobService.updateJobStatus
  (job.service.ts:232)
  └─ transaction
       └─ findOne(pessimistic_write) → validate → save
            (job.service.ts:238)
            — serializes terminal transitions before completion effects run
```

Fixes #<issue-number>

## Changes

- Persist `started_at` after a successful guest `Container.Start` in the same BoxLite state row as lifecycle state and PID.
  - Preserve it as evidence for the most recently ended lifecycle.
  - Clear it when a new lifecycle publishes a PID or recovery adopts a different shim PID.
- Expose `started_at` through Rust `BoxInfo`, the C ABI, Go, Node.js, Python, and CLI inspection output.
  - REST-backed metadata leaves the field unset because the control plane does not provide this evidence.
- Read the runner’s local state and startup evidence from one `ListInfo` snapshot.
- Query `CREATING` and `STARTING` boxes as reconciliation candidates without disrupting the existing `STARTED` reconciliation path.
- Report a transitional box as `STARTED` only when BoxLite reports both local `STARTED` state and durable start evidence.
- After the configurable stall window—60 seconds by default—complete the matching in-progress `CREATE_BOX` or `START_BOX` job through the existing completion handler.
- Claim pending jobs with a conditional `UPDATE … WHERE status = PENDING RETURNING *`.
- Serialize job status transitions with a transaction and pessimistic row lock.
- Add core, SDK, runner, API, CLI, compatibility, lifecycle, and concurrency coverage.
- Update the SDK and CLI reference documentation for the new timestamp.

## How to verify

```bash
make fmt:check
make lint
make test:unit:core
make test:unit:sdk
make test:apps

# Requires KVM on Linux or Hypervisor.framework on macOS
make test:integration:rust NEXTEST_FILTER_EXPR='binary(started_at)'
```

## Risks / rollout

- No database migration is required. `started_at` is optional and serde-defaulted in the existing BoxLite state row; legacy rows therefore load without evidence until a later successful start.
- `CBoxInfo` gains a tail field. Existing field offsets remain unchanged, but native libraries and generated SDK bindings should be rebuilt and released together.
- No strict API/runner deployment order is required:

  | Runner | Old API | New API |
  |---|---|---|
  | Old runner | Existing reconciliation only | Existing reconciliation only |
  | New runner | Existing `STARTED` reconciliation remains; lost-callback confirmation is skipped | Full startup repair |

- The default 60-second stall window delays recovery intentionally so it does not race a completion callback that is still progressing. It is configurable through `BOX_SYNC_START_CONFIRMATION_STALL_SECONDS`.
- Any failure of the additional transitional-box query is treated as “confirmation unavailable this cycle.” This preserves ordinary reconciliation but delays startup recovery until a later successful cycle.
- Recording `started_at` holds the BoxLite lifecycle-state lock while synchronously saving to SQLite, so database contention can delay other state readers.
- The PostgreSQL conditional-claim and pessimistic-lock paths are covered by mock/stub unit tests in this PR, not by a real PostgreSQL integration test.

### DIFF
--- a/apps/api/src/box/services/box.service.ts
+++ b/apps/api/src/box/services/box.service.ts
@@ -55,6 +55,9 @@ import { customAlphabet as customNanoid, nanoid, urlAlphabet } from 'nanoid'
 import { WithInstrumentation } from '../../common/decorators/otel.decorator'
 import { validateMountPaths, validateSubpaths } from '../utils/volume-mount-path-validation.util'
 import { BoxRepository } from '../repositories/box.repository'
+import { Job } from '../entities/job.entity'
+import { JobService } from './job.service'
+import { JobStatus, JobType, ResourceType } from '../dto/job.dto'
 import { PortPreviewUrlDto, SignedPortPreviewUrlDto } from '../dto/port-preview-url.dto'
 import { RegionService } from '../../region/services/region.service'
 import { BoxCreatedEvent } from '../events/box-create.event'
@@ -110,6 +113,9 @@ export class BoxService {
     private readonly regionService: RegionService,
     private readonly boxLookupCacheInvalidationService: BoxLookupCacheInvalidationService,
     private readonly boxActivityService: BoxActivityService,
+    @InjectRepository(Job)
+    private readonly jobRepository: Repository<Job>,
+    private readonly jobService: JobService,
   ) {}
 
   protected getLockKey(id: string): string {
@@ -553,15 +559,63 @@ export class BoxService {
     return this.getExpectedDesiredStateForState(state) !== undefined
   }
 
+  private getStartupJobTypeForState(state: BoxState): JobType | undefined {
+    switch (state) {
+      case BoxState.CREATING:
+        return JobType.CREATE_BOX
+      case BoxState.STARTING:
+        return JobType.START_BOX
+      default:
+        return undefined
+    }
+  }
+
+  /**
+   * The box's own startup job, if it was claimed by its runner and has since
+   * stopped making progress.
+   *
+   * "Stalled" is measured from when the runner claimed the job, because that
+   * is what a lost completion callback looks like from here: claimed, never
+   * closed. A job still within the window is assumed to be running normally,
+   * and an unclaimed (PENDING) job has no runner to have lost anything.
+   */
+  private async findStalledStartupJob(box: Box): Promise<Job | null> {
+    const startupJobType = this.getStartupJobTypeForState(box.state)
+    if (!startupJobType || !box.runnerId) {
+      return null
+    }
+
+    const stallSeconds = this.configService.getOrThrow('boxSync.startConfirmationStallSeconds')
+    const claimedBefore = new Date(Date.now() - stallSeconds * 1000)
+
+    return this.jobRepository.findOne({
+      where: {
+        runnerId: box.runnerId,
+        resourceType: ResourceType.BOX,
+        resourceId: box.id,
+        type: startupJobType,
+        status: JobStatus.IN_PROGRESS,
+        startedAt: LessThan(claimedBefore),
+      },
+      order: { createdAt: 'DESC' },
+    })
+  }
+
   async findByRunnerId(runnerId: string, states?: BoxState[], skipReconcilingBoxes?: boolean): Promise<Box[]> {
     const where: FindOptionsWhere<Box> = { runnerId }
     if (states && states.length > 0) {
-      // Validate that all states have corresponding desired states
-      states.forEach((state) => {
-        if (!this.hasValidDesiredState(state)) {
-          throw new BadRequestError(`State ${state} does not have a corresponding desired state`)
-        }
-      })
+      // Only the skip filter needs a state to have a corresponding desired
+      // state — it is defined as "state already matches desired state". Asking
+      // for a transitional state is a legitimate query on its own (a runner
+      // reconciling a startup whose job completion was lost does exactly
+      // that), so the requirement belongs to the filter, not to the parameter.
+      if (skipReconcilingBoxes) {
+        states.forEach((state) => {
+          if (!this.hasValidDesiredState(state)) {
+            throw new BadRequestError(`State ${state} does not have a corresponding desired state`)
+          }
+        })
+      }
       where.state = In(states)
     }
 
@@ -1272,7 +1326,31 @@ export class BoxService {
 
     //  only allow updating the state of started | stopped boxes
     if (![BoxState.STARTED, BoxState.STOPPED].includes(box.state)) {
-      throw new BadRequestError('Box is not in a valid state to be updated')
+      // One exception: a runner reporting STARTED for a box we still show as
+      // coming up. That means its startup job finished on the runner but the
+      // completion never reached us. Believe it only once the job has stopped
+      // making progress on its own — until then the normal callback is still
+      // the better answer, and we say nothing rather than reject a runner that
+      // is telling the truth.
+      if (newState !== BoxState.STARTED || box.desiredState !== BoxDesiredState.STARTED) {
+        throw new BadRequestError('Box is not in a valid state to be updated')
+      }
+
+      const stalledStartupJob = await this.findStalledStartupJob(box)
+      if (!stalledStartupJob) {
+        this.logger.debug(`Box ${boxId} start not yet confirmable; startup job still progressing`)
+        return
+      }
+
+      // Completing the job is what moves the box: the normal completion
+      // handler owns the STARTED transition, the pending flag, the activity
+      // stamp, the state-change event, and releasing the box's Redis lock.
+      // Doing any of that here instead would fork that logic.
+      this.logger.warn(
+        `Completing stalled startup job ${stalledStartupJob.id} for box ${boxId} from runner-reported state`,
+      )
+      await this.jobService.updateJobStatus(stalledStartupJob.id, JobStatus.COMPLETED)
+      return
     }
 
     if (box.desiredState == BoxDesiredState.DESTROYED) {
--- a/apps/api/src/box/services/job.service.ts
+++ b/apps/api/src/box/services/job.service.ts
@@ -235,33 +235,38 @@ export class JobService {
     errorMessage?: string,
     resultMetadata?: string,
   ): Promise<Job> {
-    const job = await this.findOne(jobId)
-    if (!job) {
-      throw new NotFoundException(`Job with ID ${jobId} not found`)
-    }
+    const updatedJob = await this.jobRepository.manager.transaction(async (manager) => {
+      const job = await manager.findOne(Job, {
+        where: { id: jobId },
+        lock: { mode: 'pessimistic_write' },
+      })
+      if (!job) {
+        throw new NotFoundException(`Job with ID ${jobId} not found`)
+      }
 
-    if (!this.isValidStatusTransition(job.status, status)) {
-      throw new ConflictException(`Invalid job status transition from ${job.status} to ${status} for job ${jobId}`)
-    }
+      if (!this.isValidStatusTransition(job.status, status)) {
+        throw new ConflictException(`Invalid job status transition from ${job.status} to ${status} for job ${jobId}`)
+      }
 
-    job.status = status
-    if (errorMessage) {
-      job.errorMessage = errorMessage
-    }
+      job.status = status
+      if (errorMessage) {
+        job.errorMessage = errorMessage
+      }
 
-    if (status === JobStatus.IN_PROGRESS && !job.startedAt) {
-      job.startedAt = new Date()
-    }
+      if (status === JobStatus.IN_PROGRESS && !job.startedAt) {
+        job.startedAt = new Date()
+      }
 
-    if (status === JobStatus.COMPLETED || status === JobStatus.FAILED) {
-      job.completedAt = new Date()
-    }
+      if (status === JobStatus.COMPLETED || status === JobStatus.FAILED) {
+        job.completedAt = new Date()
+      }
 
-    if (resultMetadata) {
-      job.resultMetadata = resultMetadata
-    }
+      if (resultMetadata) {
+        job.resultMetadata = resultMetadata
+      }
 
-    const updatedJob = await this.jobRepository.save(job)
+      return manager.save(Job, job)
+    })
     this.logger.debug(`Updated job ${jobId} status to ${status}`)
 
     // Handle job completion for v2 runners - update box, artifact, or backup state.
@@ -469,24 +474,35 @@ export class JobService {
       return []
     }
 
-    // Update jobs to IN_PROGRESS
     const now = new Date()
+    // A single conditional UPDATE prevents a later-row error from leaving
+    // earlier claims committed. The status predicate retains the compare-and-swap,
+    // so a concurrent poll may still win a subset of these candidates.
+    const claim = await this.jobRepository.update(
+      {
+        id: In(jobs.map((job) => job.id)),
+        status: JobStatus.PENDING,
+      },
+      {
+        status: JobStatus.IN_PROGRESS,
+        startedAt: now,
+        updatedAt: now,
+      },
+      { returning: ['id'] },
+    )
+    const claimedIds = new Set((claim.raw as Array<Pick<Job, 'id'>>).map((row) => row.id))
     const claimedJobs: JobDto[] = []
 
     for (const job of jobs) {
-      try {
-        job.status = JobStatus.IN_PROGRESS
-        job.startedAt = now
-        job.updatedAt = now
-
-        // save() with @VersionColumn will automatically check version and throw OptimisticLockVersionMismatchError if changed
-        const savedJob = await this.jobRepository.save(job)
-
-        claimedJobs.push(new JobDto(savedJob))
-      } catch (error) {
-        // If optimistic lock fails, job was already claimed by another runner - skip it
-        this.logger.debug(`Job ${job.id} already claimed by another runner (version mismatch)`)
+      if (!claimedIds.has(job.id)) {
+        this.logger.debug(`Job ${job.id} was already claimed by a concurrent poll`)
+        continue
       }
+
+      job.status = JobStatus.IN_PROGRESS
+      job.startedAt = now
+      job.updatedAt = now
+      claimedJobs.push(new JobDto(job))
     }
 
     if (claimedJobs.length > 0) {
--- a/apps/runner/pkg/services/box_sync.go
+++ b/apps/runner/pkg/services/box_sync.go
@@ -6,15 +6,26 @@ package services
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"log/slog"
+	"os"
+	"path/filepath"
 	"time"
 
 	apiclient "github.com/boxlite-ai/boxlite/libs/api-client-go"
+	sdkboxlite "github.com/boxlite-ai/boxlite/sdks/go"
 	runnerapiclient "github.com/boxlite-ai/runner/pkg/apiclient"
 	blclient "github.com/boxlite-ai/runner/pkg/boxlite"
 	"github.com/boxlite-ai/runner/pkg/models/enums"
 )
+
+// boxStateReader is the slice of the BoxLite client this service needs, kept
+// narrow so the sync loop can be exercised without a live runtime.
+type boxStateReader interface {
+	ListInfo(ctx context.Context) ([]sdkboxlite.BoxInfo, error)
+	GetBoxState(ctx context.Context, boxID string) (enums.BoxState, error)
+}
 
 type BoxSyncServiceConfig struct {
 	Logger   *slog.Logger
@@ -24,26 +35,37 @@ type BoxSyncServiceConfig struct {
 
 type BoxSyncService struct {
 	log      *slog.Logger
-	boxlite  *blclient.Client
+	boxlite  boxStateReader
+	homeDir  string
 	interval time.Duration
 	client   *apiclient.APIClient
+}
+
+// localContainerState pairs a box's local runtime state with the durable
+// evidence that the current shim actually launched its init.
+type localContainerState struct {
+	state enums.BoxState
+	// startedAt is non-nil only when BoxLite recorded a successful
+	// Container.Start for the shim this box is running right now.
+	startedAt *time.Time
 }
 
 func NewBoxSyncService(config BoxSyncServiceConfig) *BoxSyncService {
 	return &BoxSyncService{
 		log:      config.Logger.With(slog.String("component", "box_sync_service")),
 		boxlite:  config.Boxlite,
+		homeDir:  config.Boxlite.HomeDir(),
 		interval: config.Interval,
 	}
 }
 
-func (s *BoxSyncService) GetLocalContainerStates(ctx context.Context) (map[string]enums.BoxState, error) {
+func (s *BoxSyncService) GetLocalContainerStates(ctx context.Context) (map[string]localContainerState, error) {
 	boxes, err := s.boxlite.ListInfo(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to list boxes: %w", err)
 	}
 
-	boxStates := make(map[string]enums.BoxState)
+	boxStates := make(map[string]localContainerState, len(boxes))
 	for _, box := range boxes {
 		boxId := box.Name
 		if boxId == "" {
@@ -56,13 +78,92 @@ func (s *BoxSyncService) GetLocalContainerStates(ctx context.Context) (map[strin
 			continue
 		}
 
-		boxStates[boxId] = state
+		// The record lives under BoxLite's own id for the box, which is not
+		// always the key the control plane uses (that one prefers the name).
+		boxStates[boxId] = localContainerState{
+			state:     state,
+			startedAt: readStartedRecord(s.homeDir, box.ID, box.PID),
+		}
 	}
 
 	return boxStates, nil
 }
 
+// startedRecord mirrors StartedRecord in src/boxlite/src/runtime/layout.rs,
+// written by BoxLite once a box's guest Container.Start returns success.
+type startedRecord struct {
+	PID      int   `json:"pid"`
+	AtUnixMs int64 `json:"at_unix_ms"`
+}
+
+// readStartedRecord returns when BoxLite last confirmed a successful
+// Container.Start for the shim currently running this box, or nil when there
+// is no such evidence.
+//
+// The PID cross-check is what makes the record trustworthy: BoxLite clears it
+// at the start of every new lifecycle, and pairing it with the box's live PID
+// closes the remaining window where a record could survive the shim it names.
+func readStartedRecord(homeDir string, boxID string, livePID int) *time.Time {
+	if homeDir == "" || boxID == "" || livePID <= 0 {
+		return nil
+	}
+
+	contents, err := os.ReadFile(filepath.Join(homeDir, "boxes", boxID, "started"))
+	if err != nil {
+		return nil
+	}
+
+	var record startedRecord
+	if err := json.Unmarshal(contents, &record); err != nil {
+		return nil
+	}
+	if record.PID != livePID || record.AtUnixMs <= 0 {
+		return nil
+	}
+
+	startedAt := time.UnixMilli(record.AtUnixMs)
+	return &startedAt
+}
+
 func (s *BoxSyncService) GetRemoteBoxStates(ctx context.Context) (map[string]apiclient.BoxState, error) {
+	remoteBoxes, err := s.fetchRunnerBoxes(ctx, string(apiclient.BOXSTATE_STARTED), true)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get boxes from API: %w", err)
+	}
+
+	// Boxes the control plane still believes are coming up. Fetching them is
+	// what lets the existing mismatch loop below repair a startup whose job
+	// completion never landed.
+	//
+	// An API that predates this query rejects the transitional states with a
+	// 400. That must not cost us the STARTED reconciliation above, which has
+	// worked on its own for far longer — so the failure degrades to "no
+	// candidates this tick" instead of failing the whole sync.
+	transitional, err := s.fetchRunnerBoxes(
+		ctx,
+		string(apiclient.BOXSTATE_CREATING)+","+string(apiclient.BOXSTATE_STARTING),
+		false,
+	)
+	if err != nil {
+		s.log.DebugContext(
+			ctx,
+			"Transitional box query unavailable; skipping start confirmation this cycle",
+			"error", err,
+		)
+		return remoteBoxes, nil
+	}
+	for boxId, state := range transitional {
+		remoteBoxes[boxId] = state
+	}
+
+	return remoteBoxes, nil
+}
+
+func (s *BoxSyncService) fetchRunnerBoxes(
+	ctx context.Context,
+	states string,
+	skipReconcilingBoxes bool,
+) (map[string]apiclient.BoxState, error) {
 	if s.client == nil {
 		client, err := runnerapiclient.GetApiClient()
 		if err != nil {
@@ -70,16 +171,17 @@ func (s *BoxSyncService) GetRemoteBoxStates(ctx context.Context) (map[string]api
 		}
 		s.client = client
 	}
+
 	boxes, _, err := s.client.BoxAPI.GetBoxesForRunner(ctx).
-		States(string(apiclient.BOXSTATE_STARTED)).SkipReconcilingBoxes(true).
+		States(states).SkipReconcilingBoxes(skipReconcilingBoxes).
 		Execute()
 	if err != nil {
-		return nil, fmt.Errorf("failed to get boxes from API: %w", err)
+		return nil, err
 	}
 
-	remoteBoxes := make(map[string]apiclient.BoxState)
+	remoteBoxes := make(map[string]apiclient.BoxState, len(boxes))
 	for _, box := range boxes {
-		if box.Id != "" {
+		if box.Id != "" && box.State != nil {
 			remoteBoxes[box.Id] = *box.State
 		}
 	}
@@ -110,7 +212,7 @@ func (s *BoxSyncService) PerformSync(ctx context.Context) error {
 	}
 
 	syncCount := 0
-	for boxId, localState := range localStates {
+	for boxId, local := range localStates {
 		remoteState, exists := remoteStates[boxId]
 		if !exists {
 			continue
@@ -118,10 +220,14 @@ func (s *BoxSyncService) PerformSync(ctx context.Context) error {
 
 		convertedRemoteState := s.convertFromApiState(remoteState)
 
-		if localState != convertedRemoteState {
-			s.log.InfoContext(ctx, "State mismatch for box", "boxId", boxId, "localState", localState, "remoteState", convertedRemoteState)
+		if local.state != convertedRemoteState {
+			if !s.canReport(local, remoteState) {
+				continue
+			}
 
-			err := s.SyncBoxState(ctx, boxId, localState)
+			s.log.InfoContext(ctx, "State mismatch for box", "boxId", boxId, "localState", local.state, "remoteState", convertedRemoteState)
+
+			err := s.SyncBoxState(ctx, boxId, local.state)
 			if err != nil {
 				s.log.ErrorContext(ctx, "Failed to sync state for box", "boxId", boxId, "error", err)
 				continue
@@ -135,6 +241,27 @@ func (s *BoxSyncService) PerformSync(ctx context.Context) error {
 	}
 
 	return nil
+}
+
+// canReport decides whether this runner may state its local view to the
+// control plane for a given remote state.
+//
+// A box the control plane still shows as coming up is the one case where the
+// local view is not self-evidently authoritative: BoxLite publishes Running
+// once the VM is up, which happens before the guest's separate Container.Start
+// runs. Reporting STARTED off that alone would call a box ready whose init was
+// never launched — so a transitional box additionally needs BoxLite's durable
+// record that Container.Start did succeed for the shim now running it.
+//
+// Every other remote state keeps the long-standing behaviour: the local view
+// wins unconditionally.
+func (s *BoxSyncService) canReport(local localContainerState, remoteState apiclient.BoxState) bool {
+	switch remoteState {
+	case apiclient.BOXSTATE_CREATING, apiclient.BOXSTATE_STARTING:
+		return local.state == enums.BoxStateStarted && local.startedAt != nil
+	default:
+		return true
+	}
 }
 
 func (s *BoxSyncService) StartSyncProcess(ctx context.Context) {
--- a/apps/runner/pkg/services/box_sync.go
+++ b/apps/runner/pkg/services/box_sync.go
@@ -6,11 +6,8 @@ package services
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"log/slog"
-	"os"
-	"path/filepath"
 	"time"
 
 	apiclient "github.com/boxlite-ai/boxlite/libs/api-client-go"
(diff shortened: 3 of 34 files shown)


src/boxlite/src/runtime/layout.rs
```diff
@@ -1,8 +1,41 @@
...
 use boxlite_shared::errors::{BoxliteError, BoxliteResult};
 use boxlite_shared::layout::{SharedGuestLayout, dirs as shared_dirs};
+use serde::Serialize;
 use std::path::{Path, PathBuf};
 
+/// Contents of [`BoxFilesystemLayout::started_file_path`] — written by the host
+/// once the guest's `Container.Start` returns success for the current physical
+/// lifecycle.
+///
+/// It crosses a process boundary — BoxLite writes it, the cloud runner reads it
```
DorianZheng: The cloud runner should not have knowledge about the inner layout of the box. Track the container's started status through a file is not a good design. Normally we should inspect box's status through `BoxInfo`?

ltstriker:
Follow-up stacked on this branch: #1104 — `fix/usage_failure_lt_v2.0` → `fix/usage_failure_lt`.

It fixes usage-period drift (the daily roll-over copying stale resources forward, and boxes that never got a period at all). The dependency is branch-level only: it touches six files under `apps/api/src/usage/`, disjoint from the fourteen here, and calls nothing this PR adds. Merge this one first and GitHub will retarget #1104 to `main`.

DorianZheng:
### e2e, simplified

**Before**

```
runner   bx.Start ✓  (box is now genuinely running)
         └─ POST /job/{id}/status COMPLETED ──✗ lost
                                              (4xx → non-retryable, gives up at once;
                                               or retry budget exhausted)

API      job stays IN_PROGRESS
         boxInfo → inferStateFromJob → CREATING / STARTING
                   ↑ state is *derived from the job*, so it can never read STARTED
         box-start.action → checkTimeoutError: 5 min (STARTING) / 15 min (CREATING)
         └─ box = ERROR, recoverable: false      ← a healthy, running box marked broken
```

**After**

```
runner   bx.Start ✓
         ├─ writes {box_dir}/started {pid}
         └─ POST … COMPLETED ──✗ lost

         BoxSync (every 10s): box is CREATING/STARTING
                              AND started-record pid == live shim pid
         └─ PUT /box/{id}/state STARTED

API      startup job claimed and stalled ≥ 60s?
         └─ complete the job → handleJobCompletion → box = STARTED ✓
```

Two notes on the framing, from tracing the pre-PR paths:

- **"stays in `CREATING` or `STARTING` forever" doesn't match main.** The box does leave the transitional state — `box-start.action.ts:141/151` ERRORs it after 5 min (STARTING) or 15 min (CREATING), and `handleStaleJobs` independently fails the job at 10 min. So the bug being fixed is *resolves to the wrong answer* (a running box marked `ERROR`, `recoverable: false`), not *never resolves*. That's arguably the stronger motivation, since it can tear down a working VM.
- **One lost-callback case already self-heals:** if the runner restarts, `poller.go:53-66` replays its `IN_PROGRESS` jobs before the poll loop, and `CREATE_BOX` is idempotent via `GetOrCreate`. The gap this PR actually closes is the *runner keeps running* case — that startup replay never fires again, so nothing revisits the job.

Neither changes the verdict on the mechanism; they'd just make the problem statement match what main does.

src/boxlite/src/litebox/state.rs
```diff
@@ -232,6 +232,23 @@ pub struct BoxState {
...
+    /// exited container.
+    ///
+    /// **Invariant — the value belongs to the PID in the same row.** It is
+    /// cleared in the one write that publishes a new lifecycle's PID (see
+    /// `BoxImpl::init_live_state`) and in the one recovery branch that can
+    /// adopt a PID the row did not publish, so a reader never has to pair it
+    /// with anything. Serde default keeps existing DB rows readable without
+    /// migration.
+    #[serde(default)]
+    pub container_started_at: Option<DateTime<Utc>>,
```
DorianZheng: just `started_at` to match the `created_at` naming style

src/boxlite/src/runtime/layout.rs
ltstriker: fixed, changed to use ```BoxInfo```.

src/boxlite/src/litebox/state.rs
```diff
@@ -331,9 +348,33 @@ impl BoxState {
...
+    /// Adopt a live shim that recovery found on disk.
+    ///
+    /// The one place a PID can enter this row without having been published by
+    /// the boot that owns it: a crash between the shim writing its PID file and
+    /// the row being saved leaves the two naming different lifecycles. A
+    /// `container_started_at` written for the PID we are replacing cannot
+    /// describe the one we are adopting, so it goes with it — that is what lets
+    /// every reader take the timestamp at face value.
+    pub fn adopt_recovered_shim(&mut self, pid: u32) {
+        if self.pid != Some(pid) {
```
DorianZheng, line 377: please clarify when this kind of case happen.

sdks/c/include/boxlite.h
```diff
@@ -316,6 +316,11 @@ typedef struct CBoxInfo {
   uint32_t auto_delete;
   int auto_resume;
   int64_t created_at;
+  // Unix milliseconds of the successful guest `Container.Start` for the
+  // lifecycle that owns [`Self::pid`]; `0` when this box's init was never
+  // launched. Milliseconds — not `created_at`'s seconds — because a caller
+  // comparing this against a job's timeline needs sub-second resolution.
+  int64_t container_started_at_unix_ms;
```
DorianZheng: should be consistent to `created_at`

apps/api/src/box/services/job.service.ts
```diff
@@ -235,33 +235,38 @@ export class JobService {
     errorMessage?: string,
     resultMetadata?: string,
   ): Promise<Job> {
-    const job = await this.findOne(jobId)
-    if (!job) {
-      throw new NotFoundException(`Job with ID ${jobId} not found`)
-    }
+    const updatedJob = await this.jobRepository.manager.transaction(async (manager) => {
```
DorianZheng, line 238: seems too costly to do this in a transaction

DorianZheng:
### `boxStateReader`: the second method can go — and dropping it closes a hole

Line numbers are `85bafad` (post-refactor).

**Today**

```
PerformSync                                   box_sync.go:176
└─ GetLocalContainerStates                    box_sync.go:57
   ├─ ListInfo(ctx)                           box_sync.go:58   → 1 FFI call
   │    returns []BoxInfo{ ID, Name, State, PID, StartedAt }   ← State is already here
   │
   └─ for each box:                           box_sync.go:64
      ├─ GetBoxState(ctx, boxId)              box_sync.go:70   → +1 FFI call per box
      │    └─ client.go:372
      │        ├─ getOrFetchBox → runtime.Get
      │        ├─ bx.Info(ctx)                ← re-fetches the State we already had
      │        └─ switch info.State → enums.BoxState   (pure mapping, no I/O)
      │
      └─ boxStartedAt(box)                    box_sync.go:78,93  ← taken from the ListInfo snapshot
```

So `state` and `startedAt` are read **at two different moments**, then combined in one decision:

```
canReport   box_sync.go:231
└─ return local.state == enums.BoxStateStarted && local.startedAt != nil     :234
              ↑ from GetBoxState, at T1        ↑ from ListInfo, at T0
```

**Why that matters now specifically.** The refactor's own comment (`box_sync.go:88-91`) says the reader "has nothing to cross-check", because BoxLite voids `StartedAt` in the same write that publishes a new lifecycle's PID. That holds — but only for values read *together*. Since `state` comes from a later, separate call, there's still a window: read `StartedAt` at T0, the box restarts, then `GetBoxState` at T1 reports the new lifecycle as running. Stale evidence + fresh state ⇒ `canReport` says yes for a lifecycle whose init hasn't launched. The old on-disk version caught this with the PID cross-check; the new one relies on atomicity that the split read gives away.

**Suggested change** — take `state` from the same `BoxInfo`:

```
GetLocalContainerStates
└─ ListInfo(ctx)                        1 FFI call, total
   └─ for each box:
      ├─ boxStateFromInfo(box.State)    pure func, no I/O
      └─ boxStartedAt(box)              same snapshot ⇒ atomic by construction
```

which reduces the interface to the one method that actually needs the runtime:

```go
type boxStateReader interface {
    ListInfo(ctx context.Context) ([]sdkboxlite.BoxInfo, error)
}
```

Three things fall out: `1+N` FFI round-trips become `1`; `(state, StartedAt, PID)` become one consistent snapshot, so the atomicity the comment claims is real end-to-end; and the `GetBoxState` error branch at `:71-74` disappears, along with the case where a box that vanished mid-loop is silently skipped.

`Client.GetBoxState` itself stays — `services/box.go:28` and the `backend.Backend` interface (`backend.go:27`) still use it. Only this call site goes.

One judgement call if you take this: `client.go:372`'s switch maps `configured`/`running`/`stopped` and sends everything else to `Unknown`, so `StateStopping` (`sdks/go/info.go:20`) currently lands on `Unknown` even though `enums.BoxStateStopping` exists. Mirroring that exactly keeps this a pure refactor; fixing it is a separate change.

The narrow interface itself is right — `*blclient.Client` needs a live libkrun runtime, so without it `PerformSync` would only be reachable from a VM-booting test. This is just about making it narrower.

DorianZheng:
### `started_at`: three clear-sites, two different meanings

Line numbers are `85bafad`.

```
who writes it
  mark_started              state.rs:356          set          ← only writer
  └─ record_started         box_impl.rs:1019      after Container.Start

who clears it
  adopt_recovered_shim      state.rs:370-373      if pid changed
  init_live_state           box_impl.rs:1157-1159 if !adopting_running
  reset_for_reboot          state.rs:454          always

who nulls the pid but keeps it
  mark_stop                 state.rs:422-426      pid = None, started_at kept
  mark_failed               state.rs:436-441      pid = None, started_at kept
```

**1. The doc and the code disagree.** `state.rs:243-247` says:

> **the value belongs to the PID in the same row.** It is cleared in the one write that publishes a new lifecycle's PID … and in the one recovery branch …

Two sites claimed, three exist — `reset_for_reboot` is neither of the two named. And after `mark_stop` / `mark_failed` there is no PID in the row for it to belong to: `pid: None`, `started_at: Some`.

**2. Two models are shipping at once.** `mark_stop`/`mark_failed` keep the timestamp (docker's `StartedAt`, which the doc explicitly cites — *"Survives a stop the way docker keeps StartedAt on an exited container"*). `reset_for_reboot` clears it (liveness evidence). Same event in both cases: this lifecycle's process is gone.

Simplest fix: keep the docker model and **delete the `reset_for_reboot` clear**. It is already redundant — a rebooted box is `Stopped`, so `canReport`'s `state == Started` fails anyway, and the next boot voids via `init_live_state`. Whichever way it goes, one of the two behaviours should change so the field has one meaning.

**3. One rule, three hand-maintained copies.** All three clears are the same predicate — *pid changed ⇒ the timestamp is stale*. It can live in the setter instead:

```rust
pub fn set_pid(&mut self, pid: Option<u32>) {
    if self.pid != pid {
        self.started_at = None;   // it described the lifecycle we're replacing
    }
    self.pid = pid;
    self.last_updated = Utc::now();
}
```

Then `adopt_recovered_shim` is just `set_pid` + `set_status` (the conditional and most of its comment go), and `box_impl.rs:1151-1159` — 9 lines of comment + guard — deletes entirely, because `init_live_state` already calls `set_pid` five lines above at `:1144`, with exactly the right semantics: adopting-running passes the same PID (no clear), a fresh shim passes a new one (clear).

This is safe to centralise: every raw `state.pid = …` outside `state.rs` is inside `#[cfg(test)]` (`box_impl.rs:1591/1845`, `rt_impl.rs:1971`; those test modules start at `:1481` / `:1822`). Production only goes through `set_pid` / `adopt_recovered_shim`.

If you want the invariant enforced by the compiler rather than by convention, the bigger version is to make the pair one value — `Option<Lifecycle { pid, started_at }>` — so "belongs to the PID in the same row" stops being a comment. That has serde/persisted-row implications, so probably a follow-up rather than this PR.

---

Two things I looked at that are **not** problems, for the record:

- The repeated `PerTestBoxHome` + `BoxliteRuntime::new` setup in `tests/started_at.rs` matches 20 other test files — house style, not worth changing here.
- `record_started` cloning `BoxState` to drop the lock before `save_box` (`box_impl.rs:1025-1033`) deviates from `box_impl.rs:1277-1279` and `:1294-1296`, which hold the write lock across the disk write — but it deviates in the safer direction, at the cost of one clone per successful start. Worth keeping; the older sites are the ones out of step.

DorianZheng:
### Correction to my previous comment — §3 (the `set_pid` consolidation) is wrong

Apologies, I need to retract the main suggestion in [my last comment](https://github.com/boxlite-ai/boxlite/pull/1091#issuecomment-5141988752). I claimed moving the clear into `set_pid` would be behaviour-preserving because *"adopting-running passes the same PID (no clear), a fresh shim passes a new one (clear)"*. **That is not guaranteed, and the refactor as I described it would break the case your comment deliberately protects.**

The two values come from different sources:

```
adopting_running  box_impl.rs:1076   = (state.status == BoxStatus::Running)   ← the row's STATUS
pid               box_impl.rs:1140-41 = PidFileReader::at(&pid_path).read()   ← the PID FILE
```

They are not tied together, and the codebase says outright that they can disagree — `litebox/init/tasks/vmm_attach.rs:6-9`:

> Identity is read from the canonical PID file (`shim.pid`) and verified via start-time fingerprint. **`state.pid` from the DB is a cache that could lag (PID reuse, external kill)**; the file + ProcessIdentity is the trust anchor.

So a box can be `Running` (⇒ `adopting_running == true`, and today `started_at` is correctly kept) while the row's cached `pid` lags behind the PID file. Under my proposed `set_pid`, that mismatch would clear `started_at` — exactly what `box_impl.rs:1154-1156` says must not happen:

> Adopting a *running* box is not a new lifecycle — its record still names the live shim and must survive.

`state.rs:363-369` makes the same point for the recovery path. So the current explicit `if !adopting_running` guard is keyed on the right thing and my "deletes entirely" was wrong. Sorry for the noise.

I also had two citation errors in that comment: `record_started` is at **`box_impl.rs:1023`**, not `:1019`; and `set_pid` is **13 lines / 3 statements** above the guard, not "five lines".

One thing the correction surfaces that is worth keeping: I only audited raw `state.pid = …` writes, not production **`set_pid` callers**. There is a fourth — `rt_impl.rs:950` (`remove_box`, force path) calls `set_pid(None)` then `save_box` at `:951` — which any setter-based rule would also silently change. That reinforces the conclusion: the rule genuinely does not belong in `set_pid`.

---

**§1 and §2 of that comment still stand** — I re-verified them:

- `state.rs:243-247` claims two clear-sites; there are three, and `reset_for_reboot` (`state.rs:454`) is neither of the two named.
- `mark_stop` (`state.rs:422-426`) and `mark_failed` (`state.rs:436-441`) leave `pid: None, started_at: Some`, so the stated invariant "the value belongs to the PID in the same row" does not hold after either.
- Two models are still shipping at once: docker's survives-death (`state.rs:241-242`, *"Survives a stop the way docker keeps StartedAt on an exited container"*) and liveness-evidence (`reset_for_reboot`). Picking one — and dropping the now-redundant `reset_for_reboot` clear — is still the suggestion.

The `Option<Lifecycle { pid, started_at }>` idea also still stands, and the PID-file/row divergence above is actually an argument *for* it: it makes the pairing explicit at the one place the two values are reconciled, instead of leaving it to a status flag and a cached field agreeing.

DorianZheng:
### Blocker: `85bafad` does not compile under `--all-targets`

```
$ BOXLITE_DEPS_STUB=1 cargo check -p boxlite-cli --all-targets

error[E0063]: missing field `started_at` in initializer of `boxlite::BoxInfo`
   --> src/cli/src/commands/inspect.rs:237:20
    |
237 |         let info = BoxInfo {
    |                    ^^^^^^^ missing `started_at`

error: could not compile `boxlite-cli` (bin "boxlite" test) due to 1 previous error
```

`BoxInfo` (`runtime/types.rs:349`) is `#[derive(Debug, Clone, Serialize, Deserialize)]` — no `Default`, no `#[non_exhaustive]` — so adding a required field breaks every exhaustive struct literal. The `#[serde(default)]` added alongside it only affects *deserialisation*; it does nothing for struct literals.

Three literals were missed. All are exhaustive, none uses a `..` escape, none mentions `started_at`:

| file:line | |
|---|---|
| `src/cli/src/commands/inspect.rs:237` | confirmed by the compile above |
| `sdks/python/src/info.rs:374` | `fn core_info(...) -> BoxInfo` |
| `sdks/node/src/info.rs:259` | `fn core_info(...) -> BoxInfo` |

All three are `#[cfg(test)]`, which is why a plain `cargo build` stays green — but `make/quality.mk:207,210` runs

```
cargo clippy --workspace --all-targets --all-features -- -D warnings
```

and `--all-targets` compiles test code. The Rust CI jobs had not reported on this head when I checked (only CodeQL and the review bots had), so this is very likely to go red once they do.

Fix is mechanical — `started_at: None` in each. **But greening CI that way leaves the more interesting gap**, which is why I'd raise it in the same breath:

`PyBoxInfo` (`sdks/python/src/info.rs:277-303`) and `JsBoxInfo` (`sdks/node/src/info.rs:172-213`) are *separate* binding structs, so they compile fine and simply **drop** the new field — the Python and Node SDKs silently do not expose `started_at`. Same for `src/cli/src/commands/inspect.rs:46-57` `InspectStatePresenter`, which is deliberately docker-shaped (`Status`/`Running`/`Pid`/`ExitCode`) — and docker's own `State` is `{Status, Running, Pid, ExitCode, StartedAt, FinishedAt}`, so `StartedAt` is precisely the missing member of the set it is otherwise mirroring. `BoxStateInfo` (`runtime/types.rs:470-482`), which both bindings expose as `.state`, is unextended too.

So the compile error is really flagging three places that each need a decision (expose it, or deliberately don't), not three places that need `None` pasted in.

DorianZheng:
Two fixes to [the comment above](https://github.com/boxlite-ai/boxlite/pull/1091#issuecomment-5142110193). The compile blocker itself is unaffected — I re-ran all three checks and they reproduce.

- **`BoxStateInfo` is at `runtime/types.rs:484-496`**, not `:470-482`. I quoted main's line numbers; the 14-line `started_at` insertion shifts everything below it, so `:470-482` at this head is the tail of `impl PartialEq for BoxInfo`. The point stands — `BoxStateInfo` is `{status, running, pid, exit_code}` and unextended.

- **I overstated docker's `State`.** I wrote it as `{Status, Running, Pid, ExitCode, StartedAt, FinishedAt}`; that is a subset, not the whole set. Actual (docker 29.4.3, `docker inspect --format '{{json .State}}'`): `Dead, Error, ExitCode, FinishedAt, OOMKilled, Paused, Pid, Restarting, Running, StartedAt, Status`. The argument only needs the weaker true form — docker's `State` does carry `StartedAt` alongside `ExitCode`, so `StartedAt` is the member `InspectStatePresenter` is missing from the set it is otherwise mirroring.

src/boxlite/src/litebox/state.rs
ltstriker: fixed to started_at

sdks/c/include/boxlite.h
ltstriker: changed to started_at

apps/runner/pkg/services/box_sync_test.go
```diff
@@ -0,0 +1,228 @@
...
+			got := boxStartedAt(tt.info)
+
+			switch {
+			case tt.want == nil && got != nil:
+				t.Fatalf("boxStartedAt() = %s, want nil", got)
+			case tt.want != nil && got == nil:
+				t.Fatalf("boxStartedAt() = nil, want %s", tt.want)
+			case tt.want != nil && !got.Equal(*tt.want):
+				t.Fatalf("boxStartedAt() = %s, want %s", got, tt.want)
+			}
```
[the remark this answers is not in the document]
ltstriker: time.Time has a String() method, not need to make it.

apps/runner/pkg/services/box_sync.go
```diff
@@ -37,49 +54,101 @@ func NewBoxSyncService(config BoxSyncServiceConfig) *BoxSyncService {
...
+		false,
+	)
+	if err != nil {
+		s.log.DebugContext(
+			ctx,
+			"Transitional box query unavailable; skipping start confirmation this cycle",
+			"error", err,
+		)
+		return remoteBoxes, nil
+	}
```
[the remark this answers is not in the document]
ltstriker: changed to warn.

apps/api/src/box/services/job.service.ts
ltstriker: just select and update one line and no conflict when work normally. 
Or change it to a update sql, but it will change more code. I think just use transaction first util the cost happened.

src/boxlite/src/litebox/state.rs
ltstriker: This happens when the runner's `POST /job/{id}/status COMPLETED` callback is lost:
the executor gives up immediately on 4xx (non-retryable) or after exhausting its
retry budget on 5xx/timeouts. If the runner then keeps running, nothing revisits
the job — the startup replay in `poller.go` only fires on runner restart. The box
is genuinely running, but the API times the job out and marks the box
`ERROR, recoverable: false`.

sdks/node/src/info.rs
```diff
@@ -273,6 +273,7 @@ mod tests {
             auto_resume: true,
             health_status: HealthStatus::default(),
             exit_code: None,
+            started_at: None,
```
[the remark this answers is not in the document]
ltstriker: added.

src/boxlite/src/litebox/box_impl.rs
```diff
@@ -994,9 +1000,45 @@ impl BoxImpl {
                 Ok::<(), BoxliteError>(())
             })
             .await?;
+        if started_here {
+            self.record_started();
```
DorianZheng, line 1004: should record this when container initialization finished

ltstriker: just keep the same behavior as normal which records it after container started. Maybe left it future to change the positon together.

DorianZheng (dismissed):

apps/api/src/box/services/job.service.ts
```diff
@@ -469,24 +474,40 @@ export class JobService {
...
+
+      const claimedRow = (claim.raw as Job[])[0]
+      if (!claim.affected || !claimedRow) {
+        this.logger.debug(`Job ${job.id} was already claimed by a concurrent poll`)
+        continue
       }
+
+      // RETURNING * yields a plain pg row; hydrate it into a real Job so JobDto
+      // reads entity properties rather than whatever the driver handed back.
+      claimedJobs.push(new JobDto(this.jobRepository.create(claimedRow)))
```
DorianZheng: **This throws on every job we successfully claim.**

`create(row)` does not just copy fields onto a plain object. TypeORM builds the entity first and merges after: `Repository.create` -> `EntityManager.create` -> `EntityMetadata.create`, and that last one does `new this.target()` with **no arguments**.

Our `Job` constructor takes a required `params` and reads `params.id` on its very first line (`job.entity.ts:129`), so it throws `TypeError: Cannot read properties of undefined (reading 'id')`.

The `UPDATE` above has already committed by then, so:

- the row is left at `IN_PROGRESS`
- the runner gets a 500 and no job
- no later poll can pick it up again, because the `WHERE` only matches `PENDING`
- it only clears when `handleStaleJobs` fails it minutes later

This is the first code in the repo that ever calls `Repository.create()` on `Job`, which is why the constructor never mattered before. Upstream declares it as `params?` with `if (!params) return` on the first line; that guard was dropped when `apps/api` was first vendored here.

`job.service.claim.spec.ts` passes because it mocks `create` as an identity function, so the real constructor never runs.

Suggestion: skip the hydration entirely (re-`find()` the claimed ids), or put the `if (!params) return` guard back on the constructor.

apps/api/src/box/services/job.service.ts
```diff
@@ -469,24 +474,40 @@ export class JobService {
...
-        // If optimistic lock fails, job was already claimed by another runner - skip it
-        this.logger.debug(`Job ${job.id} already claimed by another runner (version mismatch)`)
+      const claim = await this.jobRepository
+        .createQueryBuilder()
+        .update(Job)
+        .set({ status: JobStatus.IN_PROGRESS, startedAt: now, updatedAt: now })
+        .where('id = :id', { id: job.id })
+        .andWhere('status = :pending', { pending: JobStatus.PENDING })
+        .returning('*')
+        .execute()
```
DorianZheng: **Each claim commits on its own — nothing wraps the loop in a transaction.**

If anything throws part way through (a later iteration, the DTO conversion, a DB hiccup), the jobs already claimed stay `IN_PROGRESS` in the database but are never returned. The method rejects as a whole, so the caller gets nothing back.

Those jobs are then invisible to polling, which only looks for `PENDING`. They sit until `handleStaleJobs` fails them.

Today the crash on line 510 hides this, because it fires on the first claim of any batch. It is still a separate bug, and it becomes the live one as soon as that is fixed.

Suggestion: claim the whole batch in one statement — `UPDATE job SET ... WHERE id = ANY($1) AND status = 'PENDING' RETURNING *` — or wrap the loop in a transaction so a failure rolls all of it back.

src/boxlite/src/litebox/box_impl.rs
```diff
@@ -1104,6 +1146,15 @@ impl BoxImpl {
...
             state.exit_code = None;
+            // Same reasoning for the start record, and this is the write that
+            // makes it safe: publishing the new PID and voiding the previous
+            // lifecycle's evidence happen in one `save_box`, so no reader can
+            // ever see this shim paired with the last one's start. Adopting a
+            // *running* box is not a new lifecycle — its record still names the
+            // live shim and must survive.
+            if !adopting_running {
+                state.started_at = None;
+            }
```
DorianZheng, line 1157: **This treats `adopting_running` as "init already ran". A bare `attach()` makes that untrue.**

`attach(None)` boots the box and creates the container but deliberately does **not** run `Container.Start` — and it still persists `Running` (see the comment at :281, and the new `tests/started_at.rs` case asserting `is_running()` with `started_at == None`).

For a box in that state, if the host process dies while the shim survives:

1. recovery adopts `Running`
2. `adopting_running` is true at :1074 — it reads `state.status` and nothing else
3. :1105-1106 presets `container_start`
4. the next `start()` reaches `get_or_try_init` at :992 and short-circuits
5. `Container.Start` is never sent, `record_started()` never runs, `start()` returns `Ok(())`

The workload silently never runs and `started_at` stays `None`. No error, no log, no retry.

To be clear, this block is byte-identical on `main`, so the PR does not introduce it. But the PR's own doc comment on the new field says "Running alone cannot say whether a box's init was ever launched" — and the gate at :1074 still decides from status alone.

Suggestion: preset the cell only when a start is actually proven (e.g. gate on `started_at.is_some()`), not on status. Worth an attach -> host crash -> recover -> start regression test; nothing covers that path today.

apps/runner/pkg/services/box_sync.go
```diff
@@ -137,6 +210,28 @@ func (s *BoxSyncService) PerformSync(ctx context.Context) error {
...
+// never launched — so a transitional box additionally needs BoxLite's durable
+// record that Container.Start did succeed for the shim now running it. Both
+// facts come from one `BoxInfo`, so they cannot describe two lifecycles.
+//
+// Every other remote state keeps the long-standing behaviour: the local view
+// wins unconditionally.
+func (s *BoxSyncService) canReport(local localContainerState, remoteState apiclient.BoxState) bool {
+	switch remoteState {
+	case apiclient.BOXSTATE_CREATING, apiclient.BOXSTATE_STARTING:
+		return local.state == enums.BoxStateStarted && local.startedAt != nil
```
DorianZheng, line 229: **This requires the box to still be `Started` right now, but `started_at` outlives the run.**

`started_at` is kept on purpose after a workload exits: `mark_stop()` / `mark_failed()` in `state.rs` clear the PID and set the status but never touch it. And the watcher calls `mark_stop()` by itself when the shim dies, so this is not limited to a user-requested stop.

So a box that started fine and then exited quickly, before the API heard about it, is now locally `Stopped` or `Error`. `local.state == enums.BoxStateStarted` is false, this returns false, and `PerformSync` hits `continue` at :191 — **before** the "State mismatch" log. Nothing is reported and nothing is logged, and it repeats on every tick, since neither field ever changes again. The startup job stays open until it times out.

Worth saying this is still a clear improvement: before this PR, `GetRemoteBoxStates` did not fetch `CREATING`/`STARTING` boxes at all, so none of them were ever reconciled. This just stops short of the fast-exit case.

Suggestion: treat `startedAt != nil` as the start evidence on its own, then report whatever the current state actually is. `TestPerformSyncConfirmsTransitionalBoxOnlyWithAConfirmedStart` only covers `StateRunning` with and without `StartedAt`; a `Stopped`/`Failed` box that still carries `StartedAt` is not covered.

apps/api/src/box/services/box.service.ts
```diff
@@ -1272,7 +1326,31 @@ export class BoxService {
...
+      }
+
+      // Completing the job is what moves the box: the normal completion
+      // handler owns the STARTED transition, the pending flag, the activity
+      // stamp, the state-change event, and releasing the box's Redis lock.
+      // Doing any of that here instead would fork that logic.
+      this.logger.warn(
+        `Completing stalled startup job ${stalledStartupJob.id} for box ${boxId} from runner-reported state`,
+      )
+      await this.jobService.updateJobStatus(stalledStartupJob.id, JobStatus.COMPLETED)
```
DorianZheng, line 1352: **The job is marked done before the box is actually repaired, and it cannot be retried afterwards.**

`updateJobStatus` commits `COMPLETED` inside its transaction, then fires `handleJobCompletion` as fire-and-forget (`job.service.ts:275`). The handler that does the real work, `handleStartBoxJobCompletion`, catches its own errors and only logs them (`job-state-handler.service.ts:163`) — it never rethrows.

So if that handler fails, or the process dies between the commit and the handler running, the box never reaches `STARTED`.

There is no second chance. `findStalledStartupJob` only matches `status: IN_PROGRESS` (:597), and the job is `COMPLETED` now, so the next sweep will not find it. The runner keeps reporting a box the API will never fix. It is stuck permanently.

Suggestion: keep the job reconcilable until the box change is durable — complete the job and move the box in one transaction, or leave a retry/outbox marker a later sweep can pick up.

DorianZheng (commented):
Five findings from an adversarial review pass, verified against this branch. Ordered by severity: the first is a hard blocker on the normal polling path.

A sixth finding (appending `started_at` to `CBoxInfo` changes `sizeof` for old C list clients) was checked and dropped — no SONAME/dylib versioning, the Go SDK links statically against a version-pinned header+lib pair, and #1009/#1031 set the same precedent already. Not raising it.

apps/api/src/box/services/box.service.ts
ltstriker: The commit-before-handler window is real, but after tracing the existing reconciliation path top-down, I don’t think “permanently stuck” or “no second chance” is accurate.

`findStalledStartupJob()` is not the only recovery mechanism. The existing desired-state reconciler can converge the box after the job has already been committed:

1. `job.service.ts:238-269` commits the job as `COMPLETED`.
2. `job.service.ts:272-277` invokes `handleJobCompletion()` after the transaction. If that handler is lost or transiently fails, the box remains `CREATING`/`STARTING` with `desiredState = STARTED`.
3. `box.manager.ts:227-249` runs `syncStates()` every 10 seconds and selects boxes whose `state != desiredState`.
4. `box.manager.ts:271-275,314,349-352` routes such a box through `syncInstanceState()` and then `BoxStartAction`.
5. `box-start.action.ts:124-131` calls `runnerAdapter.boxInfo()` and updates the box to `STARTED` when the adapter reports that state.
6. For a v2 runner, `runnerAdapter.v2.ts:61-92` first looks for an incomplete job. Since this job now has `completedAt`, it falls through to the latest job.
7. `runnerAdapter.v2.ts:100-106` maps a completed `CREATE_BOX` or `START_BOX` job to `BoxState.STARTED`.
8. The subsequent repository update applies the normal box invariants and clears `pending` once `state == desiredState` (`box.entity.ts:260-267`).

Therefore, in the failure window described here, the system is eventually consistent: the box normally converges to `STARTED` on the next reconciliation tick—typically within 0–10 seconds plus scheduling and lock delays. A process restart or stale Redis lock can extend that delay, and a persistent storage failure would still prevent convergence, but this is not a permanent liveness failure.

There is also an idempotent replay path: `COMPLETED → COMPLETED` is accepted at `job.service.ts:380-383`, which triggers the completion handler again if the runner retries the status callback.

So I agree that the asynchronous boundary deserves explicit recovery coverage, but I would characterize this as a temporary inconsistency / missing regression test rather than an unrecoverable state. A useful test would commit the job as `COMPLETED`, deliberately skip or fail the completion handler, run `syncStates()`, and assert that the box converges to `STARTED`.

apps/runner/pkg/services/box_sync.go
ltstriker: This problem should not be handle here. This pr just handle one problem: the box which long-term running but never billing for it's `started msg` have not been sent and handle by API. I keep the state reconciliation finally, in fact it will late, not immediately which should be made by normal path.

src/boxlite/src/litebox/box_impl.rs
ltstriker: It is right. It may prevent the billing, but i think it can be fixed by next action which will move the `started_at` mark into the `boot` phase which will make the started just before the mark running and billing what you said.

And also, i think it is an important bug, if we do the `started_at` mark move. Because we will bill the box, but in fact it is not really started for the user.

apps/api/src/box/services/job.service.ts
ltstriker: fixed. old test use a mock for it so i have not find it.

apps/api/src/box/services/job.service.ts
ltstriker: fixed，  i use one sql to `update` them together.

apps/api/src/box/services/job.service.ts
```diff
@@ -469,24 +474,35 @@ export class JobService {
...
+        status: JobStatus.PENDING,
+      },
+      {
+        status: JobStatus.IN_PROGRESS,
+        startedAt: now,
+        updatedAt: now,
+      },
+      { returning: ['id'] },
+    )
+    const claimedIds = new Set((claim.raw as Array<Pick<Job, 'id'>>).map((row) => row.id))
```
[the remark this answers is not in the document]
DorianZheng: @ltstriker PTAL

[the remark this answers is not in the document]
ltstriker: we use `TypeORM 0.3.31`, it can support `UpdateOptions.returning`.  look apps / package.json.   @DorianZheng

DorianZheng (approved):
LGTM

PR #1097, opened 2026-07-30
feat(infra): land runner binary upgrades via SST, one host at a time

DorianZheng:
## Summary

The Runner EC2s carry `ignoreChanges: ['ami','userDataBase64']`, so a Cargo.toml version bump was detected but never acted on. The only remedy was `scripts/deploy/runner-update-binary.sh` — wired to nothing, and matching the default runner's tag alone, so extra runners drifted. This moves that script into `apps/infra` as an SST resource: one `dependsOn`-chained `UpgradeRunnerBinary-*` command per Runner, so the dependency graph rather than a shell script sequences the restarts.

## Changes

- `scripts/deploy/runner-update-binary.sh` → `apps/infra/scripts/runner-update-binary.mjs`, declared per-Runner in `sst.config.ts` and reachable out of band via `npm run runner:update`.
- Corrects two faults in the ported script: the Go runner parses no CLI args, so `boxlite-runner --version` made a *successful* upgrade report SSM `Failed`; and a failed run discarded its own diagnostics, because the unguarded `aws ssm wait` aborted under `set -e` before the log fetch. The health route is now the version oracle, and status is decided after the logs print.
- Converges rather than reinstalls — a host already on the target, still bootstrapping, or serving something newer is left untouched, so this is free to carry on every deploy. `ALLOW_DOWNGRADE=1` forces a rollback.
- Reuses `resolveRunnerReleaseAssets` and `readWorkspaceVersion` instead of re-deriving URLs and parsing Cargo.toml. This makes the `.sha256` sidecar mandatory and name-matched (previously warn-and-proceed) and rejects prerelease targets.
- Replaces the README's runner-upgrade and decommission prose, which described a control-plane drain flow that no code implements.

Two commits are mechanical and reviewable at a glance: `chore(apps): apply prettier…` (the whole-`apps` formatter runs on any commit touching that tree) and `chore(apps): record go.work.sum…` (the workspace Go tests append these, then fail the pre-push hook).

## How to verify

```bash
cd apps/infra && npm test        # 88 pass; 20 cover the new script + SST shape
```

The interesting tests execute the *emitted* payload rather than a copy: the checksum `awk` runs against six manifest shapes, and the version comparison against six semver cases. `sst-config-contract.test.mjs` pins the `dependsOn` chain, which is otherwise only observable on a real deploy.

Against a disposable EC2 runner (since terminated), seven host-side paths were exercised: the bootstrapping guard, skip, asset-404-before-stop, both swap variants, a readiness-timeout rollback restoring the exact prior binary hash, and abort-the-roll across two targets. The downgrade guard was checked against the live dev runner, which it refused, leaving the host byte-identical.

## Risks / rollout

- **Unobserved through Pulumi.** No `sst diff` completed — `apps/infra/.env` is absent from every checkout, and with 52 `envOr` defaults a reconstructed one would silently propose reverting live config. So the `dependsOn` sequencing, the no-EC2-replacement expectation, the cwd-relative `create`, and the Output-valued `INSTANCE_IDS` rest on in-repo precedent (`RegisterExtraRunners`) plus the contract test. Worth a `sst diff` from someone holding the real `.env` before merge.
- **Runners are not drained before restart.** Boxes on the host being upgraded take the restart; `TimeoutStopSec=60` only buys a graceful VM shutdown. Cordoning needs a control-plane runner id, an operator key, and the organization-infrastructure flag, none of which the deploy path has.
- **`--exclude Runner` no longer suppresses the binary upgrade** — the commands are separate components. It fails closed (a missing asset stops the download before the unit is touched) but turns a Runner-free rollout into a failed deploy. `--target Api` is unaffected. Documented with the workaround.
- Deploying this converges dev's runner from its current out-of-band build down to the declared `Cargo.toml` version and restarts it.

Unrelated, found while verifying the docs and left alone: the admin UI's drain button PATCHes `/admin/runners/{id}/draining`, but the admin controller exposes no such route — the only one is on `@Controller('runners')`.

## Summary by CodeRabbit

* **New Features**
  * Added automated, sequential Runner binary upgrades with version checks, health validation, checksum verification, and rollback support.
  * Runner upgrades now rerun automatically when the workspace version or Runner instance changes.
  * Added safeguards against unintended downgrades and unnecessary upgrades.

* **Documentation**
  * Updated deployment guidance for upgrades, health checks, rollback behavior, and Runner decommissioning.

* **Tests**
  * Added comprehensive coverage for upgrade targeting, version handling, command execution, safeguards, and infrastructure wiring.


DorianZheng:
All nine findings addressed in `f8c8c201`. Each was re-checked against the tree first; none named stale code.

**Actionable**

- **`runnerPort` unvalidated** — rejected on the deployer now (`/^[0-9]+$/`), so a typo fails there rather than inside a curl on the host.
- **Stale `lastFailure`** — cleared on a successful poll, so a transient `InvocationDoesNotExist` can no longer outrank ten minutes of later `InProgress` and blame an error that had already resolved.
- **Unguarded section marker** — fixed as a class, not a site. Every marker lookup in `sst-config-contract.test.mjs` now goes through one `sliceBetween(text, label, start, end)` that asserts each marker and names the missing one; the same shape in `runner-update-binary.test.mjs` is guarded too. Worth noting one of those was a silent green, not just an unhelpful red: `assert.doesNotMatch(appSource, /buildx-builder/)` passes vacuously on an empty slice.
- **Multi-Runner exclude example** — the example now names every generated command, with `RUNNERS=1` and `RUNNERS=3` forms, and points at `--target Api` as the option with no per-Runner bookkeeping.
- **Cordon before counting boxes** — the decommission ceremony now starts by setting `unschedulable`, and says why `draining` alone is insufficient: `findAvailableRunners` filters both flags, but the warm-pool query (`box-warm-pool.service.ts:91`) consults only `unschedulable`.

**Nitpicks**

- **Entrypoint check** — the most consequential of the set, and it had a second trigger neither of us named. `pathToFileURL` alone is not enough: `import.meta.url` is *also* symlink-resolved, so on macOS `argv[1]` of `/var/folders/…` never matches `/private/var/folders/…`. Now `pathToFileURL(realpathSync(argv[1]))`. Both failure modes are covered by a test that runs the real script from a spaced temp path and requires it to reach the missing-CLI error; reverting to either intermediate form turns it red.
- **Throttling** — `ThrottlingException`, `TooManyUpdates` and `RequestLimitExceeded` retry alongside `InvalidInstanceId`. The retry log line was reworded, since "not reachable via SSM yet" is wrong for a throttle.
- **Discovery order** — sorted, so a fleet-wide roll visits hosts in the same sequence each run.
- **Assert the `send-command` shape** — declined. The argument list is already pinned indirectly by the tests that decode the base64 payload out of it, and asserting the literal argv would mostly restate the implementation.

Suite is 92 passing. Each behavioural fix has a test that fails when that fix alone is reverted.

One sibling left deliberately: `release-deployment-safety.test.mjs:94` has the same `indexOf` shape, but this branch doesn't otherwise touch that file and its assertion fails loudly rather than losing coverage. Worth a follow-up sweep.

DorianZheng:
Correcting myself on one item above, and closing it properly in `81870fcc`.

I declined the "assert the `send-command` argument shape" nitpick on the grounds that "the tests already decode the base64 payload out of those arguments." **That was false.** No test decoded anything — every `run` stub was a zero-argument arrow, so the argv was never even received. The argument list was pinned by nothing.

So it is asserted now: `ssm send-command`, `--document-name AWS-RunShellScript`, `--instance-ids`, `--query Command.CommandId`, and the `commands=["echo <base64> | base64 -d | bash"]` shape.

Writing it turned up something worth keeping: the obvious assertion — decode the base64 and compare to the script — is too weak. Mutating the code to emit `${payload}X` leaves it green, because Node's base64 decoder tolerates trailing junk and still yields the original bytes. The test compares the base64 string exactly instead, and that mutation now turns it red.

Same commit also covers `.sort()` on discovered instances, which I had claimed was test-backed and wasn't. `resolveTargets` gained a `describe` seam matching the `run`/`sleep` seams already in the file, and three tests: an explicit `INSTANCE_IDS` must not consult AWS at all, discovery is sorted while an explicit list keeps the operator's order, and an empty discovery throws rather than silently doing nothing.

Suite is 95 passing. Every behavioural change on this branch now has a mutation that turns exactly one test red.

DorianZheng:
Retracting the last line of my previous comment. "Every behavioural change on this branch now has a mutation that turns exactly one test red" is **false**, and I should have checked it before writing it.

Deleting any of these from the emitted payload leaves the suite at 95 passing:

- the downgrade refusal (`if [ "${ALLOW_DOWNGRADE:-}" != "1" ] && live_is_newer; then … exit 0; fi`)
- the backup and rollback (`cp -a … boxlite-runner.bak`, and the `mv -f` restore)
- the readiness gate (`&& wait_for_target` in the install/start guard)

`grep` in `runner-update-binary.test.mjs` finds no reference to `refusing to downgrade`, `boxlite-runner.bak`, `wait_for_target`, `HAD_PREVIOUS`, `rolling back`, or `systemctl is-active`. The `ALLOW_DOWNGRADE` test pins the variable assignment only, and `live_is_newer` is tested as a function nothing requires the payload to call.

What is genuinely mutation-covered: the bootstrapping guard, the checksum comparison, the awk filename escaping, the entrypoint guard, port validation, the `lastFailure` clearing, the throttling retry, discovery ordering, the `send-command` argv, empty discovery, and the four `sst.config.ts` wirings.

What is not: most of the remote payload's control flow — rollback in particular, which is the safety net. Those paths were exercised by hand against a real EC2 runner (swap, readiness timeout with the prior binary restored, three integrity aborts, abort-the-roll) and with a stubbed-host harness, but that harness lives outside the repo, so nothing guards them in CI.

Porting it in is the obvious follow-up. Flagging rather than folding it into this PR, which is already carrying more than the move it started as.

PR #1101, opened 2026-07-31
feat(archive): shared layer store with reference-counted sweep

G4614:
Adds a shared layer store — `archives/<name>.json` manifests over one content-addressed `layers/` pool — where archives of different boxes share their common layers, manifests are the references, and `ArchiveStore::gc` sweeps unreferenced objects with a grace period plus publish-side repair to close the publish/sweep race.

Stacked on #1080: the first 10 commits are that PR; only the last commit is new here.

Test plan:
- [x] `cargo nextest run -p boxlite --features krun --test clone_export_import --profile vm` — 11/11 on real VMs, including `test_store_export_import_roundtrip` (publish into a store → import from the manifest path → boot → remove + gc empties the pool)
- [x] `cargo nextest run -p boxlite --no-default-features --lib` — 976/977; the one failure (`seatbelt` sandbox) reproduces on a pristine tree
- [x] `a_layer_stays_while_any_archive_references_it` — two manifests sharing a layer: removing one frees only its unique layer; the pool empties only after the last reference is gone
- [x] `a_store_frees_a_layer_only_with_its_last_reference` — same property through the real export path: two publishes of one chain share every object (verified by mtime), gc frees nothing until both names are removed
- [x] `a_corrupt_manifest_aborts_the_sweep` — unparseable manifest ⇒ gc refuses and deletes nothing (fail closed)
- [x] `an_unreferenced_object_inside_the_grace_period_is_kept` — young orphans survive the sweep; `gc(0)` takes them
- [x] `foreign_files_in_the_pool_are_left_alone` — only content-named `.zst` and stale `.partial` files are sweep candidates
- [x] `a_single_archive_directory_is_refused` — a root `manifest.json` cannot be opened as a store
- [x] `a_path_escaping_archive_name_is_refused` — `../evil`, `a/b`, `.hidden`, empty
- [x] `cargo fmt --all -- --check`; `cargo clippy -p boxlite --no-default-features --all-targets -- -D warnings`; `cargo check -p boxlite -p boxlite-python -p boxlite-node`
- [ ] Concurrent publish/sweep race not exercised by a stress test; covered by design (grace period + post-manifest repair, both unit-tested individually)


G4614:
Folded into #1080 — the store work is stacked on that format and the review decision was to land them as one PR. All commits from this branch are now on `feat/layered-archive` (through 0a2ae993), including the import hardening merge.

PR #890, opened 2026-06-30
feat(runner): add drain guard for rollouts

G4614:
Add runner drain mode so rollout tooling can reject new box operations while keeping health checks observable.

Test plan:
- [x] gofmt -w apps/runner/cmd/runner/main.go apps/runner/pkg/api/controllers/box.go apps/runner/pkg/api/controllers/boxlite_exec.go apps/runner/pkg/api/controllers/boxlite_exec_attach.go apps/runner/pkg/api/controllers/boxlite_exec_attach_test.go apps/runner/pkg/api/controllers/boxlite_exec_test.go apps/runner/pkg/api/controllers/drain.go apps/runner/pkg/api/controllers/drain_test.go apps/runner/pkg/api/server.go apps/runner/pkg/drain/drain.go apps/runner/pkg/runner/v2/healthcheck/healthcheck.go
- [x] git diff --check origin/main...HEAD
- [x] go test ./runner/pkg/drain ./runner/pkg/api/controllers ./runner/pkg/runner/v2/healthcheck

## Summary by CodeRabbit

* **New Features**
  * Added admin controls to view and enable or disable runner draining.
  * Drain status includes the number of active attach sessions.
  * New workloads and execution sessions are rejected with HTTP 503 while draining.
  * Existing attach sessions continue to be tracked until they close.

* **Behavior Changes**
  * Health checks and runtime shutdown are skipped while draining.
  * Added coverage for drain status, request rejection, and active session tracking.

PR #254, opened 2026-02-13
feat(api): add OpenAPI 3.1.1 spec for cloud sandbox REST API

DorianZheng:
## Summary

- Add OpenAPI 3.1.1 specification for the BoxLite cloud sandbox REST API at `boxlite/openapi/rest-sandbox-open-api.yaml`
- API design follows industry-standard REST patterns (versioned prefix, multi-tenancy, capability discovery, idempotency keys, standard error model, pagination)
- 24 endpoints across 7 resource groups, all schemas mapping 1:1 to existing Rust types

## API Design

| Group | Endpoints | Description |
|-------|-----------|-------------|
| Configuration | `GET /v1/config` | Server capability discovery |
| Auth | `POST /v1/oauth/tokens` | OAuth2 client credentials |
| Boxes | 7 endpoints | CRUD + start/stop lifecycle |
| Execution | 6 endpoints | Async exec, SSE streaming, stdin, signals, TTY (WebSocket) |
| Files | 2 endpoints | Upload/download via tar streams |
| Metrics | 2 endpoints | Runtime-wide + per-box metrics |
| Images | 4 endpoints | Pull, list, get, exists check |

### Key Design Decisions

- **Multi-tenancy**: `{prefix}` path segment = organization/workspace ID
- **Execution**: Async only — `POST /exec` returns `execution_id`, stream via SSE
- **TTY**: WebSocket at `/exec/tty` for interactive terminal sessions
- **Files**: Binary tar stream (matches existing `FilesInterface`)
- **Errors**: Standard `{message, type, code}` model mapping all `BoxliteError` variants

## Test plan

- [x] YAML syntax validated (Python yaml.safe_load)
- [x] All `$ref` references resolve correctly (24 schemas, 7 parameters)
- [x] All `BoxliteError` variants mapped to error types
- [x] All `BoxOptions` fields represented in `CreateBoxRequest`
- [x] All `BoxInfo` fields represented in `Box` response schema
- [ ] Lint with `npx @redocly/cli lint` (optional, CI)
- [ ] Generate preview docs with `npx @redocly/cli preview-docs` (optional)

PR #1102, opened 2026-07-31
test(e2e): pin REST host bind mounts as rejected, not silently dropped

G4614:
Test-only: pins the REST host-bind-mount contract as an explicit rejection instead of a silent drop, which is what #1192 actually made it.

Test plan:
- [x] `nx run api:test --testPathPatterns=box-to-box.mapper.spec.ts` — 11/11 pass, including `rejects source values without a supported scheme`, the unit-level proof that a bare host path 400s at the mapper
- [x] `python3 -m py_compile apps/e2e/cases/test_volume_host_bind_mount.py`
- [ ] The e2e case itself — needs a live API (the `rt` fixture goes through `auth_context()`); has to go green in CI

Note: `read_only?: false` with `@IsIn([false])` means REST cannot express a read-only managed volume at all, while the FFI and Rust layers can. Real gap, tracked separately — not this PR.

Previously this PR carried the `source` → `volume_id` rename. Those 6 commits netted to a zero diff against main after the final revert; they are archived at `G4614/boxlite@archive/pr1102-volume-id-rename` and remain fetchable from `refs/pull/1102/head`.


G4614:
Supersedes #890 after rebasing the original change onto the latest `main` from the G4614 fork.

G4614:
Both fixed in 9410816:
1. `CreateBoxVolumeSpec::from` now strips a leading `volume://` before assigning `volume_id` — confirmed the Node/Python SDK bindings do keep their `source` field's scheme verbatim in `VolumeSpec.host_path` (their own public contract predates this rename), so this was a real regression for SDK callers, not just a CLI-only concern. Two-side verified with a new test.
2. `VolumeSpec`'s OpenAPI schema now has `anyOf` (volume_id or host_path), `minLength: 1` on volume_id, and a `volume://` pattern on host_path, matching the constraints the DTO already enforces.

G4614:
Design revision (f20fdfbb) based on further discussion: `host_path` no longer doubles as a deprecated `volume_id` alias. Traced its git history — it entered this DTO in #254 genuinely meaning "path on the host filesystem" (matching the Rust core/Node SDK's own field of the same name); #1192 repurposed the same field name into a volume reference without renaming it. Since neither form has ever shipped in a release, this reverts `host_path` to its original meaning and rejects it outright (not implemented for REST — a remote runner has no notion of "the caller's local host"). `volume_id` is now the sole, required way to reference a managed volume. See the updated PR description for the full before/after.

openapi/box.openapi.yaml
```diff
@@ -1820,26 +1820,21 @@ components:
...
       type: object
-      description: |
-        Filesystem mount attached to a box. One of `source` or the deprecated
-        `host_path` is required.
-      required: [guest_path]
+      description: Managed-volume mount attached to a box.
+      required: [volume_id, guest_path]
       properties:
-        source:
+        volume_id:
```
DorianZheng: how about just volume and it accepts both id_or_name

G4614: good idea, i will modify as this

G4614:
Good call — renamed `volume_id` to `volume` in 122b476a. To be precise about the part that matters: it already accepted both id and name at the validation layer (`VolumeService.validateVolumes` resolves either against the org's volumes), so the old field name was overclaiming "id-only" rather than the field being id-only in practice. The rename just makes the name honest about that. Renamed end to end — DTO, OpenAPI schema, and the Rust wire type — so the JSON key matches everywhere.

PR #1105, opened 2026-07-31
fix(gvproxy)!: apply allow_net to the host alias IP

ltstriker:
## Summary

`192.168.127.254` — the built-in host alias that gvproxy NATs to host loopback — was exempt from `allow_net`, so a box with a restrictive allowlist could still reach every service bound to `127.0.0.1` on the host, over both TCP and UDP. It is now an ordinary egress destination.

## Changes

- Drop `config.HostIP` from the filter's always-allow set. The alias is a NAT to host loopback, not a virtual-network endpoint like the gateway and guest addresses, so it belongs under policy.
- Build the filter through a single `newAllowNetFilter(config)`. The Go test harness previously duplicated the argument list, which is why a one-line policy change was invisible to `go test`; all test construction now routes through the production helper.
- Resolution is unchanged: the built-in DNS record for `host.boxlite.internal` is still served unconditionally — resolving the alias is not egress. Reaching it requires listing `"192.168.127.254"` or a covering CIDR.
- Tests: tap-level reproducers for both transports plus a routing guard that pins pre-NAT policy matching against post-NAT dialing; the four Rust integration cases now cover blocked-and-listed on both TCP and UDP.
- Docs: four sites stated the alias was not governed by `allow_net`.

## How to verify

```
make test:unit:gvproxy FILTER='HostAlias'
make test:integration:rust FILTER='host_alias'
```

Manually: start a box with `allow_net: ["example.com"]`, then `wget http://host.boxlite.internal:<port>/` against a host loopback listener — the connection is refused while `nslookup host.boxlite.internal` still returns `192.168.127.254`. Adding `"192.168.127.254"` to `allow_net` restores it.

## Risks / rollout

Breaking change. Existing callers that set a non-empty `allow_net` and rely on `host.boxlite.internal` must add `"192.168.127.254"`. Boxes with an empty or omitted `allow_net` are unaffected.

Granularity is address-level: allowing the alias opens every port bound to host loopback. `allow_net` has no port dimension, so this is not expressible more narrowly today.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -303,8 +303,8 @@ See [Configuring Networking](./guides/README.md#configuring-networking) for deta
 2. **Use host network:**
    - Box A exposes port
    - Box B connects to `host.boxlite.internal:port`
-   - This bypasses `allow_net`; if networking is enabled, host loopback
-     services are reachable from inside the box
+   - With an empty `allow_net`, host loopback services are reachable from
+     inside the box; a non-empty `allow_net` must list `"192.168.127.254"`
 
 3. **External service:**
    - Both boxes connect to Redis/database on host or network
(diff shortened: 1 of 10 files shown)


ltstriker:
Superseded by #1106 — same commit, branch renamed from `fix/egree_disble_local_default` to `fix/allow-net-host-alias`.

PR #1107, opened 2026-07-31
docs(contributing): mandate before/after call graph in PR descriptions

DorianZheng:
## Summary

Every PR description must now carry the end-to-end call graph it touches, before and after — the same shape already used to explain code in this repo. Bug fixes additionally mark the faulty hop and link the issue. The rule is documented in three places and mechanically enforced by the existing PR preflight hook.

## Call graph

Before — line numbers as of `origin/main`

```text
gh pr create --body …
  └─ matcher            (preflight-pr-review.sh:52)   — gh pr create|edit|ready
       └─ title check   (preflight-pr-review.sh:103)  — Conventional-Commit ≤72
            └─ ack gate (preflight-pr-review.sh:189)  — user-typed `reviewed:`
                                                        body never inspected
```

After — line numbers as of this branch

```text
gh pr create --body …
  └─ matcher            (preflight-pr-review.sh:57)   — gh pr create|edit|ready
       └─ title check   (preflight-pr-review.sh:108)  — Conventional-Commit ≤72
            └─ body extraction (preflight-pr-review.sh:126)
            │                                         — --body / --body-file / heredoc
            └─ graph check     (preflight-pr-review.sh:148)
            │                                         — deny unless the body has
            │                                           `## Call graph`, Before, After,
            │                                           and a parenthesised
            │                                           path/file.ext:LOC hop inside
            │                                           each graph
            └─ fix: extras     (preflight-pr-review.sh:226)
            │                                         — `BUG:` on a hop line inside the
            │                                           Before graph, plus `Fixes #<n>`
            └─ ack gate (preflight-pr-review.sh:341)  — unchanged
```

The per-section hop requirement carries the weight: an unfilled template has the headings but zero hop lines, a graph reduced to prose cannot borrow the other one's hops, and an incidental file mention in a sentence is not a hop. Commands with no body flag (`gh pr ready`) are untouched; `--fill` is denied, because commit text is not a PR description.

## Changes

- `CONTRIBUTING.md` — the rule plus a worked example under **Commit & PR messages**; single source of truth for the format.
- `.github/pull_request_template.md` — a `## Call graph` skeleton (Before / After / `Fixes #`), pointing at CONTRIBUTING for the example rather than repeating it.
- `CLAUDE.md` — one pointer bullet under **Communication**.
- `.claude/hooks/preflight-pr-review.sh` — the body check above. Additive: it runs before the ack gate and does not consume the ack marker.
- `.claude/hooks/preflight-pr-review.test.sh` — 24 body cases covering prose bodies, a missing *After*, labels without `file:LOC`, a prose-only graph, hops outside both graphs, the unfilled template, `--fill`, the bug-fix omissions, and the shapes that should pass.

### Review rounds

**`6a4a6b1b`** — the bug marker was accepted anywhere in the body, so a `fix:` PR passed with it sitting in prose or in the *After* graph, and `debug:` satisfied it outright. It now has to sit on a hop line inside *Before*. The arrow glyph is deliberately not policed — `←`, `<-` and a bare `BUG:` are all accepted, since mandating one Unicode character is typing friction rather than signal; `CONTRIBUTING.md` still prescribes `←`, so the style guide stays stricter than the linter.

**`c82ef77a`** — hops were counted body-wide, so two *Before* hops carried a prose-only *After*, and a pair of hop-free graphs passed on references sitting in a later section. Both graphs are now extracted and each must carry a hop of its own. Only the heading bound is fence-aware: an earlier cut gated the graph labels on fences too, which made CONTRIBUTING's own example — it wraps both labels and all hops in one fence — extract to nothing. The suite now lifts that example out of the doc at run time, so the documented shape and the enforced shape cannot drift.

**`43f7ae0f`** — the hop pattern matched a bare `path/file.ext:LOC`, which turns up mid-sentence, so `Before` / `After` labels over two lines of prose counted as two graphs. The reference must now sit between a `(` and a `)`. Not a balanced group: the left side stays permissive because a Type half carries its own parens in `(fn(u32) -> u32 · src/x.rs:5)`, and refusing to cross them denied a hop conforming to the documented shape. Nothing here can tell a real graph from a fabricated one — the typed ack is that control, and the code now says so instead of leaving the deny text to imply otherwise.

## How to verify

```
bash .claude/hooks/preflight-pr-review.test.sh   # 54 passed, 0 failed
```

Two-side check: with `preflight-pr-review.sh` restored from the preceding commit and only the test file modified, each round's new deny cases fail with `got=passthrough expected=deny`. Restored, 54/54.

## Risks / rollout

- A body containing a double quote truncates extraction and fails the checks — closed, not open, so the failure mode is a false denial rather than a bypass. Documented at the extraction branch; not covered by a test.
- An *unclosed* ``` fence suppresses the heading bound, letting a prose-only *After* borrow hops from a later section. Not a regression — the body-wide count accepted the same body — and it needs markdown that GitHub would render as a code block swallowing the rest of the description.
- `--body-file` is matched anywhere in the command and wins over `--body`, so chaining `git commit -F msg.txt && gh pr create --body "…"` reads the commit-message file as the PR body and denies. Same fail-closed trade-off already documented for `--draft`.
- Graph labels are matched at line start anywhere in the body, so a prose line beginning `Before ` or `After ` re-partitions the graphs. Pre-existing: the label presence checks have the identical looseness.
- The hop pattern is not a balanced group — an unrelated parenthetical straddling the reference satisfies it. Accepted so that a Type half carrying its own parens is not denied; no pattern can tell a real graph from a fabricated one, which is what the typed ack is for.
- `--draft` continues to skip the hook entirely, unchanged.

### DIFF
--- a/.claude/hooks/preflight-pr-review.sh
+++ b/.claude/hooks/preflight-pr-review.sh
@@ -157,26 +157,50 @@ if (( body_supplied )); then
   has_line '^[[:space:]]*#{2,}[[:space:]]*call[[:space:]]+graph[[:space:]]*$' \
     || missing+="
   - a '## Call graph' section"
-  has_line '^[[:space:]]*before([[:space:]]|:|$)' \
+  # Shared with the Before-section extractor below, so the header shape can
+  # never drift between "is there a Before label" and "where does Before end".
+  before_re='^[[:space:]]*before([[:space:]]|:|$)'
+  after_re='^[[:space:]]*after([[:space:]]|:|$)'
+  has_line "$before_re" \
     || missing+="
   - a 'Before' graph"
-  has_line '^[[:space:]]*after([[:space:]]|:|$)' \
+  has_line "$after_re" \
     || missing+="
   - an 'After' graph"
 
   # Hops must be real: `fn_name (Type · path/file.ext:LOC)`. Requiring two
   # file:LOC references is what an unfilled template cannot fake — its Before
   # and After labels are literal text, but its hop lines are HTML comments.
-  hop_lines="$(grep -cE '[A-Za-z0-9_./-]+\.[A-Za-z]+:[0-9]+' <<<"$pr_body" || true)"
+  hop_re='[A-Za-z0-9_./-]+\.[A-Za-z]+:[0-9]+'
+  hop_lines="$(grep -cE "$hop_re" <<<"$pr_body" || true)"
   (( hop_lines >= 2 )) \
     || missing+="
   - 2+ hop lines carrying 'path/file.ext:LOC' (found ${hop_lines})"
 
   # Bug-fix extras — only decidable when --title was inspectable above.
   if [[ "$pr_title" =~ ^fix(\([^\)]+\))?!?: ]]; then
-    has_line 'bug:' \
+    # "Mark the faulty hop" is literal: the marker has to sit on a hop line
+    # inside the Before graph. A `BUG:` loose in prose, or down in After, marks
+    # nothing. The arrow is deliberately not required — `←`, `<-` and a bare
+    # `BUG:` all read the same, and mandating one Unicode glyph is typing
+    # friction, not signal. The word boundary keeps `debug:` from qualifying.
+    # Rule order is load-bearing: reset on After, then print, then set on
+    # Before — that sequence is what keeps the two header lines themselves out
+    # of the captured section, so a marker on the `Before` line marks no hop.
+    before_graph="$(awk -v before_re="$before_re" -v after_re="$after_re" '
+      $0 ~ after_re  { in_before = 0 }
+      in_before      { print }
+      $0 ~ before_re { in_before = 1 }
+    ' <<<"$body_lc")"
+    marker_re='(^|[^[:alnum:]_])bug:'
+    # Same line, either order: `hop … ← BUG: why` or `← BUG: why … hop`.
+    marked_hop() {
+      grep -qE "${hop_re}.*${marker_re}" <<<"$before_graph" ||
+        grep -qE "${marker_re}.*${hop_re}" <<<"$before_graph"
+    }
+    marked_hop \
       || missing+="
-  - fix: PR — '← BUG: <what goes wrong>' on the faulty hop in Before"
+  - fix: PR — '← BUG: <what goes wrong>' on a hop line inside the Before graph"
     has_line '(fixes|closes|resolves)[[:space:]]+#[0-9]+' \
       || missing+="
   - fix: PR — an issue link, 'Fixes #<n>'"
--- a/.claude/hooks/preflight-pr-review.sh
+++ b/.claude/hooks/preflight-pr-review.sh
@@ -26,6 +26,11 @@
 #   arbitrary string is rare and the failure mode is "let one PR through
 #   without ack," not a destructive action.
 #
+# * Two deterministic content checks run before the ack gate — a Conventional-
+#   Commit `--title`, and a `--body` carrying the before/after call graph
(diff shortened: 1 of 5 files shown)


DorianZheng:
Fixed in `6a4a6b1b`, with one part declined.

**Taken:** the marker check was body-wide, so a `fix:` PR passed with `BUG:` sitting in prose or in the *After* graph — neither marks a faulty hop — and `debug:` satisfied it outright. The hook now extracts the Before section (`awk`, sharing the same header regexes as the presence checks so the two cannot drift) and requires the marker on a line that also carries a `path/file.ext:LOC` hop, word-bounded. Three new deny cases pin it: marker in *After*, marker off any hop line, and `debug:` alone. Verified two-side — with the hook restored from `origin/main` and only the test file changed, those three fail `got=passthrough expected=deny`; suite is 47/47 with the fix.

**Declined:** requiring the literal `←` and rejecting `<- BUG:`. Mandating one Unicode glyph is typing friction, not signal — what carries meaning is the marker's *position*, which is now enforced exactly. `←`, `<-` and a bare `BUG:` are all accepted, each pinned by its own test. `CONTRIBUTING.md` still prescribes `←` as canonical, so the style guide stays stricter than the linter and no doc-conformant body is ever rejected.

DorianZheng:
Fixed in `c82ef77a`. Both reports were the same defect and it reproduced: `hop_lines` counted `path/file.ext:LOC` matches over the whole `$pr_body`, so two Before hops carried a prose-only After — and a pair of hop-free graphs passed on references sitting in a later `## Changes` section.

Both graphs are now extracted and each must carry a hop of its own. The Before extractor that already existed for the bug-marker check became a shared `section` helper, bounded by the other graph's label or the next markdown heading. Three deny cases pin it: prose-only After, prose-only Before, and hops outside both graphs — all three fail `got=passthrough expected=deny` against the previous commit.

Two things worth flagging from working through it:

- Only the **heading** bound is fence-aware. An earlier cut of this fix gated the graph labels on fences too, which made `CONTRIBUTING.md`'s own worked example — it wraps both labels and all hops in one ```text fence — extract to nothing and deny "found 0 in Before, 0 in After". The suite now lifts that example straight out of the doc at run time and drives it through the hook, so the documented shape and the enforced shape cannot drift apart.
- Known residual, not a regression: an *unclosed* fence suppresses the heading bound, letting a prose-only After borrow hops from a later section. The previous body-wide count accepted that body too. It needs markdown that GitHub would render as a code block swallowing the rest of the description, so I've left it rather than guess what an unclosed fence means at EOF.

Suite is 52/0.

DorianZheng:
Fixed in `43f7ae0f`. Reproduced first: the prose body you pasted cleared the gate.

`hop_re` matched a bare `path/file.ext:LOC`, which turns up mid-sentence, so `Before` / `After` labels over two lines of prose counted as two graphs. It now requires the reference to sit between a `(` and a `)` — the shape every documented example writes. Your exact repro is now a deny case in the suite, and it fails `got=passthrough expected=deny` against the previous commit.

Two deliberate limits, both now stated in the code rather than left to be inferred:

- **Not a balanced group.** The left side stays permissive because a Type half carries its own parens — `(fn(u32) -> u32 · src/x.rs:5)` — and refusing to cross them denied a hop conforming to the documented shape. That case is pinned by its own allow-test. The cost is that an unrelated parenthetical straddling the reference also passes.
- **Your second point stands and cannot be fixed by a pattern.** No regex distinguishes a real graph from a fabricated one. The comment now says so and names the typed `reviewed:` ack as the control that can, so the deny text is no longer the only thing speaking to it.

Suite is 54/0.

PR #1115, opened 2026-08-02
refactor(agents): move gate implementations out of .claude into .agents

DorianZheng:
## Summary

Moves the gate implementations out of `.claude/` into `.agents/`, so the vendor directories name the gates rather than holding them. Runtime markers move with them. Because one installed hook can serve worktrees sitting on either side of the move, four compat mechanisms keep both sides working.

## Call graph

Before

```text
git commit                (git · .githooks/pre-commit:43)
  └─ preflight-commit-push (gate · .claude/hooks/preflight-commit-push.sh:67)   — path hardcoded
       └─ writes marker    (gate · .claude/hooks/preflight-commit-push.sh:144)  — .claude/.last-audit-handoff.json, no mkdir
            └─ commit-msg  (git · .githooks/commit-msg:16)                      — reads .claude/.last-audit.json
```

After

```text
git commit                (git · .githooks/pre-commit:58)
  └─ resolve_gate         (git · .githooks/pre-commit:34)                       — .agents/hooks first, .claude/hooks fallback, fails closed
       └─ preflight-commit-push (gate · .agents/hooks/preflight-commit-push.sh:67)
            └─ writes marker     (gate · .agents/hooks/preflight-commit-push.sh:156) — .agents/state/, mkdir first
                 └─ mirror_to_legacy (gate · .agents/hooks/preflight-commit-push.sh:80) — copy for a pre-move commit-msg
                      └─ resolve_marker (git · .githooks/commit-msg:20)         — reads whichever location the gate wrote
```

## Changes

- `.agents/hooks/` holds the 14 gate files; `.claude/hooks/*.sh` become committed symlinks (mode `120000`), so a hook from before this change still resolves the gate it names.
- `pre-commit`/`pre-push` gain `resolve_gate`, and `commit-msg` gains `resolve_marker`. Both prefer `.agents/`, fall back to `.claude/`, and fail closed — a missing gate aborts rather than running unguarded.
- The gate copies its markers to the legacy path for a `commit-msg` that predates the move. Both copies stay gitignored: the verdict gate keys its dossier to a tree hash that skips ignored paths, so an unignored marker would invalidate the audit that just passed.
- `write_command_handoff` wrote its marker without creating the directory. Harmless while the parent was `.claude/` (always present); fatal once it is a gitignored `.agents/state/`.
- The skill moves to `.agents/skills/`, reached via a `.claude/skills/<name>` symlink.

## How to verify

```bash
bash .githooks/githooks.test.sh                        # 52 passed (43 + 9 compat cases)
bash .agents/hooks/preflight-commit-push.test.sh       # 14 passed
bash .agents/hooks/preflight-verdict-check.test.sh     # 53 passed
bash .agents/hooks/preflight-pr-review.test.sh         # 54 passed
```

The 7 `githooks` and 17 `preflight-commit-push` failures are pre-existing and environmental — the failure sets are byte-identical to `main`. Both suites defer when `core.hooksPath` ends in `.githooks` and the delegate exists.

## Risks / rollout

- **Nothing ends up ungated.** Every hook-version × tree-version combination was exercised for commit and push, audited and unaudited, including `commit-msg`; a tree with the gate absent from both locations blocks rather than passing.
- **The compat layer serves a configuration `install.sh` does not create.** `install.sh` sets `core.hooksPath` *relative*, under which each worktree runs its own hooks and the two versions never cross. Only a clone configured *absolute* needs any of this. All four mechanisms can be deleted once every checkout is past the move.
- **One documented residual:** a `commit-msg` from before the move consumes only the legacy copy, so on that one combination the audit survives its commit for the rest of its 600s window. It cannot admit unaudited content — the binding to branch, HEAD, diff and subject is unchanged. Noted at `.agents/hooks/preflight-commit-push.sh:295`.
- Stale `.claude/.last-verdict.json` files left over from before the move become untracked once this lands; one `rm` clears them.

## Summary by CodeRabbit

- **New Features**
  - Added automated checks for commits, pushes, pull request actions, and session completion.
  - Added audit validation for changes, review metadata, verdicts, and required acknowledgments.
  - Added periodic workflow reminders and adversarial review guidance.

- **Bug Fixes**
  - Improved handling of stale, mismatched, incomplete, or invalid audit results.
  - Added compatibility support during migration to the new workflow location.

- **Documentation**
  - Updated setup and workflow guidance to reflect the new configuration paths and review procedures.


DorianZheng:
@coderabbitai review again

PR #1122, opened 2026-08-02
feat(infra): source the API and Runner from one artifact selector

DorianZheng:
The API could only be built from the deployed checkout and the Runner could
only be installed from a published release. So an unreleased Runner change
could not be tested anywhere, and no API artifact could be promoted rather
than rebuilt. One resolver now answers "release or build?" for both.

- artifact-source.mjs picks the source. Unset keeps each component's prior
  default, so both new corners are opt-in.
- runner-artifact.mjs resolves either Runner source to one URL+checksum
  shape, leaving the host-side verify/swap/rollback untouched.
- Release mode deploys an immutable ECR image; publish-api-image.yml builds
  it once from the tag and promotes by retag, never rebuilding.
- Build mode stages a commit-keyed, write-once S3 object. Its identity
  carries the commit, so two builds of one version stay distinguishable.
- The stage bootstrap owns the ECR repository and artifact bucket: CI must
  publish before a deploy reads them, so the consumer cannot create its own
  input.

A build deploy now refuses a ref that is not the checkout it would build the
API from, since only the Runner is addressed by ref and nothing downstream
compares the two.

## Summary by CodeRabbit

* **New Features**
  * Added commit-based preview deployments for API and Runner artifacts.
  * Added release deployments for development and production stages with version validation and optional downgrade support.
  * Added immutable API image publishing and promotion.
  * Added staged, checksum-verified Runner artifacts with persistent cloud storage.
  * Added configurable artifact sources for API and Runner deployments.

* **Bug Fixes**
  * Improved deployment validation, artifact integrity checks, and protection against mismatched or incomplete artifacts.

* **Documentation**
  * Expanded deployment setup, workflow, artifact-selection, and operational guidance.


DorianZheng:
Reviewed with a focus on the deploy path, since this lands on top of #1118 and #1120 — the fixes behind run 30746528206, the first `deploy-infra` run whose `Deploy the full stack` step ran to success rather than being skipped.

## Blocking: the config does not type-check

`SST deployment tests` fails on the PR's own code:

```
sst.config.ts(993,9):   error TS2322
sst.config.ts(1090,15): error TS2322
  Type '{ fetch: string; tarballName: string; checksumName: string; tarballUrl: string; checksumUrl: string; }'
    is not assignable to
       '{ tarballName: string; tarballUrl: string; checksumUrl: string; fetch: "https" | "s3" }'
  Types of property 'fetch' are incompatible.
    Type 'string' is not assignable to type '"https" | "s3"'.

make: *** [make/test.mk:378: test:apps:infra-config] Error 2
```

The producer returns `{ ...resolveRunnerReleaseAssets(source.version), fetch: 'https' }` and `fetch: 's3'` from a `.mjs`; the consumer declares `fetch: 'https' | 's3'`. The two shapes are confirmed — the widening mechanism is not isolated, so `as const` or a JSDoc annotation on the resolver is where I would look first rather than a guaranteed fix.

`Lint (conclusion)` also fails, with `infra` reporting `failure` and every other component skipped. `CodeQL` fails in 3s.

## The branch predates the deploy fixes

At `9372dc08`, `ci/github-deploy-role.yaml` has 0 occurrences of `cloudfront-keyvaluestore` and `scripts/sst-config-contract.test.mjs` has 0 of `cli.paths.root`. Both are on `main` at `ff2f6a5d`.

The merge itself is safe — `git merge-tree --write-tree ff2f6a5d 9372dc08` produces a tree that retains both:

```
github-deploy-role.yaml        11202 bytes   cloudfront-keyvaluestore: 1
sst-config-contract.test.mjs   18152 bytes   cli.paths.root:           1
```

The cost is signal rather than correctness: CI here ran against a tree without either fix, while this PR rewrites `deploy-infra.yml`, `sst.config.ts`, and `sst-with-cloudflare.mjs`. A rebase onto `ff2f6a5d` would make the checks say something about what actually merges.

## What holds up well

- One resolver answering "release or build?" for both components, with defaults kept per-component rather than derived from the stage, so neither new corner changes an existing `npm run deploy --stage dev`.
- `requireCheckoutMatchesArtifactRef` is guarded by `else if (apiSource.ref)`, so build mode without a ref skips rather than throwing on an undefined comparison — and a source-contract test pins the live call site, which is the same shape as the `dir: $cli.paths.root` contract.
- `release-deployment-safety.test.mjs` asserts both `ApiImagesRepository` and `RunnerArtifactsBucket` against the template.
- The bootstrap-before-deploy ordering is stated in the README, which matters because the deploy now reads stores it cannot create.
- `boxlite-<stage>-artifacts-<account>` falls under the runtime boundary's existing `boxlite-<stage>-*` grant, so Runners can read it without widening the boundary.

## Scope

Read closely: `artifact-source.mjs`, the guard and its call site, `ci/github-deploy-role.yaml`, the regression-test coverage, and the interaction with #1118/#1120. Not read line by line: `publish-api-image.yml`, `deploy-release.yml`, `dev-artifact.Dockerfile`, and the internals of `runner-artifact-build.mjs` — roughly 617 of the 3049 added lines.

PR #1158, opened 2026-08-06
fix(runtime): make guest-ready timeout configurable

### LINKED ISSUE #1111: GUEST_READY_TIMEOUT is a hard-coded 30s with no override
## Summary

`GUEST_READY_TIMEOUT` is a hard-coded 30 s with no environment or config override, so a box
that boots slowly on a contended host fails permanently with no available mitigation:

```
Box uPKTB5hOHtZf failed to start: timeout after 30s

Evidence at T+30s:
• shim_alive          = true
• console_bytes       = 0
• ready_socket_exists = true
• likely_cause        = guest agent never wrote to console (init or vsock plumbing broken)
```

`src/boxlite/src/litebox/init/tasks/guest_connect.rs:110`:

```rust
const GUEST_READY_TIMEOUT: Duration = Duration::from_secs(30);
```

It has exactly two references — the definition and the call site at `:85`. There is no env var,
no `AdvancedOptions` field, and no per-box knob, so a caller on a loaded machine has no way to
raise it. The doc comment says it is "exposed as a constant so tests can call
`wait_for_guest_ready` with a short timeout", i.e. tests can shorten it but production callers
cannot lengthen it.

## How it showed up

Running the full Python SDK integration suite on macOS (`make test:integration:python`, which
the pre-push hook runs), `tests/test_tcp_filter.py::TestEdgeCases::test_multiple_allowed_hosts`
intermittently fails with the message above. It failed 3 times during full-suite runs.

## Reproduction attempts — what does and does not trigger it

I could not reproduce it in isolation. Recording the negative results so the next person does
not repeat them:

| Condition | Result |
|---|---|
| Standalone, idle machine (3 consecutive runs) | pass — 15.9 s, 15.6 s, 17.6 s |
| Standalone, 24 CPU spinners (load avg 20.6 / 18 cores) | pass — 28.4 s |
| Standalone, 40 CPU spinners + 4 `dd` I/O loops (load avg 52.9) | pass — 69.2 s |
| Whole `test_tcp_filter.py` file, 21 boxes back to back | pass — 21/21, 179 s |
| During / immediately after the **full** integration suite | **fails**, 3 occurrences |

Two things worth noting:

- Wall time degrades sharply with host load (15.9 s idle → 28.4 s → 69.2 s), so box startup is
  clearly contention-sensitive; but pure CPU/IO pressure alone did not push the *guest-ready
  handshake* past 30 s here.
- The failure signature is `console_bytes = 0` — the guest wrote *nothing*, which does not look
  like ordinary slowness. Something about running the whole suite (hundreds of boxes) is the
  trigger; I did not isolate what.

## Why the constant is worth fixing regardless

Whatever the underlying trigger turns out to be, a user or CI job that hits it today has no
escape hatch: they cannot raise the deadline to confirm "slow" versus "hung", and the same
30 s applies to a warm laptop and a heavily loaded CI box alike.

Suggested: make it overridable (env var and/or an `AdvancedOptions` field) with the current
30 s as the default, keeping the existing diagnostic block on expiry.

## Related

- #469 — the Node SDK hits the same class of failure under parallel test execution. That one is
  a *vitest* 120 s hook timeout, i.e. a test-framework limit; this issue is about boxlite's own
  internal deadline, which no caller can configure.

## Environment

- macOS 27.0.0, arm64, 18 cores
- boxlite 0.9.7, local (non-REST) runtime, `sdks/python` built from source

tiammomo:
## Summary
Allow slow hosts to extend the guest-ready handshake timeout while preserving the 30-second default.

## Call graph
Before
GuestConnectTask::run (src/boxlite/src/litebox/init/tasks/guest_connect.rs:27) — passes a fixed 30-second deadline ← BUG: slow hosts cannot extend it
└─ wait_for_guest_ready (src/boxlite/src/litebox/init/tasks/guest_connect.rs:144) — times out at that deadline

After
GuestConnectTask::run (src/boxlite/src/litebox/init/tasks/guest_connect.rs:27) — selects a validated override or the 30-second default
├─ guest_ready_timeout (src/boxlite/src/litebox/init/tasks/guest_connect.rs:114) — reads BOXLITE_GUEST_READY_TIMEOUT_SECS
└─ wait_for_guest_ready (src/boxlite/src/litebox/init/tasks/guest_connect.rs:144) — uses the selected deadline

Fixes #1111

## Changes
- Allow slow hosts to extend the guest-ready handshake timeout while preserving the 30-second default.

## How to verify
- `CARGO_HOME=.git/starfix-cargo CARGO_NET_OFFLINE=true UV_OFFLINE=1 make test:unit:rust FILTER=guest_ready_timeout`
- `CARGO_HOME=.git/starfix-cargo CARGO_NET_OFFLINE=true UV_OFFLINE=1 make fmt:check`
- `CARGO_HOME=.git/starfix-cargo CARGO_NET_OFFLINE=true UV_OFFLINE=1 make lint`

## Summary by CodeRabbit

* **New Features**
  * Guest readiness timeout can now be configured with `BOXLITE_GUEST_READY_TIMEOUT_SECS`.
  * The timeout defaults to 30 seconds and accepts positive integer values.
* **Bug Fixes**
  * Invalid, non-positive, or unsupported timeout values now produce a clear configuration error.
* **Documentation**
  * Added configuration details, validation requirements, default behavior, and usage examples for the new environment variable.

### DIFF
--- a/docs/reference/README.md
+++ b/docs/reference/README.md
@@ -460,6 +460,18 @@ export BOXLITE_TMPDIR=/custom/tmp
 python script.py
 ```
 
+#### `BOXLITE_GUEST_READY_TIMEOUT_SECS`
+
+Set the guest-ready handshake timeout in seconds. The value must be a positive integer.
+
+**Default:** `30`
+
+**Example:**
+```bash
+export BOXLITE_GUEST_READY_TIMEOUT_SECS=90
+python script.py
+```
+
 ## Error Codes & Handling
 
 ### Error Types
--- a/src/boxlite/src/litebox/init/tasks/guest_connect.rs
+++ b/src/boxlite/src/litebox/init/tasks/guest_connect.rs
@@ -16,6 +16,7 @@ use crate::vmm::exit_info::ExitErrorKind;
 use async_trait::async_trait;
 use boxlite_shared::BoxTransport;
 use boxlite_shared::errors::{BoxliteError, BoxliteResult};
+use std::num::NonZeroU64;
 use std::path::Path;
 use std::time::Duration;
 
@@ -74,7 +75,12 @@ impl PipelineTask<InitCtx> for GuestConnectTask {
         if skip_guest_wait {
             tracing::debug!(box_id = %box_id, "Skipping guest ready wait (reattach)");
         } else {
-            tracing::debug!(box_id = %box_id, "Waiting for guest to be ready");
+            let timeout = guest_ready_timeout()?;
+            tracing::debug!(
+                box_id = %box_id,
+                timeout_secs = timeout.as_secs(),
+                "Waiting for guest to be ready"
+            );
             wait_for_guest_ready(
                 &ready_transport,
                 shim_pid,
@@ -82,7 +88,7 @@ impl PipelineTask<InitCtx> for GuestConnectTask {
                 &console_log,
                 &stderr_file,
                 box_id.as_str(),
-                GUEST_READY_TIMEOUT,
+                timeout,
             )
             .await
             .inspect_err(|e| log_task_error(&box_id, task_name, e))?;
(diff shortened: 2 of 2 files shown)


tiammomo:
Updated the timeout tests to exercise the environment-backed entry point for both the unset default and a valid override. Focused tests, formatting, and lint pass.

PR #1159, opened 2026-08-06
feat(infra): run the commerce billing service on the dev stack

DorianZheng:
The dashboard's billing client calls 20 endpoints on config.billingApiUrl and
nothing has ever served them. boxlite-ai/boxlite-commerce now does; run it in
the cluster behind commerce.<stackDomain>.

The image comes from this account's private ECR, pushed by that repository's
own workflow through GitHub OIDC — not a public registry, and not built here,
since the source lives elsewhere. commerce-artifact.mjs composes and validates
the reference the way api-artifact.mjs does for the Api, so a tag ECR would
reject fails on the deployer rather than as an ECS pull failure mid-deploy.

The service is declared only for a stage that has a published image. A stage
whose workflow never pushed one cannot build it locally either, so with
`wait: true` an unresolvable reference would hang that stage's whole deploy.

The ALB probes /api/billing/health, the service's only unauthenticated route;
a default '/' probe would 401 and fail every task.

BILLING_API_URL is deliberately left unset rather than pointed at the new
service. It is not only a dashboard hint: with it set, OrganizationService
creates every non-default organization suspended with 'Payment method
required', which a mock reporting creditCardConnected: false can never clear.
Wiring the dashboard to it stays an explicit opt-in, along with the CORS and
public-OIDC-issuer gaps noted at the service definition.

Before
  BillingApiClient       (class · apps/dashboard/src/billing-api/billingApiClient.ts:21) — 20 calls
    -> config.billingApiUrl (env · apps/api/src/config/configuration.ts:162) — unset
       -> (no server)      ← BUG: nothing has ever served this contract

After
  run()                  (fn · apps/infra/sst.config.ts:126) — stack composition
    -> commerceImageReference  (fn · apps/infra/scripts/commerce-artifact.mjs:46) — validated ECR ref
       -> commerceImageRepository (fn · apps/infra/scripts/commerce-artifact.mjs:30) — repo name
          -> awsResourceName      (fn · apps/infra/scripts/resource-name.mjs:45) — the one grammar
    -> sst.aws.Service('Commerce') (resource · apps/infra/sst.config.ts:833) — declared iff image
       -> httpHealth('/api/billing/health') (fn · apps/infra/sst.config.ts:81) — only public route
       -> BOXLITE_API_URL = api.url — membership lookups
    -> Api env BILLING_API_URL     (passthrough · apps/infra/sst.config.ts:719) — still opt-in

## Summary by CodeRabbit

- **New Features**
  - Added Commerce service deployment support with health checks and environment-specific image configuration.
  - Added Wallet, Spending, and Limits pages when billing is enabled.
  - Billing navigation is now limited to organization owners.

- **Bug Fixes**
  - Organization suspension now depends on the explicit payment-method requirement setting, rather than billing service availability.

- **Documentation**
  - Updated infrastructure setup guidance, billing configuration, architecture, repository setup, and cost estimates.

### DIFF
--- a/apps/api/src/config/configuration.ts
+++ b/apps/api/src/config/configuration.ts
@@ -275,6 +275,12 @@ const configuration = {
     apiKey: process.env.ADMIN_API_KEY,
   },
   skipUserEmailVerification: process.env.SKIP_USER_EMAIL_VERIFICATION === 'true',
+  // Whether a newly created non-default organization starts suspended until a
+  // payment method exists. Separate from billingApiUrl on purpose: pointing the
+  // dashboard at a billing service and demanding payment up front are different
+  // decisions, and conflating them means any billing service — including one that
+  // cannot register a card — locks every new organization out permanently.
+  requirePaymentMethod: process.env.REQUIRE_PAYMENT_METHOD === 'true',
   apiKey: {
     prefix: process.env.API_KEY_PREFIX || 'blk',
     validationCacheTtlSeconds: parseInt(process.env.API_KEY_VALIDATION_CACHE_TTL_SECONDS || '10', 10),
--- a/apps/api/src/organization/services/organization.service.ts
+++ b/apps/api/src/organization/services/organization.service.ts
@@ -440,7 +440,11 @@ export class OrganizationService implements OnModuleInit, TrackableJobExecutions
       organization.suspended = true
       organization.suspendedAt = new Date()
       organization.suspensionReason = 'Please verify your email address'
-    } else if (this.configService.get('billingApiUrl') && !defaultForCreator) {
+    } else if (this.configService.get('requirePaymentMethod') && !defaultForCreator) {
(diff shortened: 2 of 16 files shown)


DorianZheng:
Triage of the CodeRabbit comment on `apps/dashboard/CLAUDE.md`: **not fixed, deliberately.**

`afterwards` → `afterward` is flagged because the line shows as changed, but `870e19a4` altered only the count on that line (`13` → `10`); the wording predates this PR (present at merge-base `924c7ff0`). Per `.agents/watch/escalation-policy.md` — fix what the change introduced, leave the pre-existing — it is out of scope here rather than overlooked.

Deploy status for reviewers: run 31135557840 ended `failure`, but only the OtelCollector image rebuild failed (Corepack fetching yarn hit `ECONNRESET`). Commerce and Api applied and are live on dev — Commerce task-def rev 6 (`500a16f`), Api rev 235 (`870e19a4`). A re-dispatch is still wanted so the run goes green.

PR #1182, opened 2026-08-09
fix(disk): silent debugfs failures, xattr loss, image capacity

### LINKED ISSUE #1181: Guest rootfs build can silently fail: debugfs errors ignored, image disks lack headroom
## Bug

Building a guest rootfs ext4 image can silently produce a corrupt or
incomplete image, or fail outright with no clear signal, for two related
reasons in the `mke2fs`/`debugfs` pipeline (`src/boxlite/src/disk/ext4.rs`,
`src/boxlite/src/images/image_disk.rs`):

1. **Silent debugfs failures.** Both `debugfs -w -f -` call sites
   (`normalize_inodes_with_debugfs`, `inject_file_into_ext4`) only checked
   the process exit code. e2fsprogs logs a per-command failure (`com_err`)
   to the same stderr stream as the one-line startup banner and continues
   to the next command rather than aborting — so a failing `sif`/`write`
   never surfaces, and the image is left with wrong inode metadata or a
   guest binary that was never actually written.

2. **No capacity headroom.** `ImageDiskManager` sizes an image disk purely
   from OCI layer content. `GuestRootfsManager` later copies that disk and
   injects the `boxlite-guest` binary into it — but a near-empty image
   floors at 256 MiB, leaving ~223 MiB free after ext4/journal overhead,
   well under an unstripped debug guest binary (~231 MiB). Combined with
   (1), this failure was invisible: `debugfs write` would run out of
   blocks and the process would still exit 0.

## Impact

A box can fail to start, or start with a guest binary that was silently
never injected, with no error surfaced to the caller.

DorianZheng:
## Summary

Fixes two related bugs in the guest-rootfs ext4 build pipeline: `debugfs` errors that never surfaced because only the process exit code was checked, and — uncovered once that check was fixed — an image disk with no capacity headroom for the guest binary injected into it afterward.

## Call graph

Before
```
RuntimeImpl::initialize             (RuntimeImpl · src/boxlite/src/runtime/rt_impl.rs:237)
  └─ ImageDiskManager::new          (ImageDiskManager · src/boxlite/src/images/image_disk.rs:65)
ImageDiskManager::build_and_install (ImageDiskManager · src/boxlite/src/images/image_disk.rs:98)
  └─ create_ext4_from_dir           (fn · src/boxlite/src/disk/ext4.rs:354)  — sized from OCI content only
GuestRootfsManager::build_and_install (GuestRootfsManager · src/boxlite/src/rootfs/guest.rs:372)
  └─ inject_file_into_ext4          (fn · src/boxlite/src/disk/ext4.rs:580)  ← BUG: staged image
     has no headroom (256 MiB floor, ~223 MiB free) for the ~231 MiB debug guest binary; debugfs
     write runs out of blocks but only `output.status.success()` is checked, which is still 0 —
     e2fsprogs logs the failure to stderr and continues, so it never surfaces
```

After
```
RuntimeImpl::initialize             (RuntimeImpl · src/boxlite/src/runtime/rt_impl.rs:237)
  └─ ImageDiskManager::new          (ImageDiskManager · src/boxlite/src/images/image_disk.rs:65)  — takes a fixed 512 MiB headroom budget (src/boxlite/src/runtime/rt_impl.rs:332)
ImageDiskManager::build_and_install (ImageDiskManager · src/boxlite/src/images/image_disk.rs:98)
  └─ disk_path                      (ImageDiskManager · src/boxlite/src/images/image_disk.rs:170)  — cache key includes the budget
  └─ create_ext4_from_dir           (fn · src/boxlite/src/disk/ext4.rs:354)  — budget reserved in the image size
GuestRootfsManager::build_and_install (GuestRootfsManager · src/boxlite/src/rootfs/guest.rs:372)
  └─ inject_file_into_ext4          (fn · src/boxlite/src/disk/ext4.rs:580)
       └─ check_debugfs_output      (fn · src/boxlite/src/disk/ext4.rs:460)  — exit code AND any
          stderr beyond the one-line startup banner; shared by this call site and
          normalize_inodes_with_debugfs (src/boxlite/src/disk/ext4.rs:493)
```

Fixes #1181

## Changes

- `check_debugfs_output` (`src/boxlite/src/disk/ext4.rs:460`) replaces a bare `output.status.success()` check at both `debugfs -w -f -` call sites — a failing per-command error now aborts the build instead of leaving a partially-normalized or partially-injected image.
- `OverrideStat::read_xattr` (`src/boxlite/src/images/archive/override_stat.rs`) now returns `Err` for a present-but-malformed `user.containers.override_stat` value instead of `Ok(None)` — a layer's declared ownership is the only copy of that data, so a corrupt record must not be treated the same as no record at all.
- The image-disk headroom budget (`IMAGE_DISK_GUEST_BINARY_HEADROOM_BYTES`, `src/boxlite/src/runtime/rt_impl.rs:215`) is a fixed constant rather than derived from the guest binary's live size, and is folded into `ImageDiskManager`'s cache key — deriving it dynamically would change the cache key on every `make guest` rebuild, and `ImageDiskManager` has no GC (unlike `GuestRootfsManager`'s paired `version_key` + `gc()`) to reclaim the orphaned copies that would leave behind.
- `.githooks/pre-push`'s `new_ref_base_sha` picked the wrong fork point for a new branch: the first candidate in alphabetical ref order rather than the closest by commit distance, and `--is-ancestor` breaks once the default branch advances past the fork point. Rewritten to use `merge-base` and pick the candidate needing the fewest commits.

## How to verify

- `make fmt:check:rust && make lint:rust`
- `make test:unit:rust` (988 boxlite + 49 boxlite-shared)
- `make test:integration:rust` (254 passed, 11 skipped)
- `make test:integration:c` — exercises the real box-creation path this PR fixes
- `bash .githooks/githooks.test.sh` (85 tests)

## Summary by CodeRabbit

- **New Features**
  - Added reserved capacity for guest binaries during disk image creation, helping ensure post-build injection succeeds.
  - Improved disk-image caching so images with different capacity requirements remain isolated.

- **Bug Fixes**
  - Improved detection of filesystem image creation and file-injection failures.
  - Malformed extended attributes now report an error instead of appearing absent.
  - Improved pushed-diff detection for branch histories and merge scenarios.
  - Corrected permission handling when restoring filesystem ownership.

### DIFF
--- a/.githooks/githooks.test.sh
+++ b/.githooks/githooks.test.sh
@@ -97,9 +97,17 @@ write_ref_updates_file() {  # repo remote local_ref remote_ref output_file
   printf '%s %s %s %s\n' "$local_ref" "$local_sha" "$remote_ref" "$remote_sha" > "$output_file"
 }
 
+# Extracted from the real pre-push, not hand-duplicated: a second copy of this
+# selection algorithm is exactly how it went stale before (see new_ref_base_sha's
+# own history) — a fix applied to only one copy leaves the other asserting
+# against a base the real hook no longer picks.
+eval "$(sed -n '/^new_ref_base_sha() {/,/^}/p' "$REPO_ROOT/.githooks/pre-push" | \
+  sed -e 's|\${repo_root}|${repo}|g' -e 's|\$repo_root|\$repo|g' \
+      -e 's|\${remote_name}|${remote}|g' -e 's|\$remote_name|\$remote|g')"
+
 pushed_diff_hash_for() {  # repo remote ref_updates_file
   local repo="$1" remote="$2" ref_updates_file="$3"
-  local empty_tree local_ref local_sha remote_ref remote_sha base_sha candidate
+  local empty_tree local_ref local_sha remote_ref remote_sha base_sha
   empty_tree="$(git -C "$repo" hash-object -t tree /dev/null)"
   while read -r local_ref local_sha remote_ref remote_sha; do
     [[ -n "${local_ref:-}" ]] || continue
@@ -107,14 +115,7 @@ pushed_diff_hash_for() {  # repo remote ref_updates_file
     if [[ "$local_sha" =~ ^0+$ ]]; then
       printf 'deleted %s at %s\n' "$remote_ref" "$remote_sha"
     elif [[ "$remote_sha" =~ ^0+$ ]]; then
-      base_sha=""
-      while IFS= read -r candidate; do
-        [[ -n "$candidate" ]] || continue
-        if git -C "$repo" merge-base --is-ancestor "$candidate" "$local_sha" 2>/dev/null; then
-          base_sha="$candidate"
-          break
-        fi
-      done < <(git -C "$repo" for-each-ref --format='%(objectname)' "refs/remotes/${remote}" 2>/dev/null)
+      base_sha="$(new_ref_base_sha "$local_sha" || true)"
       if [[ -n "$base_sha" ]]; then
         git -C "$repo" log --format='commit-subject %H %s' "$base_sha..$local_sha" 2>/dev/null || true
         git -C "$repo" diff --no-ext-diff "$base_sha" "$local_sha" 2>/dev/null || true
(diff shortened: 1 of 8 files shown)


DorianZheng:
Addressed in 536e31b0:

- **Major finding** (record_owner swallowing the malformed-xattr `Err`): fixed, plus the identical sibling in `fix_rootfs_permissions`'s `set_xattr_recursive` (same bug, and worse there — it was overwriting the original malformed xattr with a fresh 0:0 record via `xattr::set`, permanently destroying it rather than just discarding it in memory for one build).
- `new_ref_base_sha` eval-extraction guard in `githooks.test.sh`: fixed.
- `aaa-early`/`zzz-late` test's `$R/err.txt` hazard: fixed, same pattern as the existing `moved_on_err` test.

Left as-is, not bundled into this PR:

- `OverrideStat::parse`'s "at least 4 fields" vs "exactly 4 fields" leniency: this is pre-existing behavior — no commit in this PR touches `parse` itself, only `read_xattr`'s handling of what `parse` returns. Leaving it out to keep this PR's diff scoped to what it actually changed.
- `.githooks/pre-push`'s default-branch fast path not comparing against the fallback scan for stacked branches: this is a deliberate trade-off from the commit that introduced it (`af400a56`), documented in its own comment — optimize for the common case (branches cut from `main`) at the cost of a rare stacked-branch edge case. Seeding the fast path from the scan would mean always running the full scan anyway, defeating the point of having a fast path. Rated Trivial/low-value by the review itself; leaving it.
- Duplicated "fake codex" heredoc in `githooks.test.sh` (lines ~538-564 vs ~604-630): pure DRY cleanup with no correctness impact, rated Trivial/low-value. Not bundling an unprompted refactor into a bug-fix PR.

PR #1185, opened 2026-08-10
fix(exec): signal executions through a verifiable exit claim

BatmanByte:
## Summary

Use a `/proc/<pid>/stat` start-time fingerprint to avoid signalling a later process that reused an exec PID. This replaces the reaper-coupled ExitClaim signal path with one identity check shared by direct Kill, timeouts, and registry shutdown.

## Call graph

Before

`spawn_execution` / container init → execution state, timeout, or shutdown → numeric PID signal ← BUG: a recycled PID can name a different process

After

`spawn_execution` / container init → `ProcessInstance::capture(pid)` → execution state, timeout, or shutdown → compare `/proc` start time → signal only the matching process

`SIGCHLD` → `Reaper::deliver` → `ExitSlot::get` — exit delivery remains separate from process identity.

## Changes

- Capture a PID start-time fingerprint for exec and container-init processes.
- Verify that fingerprint before direct and process-group signals.
- Keep the reaper responsible only for exit slots and exit callbacks.
- Log when the identity cannot be captured and signalling will be skipped.

## How to verify

- `make fmt:check:rust`
- `make guest`
- `make test:unit:guest` on Linux

## Risks / rollout

- The `/proc` comparison and `kill` are not kernel-atomic, so a PID can still recycle after a successful comparison. Eliminating that residual window requires creation-time pidfd handling and is intentionally outside this PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved process shutdown and timeout handling by verifying process identity before sending signals.
  - Prevented stale or reused process IDs from receiving termination signals.
  - Improved escalation from graceful termination to forced termination only when processes remain active.
  - Ensured exited process statuses remain available consistently across readers.
  - Improved cleanup of managed executions during resource release and shutdown.
  - Added safeguards for invalid process-group targets and clearer error reporting.


BatmanByte:
Thanks — two of the seven are acted on in 505a8af9, and one of them was the useful catch.

**Acted on**

`This test passes even without the settled_at guard` — correct, and it was the more valuable of the two because it is a class, not an instance. `signal_leader_if_live` answers `Ok(false)` both when the claim check refuses and when `kill` reports ESRCH, so any test signalling a fabricated pid cannot tell the two apart. I reproduced it before fixing: deleting `slot.settled_at.is_none()` left `signal_leader_refuses_a_settled_claim` green. It now signals a real child and asserts the child survives, and the same mutation turns it red. Swept the rest of the signalling tests for the same shape — the surviving `pid(2_001)` case asserts registry state via `claim_is_live`, which has no ESRCH ambiguity.

`Assert the exact signal that reached the child` — taken. `status.code() == None` held for death by any signal; `signal_leader_terminates_the_live_claimed_process` now pins `SIGTERM`, matching what `state.rs` and `timeout.rs` already do. Also dropped `ExitClaim::pid`, dead once the fabricated-pid assertion went away.

**Not in this PR**

- *Drop `async` from the claim registry methods* — agreed they no longer await; the signature churn reaches call sites in two more files, so it is a follow-up rather than part of a signalling fix.
- *Log the `leader_pid != claim.pid` mismatch* — that branch is a refusal, not a swallowed error; the caller gets `false`. Left as is.
- *Document `claim_is_live` as advisory* — fair, follow-up.
- *Extract the shared claim-creation block* — three call sites of a small struct literal; leaving the duplication local rather than adding an abstraction that hides which role each constructor picks.
- *Confirm the group-kill path stays unguarded on purpose* — yes, deliberate and already disclosed in the code at the branch (`state.rs`) and in the PR description. Closing it needs a claim-guarded group primitive on `Reaper`, which is a different change.

DorianZheng:
**Alternative considered: the pre-pidfd identity classic — capture `(pid, /proc/<pid>/stat starttime)` at spawn, compare before signalling.** This is what runc does for container init (`init_process_start` in state.json, checked before `unix.Kill`), so it has direct upstream precedent for this codebase's lineage.

It solves the *identity* half more cheaply than the token: one `u64` captured at spawn under the already-held `reap_fence()`, one compare in the kill helper — no registration, mint, or release choreography, and `shutdown_all`/the timeout watcher keep their shape. But it doesn't replace the claim design:

1. Unfenced, it is still check-then-act: the process can exit, be reaped, and the pid recycle between the `/proc` read and the `kill`. runc ships that residual window; closing it kernel-tight still requires holding the reap fence across compare-and-signal, so the fence dependency is unchanged — only the identity witness moves.
2. The claim also carries ownership semantics starttime cannot express: init's signal-only role, released-claim ⇒ skip in `shutdown_all`, and the shutdown wait loop polling claim liveness instead of `kill(pid, None)`. Dropping the claim regrows those as loose same-typed flags on `ExecutionState` — the transposition hazard the role enum exists to prevent.
3. Comparing starttime means parsing `/proc/<pid>/stat` around `comm`, which the sandboxed workload controls (the embedded-`)` pitfall) — untrusted bytes entering the kill path's parser.

Where it may earn its keep: the declared `process_group=true` residual. `kill(-pgid)` has no reaper slot to claim, but verifying the group leader's starttime under the fence would close that gap with runc-grade precedent — a candidate shape for the follow-up group primitive on `Reaper`.

src/guest/src/service/exec/identity.rs
```diff
@@ -0,0 +1,187 @@
...
+
+/// A process identity that distinguishes a PID's current owner from a prior
+/// process that used the same number.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) struct ProcessIdentity {
+    pid: Pid,
+    start_time: u64,
+}
+
+impl ProcessIdentity {
```
DorianZheng: there is already a `ProcessIdentity` in this repo. you could move it to the shared util dir so guest can reuse it

BatmanByte: Checked this: the existing `ProcessIdentity` is the host-side PID-file validation enum (`Verified` / `Legacy` / `Absent`), while the guest type stores a PID/start-time fingerprint and validates signal delivery. They are not a drop-in shared abstraction. I renamed the guest type to `ProcessInstance` to remove the ambiguity; extracting a shared PID/start-time helper would broaden host/guest platform dependencies without another consumer.

src/guest/src/service/exec/state.rs
```diff
@@ -335,24 +354,52 @@ impl ExecutionState {
...
+        process_group: bool,
+    ) -> Result<bool, nix::errno::Errno> {
+        if self.inner.lock().await.released {
+            return Ok(false);
+        }
+        let Some(process) = self.process else {
+            return Ok(false);
+        };
+        process.signal(signal, process_group)
+    }
```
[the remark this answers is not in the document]
BatmanByte: Fixed in f8cfe025. `signal_if_current` now retains the execution-state mutex from the `released` check through `ProcessInstance::signal`, so `release_resources` and signaling are ordered. Verified with `make fmt:check:rust` and `make guest`.

DorianZheng (approved):

PR #1189, opened 2026-08-10
feat(runner): execute the box migration jobs

DefeatMan:
## Summary

A box migration moves a stopped box between runners through a portable `.boxlite`
archive in the object store. The control plane owns the state machine; this PR
gives the runner the five job handlers it drives — `EXPORT_BOX`, `IMPORT_BOX`,
`ROLLBACK_EXPORT_BOX`, `ROLLBACK_IMPORT_BOX`, `FINISH_PREVIOUS_BOX` — plus the
S3 archive store and the boxlite export/import calls they sit on.

## Call graph

```text
Before
  Execute                      (Executor · apps/runner/pkg/runner/v2/executor/executor.go:58)
    └─ executeJob              (Executor · apps/runner/pkg/runner/v2/executor/executor.go:108)   — switch on job.Type
         ├─ …                                                                                    — box lifecycle handlers
         └─ default            (Executor · apps/runner/pkg/runner/v2/executor/executor.go:146)   — every migration type ends here: "unknown job type"
  BoxBackend                   (interface · apps/runner/pkg/backend/backend.go:18)               — no export / import / destroy-if-present
  (no object-store client on the runner)

After
  run                          (main · apps/runner/cmd/runner/main.go:40)
    ├─ migrationArchiveStore   (main · apps/runner/cmd/runner/main.go:267)                        — nil when AWS_DEFAULT_BUCKET is unset
    │    └─ NewS3ArchiveStore  (S3ArchiveStore · apps/runner/pkg/storage/archive_store.go:49)
    ├─ migrationWorkDir        (main · apps/runner/cmd/runner/main.go:291)                        — MIGRATE_WORK_DIR, else <BOXLITE_HOME>/migrate
    └─ NewExecutor             (Executor · apps/runner/pkg/runner/v2/executor/executor.go:51)

  Execute                      (Executor · apps/runner/pkg/runner/v2/executor/executor.go:68)
    └─ executeJob              (Executor · apps/runner/pkg/runner/v2/executor/executor.go:118)   — five migration cases added
         ├─ exportBox          (Executor · apps/runner/pkg/runner/v2/executor/migrate.go:39)
         │    ├─ migrateArchivePayload (Executor · apps/runner/pkg/runner/v2/executor/migrate.go:134) — arcPath is required
         │    ├─ requireMigrationSetup (Executor · apps/runner/pkg/runner/v2/executor/migrate.go:147) — explicit config error, not a refusal to start
         │    ├─ ExportBox     (BoxliteAdapter · apps/runner/pkg/backend/boxlite_adapter.go:61)
         │    │    └─ ExportBox (boxlite.Client · apps/runner/pkg/boxlite/archive.go:17)          — packs the box into destDir
         │    └─ Upload        (S3ArchiveStore · apps/runner/pkg/storage/archive_store.go:93)     — echoes arcPath back to the control plane
         ├─ importBox          (Executor · apps/runner/pkg/runner/v2/executor/migrate.go:64)
         │    ├─ Download      (S3ArchiveStore · apps/runner/pkg/storage/archive_store.go:105)
         │    └─ ImportBox     (boxlite.Client · apps/runner/pkg/boxlite/archive.go:44)           — adopts a box already present
         ├─ rollbackExportBox  (Executor · apps/runner/pkg/runner/v2/executor/migrate.go:93)
         │    └─ Remove        (S3ArchiveStore · apps/runner/pkg/storage/archive_store.go:120)
         ├─ rollbackImportBox  (Executor · apps/runner/pkg/runner/v2/executor/migrate.go:106)
         │    └─ DestroyIfPresent (boxlite.Client · apps/runner/pkg/boxlite/archive.go:74)        — absent box counts as done
         └─ finishPreviousBox  (Executor · apps/runner/pkg/runner/v2/executor/migrate.go:114)
              ├─ Remove        (S3ArchiveStore · apps/runner/pkg/storage/archive_store.go:120)    — archive first
              └─ DestroyIfPresent (boxlite.Client · apps/runner/pkg/boxlite/archive.go:74)        — the box is dropped last
```

## Changes

- **Five migration handlers** (`executor/migrate.go`). Each converges on
  redelivery: object deletes are idempotent, an import adopts a box that is
  already present, destroying an absent box counts as done. They report failures
  verbatim rather than through `common.FormatRecoverableError` — that marker asks
  the control plane to destroy and rebuild the box, which is exactly the copy a
  migration exists to preserve.
- **`arcPath` is assigned by the control plane** and travels in the job payload
  (`executor/types.go`), so a job redelivered to a restarted runner acts on the
  object the control plane recorded for the box. A bare key resolves against the
  runner's own bucket; an `s3://bucket/key` URI names its own, which is what lets
  one migration span two runners with different buckets.
- **`ArchiveStore`** (`pkg/storage/archive_store.go`) — minio-backed upload /
  download / remove over the runner's own S3 credentials. The archive is
  infrastructure state moved between runners, not org-scoped user data, so it
  does not go through the control plane's push-access grant.
- **`ExportBox` / `ImportBox` / `DestroyIfPresent`** on `BoxBackend` and the
  boxlite client (`pkg/boxlite/archive.go`). `ImportBox` names the restored box
  after the control plane's box id, which is how every other runner call
  addresses it.
- **Migration is optional.** A runner with no `AWS_DEFAULT_BUCKET` logs a warning,
  serves every other job type as before, and fails a migration job with a
  configuration error instead of refusing to start. New `MIGRATE_WORK_DIR`
  defaults to `<BOXLITE_HOME>/migrate` and must hold a whole box disk.
- **Job type enum** extended across the API (`job-type.enum.ts`,
  `job-type-map.dto.ts`) and the generated Go/TS clients.

## How to verify

```bash
# arcPath / endpoint resolution — pure Go, no native runtime
go -C apps/runner test ./pkg/storage/...

# handlers + the redelivery branches against the real runtime
make dev:go
go -C apps/runner test -tags boxlite_dev ./pkg/runner/v2/executor/... ./pkg/boxlite/...
```

`migrate_test.go` drives each handler through a fake backend and a fake archive
store: arcPath echoed on export, no import when the download fails, the local
archive removed on both paths, and `FINISH_PREVIOUS_BOX` keeping the box when the
archive removal fails.

## Risks / rollout

- Additive only — no existing job type changes behavior. Runners deployed without
  a bucket keep working; migration jobs must not be scheduled to them.
- `MIGRATE_WORK_DIR` needs free space for a full box disk on both the exporting
  and importing runner.
- The control plane side of the state machine ships separately; these handlers
  are inert until it submits the jobs.

## Summary by CodeRabbit

* **New Features**
  * Added box export and import workflows, including rollback and source-discard operations.
  * Added portable archive creation, restoration, cloud storage, and cleanup support.
  * Added configurable migration staging through the `MIGRATE_WORK_DIR` environment setting.
  * Import operations safely handle boxes that already exist.
  * Added support for the new box migration job types across client interfaces.

* **Tests**
  * Added comprehensive coverage for successful, failed, rollback, cleanup, and storage scenarios.


apps/runner/pkg/backend/backend.go
```diff
@@ -26,6 +26,14 @@ type BoxBackend interface {
 	UpdateNetworkSettings(ctx context.Context, boxId string, settings dto.UpdateNetworkSettingsDTO) error
 	GetBoxState(ctx context.Context, boxId string) (enums.BoxState, error)
 
+	// Migration — move a box between runners through a portable archive.
+	// ExportBox returns the archive path the runtime wrote inside destDir.
+	ExportBox(ctx context.Context, boxId, destDir string) (string, error)
+	ImportBox(ctx context.Context, boxId, archivePath string) error
+	// DestroyIfPresent removes a box, treating an already-absent box as success
+	// so a redelivered teardown job converges.
+	DestroyIfPresent(ctx context.Context, boxId string) error
```
DorianZheng: duplicate with `Destroy`

apps/api/src/box/dto/job-type-map.dto.ts
```diff
@@ -46,6 +46,21 @@ export interface JobTypeMap {
...
+  [JobType.IMPORT_BOX]: {
+    resourceType: [ResourceType.BOX]
+  }
+  [JobType.ROLLBACK_EXPORT_BOX]: {
+    resourceType: [ResourceType.BOX]
+  }
+  [JobType.ROLLBACK_IMPORT_BOX]: {
+    resourceType: [ResourceType.BOX]
+  }
+  [JobType.FINISH_PREVIOUS_BOX]: {
```
DorianZheng: better name

apps/runner/cmd/runner/config/config.go
```diff
@@ -62,6 +62,10 @@ type Config struct {
 	GhcrToken                          string        `envconfig:"GHCR_TOKEN"`
 	DockerHubUsername                  string        `envconfig:"DOCKERHUB_USERNAME"`
 	DockerHubToken                     string        `envconfig:"DOCKERHUB_TOKEN"`
+	// MigrateWorkDir holds the .boxlite archive of a box being migrated in or
+	// out. It must live on a filesystem large enough for a whole box disk;
+	// empty means "<BoxliteHomeDir>/migrate".
+	MigrateWorkDir string `envconfig:"MIGRATE_WORK_DIR"`
```
DorianZheng: align

apps/runner/pkg/runner/v2/executor/migrate.go
```diff
@@ -0,0 +1,164 @@
...
+	payload, err := e.migrateArchivePayload(job)
+	if err != nil {
+		return nil, err
+	}
+	store, err := e.requireMigrationSetup()
+	if err != nil {
+		return nil, err
+	}
+
+	archivePath, err := e.backend.ExportBox(ctx, job.ResourceId, e.migrateWorkDir)
```
ltstriker, line 50: When err happen, some file may left in the dir, and also export file may be conflict in the same dir when running concurrently.

suggest: It will be better to use a spec dir( like <MIGRATE_WORK_DIR>/ \<job-id>-\<random>/  )  to export/import achieve, when error happened,  delete this dir for prevent some dirty file left.

Also, consider about add a clean dir job when the runner init,  for maybe terminal happened when export is doing.

apps/runner/pkg/boxlite/archive.go
```diff
@@ -0,0 +1,94 @@
...
+func (c *Client) ImportBox(ctx context.Context, boxId, archivePath string) error {
+	present, err := c.boxPresent(ctx, boxId)
+	if err != nil {
+		return err
+	}
+	if present {
+		c.logger.InfoContext(ctx, "import skipped, box already present", "box", boxId)
+		return nil
+	}
+
```
zhaoyuana777, line 53: Any existing box with the same ID is currently treated as a successfully replayed import. This cannot distinguish the expected imported copy from a stale or unrelated box left by an earlier migration. Please persist and verify migration provenance or the archive digest, and only skip the import when the existing box belongs to the same migration.

DefeatMan: The underlying API only guarantees idempotency, and the correctness of its state is guaranteed by the upper-level cloud platform.

apps/api/src/box/dto/job-type-map.dto.ts
DorianZheng: DESTROY_EXPORTED_BOX

DorianZheng (approved):

PR #1196, opened 2026-08-11
fix(rest): return immediately on volume create, drop synchronous wait

G4614:
Follow-up to #1191 (merged). `create` blocked up to 30s polling for the volume to become ready via `VolumeService.waitForReady`, risking client/gateway timeouts around that same window for what is an inherently async operation — bucket provisioning runs on a 5s reconciler tick (`VolumeManager.processPendingVolumes`), fully decoupled from the request.

**API**: `create` now returns 202 with `state=pending_create` as soon as the row exists, matching what the pre-existing classic `VolumeController` (`apps/api/src/box/controllers/volume.controller.ts`) already did — it never waited either. Removes `VolumeService.waitForReady`, now unused.

**Client-side wait**: the Rust core's `VolumeInfo`/`VolumeResponse` never carried `state` at all, so there was no way for any caller to observe readiness short of re-deserializing raw JSON. Adds a `VolumeState` enum (mirroring the API's) plus `state`/`error_reason` fields, and `VolumeHandle::wait_until_ready` — a client-side poll loop mirroring `ComputerBox`/`SkillBox`'s `waitUntilReady` in the Node SDK — that blocks until ready/error/timeout without the server holding a connection open. Wired into the CLI: `boxlite volume create --wait [--wait-timeout secs]` restores the old synchronous-feeling UX for callers who want it; `volume ls`/`get` now show a STATE column.

Callers who don't use `--wait` learn readiness via `GET /volumes/{id}` (always authoritative, no external dependency) or the existing `volume.state.updated` webhook if their org has one configured (best-effort, opt-in — not a delivery guarantee on its own).

Test plan:
- `yarn nx run api:build` / `yarn nx run api:test -- --testPathPatterns=volume` — 4 suites, 23 tests, all pass
- `cargo check -p boxlite --features rest` / `cargo build -p boxlite-cli` — clean
- `cargo test -p boxlite-cli volume::` — 2 new/updated unit tests pass
- `cargo fmt --all -- --check` — clean
- Manual E2E against a local isolated API+Postgres+MinIO stack: `boxlite volume create` returns in ~0.3s with `state=pending_create`; `boxlite volume create --wait` blocks ~2s and returns once `state=ready`; `boxlite volume ls` shows the new STATE column throughout

### DIFF
--- a/openapi/box.openapi.yaml
+++ b/openapi/box.openapi.yaml
@@ -192,19 +192,22 @@ paths:
     post:
       operationId: createVolume
       summary: Create a managed persistent volume
-      description: Creates a volume and waits until its backing object storage is ready.
+      description: |
+        Accepts the volume for creation and returns immediately (state
+        `pending_create`); provisioning the backing object storage happens
+        asynchronously. Poll `GET /volumes/{volume_id}` or subscribe to the
+        `volume.state.updated` webhook to learn when it reaches `ready`
+        (or `error`).
       tags: [Volumes]
       responses:
-        "201":
-          description: Volume created and ready for use
+        "202":
+          description: Volume accepted for creation
           content:
             application/json:
               schema:
                 $ref: "#/components/schemas/Volume"
         "400":
           $ref: "#/components/responses/BadRequestError"
-        "408":
-          $ref: "#/components/responses/RequestTimeoutError"
         "503":
           $ref: "#/components/responses/ServiceUnavailableError"
 
@@ -1294,13 +1297,6 @@ components:
               type: ImageError
               code: image_pull_failed
               request_id: req_01HZK4TNRPQSXYZ8WM6NCVT9R5
-
-    RequestTimeoutError:
-      description: The operation did not reach its target state before the timeout
-      content:
-        application/json:
-          schema:
-            $ref: "#/components/schemas/ErrorResponse"
 
     ServiceUnavailableError:
       description: A required backend service is not configured or available
--- a/src/boxlite/src/runtime/volumes.rs
+++ b/src/boxlite/src/runtime/volumes.rs
@@ -8,20 +8,27 @@
 //! The trait is `#[async_trait]` like the other capability backends
 //! ([`ImageBackend`](crate::runtime::images::ImageBackend),
 //! [`AuthBackend`](crate::runtime::auth::AuthBackend)) so REST backends can
-//! perform network calls. The concrete backend is not implemented yet — every
-//! operation currently returns `Unsupported`.
+//! perform network calls. The REST backend implements it against the managed
+//! `/v1/volumes` API; the local backend (`LocalRuntime`) does not have a
+//! volume store wired up yet and returns `Unsupported`.
 
 use std::sync::Arc;
+use std::time::Duration;
 
 use async_trait::async_trait;
 
-use crate::BoxliteResult;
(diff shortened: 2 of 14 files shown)


openapi/box.openapi.yaml
```diff
@@ -192,19 +192,22 @@ paths:
...
+        `pending_create`); provisioning the backing object storage happens
+        asynchronously. Poll `GET /volumes/{volume_id}` or subscribe to the
+        `volume.state.updated` webhook to learn when it reaches `ready`
+        (or `error`).
       tags: [Volumes]
       responses:
-        "201":
-          description: Volume created and ready for use
+        "202":
+          description: Volume accepted for creation
```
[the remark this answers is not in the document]
G4614: False positive — this compares two unrelated APIs that happen to share an operationId name.

`apps/api/src/box/controllers/volume.controller.ts` (classic, non-REST) returns 200 and is documented in a separate, untouched spec (`apps/libs/api-client-go/api/openapi.yaml:2818`) — both already agree with each other and neither is touched by this PR (`git diff origin/main -- apps/libs/api-client-go/api/openapi.yaml` is empty).

This PR's actual change is `apps/api/src/boxlite-rest/boxlite-volume.controller.ts` (the newer, versioned `/v1/{prefix}/volumes` REST surface) returning 202, documented in `openapi/box.openapi.yaml` — those two also already agree with each other. The oasdiff warning about a removed 201 is real but intentional: that's this PR's whole point (see PR description) — create used to synchronously wait up to 30s and return 201 once ready; it now returns 202 immediately with `state=pending_create`. There's no cross-controller Go/TS client regeneration needed since the two controllers were never sharing one contract to begin with.

src/boxlite/src/volumes/store.rs
```diff
@@ -2,14 +2,27 @@
...
+//! LocalRuntime` still returns `Unsupported` — there's no local volume store.
 
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 
+/// Lifecycle state of a managed volume, mirroring the API's `VolumeState`
+/// enum (`apps/api/src/box/enums/volume-state.enum.ts`).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum VolumeState {
```
DorianZheng, line 16: 3 states is good enough. and consistent with BoxState

src/boxlite/src/runtime/volumes.rs
```diff
@@ -70,4 +77,42 @@ impl VolumeHandle {
...
+    ///
+    /// `create()` returns as soon as the volume is accepted (state
+    /// `PendingCreate`) — provisioning the backing object storage happens
+    /// asynchronously on the server. This is a convenience for callers that
+    /// want the old "create and block until usable" behavior without the
+    /// server itself holding the connection open (which risked client/gateway
+    /// timeouts for what both `boxlite`'s runner and API treat as inherently
+    /// async). Mirrors `ComputerBox`/`SkillBox`'s `waitUntilReady` in the
+    /// Node SDK (`sdks/node/lib/computerbox.ts`, `skillbox.ts`).
+    pub async fn wait_until_ready(&self, id: &str, timeout: Duration) -> BoxliteResult<VolumeInfo> {
```
DorianZheng, line 92: remove

PR #1203, opened 2026-08-12
feat(api): mark boxes on draining runners for migration

DefeatMan:
## Summary

A box migration moves a stopped box between runners through an archive in the object store. The runner already performs one step per job (#1189); this adds the box-table half of the state machine and the loop that opens it.

## Call graph

Before

```text
handleCheckDecommissionRunners   (RunnerService · apps/api/src/box/services/runner.service.ts:643)      — waits for a draining runner to empty
  └─ count                       (BoxRepository · apps/api/src/box/repositories/box.repository.ts)      — counts boxes still owned by it
                                                                                                          nothing moves a parked box, so the count never drops
  …
exportBox                        (Executor · apps/runner/pkg/runner/v2/executor/migrate.go:40)          — never reached: no migration job is ever submitted
```

After

```text
markBoxesOnDrainingRunners       (BoxMigrationService · apps/api/src/box/services/box-migration.service.ts:50)  — cron, once a minute, behind a Redis lock
  ├─ findDrainingRunnerIds       (BoxMigrationService · apps/api/src/box/services/box-migration.service.ts:80)  — draining and not decommissioned
  └─ markParkedBoxesForExport    (BoxRepository · apps/api/src/box/repositories/box.repository.ts:263)          — one conditional UPDATE over the parked boxes
       └─ migrateState = PENDING_EXPORT, migrateUpdatedAt = updatedAt = CURRENT_TIMESTAMP
                                 (Box · apps/api/src/box/entities/box.entity.ts:180)                            — the state a later step reacts to
  …
exportBox                        (Executor · apps/runner/pkg/runner/v2/executor/migrate.go:40)                  — unchanged; still waits on a job a later PR submits
```

## Changes

- `Box` carries four migrate columns. `migrateState` names the job the migration is waiting on, never what it has produced — that is what `migrateArcPath` (an archive on the object store) and `migrateRunnerId` (a box on a second runner) record, so a rollback can reclaim either one whether or not the migration stayed valid. `migrateRunnerId` is its own column rather than a reuse of `prevRunnerId`, which the archived-box start path already clears for its own purpose.
- `BoxMigrationService` claims, every minute, the boxes parked on a draining runner — stopped, wanting to stay stopped, and not already migrating — in one conditional `UPDATE`. Concurrent starts and stops either land before it and disqualify the box, or land after it and break the timestamp equality a later validity check reads.
- `migrateUpdatedAt` and `updatedAt` are written from the same statement's `CURRENT_TIMESTAMP`, so the two start out equal and any later write from outside the migration moves `updatedAt` alone. That broken equality is how every later step tells that a box was touched underneath it.
- `COMPLETED` is claimable again, so a runner that drains a second time re-migrates what it took back.
- The schema change is additive and pre-deploy: the enum defaults to `none` and the timestamp stays null until a migration writes it.

## How to verify

```bash
cd apps
npx nx test api --testPathPatterns=box-migration.service.spec.ts --testPathPatterns=box.repository.migrate-marker.spec.ts
```

The marker's SQL is covered against a real Postgres by `box.repository.migrate-marker.integration.spec.ts`, which runs whenever `DB_HOST` is set and skips otherwise.

## Risks / rollout

- No behavior changes until a runner is flagged `draining`; on a fleet with none, the loop finds no runners and returns.
- This PR only opens the migration — nothing submits `EXPORT_BOX` yet, so a marked box sits in `pending_export` until the step that submits it lands. Rollback is dropping the migration and the service registration.

## Summary by CodeRabbit

* **New Features**
  * Added automated tracking of parked boxes that need export during runner draining.
  * Added migration states for export, import, discard, rollback, and completion.
  * Added scheduled processing with safe locking and recovery when processing fails.
  * Added database support for preserving box timestamps and migration progress.

* **Bug Fixes**
  * Improved activity updates so newer timestamps reliably take precedence while preserving existing activity data.

* **Tests**
  * Added unit and integration coverage for claiming, locking, retries, exclusions, cleanup, and activity recording behavior.

### DIFF
--- a/apps/api/src/box/enums/box-migration-state.enum.ts
+++ b/apps/api/src/box/enums/box-migration-state.enum.ts
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2026 BoxLite AI
+ * SPDX-License-Identifier: AGPL-3.0
+ */
+
+/**
+ * Where a box is in a migration from one runner to another.
+ *
+ * A box that is not migrating has no `box_migration` row at all, so there is no
+ * state for it; every value here names a job the migration is waiting on, and is
+ * left behind by that job's receiver:
+ *
+ *     (no row) ─▶ PENDING_EXPORT ─▶ PENDING_IMPORT ─▶ PENDING_DISCARD_EXPORTED ─▶ COMPLETED
+ *                       └──────────────┴──▶ PENDING_ROLLBACK ─▶ (row deleted)
+ *
+ * The state records what the migration is waiting for, never what it has
+ * produced: `box_migration.arcPath` says an archive is on the object store and
+ * `box_migration.runnerId` says a box exists on another runner. Nor does the
+ * state mark a job as in flight — a box sits in PENDING_IMPORT both before its
+ * IMPORT_BOX job is submitted and while it runs, and the per-box Redis lock is
+ * what tells those two apart.
+ */
+export enum BoxMigrationState {
+  PENDING_EXPORT = 'pending_export',
+  PENDING_IMPORT = 'pending_import',
+  PENDING_DISCARD_EXPORTED = 'pending_discard_exported',
+  PENDING_ROLLBACK = 'pending_rollback',
+  COMPLETED = 'completed',
+}
--- a/apps/api/src/box/box.module.ts
+++ b/apps/api/src/box/box.module.ts
@@ -45,15 +45,17 @@ import { RegionBoxAccessGuard } from './guards/region-box-access.guard'
 import { ProxyGuard } from './guards/proxy.guard'
 import { EventEmitter2 } from '@nestjs/event-emitter'
 import { BoxLastActivity } from './entities/box-last-activity.entity'
+import { BoxMigration } from './entities/box-migration.entity'
 import { BoxActivityService } from './services/box-activity.service'
 import { BoxStateWaiterService } from './services/box-state-waiter.service'
+import { BoxMigrationService } from './services/box-migration.service'
 
 @Module({
   imports: [
     UserModule,
     OrganizationModule,
     RegionModule,
-    TypeOrmModule.forFeature([Box, Runner, WarmPool, Volume, Region, Job, BoxLastActivity]),
+    TypeOrmModule.forFeature([Box, Runner, WarmPool, Volume, Region, Job, BoxLastActivity, BoxMigration]),
   ],
   controllers: [BoxController, RunnerController, PreviewController, VolumeController, JobController],
   providers: [
@@ -75,6 +77,7 @@ import { BoxStateWaiterService } from './services/box-state-waiter.service'
     JobStateHandlerService,
     BoxActivityService,
     BoxStateWaiterService,
+    BoxMigrationService,
     BoxAccessGuard,
     RunnerAccessGuard,
     RegionRunnerAccessGuard,
--- a/apps/api/src/box/entities/box-migration.entity.ts
+++ b/apps/api/src/box/entities/box-migration.entity.ts
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2026 BoxLite AI
+ * SPDX-License-Identifier: AGPL-3.0
+ */
+
+import { Column, Entity, Index, JoinColumn, OneToOne, PrimaryColumn } from 'typeorm'
+import { BoxMigrationState } from '../enums/box-migration-state.enum'
+import { Box } from './box.entity'
+
+/**
+ * A box migration between runners, one row per box being moved.
+ *
+ * The row is the migration: it appears when the marker claims a parked box and
+ * is deleted once the box is no longer migrating, so "not migrating" has a
+ * single representation and the table stays the size of the work in flight
+ * rather than the size of the fleet. Deleting the box takes the row with it.
+ */
+@Entity('box_migration')
+@Index('box_migration_state_idx', ['state'])
+export class BoxMigration {
+  @PrimaryColumn({ name: 'boxId' })
+  boxId: string
+
+  //  Which migration job this box is waiting on, never what the migration has
+  //  produced — that is what `arcPath` and `runnerId` record.
+  @Column({
+    type: 'enum',
+    enum: BoxMigrationState,
+  })
+  state: BoxMigrationState
+
+  //  Object key of the archive the migration put on the object store. Empty
+  //  means there is no archive to reclaim. Written whether or not the migration
+  //  stayed valid, because the object exists either way and the rollback path
+  //  needs the key to delete it.
+  @Column({ type: 'character varying', default: '' })
+  arcPath = ''
+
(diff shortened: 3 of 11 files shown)


apps/api/src/box/enums/box-migrate-state.enum.ts
```diff
@@ -0,0 +1,29 @@
...
+ * `box.migrateRunnerId` says a box exists on another runner. Nor does the state
+ * mark a job as in flight — a box sits in PENDING_IMPORT both before its
+ * IMPORT_BOX job is submitted and while it runs, and the per-box Redis lock is
+ * what tells those two apart.
+ */
+export enum BoxMigrateState {
+  NONE = 'none',
+  PENDING_EXPORT = 'pending_export',
+  PENDING_IMPORT = 'pending_import',
+  PENDING_DISCARD_SOURCE = 'pending_discard_source',
```
DorianZheng: PENDING_DISCARD_EXPORTED_BOX?

apps/api/src/box/entities/box.entity.ts
```diff
@@ -169,6 +170,43 @@ export class Box {
...
+  migrateArcPath = ''
+
+  //  The second runner a migration involves: before the commit point the runner
+  //  the box was imported onto, after it the runner still holding the original
+  //  box to discard. Null means no such box exists.
+  @Column({
+    type: 'uuid',
+    nullable: true,
+  })
+  migrateRunnerId?: string
```
DorianZheng: what if the migrate runner crash

apps/api/src/box/entities/box.entity.ts
```diff
@@ -169,6 +170,43 @@ export class Box {
   @Column({ nullable: true })
   daemonVersion?: string
 
+  //  Which migration job this box is waiting on. NONE for the boxes that are
+  //  not migrating, which is nearly all of them.
+  @Column({
```
DorianZheng: Migration-related fields should be decoupled into a dedicated table.

apps/api/src/box/enums/box-migrate-state.enum.ts
DefeatMan: Same style as box-state.enum without the _BOX suffix

apps/api/src/box/entities/migrate.entity.ts
```diff
@@ -0,0 +1,61 @@
...
+ * A box migration between runners, one row per box being moved.
+ *
+ * The row is the migration: it appears when the marker claims a parked box and
+ * is deleted once the box is no longer migrating, so "not migrating" has a
+ * single representation and the table stays the size of the work in flight
+ * rather than the size of the fleet. Deleting the box takes the row with it.
+ */
+@Entity('migrate')
+@Index('migrate_state_idx', ['state'])
+export class Migrate {
```
DorianZheng: how about `BoxMigration`

apps/api/src/box/enums/box-migrate-state.enum.ts
```diff
@@ -0,0 +1,29 @@
...
+ *                       └──────────────┴──▶ PENDING_ROLLBACK ─▶ (row deleted)
+ *
+ * The state records what the migration is waiting for, never what it has
+ * produced: `migrate.arcPath` says an archive is on the object store and
+ * `migrate.runnerId` says a box exists on another runner. Nor does the state
+ * mark a job as in flight — a box sits in PENDING_IMPORT both before its
+ * IMPORT_BOX job is submitted and while it runs, and the per-box Redis lock is
+ * what tells those two apart.
+ */
+export enum BoxMigrateState {
```
DorianZheng: `BoxMigrationState`

apps/api/src/migrations/pre-deploy/1786200000000-add-migrate-table-migration.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export class AddMigrateTable1786200000000 implements MigrationInterface {
+  name = 'AddMigrateTable1786200000000'
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    // An earlier revision of this same unreleased migration carried the migrate
+    // fields as box columns. Nothing has shipped and no migration has ever run
+    // in a deployed environment, so a database that took that revision drops
+    // them rather than backfilling the table from them.
+    await queryRunner.query(`ALTER TABLE "box" DROP COLUMN IF EXISTS "migrateUpdatedAt"`)
```
DorianZheng: do we need this

DorianZheng (dismissed):
left comments, rest LGTM

apps/api/src/box/enums/box-migration-state.enum.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+/**
+ * Where a box is in a migration from one runner to another.
+ *
+ * A box that is not migrating has no `migrate` row at all, so there is no state
+ * for it; every value here names a job the migration is waiting on, and is left
+ * behind by that job's receiver:
+ *
+ *     (no row) ─▶ PENDING_EXPORT ─▶ PENDING_IMPORT ─▶ PENDING_DISCARD_EXPORTED ─▶ COMPLETED
```
ltstriker, line 13: what is the diff between no row and COMPLETED

ltstriker (dismissed):

apps/api/src/migrations/pre-deploy/1786200000000-add-migrate-table-migration.ts
```diff
@@ -0,0 +1,32 @@
...
+    // "not migrating" state: that is the absence of a row, which is also what
+    // the cascade leaves behind when the box goes away.
+    //
+    // `arcPath` is empty, not null, when there is no archive to reclaim — the
+    // rollback path tests it for emptiness, and one sentinel keeps that a plain
+    // comparison instead of a null-or-empty pair. `updatedAt` is NOT NULL
+    // because the marker copies `box.updatedAt` into it in the statement that
+    // creates the row.
+    await queryRunner.query(
+      `CREATE TABLE "migrate" ("boxId" character varying NOT NULL, "state" "public"."migrate_state_enum" NOT NULL, "arcPath" character varying NOT NULL DEFAULT '', "runnerId" uuid, "updatedAt" TIMESTAMP WITH TIME ZONE NOT NULL, CONSTRAINT "migrate_boxId_pk" PRIMARY KEY ("boxId"), CONSTRAINT "migrate_boxId_fk" FOREIGN KEY ("boxId") REFERENCES "box"("id") ON DELETE CASCADE ON UPDATE NO ACTION)`,
```
DorianZheng: box_migration

apps/api/src/box/enums/box-migration-state.enum.ts
DefeatMan: no diff, just marked. could be removed later if everything go fine.

DorianZheng (dismissed):

PR #1214, opened 2026-08-12
feat(api): submit box migration jobs from the control plane

DefeatMan:
## Summary

The marker (#1203) opens a migration on a parked box; this adds the two control-plane loops that carry it forward, one job at a time. Stacked on #1203 — the first commit here is that PR.

## Call graph

Before

```text
markBoxesOnDrainingRunners      (BoxMigrationService · apps/api/src/box/services/box-migration.service.ts:50)   — cron, opens a migration per parked box (#1203)
  └─ markParkedBoxesForExport   (BoxRepository · apps/api/src/box/repositories/box.repository.ts:271)           — one box_migration row, state = pending_export
                                                                                                                 nothing reads that state, so the row just sits there
exportBox                       (Executor · apps/runner/pkg/runner/v2/executor/migrate.go:40)                   — never reached: no migration job is ever created
```

After

```text
submitMigrationJobs             (BoxMigrationManager · apps/api/src/box/managers/box-migration.manager.ts:104)  — cron every 10s, one replica behind a Redis lock
  ├─ findMigrations             (BoxMigrationManager · apps/api/src/box/managers/box-migration.manager.ts:175)  — pending_export / pending_import, most recently used box first
  └─ submitValidated            (BoxMigrationManager · apps/api/src/box/managers/box-migration.manager.ts:241)  — the shape both legs share
       ├─ getMigrateJobLockKey  (lock-key.util · apps/api/src/box/utils/lock-key.util.ts:21)                    — per-box-per-job lock; taken here, released by the job's receiver
       ├─ isUndisturbedBy       (BoxMigration · apps/api/src/box/entities/box-migration.entity.ts:75)           — ValidCheck under a pessimistic lock on the box row
       │    └─ markRollback     (BoxMigrationManager · apps/api/src/box/managers/box-migration.manager.ts:372)  — box was touched → turn the migration around instead
       ├─ getRandomAvailableRunner (RunnerService · apps/api/src/box/services/runner.service.ts:722)            — import leg only: a runner that is not the box's and not draining
       └─ createJob             (JobService · apps/api/src/box/services/job.service.ts:50)                      — EXPORT_BOX on the holder, then IMPORT_BOX on the target
            └─ exportBox / importBox (Executor · apps/runner/pkg/runner/v2/executor/migrate.go:40,65)           — archive to the object store, restore from it

submitRollbackJobs              (BoxMigrationManager · apps/api/src/box/managers/box-migration.manager.ts:135)  — cron every minute
  ├─ reclaim                    (BoxMigrationManager · apps/api/src/box/managers/box-migration.manager.ts:303)  — ROLLBACK_EXPORT_BOX for the archive, ROLLBACK_IMPORT_BOX for the second box
  │    └─ finishRollback        (BoxMigrationManager · apps/api/src/box/managers/box-migration.manager.ts:387)  — nothing left to reclaim → the migration row is deleted
  └─ discardExported            (BoxMigrationManager · apps/api/src/box/managers/box-migration.manager.ts:325)  — DISCARD_EXPORTED_BOX on the runner the box moved away from
```

## Changes

- The submitter hands a box the job its migration state names — `EXPORT_BOX` on the runner holding it, then `IMPORT_BOX` on a runner that is neither that one nor being drained. Neither loop advances the state on success: the receiver of the job status does, so a state only moves once the work behind it is done.
- The rollback submitter hands out the jobs that reclaim what a turned-around migration left behind — the archive on the object store, the box on the second runner — and after the commit point discards the box the migration moved away from. Both artifacts can be outstanding at once, and each is reclaimed by its own job on the runner that holds it.
- Because a migration's state names the job it waits on rather than whether that job has been submitted, a per-box-per-job Redis lock is what marks one in flight. The submitter takes it and the job-status receiver releases it, so it spans the runner's work rather than the tick that started it; its TTL (15 min) outlasts the job table's stale sweep (`JobService.handleStaleJobs`, 10 min) so an expiring lock can never put a second job on a runner still working on the first.
- Every submission is guarded by ValidCheck, taken on the box row under a pessimistic lock immediately before the irreversible step: the box has to still be parked, and `box.updatedAt` has to still equal the copy the marker took. Any write from outside the migration turns the migration around rather than letting it move a box out from under its user.
- A `PENDING_IMPORT` migration with no recorded archive cannot finish, so it is turned around rather than retried forever; rollback finds nothing to reclaim and drops the row, leaving the box for the marker to claim again.

## How to verify

```bash
cd apps
npx nx test api --testPathPatterns=box-migration.manager.spec.ts
```

`box-migration.manager.spec.ts` covers both loops against a stubbed data source: the export/import targets, ValidCheck turning a touched box around, the in-flight lock being held on success and released on failure, scan order, both artifacts reclaimed on their own runners, and the row being deleted when nothing is left.

## Risks / rollout

- Nothing is submitted until a runner is flagged `draining` and the marker opens a migration; on a fleet with none, both loops find no rows and return.
- The job-status receiver is not in this PR. Until it lands, an opened migration gets its `EXPORT_BOX` submitted and the runner will run it, but no state advances and no in-flight lock is released — so a migration stops after its first job and the box stays parked on the draining runner. Do not flag a runner `draining` on a deployment carrying this PR without the receiver.
- Rollback is dropping the two cron registrations (`BoxMigrationManager` in `box.module.ts`); no schema change of its own.

## Summary by CodeRabbit

- **New Features**
  - Added automated box migration workflows for draining runners.
  - Added support for exporting, importing, rolling back, and discarding migration archives.
  - Added migration state tracking and recovery for interrupted or invalid migrations.
  - Added configurable archive storage locations.
  - Added safeguards for changed or otherwise ineligible boxes.

- **Bug Fixes**
  - Improved activity timestamp updates while preserving newer data.
  - Added cleanup for missing or obsolete migration artifacts.

- **Tests**
  - Added comprehensive unit and integration coverage for migration workflows, locking, recovery, and activity persistence.

### DIFF
--- a/apps/api/src/box/managers/box-migration.manager.ts
+++ b/apps/api/src/box/managers/box-migration.manager.ts
@@ -1,0 +1,442 @@
+/*
+ * Copyright 2026 BoxLite AI
+ * SPDX-License-Identifier: AGPL-3.0
+ */
+
+import { Injectable, Logger, OnApplicationShutdown } from '@nestjs/common'
+import { Cron, CronExpression } from '@nestjs/schedule'
+import { InjectDataSource } from '@nestjs/typeorm'
+import { DataSource, EntityManager, FindOptionsWhere, IsNull, Not } from 'typeorm'
+import { setTimeout } from 'timers/promises'
+
+import { RedisLockProvider } from '../common/redis-lock.provider'
+import { Box } from '../entities/box.entity'
+import { BoxMigration } from '../entities/box-migration.entity'
+import { BoxMigrationState } from '../enums/box-migration-state.enum'
+import { JobType } from '../enums/job-type.enum'
+import { ResourceType } from '../enums/resource-type.enum'
+import { JobService } from '../services/job.service'
+import { RunnerService } from '../services/runner.service'
+import { getMigrateJobLockKey } from '../utils/lock-key.util'
+
+import { LogExecution } from '../../common/decorators/log-execution.decorator'
+import { WithInstrumentation } from '../../common/decorators/otel.decorator'
+import { TrackJobExecution } from '../../common/decorators/track-job-execution.decorator'
+import { TrackableJobExecutions } from '../../common/interfaces/trackable-job-executions'
+import { TypedConfigService } from '../../config/typed-config.service'
+
+// Only one API replica sweeps at a time; the TTL only bounds a replica that
+// dies mid-tick, since the happy path unlocks in `finally`.
+const SUBMIT_LOOP_LOCK_KEY = 'box-migration-submit-worker-selected'
+const ROLLBACK_LOOP_LOCK_KEY = 'box-migration-rollback-worker-selected'
+const LOOP_LOCK_TTL_SECONDS = 60
+
+// A migration job's lock outlives the tick that took it — the receiver of the
+// job status releases it. This TTL is the backstop for a status that never
+// arrives, so it has to outlast the job table's own stale sweep (10 minutes,
+// JobService.handleStaleJobs); otherwise the lock would expire and a second job
+// would be submitted while the runner is still working on the first.
+const MIGRATE_JOB_LOCK_TTL_SECONDS = 15 * 60
+
+const MIGRATIONS_PER_TICK = 100
+
+/** Mirrors the runner's MigrateArchivePayload (executor/types.go). */
+interface MigrateArchivePayload {
+  arcPath: string
+}
+
+interface MigrateJobTarget {
+  runnerId: string
+  payload?: MigrateArchivePayload
+}
+
+/**
+ * Drives the two migration loops of the control plane.
+ *
+ * Submitter (10s) hands a migration its next job — export, then import — each
+ * guarded by ValidCheck, which is what catches a user touching the box
+ * mid-flight and turns the migration around.
+ *
+ * RollBackSubmitter (60s) hands out the jobs that reclaim what a migration
+ * left behind: the archive on the object store, the box on the other runner.
+ *
+ * Neither loop advances the migration's state on success — the receiver of the
+ * job status does, so a state only moves once the work behind it is done.
+ */
+@Injectable()
+export class BoxMigrationManager implements TrackableJobExecutions, OnApplicationShutdown {
+  activeJobs = new Set<string>()
+
+  private readonly logger = new Logger(BoxMigrationManager.name)
+
+  /** Read once: the prefix is validated at boot and fixed for the process. */
+  private readonly archivePrefix: string
+
+  constructor(
+    @InjectDataSource()
+    private readonly dataSource: DataSource,
+    private readonly runnerService: RunnerService,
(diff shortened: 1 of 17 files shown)


apps/api/src/box/managers/box-migration.manager.ts
```diff
@@ -0,0 +1,398 @@
...
+  payload?: MigrateArchivePayload
+}
+
+/**
+ * The object store key this migration moves the box through. Derived from the
+ * box so a retried export overwrites its own archive instead of stranding one:
+ * a box has at most one migration in flight — one `box_migration` row — and the
+ * key is dropped once that migration commits or rolls back.
+ */
+function migrateArcPath(boxId: string): string {
```
DorianZheng: should be configurabled

apps/api/src/migrations/pre-deploy/1786200000000-add-box-migration-table-migration.ts
```diff
@@ -0,0 +1,32 @@
...
+    // "not migrating" state: that is the absence of a row, which is also what
+    // the cascade leaves behind when the box goes away.
+    //
+    // `arcPath` is empty, not null, when there is no archive to reclaim — the
+    // rollback path tests it for emptiness, and one sentinel keeps that a plain
+    // comparison instead of a null-or-empty pair. `updatedAt` is NOT NULL
+    // because the marker copies `box.updatedAt` into it in the statement that
+    // creates the row.
+    await queryRunner.query(
+      `CREATE TABLE "box_migration" ("boxId" character varying NOT NULL, "state" "public"."box_migration_state_enum" NOT NULL, "arcPath" character varying NOT NULL DEFAULT '', "runnerId" uuid, "updatedAt" TIMESTAMP WITH TIME ZONE NOT NULL, CONSTRAINT "box_migration_boxId_pk" PRIMARY KEY ("boxId"), CONSTRAINT "box_migration_boxId_fk" FOREIGN KEY ("boxId") REFERENCES "box"("id") ON DELETE CASCADE ON UPDATE NO ACTION)`,
```
DorianZheng, line 20: duplicated with previous PR

apps/api/src/box/managers/box-migration.manager.ts
```diff
@@ -0,0 +1,398 @@
...
+      if (!migration.box) {
+        this.logger.error(`Migration of box ${migration.boxId} has no box row`)
+        return false
+      }
+      return true
+    })
+  }
+
+  private async submitExport(migration: BoxMigration): Promise<void> {
+    if (!migration.box.runnerId) {
```
DorianZheng, line 220: can you elaborate on when this happens

apps/api/src/migrations/pre-deploy/1786200000000-add-box-migration-table-migration.ts
DefeatMan: this PR contains previous, review by selecting commits

PR #1215, opened 2026-08-12
feat(api): advance box migrations from reported job status

DefeatMan:
## Summary

The submitters (#1214) hand out one migration job at a time and never move a box forward on success; this adds the receiver that does, so a migration state only changes once the work behind it is done. Stacked on #1203 and #1214 — the first two commits here are those PRs.

## Call graph

Before

```text
updateJobStatus                 (JobService · apps/api/src/box/services/job.service.ts:232)                        — a runner reports the job it finished
  └─ handleJobCompletion        (JobStateHandlerService · apps/api/src/box/services/job-state-handler.service.ts:39) — switch on job.type; the five migration types match no case
                                                                                                                     the migrate state never moves and the job's in-flight lock is never released
submitMigrationJobs             (BoxMigrationManager · apps/api/src/box/managers/box-migration.manager.ts:104)     — next tick still sees pending_export, still locked: the migration stops after its first job (#1214)
```

After

```text
updateJobStatus                 (JobService · apps/api/src/box/services/job.service.ts:232)
  └─ handleJobCompletion        (JobStateHandlerService · apps/api/src/box/services/job-state-handler.service.ts:39)
       └─ isMigrationJobType    (box-migration-job-receiver.service · apps/api/src/box/services/box-migration-job-receiver.service.ts:29)  — hand it over before the box's state-change lock is taken
            └─ handleJobCompletion (BoxMigrationJobReceiver · apps/api/src/box/services/box-migration-job-receiver.service.ts:91)          — releases the job's in-flight lock whatever happened
                 ├─ receiveExport          (…:128)  — pending_export → pending_import, or pending_rollback when ValidCheck fails; records the archive either way
                 │    └─ exportedArcPath   (…:326)  — the key the control plane assigned, never the one the runner echoes back
                 ├─ receiveImport          (…:153)  — commit point: box.runnerId → the target runner, source runner becomes the copy to discard
                 │    ├─ withMigration     (…:258)  — box row locked first, migrate row second; ValidCheck read under that lock
                 │    │    └─ isUndisturbedBy (BoxMigration · apps/api/src/box/entities/box-migration.entity.ts:75)
                 │    └─ invalidate        (BoxLookupCacheInvalidationService · apps/api/src/box/services/box-lookup-cache-invalidation.service.ts:35)  — the cached box still names the runner that gave it up
                 ├─ receiveRollbackExport / receiveRollbackImport (…:214, …:221)  — each clears the field naming the artifact it freed
                 ├─ receiveDiscardExported (…:231)  — pending_discard_exported → completed
                 └─ updateMigration        (…:297)  — every write conditional on the state the job was submitted from
```

## Changes

- Export records where the archive landed — the key the control plane assigned in the job's payload, never the one the runner echoes back, because that key is later handed to a *second* runner as something to download and to delete. A mismatched echo is logged: it means the runner wrote somewhere the migration will never reclaim.
- Import moves box ownership to the runner the box was imported onto, and is the commit point: ownership moves in the same transaction that checks the box is still the migration's to move, the runner it came from becomes the copy left to discard, and the box's lookup cache is dropped because it still names a runner that no longer serves it.
- The reclaim jobs each clear the field naming what they freed, and the discard of the exported box completes the migration.
- Both legs that move real data re-run ValidCheck where they record their result, because the box may have been started while the runner worked. That turns the migration around towards rollback rather than failing it, and records the artifact either way — it exists either way, and rollback needs the field to reclaim it.
- Every write is conditional on the state the job was submitted from, so a status redelivered after a restart matches nothing and applies once. The box row is locked before the migrate row, the order the marker and the submitter take them, so the three never deadlock; locking the box is also what makes ValidCheck mean anything, since `box.updatedAt` is what anyone may move between the check and the write it guards.
- A failed job is not the receiver's to repair: the state stays where it is and the submitting loop retries. The job's lock is released either way, because that lock is what marks the job in flight — and it is the only lock a migration status touches, since the box's state-change lock belongs to a concurrent start or stop of the same box.

## How to verify

```bash
cd apps
npx nx test api --testPathPatterns=box-migration-job-receiver.service.spec.ts
npx nx test api --testPathPatterns=job-state-handler.service.spec.ts
```

`box-migration-job-receiver.service.spec.ts` drives each leg against a stubbed data source: the assigned archive key winning over the reported one, ValidCheck turning a touched box around on both legs, ownership moving with the source runner left to discard, the cache invalidation, a failed job leaving the state alone, a redelivered status applying nothing, and each reclaim job clearing its own field. `job-state-handler.service.spec.ts` covers the hand-off: a migration job reaches the receiver and never takes the box's state-change lock, a non-migration job is unaffected.

## Risks / rollout

- Nothing runs until a runner is flagged `draining` and the marker opens a migration; with none, the receiver is never reached.
- With this PR the chain is complete: an opened migration exports, imports, moves ownership, and discards the copy it left behind. #1214 alone stops after the first job, so deploy them together.
- Rollback is dropping this PR's commit: the five migration job types stop being answered and migrations park where they are, on states the submitting loops still read.
- No schema change of its own (the `box_migration` table comes from #1203).

## Summary by CodeRabbit

* **New Features**
  * Added automated migration support for parked boxes during runner draining.
  * Added export, import, rollback, and cleanup workflows with migration status tracking.
  * Added configurable, validated archive storage prefixes.
  * Added safeguards for concurrent operations, stale data, failed jobs, and migration recovery.
  * Completed migrations now update ownership and invalidate relevant cached information.

* **Bug Fixes**
  * Improved activity timestamp updates to preserve newer stored values and fill missing timestamps.

* **Tests**
  * Added comprehensive unit and integration coverage for migration workflows and activity persistence.

### DIFF
--- a/apps/api/src/box/services/job-state-handler.service.ts
+++ b/apps/api/src/box/services/job-state-handler.service.ts
@@ -16,6 +16,7 @@ import { Box } from '../entities/box.entity'
 import { RedisLockProvider } from '../common/redis-lock.provider'
 import { ResourceType } from '../enums/resource-type.enum'
 import { getStateChangeLockKey } from '../utils/lock-key.util'
+import { BoxMigrationJobReceiver, isMigrationJobType } from './box-migration-job-receiver.service'
 
 /**
  * Service for handling entity state updates based on job completion (v2 runners only).
@@ -28,6 +29,7 @@ export class JobStateHandlerService {
   constructor(
     private readonly boxRepository: BoxRepository,
     private readonly redisLockProvider: RedisLockProvider,
+    private readonly boxMigrationJobReceiver: BoxMigrationJobReceiver,
   ) {}
 
   /**
@@ -40,6 +42,15 @@ export class JobStateHandlerService {
     }
 
     if (!job.resourceId) {
+      return
+    }
+
+    // A migration job carries its own lock — taken by the loop that submitted
+    // it, released by the receiver below — and never the box's state-change
+    // lock. Handing it to the tail of this method would delete a lock a
+    // concurrent start or stop of the same box is holding.
+    if (isMigrationJobType(job.type)) {
+      await this.boxMigrationJobReceiver.handleJobCompletion(job)
       return
     }
 
--- a/apps/api/src/box/box.module.ts
+++ b/apps/api/src/box/box.module.ts
@@ -50,6 +50,7 @@ import { BoxMigration } from './entities/box-migration.entity'
 import { BoxActivityService } from './services/box-activity.service'
 import { BoxStateWaiterService } from './services/box-state-waiter.service'
 import { BoxMigrationService } from './services/box-migration.service'
+import { BoxMigrationJobReceiver } from './services/box-migration-job-receiver.service'
 
 @Module({
   imports: [
@@ -80,6 +81,7 @@ import { BoxMigrationService } from './services/box-migration.service'
     BoxStateWaiterService,
     BoxMigrationService,
     BoxMigrationManager,
+    BoxMigrationJobReceiver,
     BoxAccessGuard,
     RunnerAccessGuard,
     RegionRunnerAccessGuard,
--- a/apps/api/src/box/box.module.ts
+++ b/apps/api/src/box/box.module.ts
@@ -15,6 +15,7 @@ import { RunnerService } from './services/runner.service'
 import { Runner } from './entities/runner.entity'
 import { RunnerController } from './controllers/runner.controller'
 import { BoxManager } from './managers/box.manager'
+import { BoxMigrationManager } from './managers/box-migration.manager'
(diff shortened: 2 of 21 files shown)


apps/api/src/box/services/job-state-handler.service.ts
```diff
@@ -43,6 +45,15 @@ export class JobStateHandlerService {
       return
     }
 
+    // A migration job carries its own lock — taken by the loop that submitted
+    // it, released by the receiver below — and never the box's state-change
+    // lock. Handing it to the tail of this method would delete a lock a
+    // concurrent start or stop of the same box is holding.
+    if (isMigrationJobType(job.type)) {
```
DorianZheng, line 52: move to below switch-case clause for consistent type distribution

PR #1216, opened 2026-08-13
feat(cli): remove --mount

G4614:
Removes `--mount` (`src=volume://<id>,target=<path>`), the only way to reference a managed volume by id at box-create time. No replacement in this PR — `-v`/`--volume` keeps its existing local bind-mount / anonymous-volume semantics completely unchanged. A follow-up PR reintroduces managed-volume support through `-v` once its syntax is settled (design still under discussion — see #1208).

**Call graph**

Before:
```
--mount src=volume://<id>,target=<box> → parse_managed_mount_spec → normalize_managed_volume_source → apply_managed_to → managed volume
-v <host>:<box>                        → parse_volume_spec → apply_to → local bind mount / anonymous volume (unchanged)
```

After:
```
-v <host>:<box> → parse_volume_spec → apply_to → local bind mount / anonymous volume (only path now, byte-for-byte unchanged from before this PR)
```

Also removes `GlobalFlags::resolves_rest_runtime()` — it existed only to back the `--mount`-requires-REST-runtime check, now dead.

**Test plan:** `cargo test -p boxlite-cli --bin boxlite` — 179/179 pass.

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed support for the obsolete `--mount` volume syntax.
  * Managed volumes must now be configured using the `-v` option.
  * Removed related validation and configuration behavior for managed-volume mounts.

### DIFF
--- a/src/cli/src/cli.rs
+++ b/src/cli/src/cli.rs
@@ -314,14 +314,6 @@ impl GlobalFlags {
         }
     }
 
-    pub fn resolves_rest_runtime(&self) -> bool {
-        let stored = crate::credentials::load_named(&self.resolved_profile())
-            .ok()
-            .flatten();
-        let env_api_key = std::env::var("BOXLITE_API_KEY").ok();
-        self.resolve_rest_options(stored, env_api_key).is_some()
-    }
-
     /// Build REST connection options from the selected credential profile and
     /// the ambient `BOXLITE_API_KEY`. Returns `None` when no URL is configured
     /// (the caller then falls back to the local runtime). Pure — takes the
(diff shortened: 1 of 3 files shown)


G4614:
Fixed in 8580448: `looks_like_host_path` now treats `\\` as a path separator (Windows-relative sources like `subdir\\cache:/guest` classify as bind mounts again), added the `subdir\\cache:/guest` regression test (two-side verified: fails without the fix), ran `cargo fmt`, and added a `text` language tag to the volume-grammar fence.

DorianZheng (approved):

PR #1220, opened 2026-08-13
fix(box): persist the resolved volume id, not the caller's name, at create

G4614:
## Summary

Fixes a pre-existing bug in the managed-volume-at-create feature (#1056/#1192): creating a box with a volume referenced by *name* (`source: volume://<name>`) silently persisted the name as `volumeId` instead of the volume's real id, so the runner could never find the matching S3 bucket (named by id). Referencing by id already worked.

Found while verifying #1216 (`-v` now accepts a bare managed-volume name/id) end to end against a live deployment: the same create request succeeds and mounts with the real uuid, but 500s with the name.

## Call graph

**Before:**
```
POST /boxes {volumes:[{source:'volume://myvolume', guest_path:'/data'}]}
  -> BoxService.create()
      -> VolumeService.validateVolumes(orgId, ['myvolume'])
           checks 'myvolume' exists + is ready, returns void
      -> box.volumes = resolveVolumes(createBoxDto.volumes)
           ← BUG: volumeId stays 'myvolume' (raw caller string), never
             substituted with the volume's real id
      -> next start(): runner looks up the S3 bucket by 'myvolume'
         instead of the real id -> mount fails, box errors
```

**After:**
```
POST /boxes {volumes:[{source:'volume://myvolume', guest_path:'/data'}]}
  -> BoxService.create()
      -> VolumeService.validateVolumes(orgId, ['myvolume'])
           now returns the resolved Volume[] (existence + ready checks unchanged)
      -> builds an id/name -> Volume map from the resolved entities
      -> box.volumes stores volume.id for every entry, whether the
         caller passed an id or a name
```

## Testing

- Two new `BoxService.create()` unit tests (`box.service.spec.ts`): one confirms a name-based volume request persists the resolved id, one confirms an id-based request is unaffected.
- Two-side verified: reverted both files, confirmed the name-based test fails with `"volumeId": "my-vol"` (the bug, verbatim) instead of the real id, restored, confirmed both new tests + the rest of the file (29/29) pass.
- `VolumeService.validateVolumes`'s existing 9 tests unaffected by the return-type change (`Promise<void>` → `Promise<Volume[]>`; no test asserted on the old return value).
- Repro'd directly against a live deployment via curl before writing the fix: `source: volume://<uuid>` → box created and mounted successfully (verified by exec'ing in and reading back a written file); `source: volume://<name>` → 500. Not re-verified live post-fix (this branch isn't deployed anywhere).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed box creation with volume names so mounts are saved using the correct volume identifiers.
  * Preserved volume identifiers when volumes are provided directly.
  * Improved volume validation to consistently resolve requested volumes and prevent invalid or unavailable volumes from being attached.

### DIFF
--- a/apps/api/src/box/services/box.service.spec.ts
+++ b/apps/api/src/box/services/box.service.spec.ts
@@ -8,6 +8,7 @@ import { BoxService } from './box.service'
 import { BoxState } from '../enums/box-state.enum'
 import { BoxDesiredState } from '../enums/box-desired-state.enum'
 import { RunnerState } from '../enums/runner-state.enum'
+import { VolumeState } from '../enums/volume-state.enum'
 import { BoxEvents } from '../constants/box-events.constants'
 
 // ensureStartedForProxy touches boxRepository + eventEmitter +
@@ -305,6 +306,7 @@ describe('BoxService public defaults', () => {
         release: jest.fn().mockResolvedValue(undefined),
       }),
     }
+    const volumeService = { validateVolumes: jest.fn().mockResolvedValue([]) }
     const service = Object.create(BoxService.prototype) as BoxService
     Object.assign(service as any, {
       getValidatedOrDefaultRegion: jest.fn().mockResolvedValue({ id: 'region-1' }),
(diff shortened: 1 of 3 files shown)


G4614:
Superseded by #1350 — `VolumeService.validateVolumes` now returns `Map<selector, canonicalId>` instead of `void`, so the resolved id is persisted rather than the caller's raw selector (see the PR's call-graph writeup for the exact bug this fixes). Closing as done.

PR #1221, opened 2026-08-13
fix(box): reject UUID-shaped volume names

G4614:
## Summary

Rejects a `name` shaped like a UUID when creating a volume. Surfaced while looking at #1220: `VolumeService.findOneByIdOrName`/`validateVolumes` resolve a caller-supplied string against *both* the `id` and `name` columns in one query — nothing stops a name from colliding with a different volume's real id in the same org, making that lookup ambiguously match two rows.

## Call graph

**Before:**
```
POST /volumes {name: "<some other volume's real uuid>"}
  -> VolumeService.create(): only checks for an existing volume with that *name*
     (organizationId, name unique constraint) — never checks it against other
     volumes' ids
  -> later: findOneByIdOrName("<that uuid>", org) / validateVolumes(...)
       where: [{id: In([uuid])}, {name: In([uuid])}]
     <- BUG: matches two rows (the real owner by id, this one by name);
        which one "wins" depends on array/Map iteration order, not any
        defined rule
```

**After:**
```
POST /volumes {name: "<anything UUID-shaped>"}
  -> CreateVolumeDto.name: @Validate(IsNotUuidShapedConstraint) -> 400
     (uuid's own `validate()`, not a hand-rolled regex)
  -> the id and name namespaces stay disjoint by construction, so the
     ambiguous double-match above can no longer happen
```

Also adds the missing `@IsOptional()` on `name` — without it, omitting `name` entirely (the common case; the volume then defaults its name to its own id) fails `@IsString()` under class-validator's default undefined handling, confirmed with a probe test before writing the real one.

## Testing

- 3 new tests (`create-volume.dto.spec.ts`): rejects a UUID-shaped name, accepts an ordinary name, accepts an omitted name.
- Two-side verified: reverted the file, confirmed 2 of the 3 tests fail (the UUID-rejection one accepts the malicious name; the omitted-name one now genuinely needs `@IsOptional()` back), restored, confirmed all 3 pass.

### DIFF
--- a/apps/api/src/box/dto/create-volume.dto.spec.ts
+++ b/apps/api/src/box/dto/create-volume.dto.spec.ts
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2025 Daytona Platforms Inc.
+ * Modified by BoxLite AI, 2025-2026
+ * SPDX-License-Identifier: AGPL-3.0
+ */
+
+import 'reflect-metadata'
+import { validate } from 'class-validator'
+import { plainToInstance } from 'class-transformer'
+import { CreateVolumeDto } from './create-volume.dto'
+
+// VolumeService.findOneByIdOrName/validateVolumes match ids and names
+// through the same lookup, so a name shaped like a UUID could collide with
+// another volume's real id in the same organization and make that lookup
+// ambiguous. Rejecting the UUID shape at the request boundary keeps the two
+// namespaces disjoint by construction.
+describe('CreateVolumeDto name', () => {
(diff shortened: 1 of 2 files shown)


G4614:
Superseded by #1350 — rather than rejecting UUID-shaped names at creation, `validateVolumes` now queries both `id` and `name` columns and resolves collisions with an explicit id-wins-over-name precedence rule, eliminating the ambiguity this PR guarded against. Closing as no longer needed.

PR #1224, opened 2026-08-13
feat(rest): accept name on POST /v1/volumes

G4614:
## Summary

`POST /v1/volumes` now honors a caller-supplied `name` instead of silently ignoring the request body. Split out of #1208, which bundled this with two unrelated concerns — see #1220/#1223/#1102/#1216/#1221 for related work.

## Call graph

**Before:**
```
POST /v1/volumes {name: "my-vol"}
  -> BoxliteVolumeController.create(authContext) — no @Body() at all
  -> this.volumeService.create(authContext.organization, {})
       ← BUG: the caller's body is silently ignored; name always
         defaults to the server-assigned id
```

**After:**
```
POST /v1/volumes {name: "my-vol"}
  -> BoxliteVolumeController.create(authContext, @Body() body: CreateVolumeDto)
  -> this.volumeService.create(authContext.organization, body)
       name is honored, still defaults to the id when omitted
       (VolumeService.create(), unchanged)
```

`CreateVolumeDto` also rejects a UUID-shaped name — same guard as #1221's fix to the classic `POST /volumes` DTO. `VolumeService.findOneByIdOrName` matches a caller's string against both the `id` and `name` columns in one lookup, so an unguarded name shaped like a UUID could collide with a different volume's real id in the same organization. This endpoint would have reopened that exact gap without it.

## Testing

- `yarn nx run api:test --testPathPatterns=create-volume.dto.spec.ts` — 4/4 pass. Two-side verified: removing the UUID-shape decorator makes exactly that one test fail (the other 3 unaffected), restoring it makes all 4 pass.
- `yarn nx run api:test --testPathPatterns=boxlite-volume.controller.spec.ts` — 8/8 pass
- `npx prettier --check` on the new/changed files

### DIFF
--- a/apps/api/src/boxlite-rest/boxlite-volume.controller.spec.ts
+++ b/apps/api/src/boxlite-rest/boxlite-volume.controller.spec.ts
@@ -36,10 +36,13 @@ describe('BoxliteVolumeController', () => {
     const organization = { id: 'org-1' }
 
     await expect(
-      controller.create({
-        organization,
-        organizationId: organization.id,
-      } as never),
+      controller.create(
+        {
+          organization,
+          organizationId: organization.id,
+        } as never,
+        {},
+      ),
     ).resolves.toEqual({
       id: volume.id,
       name: volume.name,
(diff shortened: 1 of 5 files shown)


G4614:
Superseded by #1350 — `CreateVolumeDto` now accepts `name?: string` with Docker-style pattern validation (`VOLUME_NAME_PATTERN`), and `boxlite volume create --name my-data` ships in the CLI. Closing as done.

PR #1231, opened 2026-08-13
fix(runtime): derive RLIMIT_FSIZE from disk_size_gb

### LINKED ISSUE #1152: RLIMIT_FSIZE defaults to 1 GiB regardless of the disk size accepted in the same options, killing the VM with no crash record
Running at rev \`8803834036205cf2cac5cfca98bb3875812c897a\`.

## What happens

A box created with an 8 GiB disk dies the instant its guest has written 1 GiB. The client sees:

```
Attach stream error: status: Unknown, message: "h2 protocol error: error reading a body from connection"
```

and the exec wait RPC reports \`exit_code: -1\`. Both are the client watching the whole microVM disappear, not a stream fault.

## Root cause

The VM monitor is jailed behind \`RLIMIT_FSIZE\` with a default ceiling of 1 GiB, while the same options struct accepts a much larger \`disk_size_gb\` backed by a single growing qcow2 file on the host. When the qcow2 crosses the ceiling the monitor takes \`SIGXFSZ\`, and the shim installs no handler for it, so the process dies leaving no exit file and no crash record.

## Evidence

- \`disk.qcow2\` stops at exactly \`1073741824\` bytes, and stops at the same byte with \`disk_size_gb: 32\`.
- Replacing the workload (an npm install) with plain \`dd\` reproduces it identically, so it is write volume, not the tool.
- 20 seconds after the death every further exec returns \`Connection refused\` — the VM is gone.
- The console log ends mid-normal-operation: no OOM kill, no panic, no kernel oops.
- Disk and memory were both healthy 5 s before death (\`2371 used / 5644 free\` MiB disk, \`288 used / 1698 avail\` MiB memory).

## Why it matters

Any guest workload that writes more than a gibibyte hits this, which is ordinary for a package install or a build. Because the monitor leaves no record, it surfaces as an unrelated-looking transport error, so the time to diagnose is long.

## Suggested fixes

1. Derive the file ceiling from the disk size the caller asked for, rather than a fixed default that can contradict it.
2. Handle \`SIGXFSZ\` in the shim and write a crash record, so the failure names itself instead of appearing as an h2 error.
3. Consider whether \`exit_code: -1\` should be distinguishable from a real exit status, since it currently means "no exit status was ever observed".

We worked around 1 by sizing the ceiling from the requested disk plus headroom on our side, but the default contradicting the accepted disk size looks worth fixing upstream.

G4614:
Fixes #1152

Derives the jailed shim's `RLIMIT_FSIZE` from the box's own `disk_size_gb` instead of the fixed 1 GiB in `SecurityOptions::default()`, which cannot see it.

## Where

In `BoxOptions::sanitize` (`src/boxlite/src/runtime/options.rs`), next to the field it derives from:

```rust
self.advanced.security.resource_limits.max_file_size = Some(
    self.disk_size_gb
        .unwrap_or(DEFAULT_DISK_SIZE_GB)
        .saturating_mul(FSIZE_DISK_MULTIPLIER)
        .saturating_mul(1024 * 1024 * 1024),
);
```

`spawn.rs` and `advanced_options.rs` are untouched by this PR now. The earlier revision's `min_file_size()`, its qcow2 header reads, its margin constant, and the `ResourceLimits::ensure_min_file_size` helper with its three tests are all gone.

## Why loose slack rather than an exact size

`RLIMIT_FSIZE` is a runaway-write backstop, not a capacity policy. Capacity is the qcow2 virtual size, where a guest that fills its disk gets a clean `ENOSPC` inside the box; enforcing capacity through the rlimit instead degrades that into a host-side `SIGXFSZ` that kills the VM and reaches the client as an opaque h2 transport error. Slack costs nothing here. Being a few hundred MiB short costs the box.

No other project ships a fixed `fsize` default for the VMM process — Firecracker's jailer leaves `file_size: None`, gVisor uses `RLimInfinity`, runc/OCI is pass-through. Whether boxlite should keep this ceiling at all is a fair question; this PR keeps it and makes the number follow the disk.

## `sanitize` now takes `&mut self`

Two consequences worth reviewing:

1. One runtime call site changes (`rt_impl`'s create path, which already owns the options and returns them). The rest are tests.
2. An explicit caller-supplied `max_file_size` is **overwritten**, not respected as a floor — a plain assignment, per review. The disk-derived value is authoritative.
3. A **trusted archive import** no longer gets its security options back verbatim. `options_from_manifest` runs `sanitize` and then returns a trusted archive's options unchanged, so the file-size ceiling is now the one field that gets rewritten. `trusted_import_preserves_archive_configuration` is updated to state that exemption instead of asserting blanket preservation.

## Accepted regression

`create_cow_disk` sizes the COW disk at `max(disk_size_gb, base image size)`, so a box whose image is more than 2x its requested size still lands under the ceiling — `diskSizeGb: 4` on a 10 GiB image gets a 10 GiB disk under an 8 GiB limit and dies with the same `EFBIG`.

The right fix is at the other end: asking for a disk smaller than the image should be rejected at create rather than silently widened. Until that lands, this case fails. The unit test that covered it is removed with the header read.

## Test plan

- [x] `cargo test -p boxlite --lib sanitize_` — the two new cases pass. Two-side verified: with the multiplier at 1, both size assertions fail with the exact byte counts; restored, green.
- [x] `cargo test -p boxlite --lib` — 994 passed, 2 failed. Both failures are environmental and pre-existing: `net::socket_path::tests::vendored_libkrun_still_derives_krun_sock_suffix` reads the absent libkrun submodule, and `jailer::sandbox::seatbelt::tests::test_seatbelt_runtime_denies_write_outside_writable_path` expects a denial that does not materialise on an already-sandboxed host. CI is the authority.
- [x] `cargo clippy -p boxlite --lib --tests -- -D warnings` and `cargo fmt` — clean.
- [ ] Not re-run end to end on a real box. The earlier revision's manual repro (apt + a large npm/Playwright install, 85+ minutes clean on a default box) covers the default-size path, which this revision widens rather than narrows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved disk usage protection by applying a file-size limit based on the configured disk capacity.
  * Added a safe default limit when no disk size is specified.
  * Ensured trusted and untrusted archive imports receive appropriate disk-derived file-size protection while preserving applicable security settings.

### DIFF
--- a/src/boxlite/src/runtime/import.rs
+++ b/src/boxlite/src/runtime/import.rs
@@ -120,6 +120,14 @@ fn options_from_manifest(
         return Err(rejected_upload("host volume mounts"));
     }
     options.advanced.security = SecurityOptions::default();
+    // That reset also restores the default's hardcoded 1 GiB RLIMIT_FSIZE —
+    // the ceiling #1152 is about — because it replaces the whole struct rather
+    // than just the isolation fields it means to. The box still boots on a
+    // disk sized from its own `disk_size_gb`, so re-derive the limit.
+    // `sanitize` assigns it outright, so running it twice is idempotent.
+    options.sanitize().map_err(|error| {
+        BoxliteError::InvalidArgument(format!("invalid archive box_options: {error}"))
+    })?;
 
     Ok(options)
 }
@@ -221,6 +229,7 @@ pub(crate) fn validate_no_backing_references(disk_path: &Path) -> BoxliteResult<
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::runtime::types::Bytes;
     use tempfile::TempDir;
 
     fn v3_manifest(options: BoxOptions) -> ArchiveManifest {
@@ -366,7 +375,32 @@ mod tests {
             options_from_manifest(&v3_manifest(options), ArchiveImportPolicy::UntrustedRemote)
                 .unwrap();
 
-        assert_eq!(resolved.advanced.security, SecurityOptions::default());
+        // Everything but the file-size ceiling is the server default; that one
+        // is derived from the box's own disk (#1152), so it is the default's
+        // stale 1 GiB that must NOT come back.
+        let mut expected = SecurityOptions::default();
+        expected.resource_limits.max_file_size = Some(Bytes::from_gib(20).as_bytes());
+        assert_eq!(resolved.advanced.security, expected);
+    }
+
+    /// The server-default reset must not put the 1 GiB ceiling back: an
+    /// uploaded archive's box boots on a disk sized from its own
+    /// `disk_size_gb` and needs a limit that covers it.
+    #[test]
+    fn untrusted_import_derives_the_fsize_limit_from_the_disk() {
+        let options = BoxOptions {
+            disk_size_gb: Some(20),
+            ..BoxOptions::default()
+        };
+
+        let resolved =
+            options_from_manifest(&v3_manifest(options), ArchiveImportPolicy::UntrustedRemote)
+                .unwrap();
+
+        assert_eq!(
+            resolved.advanced.security.resource_limits.max_file_size,
+            Some(Bytes::from_gib(40).as_bytes())
+        );
     }
 
     #[test]
@@ -385,7 +419,13 @@ mod tests {
 
         assert!(resolved.advanced.nested_virtualization);
         assert!(resolved.advanced.privileged);
-        assert_eq!(resolved.advanced.security, SecurityOptions::disabled());
+
+        // `sanitize` derives RLIMIT_FSIZE from `disk_size_gb` (#1152), so the
+        // one security field a trusted import does not carry over verbatim is
+        // the file-size ceiling. Everything else is preserved.
+        let mut expected = SecurityOptions::disabled();
+        expected.resource_limits.max_file_size = Some(Bytes::from_gib(20).as_bytes());
+        assert_eq!(resolved.advanced.security, expected);
     }
 
     #[test]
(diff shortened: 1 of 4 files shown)


ltstriker:
how about untrusted import?

src/boxlite/src/runtime/import.rs
```diff
@@ -366,7 +374,32 @@ mod tests {
             options_from_manifest(&v3_manifest(options), ArchiveImportPolicy::UntrustedRemote)
                 .unwrap();
 
-        assert_eq!(resolved.advanced.security, SecurityOptions::default());
+        // Everything but the file-size ceiling is the server default; that one
+        // is derived from the box's own disk (#1152), so it is the default's
+        // stale 1 GiB that must NOT come back.
+        let mut expected = SecurityOptions::default();
+        expected.resource_limits.max_file_size = Some(20 * 1024 * 1024 * 1024);
```
ltstriker: use constant but not magic number.

src/boxlite/src/runtime/options.rs
```diff
@@ -627,12 +632,32 @@ impl BoxOptions {
...
+        //
+        // `FSIZE_DISK_MULTIPLIER` is deliberately loose: this is a
+        // runaway-write backstop, not a capacity policy. Capacity is the qcow2
+        // virtual size, where a guest that fills its disk gets a clean `ENOSPC`
+        // inside the box rather than a host-side `SIGXFSZ` that kills the VM.
+        self.advanced.security.resource_limits.max_file_size = Some(
+            self.disk_size_gb
+                .unwrap_or(DEFAULT_DISK_SIZE_GB)
+                .saturating_mul(FSIZE_DISK_MULTIPLIER)
+                .saturating_mul(1024 * 1024 * 1024),
```
ltstriker: the same magic number problem.

ltstriker (approved):

Issue #933, opened 2026-07-07
Expose container exit — box state, exit code, and resource cleanup


NavidMitchell:
## What we're building

We orchestrate boxlite micro VMs as managed workloads (long-running services and
run-to-completion jobs such as database migrations). For batch jobs the host needs to
know **that** the workload finished, **whether** it succeeded, and to release the VM's
resources when it does.

## Current behavior (0.9.7, macOS arm64, Node SDK)

When a box's entrypoint exits, nothing observable happens at the box level:

- The VM stays up indefinitely and `runtime.getInfo(id)` keeps reporting
  `state.running: true` / `status: "running"` — we polled for 60s past a 3-second
  entrypoint's exit with no transition.
- The exit code is not exposed through the API (we can see an `exit.previous` file is
  written inside the box directory, so it appears to be recorded internally).
- The only external signal that the workload ended is a subsequent `exec` failing:

  ```
  spawn_failed: Container init process exited — cannot exec.
  Original error: ... failed operation due to incompatible container status: `Stopped`
  ```

So a completed job is indistinguishable from a running one, its exit status is
unreachable, and the VM continues to hold its memory until the host calls `stop()`.
Our current workaround is to poll with `exec` and treat that specific failure as
"completed" — which is fragile, loses the exit code, and keeps the VM alive in the
meantime.

## What we'd like

When the container's init process exits:

1. **Transition the box state** so `getInfo`/`listInfo` reflect it — e.g.
   `status: "exited"`, `running: false`.
2. **Expose the exit code** on the info object (e.g. `state.exitCode`) — it appears to
   already be recorded on disk.
3. **Release the VM's resources** at that point (or offer an option to), so a finished
   job doesn't hold guest memory until the host notices.

A `wait(idOrName)`-style API resolving with the exit code would be a welcome
complement, but the state transition and exit code alone would unblock us.

## Reproduction

```typescript
import { SimpleBox, getJsBoxlite } from "@boxlite-ai/boxlite";

const runtime = getJsBoxlite().withDefaultConfig();
const box = new SimpleBox({
  image: "alpine:latest",
  name: "batch-demo",
  autoRemove: false,
  runtime,
  entrypoint: ["sh", "-c", "echo working && sleep 3"],
});
await box.getId();
const handle = await runtime.get("batch-demo");
await handle.start();

// entrypoint exits after ~3s; poll for a minute:
for (let i = 0; i < 60; i++) {
[...]

PR #1237, opened 2026-08-13
feat(sdks): expose init exit code on box info

NavidMitchell:
## Summary

The core records the init's exit code when a box stops because its command exited, but every SDK binding drops the field on the way out — so callers see a completed workload transition to `stopped` yet cannot read why it stopped. Refs #933, whose state-transition and exit-code asks landed in the core with #988; this is the binding half.

## Call graph

Before
```
on_shim_exit                  (BoxWatcher   · src/boxlite/src/litebox/watcher.rs:180)  — main command exited
  └─ record_main_command_exit (rt_impl      · src/boxlite/src/runtime/rt_impl.rs:51)   — sets state.exit_code, marks stop
       └─ BoxStateInfo::new   (BoxStateInfo · src/boxlite/src/runtime/types.rs:501)    — carries exit_code
            ├─ JsBoxStateInfo::from    (napi · sdks/node/src/info.rs:158)              — drops exit_code
            ├─ PyBoxStateInfo::from    (pyo3 · sdks/python/src/info.rs:266)            — drops exit_code
            ├─ CBoxInfo::from_box_info (ffi  · sdks/c/src/info.rs:239)                 — drops exit_code
            └─ cBoxInfoToGo            (cgo  · sdks/go/info.go:160)                    — drops exit_code
```

After
```
on_shim_exit                  (BoxWatcher   · src/boxlite/src/litebox/watcher.rs:180)  — unchanged
  └─ record_main_command_exit (rt_impl      · src/boxlite/src/runtime/rt_impl.rs:51)   — unchanged
       └─ BoxStateInfo::new   (BoxStateInfo · src/boxlite/src/runtime/types.rs:501)    — unchanged
            ├─ JsBoxStateInfo::from    (napi · sdks/node/src/info.rs:158)  — maps → info.state.exitCode
            ├─ PyBoxStateInfo::from    (pyo3 · sdks/python/src/info.rs:266) — maps → info.state.exit_code
            ├─ CBoxInfo::from_box_info (ffi  · sdks/c/src/info.rs:239)      — maps → info->exit_code + has_exit_code
            └─ cBoxInfoToGo            (cgo  · sdks/go/info.go:160)         — maps → info.ExitCode
```

## Changes

- Node and Python surface the code as a nullable field on the state object (`exitCode` / `exit_code`), beside the existing `pid`.
- C pairs `exit_code` with a `has_exit_code` flag. `pid` and `started_at` can use `0` for "not recorded", but `0` is the most common *valid* exit code, so it cannot carry the same sentinel. The alternative is a nullable `int *exit_code` matching the existing `network` pointer idiom — allocation-free won out, but say the word and I'll switch it before merge (the Go mapping moves with it).
- Go maps that pair to `ExitCode *int`, nil when no code was recorded, mirroring the existing nullable `Network *NetworkInfo`.
- All four bindings dropped the field identically, so all four are fixed here rather than just the one that surfaced it.
- `sdks/c/include/boxlite.h` is regenerated output from `make dev:c`, not a hand edit.
- Reference docs for the Node and Python `state` object list the new field.

## How to verify

- `cargo test -p boxlite-node --lib` — `box_state_conversion_preserves_recorded_exit_code` asserts a stopped box carries the code and a running one carries none.
- `make dev:go && cd sdks/go && go test -tags boxlite_dev -run TestCBoxInfoToGoTraversesExitCode ./...` — three subtests over the C→Go boundary.
- To confirm that Go test earns its keep, replace `has_exit_code` with a `code != 0` sentinel in `cBoxInfoToGo`: `zero_exit_code_is_not_absence` fails while the other two still pass.
- `cargo test -p boxlite-python --lib` cannot link (`pyo3` is pinned with `extension-module`), so the Python test module — the new case and the pre-existing one beside it — is not reached by any `make` target. Flagging rather than working around it.

## Risks / rollout

- `CBoxInfo` gains two trailing fields, so C consumers that construct or copy the struct by value need a recompile against the regenerated header.

## Summary by CodeRabbit

* **New Features**
  * Box status information now reports the process exit code when available.
  * Added optional exit-code support across the C, Go, Node.js, and Python SDKs.
  * Exit code `0` is preserved and distinguished from an unavailable value.

* **Documentation**
  * Updated Node.js and Python reference documentation to describe the new exit-code fields.

* **Tests**
  * Added coverage for running and stopped boxes, missing exit codes, and zero exit codes.


NavidMitchell:
I'm not very familiar with your code base. I used Claude to help me with this. Unfortunately this is a blocker for me. Anything else I can work through or help with I am willing to do. Thank you!

NavidMitchell:
I fixed the merge conflict with main, since the newest release.

Issue #1255, opened 2026-08-15
infra: deploy role reaches sibling stages via role/boxlite-* and secretsmanager:*


DorianZheng:
The deploy role created by `apps/infra/bootstrap/aws/github-deploy-role.yaml` is bound to one GitHub Environment, but two of its grants reach every stage in the account. Both predate the stage-configuration change that surfaced them, and neither is narrowed there because getting the resource patterns wrong fails a real deploy with `AccessDenied`, which cannot be checked without a live preview.

They matter more than they used to. That change makes the SST secret store authoritative for what a stage deploys, so cross-stage reach is now reach over another stage's configuration, not only over its runtime resources.

## 1. Sibling-stage runtime roles

`ManageBoxLiteRoles` grants `iam:PutRolePolicy`, `iam:UpdateAssumeRolePolicy`, `iam:PassRole`, `iam:DeleteRole` and friends on:

```
arn:aws:iam::<account>:role/boxlite-*
```

`boxlite-*` matches every stage's SST-created runtime roles, so a job bound to `dev` can rewrite or delete the roles `prod`'s services run as.

The deploy role's *own* role and the runtime boundary policy are already excluded by the `DenySelfPrivilegeEscalation` statement on this branch — this issue is the rest.

**Likely fix:** scope to `role/boxlite-${GitHubEnvironment}-*`.
**Why it is not done here:** it depends on every SST-created role actually carrying the stage as a name prefix. If any does not, the next deploy fails and clearing it means redeploying the bootstrap stack with a wider policy.

## 2. `secretsmanager:*` on `Resource: '*'`

The `BoxLiteAwsControlPlane` statement grants `secretsmanager:*` account-wide, so one stage can read and mutate another's secrets.

**Likely fix:** `arn:aws:secretsmanager:<region>:<account>:secret:boxlite-${GitHubEnvironment}-*`, which is exactly how the runtime permissions boundary in the same template already scopes `secretsmanager:GetSecretValue`.
**Why it is not done here:** same reason — untested against a real deploy, and SST may address secrets it creates by an ARN that does not carry the stage.

## Validating a fix

`apps/infra/deployment/release-safety.test.ts` has the harness: `deployRolePolicyStatements` collects every inline and attached policy, and the account-wide inventory in *"the grants that are NOT stage-scoped are the documented ones, and only those"* is where these two would stop being listed as exceptions.

[...]

PR #1257, opened 2026-08-15
feat(infra): read stage config from the SST secret store

DorianZheng:
A stage's deploy-time configuration lived in two control planes. The GitHub Environment secret `DEPLOY_ENV` was written out as `apps/infra/.env` for the length of a job, so changing what a stage deployed needed no AWS access at all, and offboarding meant revoking both halves. A local deploy and CI also read different files, so the two could drift.

`npm run bootstrap` now loads `.env` into the stage's SST secret store, together with a manifest of the key names it owns and a digest of their values — one `sst secret load`, because `secret load` is a read-modify-write of one document and an interrupted one leaves some keys new and some old. The manifest cannot detect that; every name it lists is still present. `deployment/sst.ts` hydrates `process.env` from that store before spawning sst, so `stack/*.ts` keeps reading plain `process.env` / `envOr()` values and needs no change.

Three rules narrow what may reach `process.env`, since `sst secret set` is a command any deployer can run against any name: the key must be named by the manifest, must be one the deploy demonstrably reads, and must not be local-only. A store that is unreadable, unbootstrapped, torn, half-written, or that would apply nothing stops the command — `diff` included, because a preview built from defaults is not a preview of the apply that follows it.

Not everything can move. The Cloudflare provider credentials cannot: reading the store initializes every provider `app()` declares, so a token kept there would be needed in order to read itself. They stay in SSM and the stage's GitHub Environment, which also holds `AWS_ACCOUNT_ID` and `AWS_REGION` — both read before any AWS credentials exist.

## Call graph

Line numbers are as of `c3c536ee` for Before and this branch's head for After.

```
Before
  Materialize stage configuration   (deploy job · .github/workflows/deploy-infra.yml:419)   ← BUG: secrets.DEPLOY_ENV written to apps/infra/.env, a second control plane
    └─ loadDeploymentEnvironment    (sst wrapper · apps/infra/deployment/sst.ts:105)        ← BUG: the whole materialized .env into process.env, stale local values included
         └─ fetchFromSsm            (sst wrapper · apps/infra/deployment/sst.ts:178)        — Cloudflare credentials, the only thing not in that file
              └─ spawn sst deploy   (sst wrapper · apps/infra/deployment/sst.ts:105)        — stack/*.ts then reads process.env / envOr()

After
  ensureStageConfig                 (bootstrap · apps/infra/bootstrap/bootstrap.ts:864)     — the one write, after every check has passed
    ├─ validateDotenvSyntax         (guard · apps/infra/deployment/validate-environment.ts:205)  — rejects only what dotenv itself would drop
    ├─ prepareStageConfigLoad       (bootstrap · apps/infra/bootstrap/environment.ts:261)   — one read of .env; values + manifest + digest decided together
    ├─ storeCloudflareCredential    (bootstrap · apps/infra/bootstrap/bootstrap.ts:398)     — GitHub Environment first, then SSM; a torn pair names --force
    └─ sst secret load              (bootstrap · apps/infra/bootstrap/bootstrap.ts:536)     — one load, so the digest describes the values beside it

  loadStageConfig                   (sst wrapper · apps/infra/deployment/sst.ts:385)        — every plan-building subcommand, `diff` included
    ├─ loadDeploymentEnvironment    (sst wrapper · apps/infra/deployment/sst.ts:150)        — .env promoted for local-only keys only, never stage config
    ├─ fetchFromSsm                 (sst wrapper · apps/infra/deployment/sst.ts:370)        — before the store read; reading it initializes that provider
    ├─ StageConfigStore.read        (store · apps/infra/deployment/stage-config.ts:203)     — sst secret list; values are never logged
    ├─ hydrateStageConfig           (store · apps/infra/deployment/stage-config.ts:156)     — manifest ∩ allowlist ∖ local-only; a real env var still wins
    └─ five refusals → exit 1       (sst wrapper · apps/infra/deployment/sst.ts:261)        — unreadable │ no manifest │ digest absent or mismatched │ missing │ applies nothing
         └─ spawn sst deploy        (sst wrapper · apps/infra/deployment/sst.ts:385)        — stack/*.ts reads process.env / envOr(), unchanged
```

## What review added

The first commit is the change above. The rest came out of reviewing it, and each is a separate commit:

- **Cross-stage IAM.** The deploy role's account-wide `s3:*`/`ssm:*` grants are scoped to this stage's state prefixes, passphrase and buckets. `/sst/bootstrap` — which names the buckets every stage shares — became read-only. Two privilege-escalation paths are denied outright: the role's own name matches the `role/boxlite-*` it can write, and so does every sibling stage's deploy role and runtime-boundary policy.
- **Honest documentation.** `docs/security.md` now states what the policy grants rather than what it ought to, including three grants that are shared by construction and one shared write. Two that reach other stages are **not** fixed here and are tracked in #1255 — narrowing them needs an `apply=false` preview against a real stage, since a wrong resource pattern fails a deploy with AccessDenied.
- **Bootstrap correctness.** Auth0 provisioning read `process.env.STACK_DOMAIN` while the store got the file's value, so a shell override made them disagree — and Auth0 provisioning is not idempotent. The `.env` syntax check was stricter than the parser it guards, rejecting `export KEY=value`, `KEY = value` and `KEY: value`; and every apostrophe was refused, which blocked bootstrap over ordinary generated passwords.

## Verification

- `make test:apps:infra` → **383 passing**, up from 319 on main
- `npx tsc --noEmit` in `apps/infra` → clean
- Each fix was checked by reverting it and confirming the guarding test fails — including the IAM statements, where removing a `Deny` or widening a resource pattern fails the policy tests

Not yet verified against AWS: the narrowed policy has never run a real deploy. Four of its grant groups came from live listings, the rest are reasoned. **Preview with `apply=false` before applying** — a gap surfaces there as AccessDenied rather than mid-apply, and clearing one means redeploying the bootstrap stack.

## Cutover

Per stage: redeploy the bootstrap stack, run `npm run bootstrap -- --stage <stage>`, then preview before applying. `DEPLOY_ENV` and the per-stage `AWS_DEPLOY_ROLE_ARN` variable can be deleted from the Environment once a deploy has succeeded — bootstrap prints the exact `gh` commands.

## Summary by CodeRabbit

* **New Features**
  * Deployment configuration is now managed through stage-specific secret storage.
  * Added a command to list configured secret names safely.
  * Cloudflare credentials are synchronized across deployment environments.

* **Security**
  * Deployment permissions are more narrowly scoped by stage and resource.
  * Deployment roles are protected from unauthorized modification.

* **Documentation**
  * Updated setup, deployment, security, and secret-management guidance.
  * Clarified required AWS account, region, reviewer, and environment configuration.

* **Bug Fixes**
  * Deployments now detect missing, incomplete, or inconsistent stage configuration before making changes.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -247,12 +247,15 @@ compiles neither component.
 
 Required GitHub configuration:
 
-- Environment variables `AWS_ACCOUNT_ID` and `AWS_REGION`, per stage. The first is what
-  the deploy workflows compose
-  `arn:aws:iam::<id>:role/boxlite-<stage>-github-deploy`. It cannot come from the
-  stage's secret store like everything else, because `configure-aws-credentials`
-  reads it before any AWS credentials exist. The region is a literal in the
-  workflows, pinned to `DEFAULT_AWS_REGION` by a test.
+- Environment variables `AWS_ACCOUNT_ID` and `AWS_REGION`, per stage. Neither can come
+  from the stage's secret store like everything else, because
+  `configure-aws-credentials` reads both before any AWS credentials exist.
+  - `AWS_ACCOUNT_ID` is **required**. The workflows compose
+    `arn:aws:iam::<id>:role/boxlite-<stage>-github-deploy` from it; only the account id
+    is unknown, since the role name follows from the stage.
+  - `AWS_REGION` is **optional**, and only for a stage outside the default. The
+    workflows fall back to `DEFAULT_AWS_REGION` as a literal, pinned to the code by a
+    test; set the variable when the stage lives elsewhere.
 
 - Environment secrets `CLOUDFLARE_API_TOKEN` and `CLOUDFLARE_DEFAULT_ACCOUNT_ID`,
   per stage. These cannot come from the SST secret store: reading it initializes the
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -245,11 +245,25 @@ deployment Environments block before the job can obtain AWS credentials.
 matching Runner GitHub Release assets, verifies both before SST runs, and
 compiles neither component.
 
-Required Environment configuration (per stage):
+Required GitHub configuration:
 
-- Variable `AWS_DEPLOY_ROLE_ARN`
-- Variable `AWS_REGION` (defaults to `ap-southeast-1`)
-- Secret `DEPLOY_ENV` containing the stage's dotenv configuration
+- Environment variables `AWS_ACCOUNT_ID` and `AWS_REGION`, per stage. The first is what
+  the deploy workflows compose
+  `arn:aws:iam::<id>:role/boxlite-<stage>-github-deploy`. It cannot come from the
+  stage's secret store like everything else, because `configure-aws-credentials`
+  reads it before any AWS credentials exist. The region is a literal in the
+  workflows, pinned to `DEFAULT_AWS_REGION` by a test.
+
+- Environment secrets `CLOUDFLARE_API_TOKEN` and `CLOUDFLARE_DEFAULT_ACCOUNT_ID`,
+  per stage. These cannot come from the SST secret store: reading it initializes the
+  Cloudflare provider, so a token kept there would be needed in order to read itself.
+
+A stage's other configuration lives in its SST secret store, seeded by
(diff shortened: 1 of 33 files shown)


DorianZheng:
Superseded by #1258, which is the same work rebased onto main plus the fixes that came out of reviewing it. This branch had a conflict with main and its history has been rewritten.

PR #1274, opened 2026-08-18
fix(disk): handle whitespace in debugfs paths

### LINKED ISSUE #1272: Rootless ext4 image creation fails for OCI paths containing whitespace
## Description

Rootless ext4 image creation fails when an OCI image contains files or directories with whitespace in their names.

`normalize_inodes_with_debugfs()` builds a `debugfs` batch script by interpolating each image path directly into `sif` commands:

```text
sif <path> uid <uid>
sif <path> gid <gid>
```

Because the path is not quoted for the `debugfs` command parser, whitespace splits one legal OCI path into multiple arguments.

## Reproduction

1. Run BoxLite as a non-root user on Linux, so ext4 inode ownership normalization is required.
2. Create a box from an image containing a path with whitespace.

The current BoxLite base runtime image provides a concrete reproduction because the Debian setuptools package contains:

```text
/usr/lib/python3/dist-packages/setuptools/script (dev).tmpl
/usr/lib/python3/dist-packages/setuptools/command/launcher manifest.xml
```

## Actual behavior

The generated commands are equivalent to:

```text
sif /usr/lib/python3/dist-packages/setuptools/script (dev).tmpl uid 0
```

`debugfs` parses too many arguments and reports:

```text
sif: Usage: set_inode <inode> <field> <value>
```

There are four such errors for the two paths above because both UID and GID are normalized. Box creation then fails with a storage error while normalizing the ext4 image.

## Expected behavior

Legal OCI/POSIX paths containing whitespace should be treated as a single `debugfs` argument and should not prevent a box from being created.

## Suggested fix

- Quote every image path passed through the `debugfs` command parser.
- Reject or otherwise safely encode path characters that cannot be represented inside a quoted batch command, especially double quotes and line breaks, rather than emitting malformed or additional commands.
- Add a regression test using a path such as `/usr/lib/launcher manifest.xml`.

This primarily affects rootless execution because the ownership normalization pass is skipped when BoxLite runs as root.

mamamiyear:
## Summary

Treat OCI paths containing whitespace as one argument when rootless ext4 inode metadata is normalized through `debugfs`. Reject path characters that cannot be safely represented in a quoted batch command and cover the boundary with a regression test.

## Call graph

Before

```text
create_ext4_from_dir              (function · src/boxlite/src/disk/ext4.rs:358) — builds an ext4 image from the merged OCI rootfs
  └─ normalize_inodes_with_debugfs (function · src/boxlite/src/disk/ext4.rs:497) ← BUG: interpolates an unquoted path into each `sif` command
     └─ debugfs batch parser       (e2fsprogs · src/boxlite/src/disk/ext4.rs:537) — splits whitespace into extra arguments and rejects the command
```

After

```text
create_ext4_from_dir              (function · src/boxlite/src/disk/ext4.rs:358) — builds an ext4 image from the merged OCI rootfs
  └─ normalize_inodes_with_debugfs (function · src/boxlite/src/disk/ext4.rs:513) — normalizes ownership and restored modes
     └─ quote_debugfs_path         (function · src/boxlite/src/disk/ext4.rs:497) — quotes whitespace paths and rejects quotes or line breaks
        └─ debugfs batch parser    (e2fsprogs · src/boxlite/src/disk/ext4.rs:555) — receives the image path as one argument
```

Fixes #1272

## Changes

- Quote every path used by the UID, GID, and restored-mode `sif` commands.
- Reject double quotes and line breaks instead of emitting a malformed or additional `debugfs` command.
- Add regression coverage for a path containing whitespace and unsafe command delimiters.

## How to verify

- `cargo fmt --all --check`
- `cargo test -p boxlite quote_debugfs_path_preserves_whitespace_and_rejects_command_delimiters --lib` — 1 test passed.

## Risks / rollout

- OCI entries containing a literal double quote or line break now fail explicitly during image preparation; previously they produced malformed batch commands.
- Root execution skips this inode-normalization pass, so its behavior is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of filesystem paths containing whitespace during ownership and permission updates.
  * Added safeguards to reject unsafe paths containing quotes or line breaks.

### DIFF
--- a/src/boxlite/src/disk/ext4.rs
+++ b/src/boxlite/src/disk/ext4.rs
@@ -487,6 +487,22 @@ fn check_debugfs_output(what: &str, output: &std::process::Output) -> BoxliteRes
     Ok(())
 }
 
+/// Quote a path passed through debugfs's command parser.
+///
+/// Whitespace is significant to that parser, so even an otherwise ordinary
+/// OCI path such as `launcher manifest.xml` must be one quoted argument. The
+/// parser does not provide an escape for a literal double quote or a line
+/// break inside a quoted argument; reject those paths instead of emitting a
+/// malformed (or additional) command.
+fn quote_debugfs_path(path: &str) -> BoxliteResult<String> {
+    if path.contains(['"', '\n', '\r']) {
+        return Err(BoxliteError::Storage(format!(
+            "debugfs cannot safely address path {path:?}"
+        )));
+    }
+    Ok(format!("\"{path}\""))
+}
+
 /// Normalize inode metadata in the ext4 image via debugfs: give every file the
 /// ownership its layer declared (0:0 when none was recorded), and restore the
 /// original mode on any entry whose owner-read bit was temporarily widened so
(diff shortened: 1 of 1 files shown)


DorianZheng:
what happen if we don't quote it

mamamiyear:
> * OCI entries containing a literal double quote or line break now fail explicitly during image preparation; previously they produced malformed batch commands.

If there are any whitespace in image file, like
`/usr/lib/python3/dist-packages/setuptools/command/launcher manifest.xml`, 
the command will be separated to two part like
`sif /usr/lib/python3/dist-packages/setuptools/command/launcher uid 0`
and
`sif manifest.xml uid 0`
but the right command is
` sif "/usr/lib/python3/dist-packages/setuptools/command/launcher manifest.xml" uid 0`.

PR #1296, opened 2026-08-20
refactor(runtime): stop mutating capabilities from privileged mode

G4614:
`set_privileged`/`normalize_privileged` mutated `capabilities` in place on enable and tried to withdraw exactly that on disable, needing an ownership bit to tell an auto-install apart from a caller's own identical-looking policy. Checked how moby resolves this (`oci/caps.TweakCapabilities`): privileged short-circuits to every capability but never writes that back into `CapAdd`/`CapDrop` — the effective set is computed fresh every time a spec is built, so there's no mutation to track ownership of. `capabilities` is never mutated by `privileged` now, in either direction; `resolve_container_security` computes the effective set the same way moby does, so `privileged` alone remains a complete, one-flag DinD enabler.

Second commit closes two related gaps: `capabilities` is now `Option<ContainerCapabilities>` so an explicitly-empty override is distinguishable from the caller never touching the field (privileged mode rejects the former, same as any other explicit non-canonical override, and allows the latter). The field is private; `set_capabilities()` is the only way to write it, and it refuses to change the policy once `resolve_container_security` has already run on the object.

Test plan:
- [x] make clippy
- [x] make fmt:check:rust
- [x] make test:unit:rust (1028 + 49 passed)
- [x] cargo nextest run -p boxlite-c (75 passed)
- [x] cargo nextest run -p boxlite-node (22 passed)
- [x] cargo nextest run -p boxlite-cli --bins (185 passed)

## Summary by CodeRabbit

- **Security & Reliability**
  - Privileged mode now consistently applies the full capability set when capabilities are unspecified.
  - Non-privileged configurations default to no capabilities unless explicitly configured.
  - Explicitly conflicting capability overrides are rejected, while valid policies remain supported.
  - Capability settings can no longer be changed after security resolution.
  - Runtime option evaluation no longer mutates configuration.

- **Tests**
  - Expanded coverage for optional capability policies, privileged-mode behavior, conflict detection, and security resolution.
  - Updated SDK, CLI, REST, and runtime validation coverage.

### DIFF
--- a/src/boxlite/src/litebox/archive.rs
+++ b/src/boxlite/src/litebox/archive.rs
@@ -41,7 +41,11 @@ pub(crate) const MAX_SUPPORTED_VERSION: u32 = PUBLISHED_PORTS_ARCHIVE_VERSION;
 pub(crate) fn archive_version_for_options(options: &crate::runtime::options::BoxOptions) -> u32 {
     if !options.ports.is_empty() {
         PUBLISHED_PORTS_ARCHIVE_VERSION
-    } else if options.advanced.capabilities.is_empty() {
+    } else if options
+        .advanced
+        .capabilities()
+        .is_none_or(|capabilities| capabilities.is_empty())
+    {
         ARCHIVE_VERSION
     } else {
         CAPABILITY_POLICY_ARCHIVE_VERSION
@@ -284,14 +288,17 @@ mod tests {
         let ordinary = crate::runtime::options::BoxOptions::default();
         assert_eq!(archive_version_for_options(&ordinary), ARCHIVE_VERSION);
 
-        let custom = crate::runtime::options::BoxOptions {
-            advanced: crate::runtime::advanced_options::AdvancedBoxOptions {
-                capabilities: crate::runtime::advanced_options::ContainerCapabilities {
+        let mut custom_advanced = crate::runtime::advanced_options::AdvancedBoxOptions::default();
+        custom_advanced
+            .set_capabilities(Some(
+                crate::runtime::advanced_options::ContainerCapabilities {
                     drop: vec!["NET_RAW".into()],
                     ..Default::default()
                 },
-                ..Default::default()
-            },
+            ))
+            .unwrap();
+        let custom = crate::runtime::options::BoxOptions {
+            advanced: custom_advanced,
             ..Default::default()
         };
         assert_eq!(
--- a/src/boxlite/src/litebox/archive.rs
+++ b/src/boxlite/src/litebox/archive.rs
@@ -38,14 +38,16 @@ pub(crate) const PUBLISHED_PORTS_ARCHIVE_VERSION: u32 = 5;
 pub(crate) const MAX_SUPPORTED_VERSION: u32 = PUBLISHED_PORTS_ARCHIVE_VERSION;
 
 /// Pick the archive format an exported box needs.
+///
+/// Any explicit policy — including an explicitly empty one — is stamped v4:
+/// `capabilities()` returning `Some` means the caller configured something,
+/// which a pre-capability importer has no way to represent and must refuse
+/// rather than silently drop. Only `None` (the caller never touched the
+/// field) is indistinguishable from what a v3 importer already does.
 pub(crate) fn archive_version_for_options(options: &crate::runtime::options::BoxOptions) -> u32 {
     if !options.ports.is_empty() {
         PUBLISHED_PORTS_ARCHIVE_VERSION
-    } else if options
-        .advanced
-        .capabilities()
-        .is_none_or(|capabilities| capabilities.is_empty())
-    {
+    } else if options.advanced.capabilities().is_none() {
         ARCHIVE_VERSION
     } else {
         CAPABILITY_POLICY_ARCHIVE_VERSION
@@ -303,6 +305,25 @@ mod tests {
         };
         assert_eq!(
             archive_version_for_options(&custom),
+            CAPABILITY_POLICY_ARCHIVE_VERSION
+        );
+
+        // An explicit, empty capability policy is still an explicit policy,
+        // not the same as never touching the field — a pre-capability
+        // importer must not decide that distinction was safe to drop.
+        let mut explicit_empty_advanced =
+            crate::runtime::advanced_options::AdvancedBoxOptions::default();
+        explicit_empty_advanced
+            .set_capabilities(Some(
+                crate::runtime::advanced_options::ContainerCapabilities::default(),
+            ))
+            .unwrap();
+        let explicit_empty = crate::runtime::options::BoxOptions {
+            advanced: explicit_empty_advanced,
+            ..Default::default()
+        };
+        assert_eq!(
+            archive_version_for_options(&explicit_empty),
             CAPABILITY_POLICY_ARCHIVE_VERSION
         );
 
--- a/src/boxlite/src/rest/types.rs
+++ b/src/boxlite/src/rest/types.rs
@@ -196,11 +196,13 @@ impl CreateBoxRequest {
             volumes,
             detach: Some(options.detach),
             tty: options.tty.then_some(true),
-            advanced: (!options.advanced.capabilities.is_empty()).then(|| {
-                CreateBoxAdvancedOptions {
-                    capabilities: options.advanced.capabilities.clone(),
-                }
-            }),
+            advanced: options
+                .advanced
+                .capabilities()
+                .filter(|capabilities| !capabilities.is_empty())
+                .map(|capabilities| CreateBoxAdvancedOptions {
+                    capabilities: capabilities.clone(),
+                }),
             // The deprecated remove-on-stop flag was never applied by the cloud
             // control-plane mapper. Keep remote defaults unchanged and only send
             // the modern lifecycle fields when callers explicitly configure them.
@@ -729,14 +731,15 @@ mod tests {
 
     #[test]
     fn test_create_box_request_carries_container_capabilities() {
+        let mut advanced = crate::AdvancedBoxOptions::default();
+        advanced
+            .set_capabilities(Some(ContainerCapabilities {
+                add: vec!["SYS_ADMIN".into()],
+                drop: vec!["CAP_NET_RAW".into()],
+            }))
+            .unwrap();
         let opts = BoxOptions {
-            advanced: crate::AdvancedBoxOptions {
-                capabilities: ContainerCapabilities {
-                    add: vec!["SYS_ADMIN".into()],
-                    drop: vec!["CAP_NET_RAW".into()],
-                },
-                ..Default::default()
-            },
+            advanced,
             ..Default::default()
         };
 
@@ -832,12 +835,11 @@ mod tests {
         use crate::runtime::advanced_options::AdvancedBoxOptions;
         use crate::runtime::options::{BoxOptions, RootfsSpec};
 
+        let mut advanced = AdvancedBoxOptions::default();
+        advanced.security = SecurityOptions::enabled();
         let opts = BoxOptions {
             rootfs: RootfsSpec::Image("alpine:latest".into()),
-            advanced: AdvancedBoxOptions {
-                security: SecurityOptions::enabled(),
-                ..Default::default()
-            },
+            advanced,
             ..Default::default()
         };
         let req = CreateBoxRequest::from_options(&opts, None);
--- a/src/boxlite/src/rest/types.rs
+++ b/src/boxlite/src/rest/types.rs
@@ -196,13 +196,15 @@ impl CreateBoxRequest {
             volumes,
             detach: Some(options.detach),
             tty: options.tty.then_some(true),
-            advanced: options
-                .advanced
-                .capabilities()
-                .filter(|capabilities| !capabilities.is_empty())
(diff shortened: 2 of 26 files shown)


ltstriker (dismissed):

src/boxlite/src/rest/types.rs
```diff
@@ -196,11 +196,13 @@ impl CreateBoxRequest {
...
-                    capabilities: options.advanced.capabilities.clone(),
-                }
-            }),
+            advanced: options
+                .advanced
+                .capabilities()
+                .filter(|capabilities| !capabilities.is_empty())
+                .map(|capabilities| CreateBoxAdvancedOptions {
+                    capabilities: capabilities.clone(),
+                }),
```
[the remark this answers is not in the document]
G4614: Confirmed and fixed 3 of the 4 sites, each reproduced with a test against the old logic first: `rest/types.rs` (was filtering `Some(empty)` out of the wire request), `rest/runtime.rs` (was only probing server support for a non-empty policy), `archive.rs` (was stamping the pre-capability archive version for `Some(empty)`). Fixed in fb979744.

Left `sdks/python/src/options.rs#L567-568` as-is — the suggested fix isn't actually implementable at that line. The distinction is already gone one layer up: `PyAdvancedBoxOptions.capabilities` is `PyContainerCapabilities` (not `Option<PyContainerCapabilities>`), populated via `.unwrap_or_default()` in the `#[new]` constructor. `AdvancedBoxOptions()` and `AdvancedBoxOptions(capabilities=None)` are already the same call from Python's side (`None` is the parameter default), and even `AdvancedBoxOptions(capabilities=ContainerCapabilities())` collapses to the identical stored value. By the time execution reaches line 567, there's nothing left to distinguish.

Closing this properly means making that field `Option<PyContainerCapabilities>` — but it's already public API on the 0.9.7 PyPI release (same version as this crate's own last publish, which predates `privileged` by ~7 weeks — the same publish-timing check I used earlier in this PR, but this field is much older, from #1047, so it's actually shipped). Changing its type now is a real breaking change for existing Python callers who read `.capabilities` expecting a concrete object, not a case where I can point to "never published" the way I could for `privileged`/`set_privileged`. Since `privileged` isn't exposed via Python at all, there's no live path today where this loss of distinction is observable — flagging it here rather than silently leaving it unaddressed, but not fixing it in this PR.

src/boxlite/src/litebox/archive.rs
```diff
@@ -38,14 +38,16 @@ pub(crate) const PUBLISHED_PORTS_ARCHIVE_VERSION: u32 = 5;
...
+/// field) is indistinguishable from what a v3 importer already does.
 pub(crate) fn archive_version_for_options(options: &crate::runtime::options::BoxOptions) -> u32 {
     if !options.ports.is_empty() {
         PUBLISHED_PORTS_ARCHIVE_VERSION
-    } else if options
-        .advanced
-        .capabilities()
-        .is_none_or(|capabilities| capabilities.is_empty())
-    {
+    } else if options.advanced.capabilities().is_none() {
```
ltstriker, line 50: `serve`'s `build_box_options` — `CreateBoxRequest.advanced` is a non-Option `#[serde(default)]` field, so a client sending {"image":"alpine:latest"} still produces `add:[], drop:[]`, and `build_box_options` wraps that in `Some(...)` unconditionally. With `archive_version_for_options` now keying off `is_none()`, every box created through boxlite `serve` — including ones that never configured a capability — is stamped v4 instead of v3, and an older importer will refuse them.

ths same like:   sdks/python/src/options.rs

check it right or not

ltstriker (changes requested):

src/boxlite/src/litebox/archive.rs
G4614: Right, confirmed with a test first — `{"image":"alpine:latest"}` was resolving to `capabilities() == Some(empty)`, so it got stamped v4 by the archive check. `sdks/python/src/options.rs` is a different case though: that one's genuinely stuck (the distinction is already lost one layer up in `PyAdvancedBoxOptions.capabilities`, which is a published, non-Option field on the 0.9.7 PyPI release — see my earlier reply on this PR). `serve`'s wire schema is this CLI's own, nothing depends on the old shape, so it was fixable: `CreateBoxAdvancedOptions.capabilities` is now `Option<ContainerCapabilitiesRequest>`, and `build_box_options` maps it straight through. Fixed in 8da2e142.

G4614:
Addressed CodeRabbit's outside-diff finding on `rt_impl.rs:551-565` (`check_options_compatibility`): a non-privileged request with explicit `capabilities.add = ["ALL"]` could reuse a stored privileged box, because both sides resolve to the same effective capabilities even though the privileged box also has cleared readonly paths and a writable `/sys` — more than the request asked for.

Fixed in 0300db22 by making the `privileged` comparison symmetric (reject on mismatch in either direction), with a reproducer test (`check_options_compatibility_rejects_non_privileged_request_reusing_a_privileged_box`) that failed before the fix and passes after.

src/boxlite/src/runtime/advanced_options.rs
```diff
@@ -669,11 +669,20 @@ fn validate_capability_names(
...
+    /// from `Some` of an empty policy: combined with `privileged`, leaving
+    /// this unspecified is the one-flag DinD case (see `privileged`'s own
+    /// doc comment), while an explicit override — even an empty one — is a
+    /// conflict `validate_privileged_capability_conflict` rejects. Private;
+    /// read via `capabilities()`, write via `set_capabilities`, which also
+    /// enforces that this can't change out from under an already-resolved
+    /// request (see `resolved`).
     #[serde(default)]
-    pub capabilities: ContainerCapabilities,
+    capabilities: Option<ContainerCapabilities>,
```
ltstriker, line 695: `AdvancedBoxOptions.capabilities` became `Option<...>` with only `#[serde(default)]` and no `skip_serializing_if`, so None serializes as "capabilities": null. The pre-Option field is a plain `ContainerCapabilities`, and field-level `#[serde(default)]` only covers a missing key — not an explicit null — so every ordinary box's exported manifest fails on older builds with invalid type: `null, expected struct ContainerCapabilities`.   

maybe a compatibility problem

ltstriker (changes requested):

src/boxlite/src/runtime/advanced_options.rs
G4614: Confirmed — reproduced standalone first: a plain (pre-Option) `ContainerCapabilities` field with only `#[serde(default)]` parses an absent key fine but rejects an explicit `"capabilities": null` with exactly `invalid type: null, expected struct ContainerCapabilities`. And `AdvancedBoxOptions::default()` does serialize to `"capabilities":null` on this branch, since `skip_serializing_if` was missing.

Fixed by adding `skip_serializing_if = "Option::is_none"` alongside the existing `default`, so `None` omits the key again — byte-for-byte what a pre-#1296 build already produces/reads, which is exactly what `archive_version_for_options` assumes when it leaves an unspecified-capabilities box at `ARCHIVE_VERSION` instead of bumping it.

Added a regression test that serializes `BoxOptions::default()` and asserts the `capabilities` key is absent (not `null`) — confirmed it fails on the pre-fix code with the exact null shape, then passes with the fix.

G4614:
Ran a broader self-audit for other backward/forward-compatibility gaps beyond what reviewers had already flagged (crates.io API surface, wire/persisted-JSON shapes, archive/DB versioning, SDK bindings, REST/CLI wire). Two more real, reproduced issues, both fixed:

1. **Legacy on-disk box configs get over-versioned on export** (e9423d3b). Before this PR, `capabilities` was a plain, always-serialized field, so every ordinary box's persisted `box_config.json` already contained `"capabilities":{"add":[],"drop":[]}`. After the field became `Option<ContainerCapabilities>`, that exact legacy shape deserializes as `Some(empty)`, not `None` — indistinguishable on the wire from a deliberate new explicit-empty override. `archive_version_for_options` would then stamp every already-existing, never-customized box as `CAPABILITY_POLICY_ARCHIVE_VERSION` instead of `ARCHIVE_VERSION` the first time it's exported after a host upgrades to this build. Added migration `v9_to_v10`: strips the `capabilities` key from a persisted row only when it's the exact empty shape, once, while that's still knowable — a privileged box's persisted `add:["ALL"]`, or any real override, is left untouched.

2. **Python SDK silently turns "unrelated advanced option" into "explicit empty capability override"** (ee0da401). `PyAdvancedBoxOptions.capabilities` was a plain (non-Option) field, and the `TryFrom<PyBoxOptions>` conversion called `set_capabilities` unconditionally whenever `advanced` was `Some(..)` — so a Python caller setting only `security=` or `health_check=` got `capabilities() == Some(default)` instead of `None`, tripping the same archive-version bump plus the REST `require_linux_capabilities_enabled` gate for a caller who never touched capabilities. Made it `Option<PyContainerCapabilities>`, mirroring the Node SDK (which already threads this correctly), and only call `set_capabilities` when the caller actually set it. Confirmed via `cargo check -p boxlite-python --tests` (clean) and manual construction of the real `PyAdvancedBoxOptions`/`PyBoxOptions` types feeding the real `TryFrom` impl — this sandbox can't link libpython for `cargo test -p boxlite-python` (pre-existing, same limitation noted in #1206's test plan), so I couldn't get an actual pytest/cargo-test run; flagging that as residual risk for CI, which does have a working Python toolchain.

ltstriker (approved):

G4614:
Ran a correctness-focused review pass (separate from the earlier compatibility audit) over this PR's own additions. Two coverage gaps confirmed and closed, both verified by temporarily breaking the guarded logic and watching the new test catch it (commit 83c5fef9):

- No test called `resolve_container_security()` and then `set_capabilities()` on the same instance — a regression to the resolved-freeze guard would have shipped silently. Added `set_capabilities_rejects_mutation_after_resolve`.
- Only the *rejection* direction of `check_options_compatibility`'s privileged logic was tested. Added the positive case — a privileged request actually reusing an existing privileged box — including the harder variant where the stored box has the legacy `capabilities.add=["ALL"]` shape instead of an unset field.

Also documented (no behavior change) why `set_privileged` is deliberately not gated by the same freeze as `set_capabilities` — the freeze protects the capability policy specifically, not `privileged` itself, and no current caller reuses one instance across two resolves. Flagging this explicitly so it reads as a decision, not an oversight, if anyone looks at it again.

Separately: the review also surfaced a real, **pre-existing** bug unrelated to this PR's diff — `RestRuntime` silently drops `privileged` entirely (never sent on the wire, no local rejection either, unlike `nested_virtualization`/`kernel`), and `RestRuntime::get_or_create` performs no compatibility check on reuse at all. Confirmed via `git show --stat` that the affected code is untouched by this PR. Filed as #1348 rather than folding into this PR's scope.

G4614:
Tried to actually boot a VM in privileged mode to check this end-to-end (real capabilities, real `/sys` mount options), not just the host-side resolution logic. Added the test (`5f35b1e3`), but a real run surfaces that it can't currently pass anywhere:

```
Unsupported("guest 0.9.7 is older than the required 0.9.8; recreate the box with the current runtime")
```

Root cause, confirmed by reading the actual code: the guest reports its version via `env!("CARGO_PKG_VERSION")`, which resolves to the workspace `Cargo.toml`'s version — currently `0.9.7`. `guest_init.rs`'s `MIN_PRIVILEGED_CONTAINER_GUEST_VERSION`/`MIN_CAPABILITY_GUEST_VERSION`/`MIN_DEVICE_GUEST_VERSION` all require `0.9.8`, and per `git log -p` have required that since #646 introduced them three months ago — the version has never been bumped. So **no guest built from current `main` can satisfy any of these three gates**: privileged boxes, explicit capability overrides, and device passthrough/nested virtualization all fail the guest handshake, every time, right now.

This isn't caused by, or fixable from, this PR — `guest_init.rs` is untouched by its diff. Filed as #1349. Left the new test in as `#[ignore]`d (linked to that issue) so it's real coverage the moment the version floor gets resolved, rather than blocking this PR's CI on an unrelated pre-existing gap.

G4614:
Followed up on the #1349 block by bypassing it locally (temporarily lowering the three version constants to 0.9.7, not committed) to check whether the actual feature works once past the gate. It does — with one wrinkle along the way.

First run past the gate hit a different error:
```
failed to decode Protobuf message: ProcessOptions.capabilities: ContainerAdvancedOptions.process: ...
```
That `ContainerAdvancedOptions.process` path is the *old*, pre-#1289 wire shape (`fix(proto): stop wrapping capabilities in a process submessage`) — the cached guest binary in this sandbox predated that fix. Rebuilt the guest fresh from this exact HEAD (`make guest PROFILE=debug`, then had to `touch build.rs` to get cargo to actually pick up the new binary — its build script caches "not found" and doesn't re-check).

With a guest actually built from current HEAD and the version gate bypassed, the real test passes:
```
test privileged_box_gets_more_capabilities_and_writable_sys ... ok
```
i.e. a privileged box's guest process gets a real, strictly larger effective capability set than a plain box's, and `/sys` is actually mounted writable for privileged / read-only for plain — confirmed via a real VM boot, not simulated. Reverted the local-only version bypass immediately after (nothing committed from this check beyond what's already in `5f35b1e3`).

So: the design and implementation in this PR are correct end-to-end; #1349 is the only thing standing between this and a real, unblocked boot.

PR #1308, opened 2026-08-21
fix(lint): satisfy clippy 1.98 across the workspace

BatmanByte:
## Summary

Stable Rust 1.98 landed on 2026-08-18 with a new `chunks_exact_to_as_chunks` lint and a stricter `result_large_err`. `rust-toolchain.toml` pins the floating `stable` channel, so CI took both immediately, and `cargo clippy` now fails on seven pre-existing sites. This clears all of them.

## Call graph

Two of the seven are real code changes; the rest are reasoned `allow`s, so only these hops move.

Before

  read_l1_table         (Qcow2 · src/boxlite/src/disk/qcow2.rs:1261)
    └─ chunks_exact(8)  — yields `&[u8]`, so every entry needs `c.try_into().unwrap()` and carries its panic path
  users_groups_by_id    (SftpSession · src/guest/src/service/ssh/sftp.rs:792)
    └─ chunks_exact(4)  — same shape, `id.try_into().expect("four-byte chunk")`

After

  read_l1_table         (Qcow2 · src/boxlite/src/disk/qcow2.rs:1261)
    └─ as_chunks::<8>() — yields `&[u8; 8]`, so `from_be_bytes` takes it directly and the conversion is gone
  users_groups_by_id    (SftpSession · src/guest/src/service/ssh/sftp.rs:792)
    └─ as_chunks::<4>() — same, no fallible conversion left to panic

## Changes

- `chunks_exact` → `as_chunks` at the two sites the new lint names. The fixed-size array removes the `try_into` and its unreachable panic, so this is the lint doing its job rather than a workaround.
- The five `result_large_err` sites all return a framework's own error type — three `OutputManager` methods answering gRPC calls with `tonic::Status`, two axum helpers returning `Response` because callers `?` them straight into one. The size is the boundary's shape, not a local choice, and boxing would be unboxed again one frame up, so each carries an `allow` with its reason. That matches how the repo already handles inapplicable lints (for example `util/process.rs:407`); there is no `[workspace.lints]` table to put them in.
- The `allow` on `impl OutputManager` is deliberate rather than three per-method copies: every fallible method in that impl returns `Status`, so the statement is exact for the whole block.

## How to verify

```bash
make clippy
```

Needs 1.98 or newer — on 1.96 the new lint does not exist, so an older toolchain reports clean either way.

Two behavioral tests cover the changed sites: `cargo test -p boxlite --lib -- qcow2` (77 tests) and the guest's `users_groups_by_id_is_advertised_and_preserves_request_order`, which parses exactly the four-byte array the `as_chunks` change rewrote.

## Risks / rollout

Worth knowing separately from this fix: the Rust clippy job was `skipped` on every recent main commit, so main has been passing `Lint (conclusion)` without running clippy at all. These seven sites had accumulated unseen. A floating `stable` channel plus a `changes` filter that can skip the lint means the next toolchain release can turn CI red on untouched code again.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal parsing efficiency for disk metadata and secure file-transfer identifiers.
  * Preserved existing validation, lookup limits, and big-endian identifier decoding.
  * Improved handling of output attachment failures, including clearer distinctions between already-attached and finalizing sessions.

* **Chores**
  * Updated internal code quality annotations and documentation to clarify expected error handling.
  * No new user-facing features or changes to command behavior.

### DIFF
--- a/src/boxlite/src/disk/qcow2.rs
+++ b/src/boxlite/src/disk/qcow2.rs
@@ -1258,8 +1258,10 @@ impl FlattenLayer {
             ))
         })?;
         let l1_table: Vec<u64> = l1_buf
-            .chunks_exact(8)
-            .map(|c| u64::from_be_bytes(c.try_into().unwrap()))
+            .as_chunks::<8>()
+            .0
+            .iter()
+            .map(|entry| u64::from_be_bytes(*entry))
             .collect();
 
         Ok((
--- a/src/cli/src/commands/serve/mod.rs
+++ b/src/cli/src/commands/serve/mod.rs
@@ -940,6 +940,9 @@ fn constant_time_eq(a: &[u8], b: &[u8]) -> bool {
 // Box Handle Cache Helper
 // ============================================================================
 
+// The error is an axum `Response` because callers `?` this straight into one, so
+// its size is the framework's shape rather than a choice made here.
+#[allow(clippy::result_large_err)]
 async fn get_or_fetch_box(state: &AppState, box_id: &str) -> Result<Arc<LiteBox>, Response> {
     // Check cache first.
     //
@@ -1071,6 +1074,7 @@ where
 /// so attaching to a finished job would silently run the job again. `attach()`
 /// already does the right thing for every status by itself: it boots a
 /// `Configured` box (which has never run) and refuses a `Stopped` one.
+#[allow(clippy::result_large_err)] // an axum `Response` error, same as above
 async fn get_or_attach_main_session(
     state: &AppState,
     box_id: &str,
--- a/src/guest/src/service/exec/output.rs
+++ b/src/guest/src/service/exec/output.rs
@@ -47,6 +47,19 @@ struct OutputState {
     stderr: StreamState,
 }
 
+/// Why an attach was refused.
+///
+/// Deliberately small: the caller maps this to its own error and never forwards
+/// a `tonic::Status` from here, so building one would size a 128-byte error only
+/// to discard it a frame later.
+#[derive(Debug, PartialEq, Eq)]
+pub(crate) enum AttachRefused {
+    /// Another consumer already holds the lease.
+    AlreadyAttached,
+    /// Terminal output is being sealed, so no live attach can start.
+    Finalizing,
+}
+
 struct ConsumerLease {
     claimed: Arc<AtomicBool>,
     updated: watch::Sender<()>,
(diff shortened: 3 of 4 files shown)


BatmanByte:
Right, and the comment was worse than the lint: `OutputManager` is not the boundary. Its only caller drops the value entirely — `state.rs`'s `map_err(|_| ExecutionError::AlreadyAttached)` — so a 128-byte `Status` was being built to be thrown away one frame later, which is exactly what `result_large_err` was pointing at. An `allow` justified by a false claim is worse than no `allow`.

Fixed in 9ab05672 by taking your first suggestion rather than the scoping one: `attach`, `attach_retained`, and `claim_consumer` now return a two-variant `AttachRefused` (`AlreadyAttached`, `Finalizing`). The lint is gone because the cause is gone, no allowance left in that impl, and the signature no longer implies a boundary that isn't there.

Behavior is unchanged — the caller still flattens both variants to `AlreadyAttached`, exactly as the `|_|` did. The two tests now assert the variant instead of a `tonic::Code` that never reached a wire.

One thing I noticed and deliberately did not change here: that flattening means a refusal caused by sealing is reported to the client as "already attached". Now that the two conditions have names, distinguishing them is a small change — but it alters what clients see, so it does not belong in a lint fix.

BatmanByte:
Superseded by #1309, which landed the same fix for the same seven sites — `as_chunks` at both `chunks_exact` call sites, and `allow(clippy::result_large_err)` on the five framework-error returns. Closing rather than rebasing, since main is already clean under 1.98.

One difference is worth carrying forward separately. For the three `OutputManager` methods, #1309's allow reads "these control-plane methods fail rarely, so boxing isn't worth it" — but the reason those errors are large is not that boxing is unattractive. Their only caller drops the value: `state.rs` does `map_err(|_| ExecutionError::AlreadyAttached)`, so a `tonic::Status` is constructed at 128-odd bytes purely to be thrown away one frame later. `OutputManager` never touches the gRPC boundary at all. CodeRabbit caught the same thing on this PR, where my own allow had claimed the boundary required the type.

So the lint was pointing at something real there, and the fix is to stop building the error: a two-variant `AttachRefused` (`AlreadyAttached`, `Finalizing`) removes the cause instead of silencing it, needs no allowance, and stops the signature implying a boundary that isn't there. Behavior is unchanged — the caller still flattens both variants. I'll offer that as its own small PR against the current main rather than reviving this one.

That change also surfaces a question for whoever owns the attach path: because the caller flattens, a refusal caused by sealing is currently reported to clients as "already attached". Naming the two conditions makes distinguishing them cheap, but it changes what clients see, so it wants its own decision.

PR #1318, opened 2026-08-23
fix(runtime): verify embedded cache contents

### LINKED ISSUE #786: runtime: embedded-runtime fast path trusts .complete without verifying boxlite-guest is present
## Summary

The embedded-runtime extractor trusts a cache directory on the presence of its `.complete` stamp alone, without verifying the expected binaries are actually in it. A version-keyed cache dir that was populated by a build whose manifest lacked `boxlite-guest` (e.g. a partial/variant build, or an SDK built in a different worktree) is then reused by a later process that *does* need the guest — and the missing-binary failure surfaces much later, at box start, instead of at extraction.

## Where (as of `07fa30f9`)

Fast path returns on stamp existence with no content check:

```rust
// src/boxlite/src/runtime/embedded.rs:104
let stamp = dir.join(".complete");
if stamp.exists() {
    let now = filetime::FileTime::now();
    let _ = filetime::set_file_mtime(&stamp, now);
    return Ok(Self { dir });   // <- trusts the dir without verifying boxlite-guest is present
}
```

Release builds key the cache dir by version only (`v{VERSION}/`, see the module header at `src/boxlite/src/runtime/embedded.rs:9-13`), so two builds sharing a version but differing in embedded contents collide on the same path. The first writer's `.complete` then satisfies every later reader.

The stamp is written *after* all manifest files within a single extraction (`embedded.rs:118-132`), so a same-build extraction is internally consistent — the gap is specifically cross-build reuse of a version-keyed dir.

## Impact

Low for published releases (all users on a version have identical binaries — the stated assumption). It bites dev / multi-worktree setups where SDKs built from different trees share `~/.local/share/boxlite/runtimes/v{VERSION}/`: a guest-less `.complete` dir is trusted and `box.start()` later fails to find `boxlite-guest`.

## Expected

Have the fast path verify the dir actually contains the expected runtime binaries (at minimum `boxlite-guest`) before trusting `.complete`; on mismatch, re-extract (or fail loudly at `get()` with a clear message) rather than returning a known-incomplete dir.

## Current workaround (downstream)

`apps/infra-local` pre-scans the runtime cache and skips any `.complete` dir missing `boxlite-guest`, pinning a known-good one via `BOXLITE_RUNTIME_DIR`:
- `apps/infra-local/boxlite_local/config.py` — `pick_runtime_dir` / `resolve_runtime_dir`.

This is exactly the verification we think belongs in the extractor's fast path.

mikemikimike:
## Summary

Validate the embedded runtime cache before trusting its `.complete` marker. If a published cache has lost one or more manifest files, quarantine it and atomically publish a complete replacement.

The original version-only cache collision described in #786 is no longer the current failure mechanism: the cache directory now includes the version, commit, and embedded manifest hash. This PR addresses the remaining post-publication cache damage case.

### Before

```text
EmbeddedRuntime::init (Option<EmbeddedRuntime> - src/boxlite/src/runtime/embedded.rs:81)
  `- extract (BoxliteResult<EmbeddedRuntime> - src/boxlite/src/runtime/embedded.rs:99)
       `- `.complete` fast path trusts the marker without validating cache contents
```

### After

```text
EmbeddedRuntime::init (Option<EmbeddedRuntime> - src/boxlite/src/runtime/embedded.rs:81)
  `- versioned_dir (BoxliteResult<PathBuf> - src/boxlite/src/runtime/embedded.rs:257)
       `- extract (BoxliteResult<EmbeddedRuntime> - src/boxlite/src/runtime/embedded.rs:99)
            |- has_all_manifest_files validates every manifest entry
            |- quarantine_incomplete_cache removes a damaged cache atomically
            `- re-extract and atomically publish a complete replacement
```

## Changes

- Validate every embedded manifest entry before accepting a stamped cache.
- Atomically move an incomplete cache to a process-unique quarantine path before re-extracting it.
- Re-check the cache after a rename race so a complete replacement published by another process is reused.
- Reject symlink entries when validating manifest files.
- Exercise the recovery behavior through `extract()` and cover incomplete-cache recovery, quarantine behavior, manifest validation, and complete extraction.
- Update the documentation and PR context to describe post-publication cache damage rather than the already-resolved version-only collision.

## How to verify

- `cargo fmt --all -- --check` passed.
- `git diff HEAD^ HEAD --check` passed.
- Focused command: `cargo test -p boxlite --lib runtime::embedded::tests --no-default-features --features embedded-runtime`.
- A Linux runner with `BOXLITE_DEPS_STUB=2`, `RUSTUP_TOOLCHAIN=1.93.1`, and `protobuf-compiler` installed reached the `boxlite` build, but stopped before running tests because the prebuilt runtime download failed with a GitHub TLS EOF.
- A separate `BOXLITE_DEPS_STUB=1` run passed 9 tests, but that stub generates an empty manifest and therefore does not provide meaningful coverage for this cache regression.

## Risks / rollout

The fast path now performs a metadata check for each embedded manifest entry. Damaged stamped caches are renamed out of the shared path and re-extracted; the existing cache format and atomic publication behavior are otherwise unchanged. The Linux focused test remains to be verified in an environment that can download the prebuilt runtime artifact.

Fixes #786


mikemikimike:
Addressed the latest CodeRabbit findings in commit `2dd40103`: the regression now drives `EmbeddedRuntime::extract_into` through the incomplete-cache fast path; manifest validation uses `symlink_metadata` and rejects symlink entries; and quarantine/losing-temp cleanup errors are propagated. `cargo fmt --all` and `git diff --check` pass. The focused Windows test remains blocked before crate compilation by the pre-existing `libkrun-sys` platform-gated build-script errors (`LIBKRUNFW_SHA256`, `LIB_DIR`, and `build`).

mikemikimike:
Addressed the remaining symlink-test review in commit `45e4850c`: every manifest entry after `MANIFEST[0]` is now materialized as a regular file, so the assertion isolates symlink rejection. `cargo fmt --all -- --check` and `git diff --check` pass.

PR #1321, opened 2026-08-24
docs: adopt the shared agent-tooling guidance block in AGENTS.md

DorianZheng:
## What

Splits this repository's root instructions into the domain half it owns and the shared engineering workflow that `boxlite-ai/agent-tooling` now owns and distributes.

- **`AGENTS.md` becomes the real file** (it was a symlink to `CLAUDE.md`): Project Overview, Tech Stack, Project Structure, Common Commands, Code Style, a Commit & PR Messages pointer, and a new **Local Exemplars** section — where the `ImageManager` facade and the `jailer/common/` stateless-utility exception now live, since the shared text must stay domain-neutral.
- **`CLAUDE.md` becomes the `@AGENTS.md` bridge** Claude Code inlines, with room for Claude-specific lines below the import.
- **The guidance block** sits between `<!-- agent-tooling:guidance:begin rev=… sha256=… -->` markers, spliced by `sync-guidance.sh` on explicit install.

Why this direction: Codex reads only `AGENTS.md` by default, Claude Code only `CLAUDE.md` (inlining the import), Copilot either — so the real content belongs in `AGENTS.md` with `CLAUDE.md` bridging. Claude Code strips block-level HTML comments before injection, so the markers cost no context.

## No policy is lost

The four rules this file gained after the shared text was extracted — verdict auditing before turn-ending assertions, treat-every-failure-as-a-class, and both commit/PR-text rules — were absorbed into the canonical guidance first (boxlite-ai/agent-tooling#7). They now arrive through the block instead of living here. The block body is byte-identical to `git show d995e40:plugins/boxlite-agent-tooling/guidance/workflow.md`, which the marker's `rev=d995e4051ace` names; the only generalization is that repo-relative CONTRIBUTING links became "the repository's CONTRIBUTING" (this repo's own pointer is preserved in its new Commit & PR Messages section).

## Drift protection

Editing inside the markers, or dropping the block, fails the pinned commit and push gates. The `agent-tooling` CI job additionally runs the installer and then `git diff --exit-code -- AGENTS.md CLAUDE.md`, so a stale block becomes a diff; `AGENTS.md`/`CLAUDE.md` are now trigger paths for that job. Updating the shared text is a PR in `agent-tooling`; consumers pick it up on their next explicit install.

## Verification

`.agent-tooling/install.sh` adopts `d995e40` here; the adopted revision's `sync-guidance.sh --check` exits 0 silently (block CURRENT, so the new CI step passes). Change went through this repo's commit and push audit gates (both PASS, no blocking findings).

Before/after: **before** — one `CLAUDE.md` carrying repo knowledge and a hand-copied workflow that silently drifted from every sibling repo. **after** — `AGENTS.md` carries repo knowledge plus a hash-verified shared block, `CLAUDE.md` bridges for Claude Code, and drift is a gate failure instead of an invisible divergence.

https://claude.ai/code/session_01Fmb7Y26648j189bHPTqjZH

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive project guidance covering development workflows, coding conventions, testing, verification, and communication practices.
  * Updated Claude-specific guidance to reference shared project instructions, keeping project documentation consistent.

* **Chores**
  * Updated validation automation to run when agent guidance changes.
  * Added checks ensuring tooling installation does not modify project guidance files.
  * Improved safeguards for maintaining consistent development instructions.


DorianZheng:
Thanks — the concern is fair, but the suggested fix (pinning `tooling.ref` to `d995e4051ace`) cannot be applied: it would fail the very CI job it is meant to protect.

`tooling.ref` names a **branch** by contract. Both the bootstrap and the profile validator reject a SHA-shaped ref explicitly:

```
agent-tooling: tooling.ref names a branch; to freeze a revision write it to .agent-tooling/hold
```

so `./.agent-tooling/install.sh` — the only step in the `agent-tooling` job — would exit 1. This repository moved from a pinned `tooling.sha` to the float deliberately in #1319; the mechanism for freezing a revision is `.agent-tooling/hold` (one full lowercase commit SHA), which freezes the fleet when committed or one machine when kept local.

On the underlying risk, "upstream changes silently alter setup behavior" is the case the design already closes:

- The adopted revision is recorded per clone in `.git/agent-tooling/current`, written only after validation passes, and every adoption is appended to `.git/agent-tooling/history.log` — so what ran here is always answerable.
- The guidance block carries a content hash (`sha256=…` in its begin marker). If upstream text changed, the CI step added in this PR (`git diff --exit-code -- AGENTS.md CLAUDE.md`) turns that into a **failing build**, not a silent edit. Detection is the point of the block, and a hand-edit inside the markers fails the local commit and push gates as well.
- Floating is the intended property: a policy fix in `boxlite-ai/agent-tooling` should reach every consumer without a pin-bump PR per repository.

Leaving `tooling.ref: main` as-is. If a specific revision ever needs freezing here, that is a one-line `.agent-tooling/hold` file rather than a profile change.

PR #1328, opened 2026-08-24
fix(guest): enable protected link sysctls

### LINKED ISSUE #1154: Guest kernel does not set fs.protected_hardlinks or fs.protected_symlinks
Running at rev `8803834036205cf2cac5cfca98bb3875812c897a`.

Inside a guest, both `/proc/sys/fs/protected_hardlinks` and `/proc/sys/fs/protected_symlinks` read `0`, and `/`, `/workspace` and the agent state directories all sit on one device.

That means an unprivileged guest process can hardlink a root-owned file it cannot write into a directory it controls. Any later root-run recursive `chown` over that directory then hands it ownership of the original inode. We hit this concretely: a uid-1000 process could link a read-only, root-owned instruction file into a writable path.

Debian sets both to 1 via `/etc/sysctl.d`, which a microVM init does not apply. Setting them in the guest kernel would close the class for anything running as root in a cell, rather than each caller having to scope its own recursive operations.

aiqubits:
## Summary

Enable Linux protected-hardlink and protected-symlink sysctls during guest agent boot, before the zygote or any workload process can start. This restores security defaults normally supplied by distro `sysctl.d` configuration in BoxLite's minimal guest init.

## Call graph

Before
  `main` (`Guest agent` · `src/guest/src/main.rs:75`) — initializes logging, then starts the zygote ← BUG: skips protected-link kernel hardening
    └─ `Zygote::start` (`Zygote` · `src/guest/src/container/zygote.rs:120`) — can spawn workloads while both sysctls remain disabled

After
  `main` (`Guest agent` · `src/guest/src/main.rs:75`) — hardens the kernel before starting the zygote
    ├─ `apply_boot_hardening` (`sysctl` · `src/guest/src/sysctl.rs:30`) — applies the protected-link policy
    │  └─ `write_setting` (`sysctl` · `src/guest/src/sysctl.rs:41`) — writes both `/proc/sys/fs/protected_*` values or fails closed
    └─ `Zygote::start` (`Zygote` · `src/guest/src/container/zygote.rs:120`) — starts only after hardening succeeds

Fixes #1154

## Changes

- Set `fs.protected_hardlinks=1` and `fs.protected_symlinks=1` at guest startup.
- Return a contextual startup error if either security setting cannot be applied.
- Cover successful application and the fail-closed path with guest unit tests, including Cargo's serial fallback suite.

## How to verify

- `make test:unit:guest`
- `make fmt:check:rust`
- `make lint:rust`

## Risks / rollout

- Guest startup now deliberately fails if the kernel does not expose or permit either protected-link sysctl, instead of running with the unsafe defaults.

## Summary by CodeRabbit

* **Security Enhancements**
  * Guest startup now enables kernel protections against unsafe hardlinks and symlinks before launching workloads.
  * Startup reports when hardening is complete and stops with a clear error if required protections cannot be applied.
  * Unprivileged workloads are protected from creating unauthorized hardlinks to protected files.

* **Testing**
  * Added coverage for successful hardening and fail-closed behavior when required system settings are unavailable.
  * Added validation that protected-link enforcement preserves file integrity and expected permissions.
  * Updated test documentation to reflect the expanded guest test coverage.

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -232,7 +232,7 @@ test\:unit\:rust:
 # nextest's process-per-test isolation is what makes the whole crate runnable:
 # several suites assert on process-global state (raw fd numbers, waitpid), so
 # they can interfere under a shared process. The cargo fallback has no such
-# isolation, so it stays on the two modules that are pure logic.
+# isolation, so it stays on the modules that are pure logic.
 test\:unit\:guest:
 	@if [ "$$(uname)" != "Linux" ]; then \
 		echo "⏭️  Guest unit tests require Linux"; \
@@ -242,7 +242,7 @@ test\:unit\:guest:
 	if command -v cargo-nextest >/dev/null 2>&1; then \
 		cargo nextest run --no-tests=fail -p boxlite-guest; \
 	else \
-		cargo test -p boxlite-guest --bins -- --test-threads=1 capabilit spec::tests; \
+		cargo test -p boxlite-guest --bins -- --test-threads=1 capabilit spec::tests sysctl::tests; \
 	fi
 
 # Keep ordinary ownership tests unprivileged. Only explicitly ignored tests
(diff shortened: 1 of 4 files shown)


DorianZheng:
Hi @aiqubits. Thanks for filing the PR. @ltstriker Could you please take a look on this PR?

ltstriker (approved):
vm test is also pass.  good job.

PR #1352, opened 2026-08-26
fix(api): reject tunnel CONNECT for a stopped box instead of silently dropping the stream

G4614:
## Summary
A stopped, non-auto-resumable box's tunnel CONNECT silently succeeded and then the stream went dead — no way for a client to tell a stopped box from a network blip. Root cause: `proxyNetworkTunnel` resolved a tunnel URI unconditionally, unlike every other proxy route which gates on box state via `autoResume`/`ensureStartedForProxy`.

Fix: the endpoint now auto-resumes when the box allows it (matching `proxyToRunner`'s policy), and otherwise returns `409` for a stopped/stopping box. A missing/destroyed box already 404s via `findOneByIdOrName`.

Promotes the strict `xfail` e2e coverage (Python + Node) for `rejects_stopped_box` to real assertions.

Fixes POL-214

## Test plan
- `npx jest src/boxlite-rest/boxlite-proxy.controller.spec.ts` — 12/12 pass, including 2 new cases (409 on stopped/non-resumable, auto-resume path).
- `apps/e2e/cases/test_sdk_tunnel.py::test_python_sdk_tunnel_rejects_stopped_box` and `test_node_tunnel.py::test_node_sdk_tunnel_rejects_stopped_box` no longer marked xfail (not run here — need live dev env).

## Summary by CodeRabbit

- **Bug Fixes**
  - Network tunnel requests now require the box to be started.
  - Requests for stopped or non-started boxes return a conflict instead of attempting auto-resume or generating a tunnel URL.
  - SDK and node tunnel behavior now consistently reports an immediate error for unavailable boxes.

- **Tests**
  - Added coverage for tunnel requests against stopped and non-started boxes.


DefeatMan (approved):
LGTM

PR #1355, opened 2026-08-26
fix(api): stop fencing runner assignment on box creation

### LINKED ISSUE #1353: 400 No available runners on concurrent box creation
## Symptom

On cloud.dev, creating several boxes concurrently makes some requests fail with HTTP `400 No available runners`, while creating the same boxes sequentially succeeds. The generated clients under `apps/libs/api-client*` carry no status-based retry policy, so the failed part of the burst is simply lost.

## Cause

The per-runner assignment lease is acquired with a non-blocking `SET NX`:

    // apps/api/src/box/common/redis-lock.provider.ts:167
    async acquireLease(key: string, ttl: number): Promise<RedisLockLease | null> {
      const owner = new LockCode(randomUUID())
      if (!(await this.lock(key, ttl, owner))) {
        return null                            // caller cannot tell "busy" from "unusable"
      }

The caller treats that `null` as unavailability and excludes the runner for the rest of its attempt:

    // apps/api/src/box/services/box.service.ts:134
    const lease = await this.redisLockProvider.acquireLease(lockKey, 30)
    if (!lease) {
      excludedRunnerIds.push(runner.id)
      continue
    }

When the (region, class) has a single schedulable runner, the next `findAvailableRunners` call returns zero rows and `getRandomAvailableRunner` throws (`apps/api/src/box/services/runner.service.ts:759`). Nothing was out of capacity — the request only lost a race whose window is the whole box insert.

## Reproduction

Two concurrent creates, one READY runner in the region and class:

    t0  req A: getRandomAvailableRunner()  -> [R1]      req B: getRandomAvailableRunner()  -> [R1]
    t1  req A: SET runner:R1:box-assignment NX -> OK    req B: SET ... NX -> nil
    t2  req A: recheck R1 -> READY                      req B: excludedRunnerIds = [R1]; attempt 2
    t3  req A: insert(box); release lease               req B: findAvailableRunners({exclude:[R1]}) -> []
    t4  req A: 201 Created                              req B: 400 No available runners

With `n` runners in the region and class, a burst of `k` concurrent creates starts failing as soon as requests collide on the same runners — the exclusion is per-request but permanent within that request, and the loop only tries three times.

Warm-pool top-up runs through the same fence (`apps/api/src/box/services/box.service.ts:176`), so background refills compete with user traffic for the same per-runner key.

## Expected

A runner should be excluded only when it is genuinely unschedulable. Losing a lock race is not a capacity condition and should not reach the client as `400`, which by HTTP semantics means the request was malformed and is therefore not retried.

DefeatMan:
## Summary

Concurrent box creation aimed at the same runner contended a non-blocking Redis `SET NX` lease; the loser could not tell "busy" from "unusable", excluded the runner, and — with one schedulable runner in that region and class — emptied the candidate set and answered `400 No available runners` while the fleet still had capacity. The per-runner interlock is removed rather than replaced: creation picks a candidate and inserts, and a drain that commits in that window leaves the box on a runner already marked draining.

## Call graph

Before
  BoxService.create                     (BoxService · apps/api/src/box/services/box.service.ts:200)
    └─ persistWithRunnerAssignmentFence (BoxService · apps/api/src/box/services/box.service.ts:126)
         ├─ getRandomAvailableRunner    (RunnerService · apps/api/src/box/services/runner.service.ts:755)
         ├─ acquireLease                (RedisLockProvider · apps/api/src/box/common/redis-lock.provider.ts:167)  ← BUG: non-blocking, so a lost race excludes a healthy runner
         ├─ findOneUncachedOrFail       (RunnerService · apps/api/src/box/services/runner.service.ts:213)         — re-read under the lease
         └─ boxRepository.insert        (BoxRepository · apps/api/src/box/repositories/box.repository.ts:55)
  …
  updateDrainingStatus                  (RunnerService · apps/api/src/box/services/runner.service.ts:737)         — waitForLease on the same per-runner key
  handleCheckDecommissionRunners        (RunnerService · apps/api/src/box/services/runner.service.ts:643)         — acquireLease, same key

After
  BoxService.create                     (BoxService · apps/api/src/box/services/box.service.ts:183)
    └─ persistOnAvailableRunner         (BoxService · apps/api/src/box/services/box.service.ts:144)
         ├─ getRandomAvailableRunner    (RunnerService · apps/api/src/box/services/runner.service.ts:778)         — the only eligibility check
         └─ boxRepository.insert        (BoxRepository · apps/api/src/box/repositories/box.repository.ts:55)      — own transaction, no runner-row lock
  …
  updateDrainingStatus                  (RunnerService · apps/api/src/box/services/runner.service.ts:752)         — FOR UPDATE, vs the decommission transition only
  handleCheckDecommissionRunners        (RunnerService · apps/api/src/box/services/runner.service.ts:643)         — FOR UPDATE, count then transition

Fixes #1353

## Changes

- `runner:<id>:box-assignment` and its three holders are gone, and nothing takes their place: `getRunnerAssignmentLockKey` is deleted, and `RedisLockProvider` stays in `RunnerService` only for the singleton cron guards (`check-runners`, `check-decommission-runners`), which are unrelated to per-runner exclusion.
- `persistOnAvailableRunner` is the whole assignment: pick a candidate, stamp `runnerId`, persist. No transaction wrapping the pair, no re-read, no candidate-exclusion retry loop. `findAvailableRunners` already filters out draining, unschedulable and non-READY rows, so it is the only eligibility check.
- **The accepted race.** A drain committing between the candidate query and the insert wins the row but not the box. Draining is soft — it stops new scheduling and waits out the boxes already there — and `handleCheckDecommissionRunners` counts assignments before every transition and resets its counter while any remain, so a box that slips in delays the drain instead of being orphaned. Nomad settles the same race the same way, re-checking node eligibility only at plan apply; Kubernetes lets the kubelet reject a pod the scheduler already bound.
- The two writers of runner state keep an interlock, but only against each other: `updateDrainingStatus` saves a whole `Runner`, so without `FOR UPDATE` its `state` read could be written back over a concurrent `DECOMMISSIONED`. Neither excludes assignment any more, so neither carries a `lock_timeout` — no create is queued behind them.
- **Status codes.** `400 No available runners` is again the only answer to an exhausted candidate set. The `408` paths and the contention reclassification are gone with the contention that motivated them; a `BadRequestError` from `getRandomAvailableRunner` propagates unchanged.
- `BoxRepository.insert` keeps its own transaction and takes no `EntityManager`; `BoxService` no longer injects `DataSource`. The generated-name retry wraps only the insert, so a name collision keeps the runner it already chose.
- Tests: `box.service.runner-assignment.integration.spec.ts` (real Postgres) covers the concurrent burst, a drain holding the runner row mid-commit, and a genuinely draining runner; the unit cases live in a `BoxService runner assignment` block and drive the assignment directly rather than through `create()`.

## How to verify

    cd apps
    DB_HOST=<host> DB_PORT=<port> DB_USERNAME=<user> DB_PASSWORD=<pass> DB_DATABASE=<db> \
    REDIS_HOST=<host> REDIS_PORT=<port> \
      npx jest --config api/jest.config.ts --rootDir api

`box.service.runner-assignment.integration.spec.ts` needs real Postgres and creates its own database; it skips when `DB_HOST` is unset. Against `main` its burst case fails with `No available runners`. Its drain case discriminates the other direction: reinstate a `pessimistic_read` read on the runner row inside an assignment transaction and `assigns onto a runner a drain is mid-commit instead of waiting for it` parks on the lock until the 20s test timeout.

## Risks / rollout

- The decommission re-count is now a snapshot rather than a fence. An insert that commits after it would leave a box on a `DECOMMISSIONED` runner. That needs a create stalled past three 10s checks between picking a runner and inserting — where the shared lock made it impossible.
- A box that lands on a draining runner holds the drain open for its own lifetime; auto-stop and auto-delete still apply, so the runner drains once the box goes away.
- `updateDrainingStatus` still resets its Redis counter and invalidates cache after the transaction commits (unchanged). A crash in that window leaves a stale `runner:draining-check:<id>`, so a later cycle can reach three checks early; the transition still re-verifies `draining` and a zero box count under `FOR UPDATE`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved concurrent box assignment reliability and distribution across available runners.
  * Prevented new assignments to runners that are actively draining.
  * Improved runner decommissioning and status updates during concurrent activity.
  * Reduced contention during runner state changes.
  * Standardized the error shown when no runners are available.

### DIFF
--- a/apps/api/src/box/repositories/box.repository.ts
+++ b/apps/api/src/box/repositories/box.repository.ts
@@ -23,6 +23,10 @@ import { BoxPublicStatusUpdatedEvent } from '../events/box-public-status-updated
 import { BoxOrganizationUpdatedEvent } from '../events/box-organization-updated.event'
 import { BoxLookupCacheInvalidationService } from '../services/box-lookup-cache-invalidation.service'
 
+// SQLSTATE lock_not_available — a statement waited longer than lock_timeout to
+// acquire a lock.
+const PG_LOCK_TIMEOUT_CODE = '55P03'
+
 // Cap how long the proxy auto-start UPDATE waits to acquire the box row's
 // write lock. Concurrent start/stop/sync go through updateWhere(), which holds
 // a pessimistic_write lock on the same row; there is no global statement/lock
@@ -32,10 +36,6 @@ import { BoxLookupCacheInvalidationService } from '../services/box-lookup-cache-
 // race-lost no-op. Aligned with the caller-side wait cap in
 // boxlite-proxy.controller.ts (PROXY_START_HINT_TIMEOUT_MS).
 const PROXY_START_LOCK_TIMEOUT_MS = 2000
-
-// SQLSTATE for `lock_not_available` — raised when a statement waits longer than
-// lock_timeout to acquire a lock.
-const PG_LOCK_TIMEOUT_CODE = '55P03'
 
 // A box the migration marker may claim, with the stamp its claim copies.
 type ParkedBox = { id: string; updatedAt: Date }
--- a/apps/api/src/box/services/box.service.runner-assignment.integration.spec.ts
+++ b/apps/api/src/box/services/box.service.runner-assignment.integration.spec.ts
@@ -1,0 +1,253 @@
+/*
+ * Copyright 2026 BoxLite AI
+ * SPDX-License-Identifier: AGPL-3.0
+ */
+
+import { randomUUID } from 'node:crypto'
+import Redis from 'ioredis'
+import { DataSource, Repository } from 'typeorm'
+import { CustomNamingStrategy } from '../../common/utils/naming-strategy.util'
+import { RedisLockProvider } from '../common/redis-lock.provider'
+import { Box } from '../entities/box.entity'
+import { BoxLastActivity } from '../entities/box-last-activity.entity'
+import { BoxMigration } from '../entities/box-migration.entity'
+import { Runner } from '../entities/runner.entity'
+import { BoxClass } from '../enums/box-class.enum'
+import { RunnerState } from '../enums/runner-state.enum'
+import { BoxRepository } from '../repositories/box.repository'
+import { BoxService } from './box.service'
+import { RunnerService } from './runner.service'
+
+const describeIfDatabase = process.env.DB_HOST ? describe : describe.skip
+
+// Its own database, not a schema in the shared one. The Runner entity's
+// `@PrimaryGeneratedColumn('uuid')` makes TypeORM issue
+// `CREATE EXTENSION IF NOT EXISTS "uuid-ossp"` on connect, and that statement is
+// not atomic — a sibling suite doing the same concurrently collides on
+// pg_extension_name_index. usage.service.integration.spec.ts also drops and
+// recreates schema public wholesale, which would take the extension with it. A
+// private database removes both couplings.
+const databaseName = `runner_assignment_${process.pid}_${randomUUID().replaceAll('-', '')}`
+const schemaName = 'public'
+const organizationId = '00000000-0000-4000-8000-0000000000ff'
+const region = 'us'
+
+// Enough concurrent creates to make the collision structural rather than a
+// timing coincidence: every caller resolves its candidate query before any of
+// them reaches the assignment step, so all of them target the single runner.
+const BURST = 8
+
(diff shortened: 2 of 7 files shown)


DorianZheng (dismissed):

DorianZheng:
LGTM

PR #1360, opened 2026-08-26
fix(api): api key with volume management

Monsteer714:
## Summary
Let a customer obtain and use a volume-scoped API key without a human in the loop: the console offers the volume permissions, `/v1/api-keys` issues keys from code, and `/v1/me` reports scopes derived from what the guards actually enforce.

## Call graph
Before
```
  Create Key dialog
    (CreateApiKeyDialog · apps/dashboard/src/components/CreateApiKeyDialog.tsx:92)
      └─ CREATE_API_KEY_PERMISSIONS_GROUPS = [Boxes]
           (constants/CreateApiKeyPermissionsGroups.ts:23)   ← BUG: the Volumes group was dropped in #715, so no console key carries a volume permission

  GET /v1/volumes
    (BoxliteVolumeController · apps/api/src/boxlite-rest/boxlite-volume.controller.ts:40)
      └─ canActivate (OrganizationResourceActionGuard · organization-resource-action.guard.ts:76) — 403


  POST /v1/api-keys → 404 — key issuance lives at `/api/api-keys`, above the `{basePath}/v1` root the REST contract is written against
  
  GET /v1/me
    (BoxliteMeController · boxlite-me.controller.ts:43)
      └─ hardcoded ["box:read", …, "image:read", "image:write", "me:read"]   ← BUG: unconnected to enforcement — claims image scopes this deployment serves no route for, never mentions volumes
```
After
```
  Create Key dialog
    (CreateApiKeyDialog · CreateApiKeyDialog.tsx:92)
      ├─ selectableGroups (lib/api-key-permissions.ts:25) — Volumes, per action
      └─ alwaysGrantedPermissions (lib/api-key-permissions.ts:39) — box access, not a choice
           └─ POST /api-keys {permissions: [write:boxes, delete:boxes, read:volumes, …]}

  GET /v1/volumes
    (BoxliteVolumeController · boxlite-volume.controller.ts:40)
      └─ canActivate (OrganizationResourceActionGuard · :76) — 200

  POST|GET /v1/api-keys, DELETE /v1/api-keys/{name}
    (BoxliteApiKeyController · boxlite-api-key.controller.ts:50)  — NEW
      ├─ ungrantablePermissions (api-key/api-key-grant.ts:50) — 403 naming the refused subset
      └─ ApiKeyService.createApiKey — same credential as the dashboard mints
      
  GET /v1/me
    (BoxliteMeController · boxlite-me.controller.ts:43)
      └─ honouredPermissions (boxlite-me.controller.ts:78) — resolves membership the guards never attached to this route
           └─ effectivePermissions (api-key/api-key-grant.ts:29) — key permissions, role permissions, or unbounded
                └─ resolveApiScopes (boxlite-rest/api-scope.ts:75)
                     └─ SCOPE_REQUIREMENTS (api-scope.ts:54) — locked to the route metadata by api-scope.spec.ts
```
### Console can issue volume keys
- `apps/dashboard/src/constants/CreateApiKeyPermissionsGroups.ts`: restore the `Volumes` group (`read`/`write`/`delete:volumes`) that #715 removed; mark `Boxes` `alwaysGranted` — box access is what a key is for, and the checkbox had one sensible answer.
- `apps/dashboard/src/lib/api-key-permissions.ts`: new module holding the selection rules — `selectableGroups`, `alwaysGrantedPermissions`, `togglePermission`, `toggleGroup`, `groupSelectionState`, `permissionActionLabel`. Selections normalise to declaration order so the request body does not depend on click order; both the offered and the always-granted sets are narrowed to what the operator holds, so the API never rejects a permission the operator never chose.
- `apps/dashboard/src/components/CreateApiKeyDialog.tsx`: replace the static "Boxes API access" blurb with per-action checkboxes; the group header toggles its actions and shows an indeterminate state; the dialog submits the operator's selection plus the always-granted set.
- `apps/dashboard/vite.config.mts`: add the vitest `test` block. Without it `@nx/vite` infers no `test` target, so the project's spec files — the pre-existing `CreateApiKeyDialog` spec included — had never run.

### Keys can be issued from code
- `apps/api/src/boxlite-rest/boxlite-api-key.controller.ts`: new controller on `v1/api-keys` and `v1/:prefix/api-keys` (create, list, revoke), delegating to `ApiKeyService`. Mirrors how `BoxliteVolumeController` pairs with the dashboard's own `VolumeController`; snake_case responses; the secret is returned once, on create.
- `apps/api/src/api-key/api-key-grant.ts`: `effectivePermissions` — the permissions the resource guard honours for a caller — and `ungrantablePermissions` built on it. An API key may pass on only what it carries, so a key issued for one resource cannot mint itself access to another; interactive sessions keep the membership rule. The dashboard's existing controller is untouched.
- `apps/api/src/boxlite-rest/dto/create-api-key.dto.ts`: request body. Names cannot contain `/` — a key is revoked by name on this surface, and a name that splits into two path segments would leave it unrevokable by the code that minted it.

### Scopes describe what the credential can reach
- `apps/api/src/boxlite-rest/api-scope.ts`: the scope vocabulary, `SCOPE_REQUIREMENTS` (what each scope costs), and `resolveApiScopes`. Image and snapshot scopes are gone — this deployment serves no such route, as `GET /v1/config` also reports. An empty requirement list means the scope is unconditional: box access rides on every key, so `box:*` is always reported.
- Every REST handler now carries `@RestApiScope` beside the permissions the guard already checks (`boxlite-box`, `boxlite-volume`, `boxlite-api-key`, `boxlite-proxy`, `boxlite-me`).
- `apps/api/src/boxlite-rest/boxlite-me.controller.ts`: derive the array instead of hardcoding it, and resolve the caller's membership first — `/v1/me` runs behind `CombinedAuthGuard` alone, so unlike every other route nothing has attached it by the time the handler runs.
- `openapi/box.openapi.yaml`: document the API-key operations, the permission vocabulary, and a `ForbiddenError` response the contract had nowhere defined; rewrite the `scopes` description around the exactness the field now has and drop its `boxes:read` example, a third spelling that existed nowhere in the system.

## How to verify

- `cd apps && npx jest --config api/jest.config.ts --rootDir api` — 608 pass. 47 are new: grant rules (9), the REST controller end-to-end behind a real Nest app (10), scope coverage and resolution (10 + 32 generated), `/v1/me` derivation (8).
- `cd apps && npx vitest run --root dashboard` — 158 pass. 24 are new, including a render test that drives the real dialog: the three volume actions appear as separate options, no box row is offered, and the submitted payload carries box access plus whatever volume actions were left ticked. One pre-existing spec fails under Node 26, whose own `localStorage` global shadows the jsdom one.
- `api-scope.spec.ts` reflects over the route metadata, so the scope map cannot drift: dropping a handler's `@RestApiScope`, changing a route's required permissions, or adding a scope no route serves each fails a specific assertion.
- Local stack (postgres + redis + dex, no runner needed):
  - `POST /v1/{prefix}/api-keys` with the three volume permissions → 201; the same key then mints a narrower one → 201, and a wider one → 403 naming `write:boxes`
  - a boxes-only key requesting `read:volumes` → 403 naming it
  - `GET /v1/api-keys` → metadata only, no `value`; `DELETE /v1/api-keys/{name}` → 204, the key then answers 401
  - `GET /v1/me` for a volume key lists `volume:read|write|delete`; for a boxes-only key lists none of them, matching `GET /v1/volumes` answering 200 and 403 respectively
  - name containing `/`, empty permissions, and an unknown permission → 400; duplicate name → 409
- Against a deployed stage: a volume-scoped key provisions volumes concurrently — five created, all `ready`, listed, then removed.

## Not in this PR
- Read-only mounts. `read_only` exists only as a DTO field pinned to `false`, with nothing plumbed to the runner; it is a mount option rather than a permission, and belongs with that feature.
- The dashboard's own `/api/api-keys` keeps its membership-based rule, so a key belonging to an owner can still mint any key there.
- `boxlite serve` carries its own hardcoded scope list, now diverged from this vocabulary.

## Summary by CodeRabbit

- **New Features**
  - Added API key creation, listing, and revocation.
  - Added configurable permissions and expiration limits for API keys.
  - Added permission-aware access scopes across REST resources.
  - Added dashboard controls for selecting individual or grouped permissions.
- **Documentation**
  - Updated the API specification with API key endpoints, schemas, permissions, and scopes.
- **Tests**
  - Expanded coverage for API key workflows, permissions, scopes, and dashboard selection.


apps/api/src/boxlite-rest/boxlite-api-key.controller.ts
```diff
@@ -0,0 +1,116 @@
...
+  @Delete(':name')
+  @HttpCode(204)
+  @Audit({
+    action: AuditAction.DELETE,
+    targetType: AuditTarget.API_KEY,
+    targetIdFromRequest: (req) => req.params.name,
+  })
+  @RestApiScope('api_key:delete')
+  async remove(@AuthContext() authContext: OrganizationAuthContext, @Param('name') name: string): Promise<void> {
+    await this.apiKeyService.deleteApiKey(authContext.organizationId, authContext.userId, name)
```
G4614: Do we need to check the permission of `deleteApiKey` before the deleting operation?
Like the `ungrantablePermissions` used while `create` in L68 in this code file.
For example, we can delete other strong keys while login with a weak key.
If this is accepted, maybe we need to instruct inside documents

apps/api/src/boxlite-rest/dto/create-api-key.dto.ts
```diff
@@ -0,0 +1,48 @@
...
+  name: string
+
+  @IsArray()
+  @ArrayNotEmpty()
+  @IsEnum(OrganizationResourcePermission, { each: true })
+  permissions: OrganizationResourcePermission[]
+
+  @IsOptional()
+  @Type(() => Date)
+  @IsDate()
```
G4614: Shall we operate a check about whether the date is valid but not only the right instance?

apps/dashboard/src/components/CreateApiKeyDialog.tsx
```diff
@@ -212,7 +265,7 @@ export const CreateApiKeyDialog: React.FC<CreateApiKeyDialogProps> = ({
               <button
                 type="submit"
                 form="create-api-key-form"
-                disabled={!name.trim() || submitting || !organizationId || availablePermissions.length === 0}
+                disabled={!name.trim() || submitting || !organizationId}
```
G4614: A member whose roles carry no grantable permissions can now click Create and will always get a server-side 400.

The previous disabled guard included `availablePermissions.length === 0`; this change drops it on the assumption that box access is always offered. But both `alwaysGrantedPermissions()` and `selectableGroups()` narrow to what the operator holds (`available`), so for a member with nothing grantable the submitted payload is `permissions: []`, which violates the API's `@ArrayNotEmpty` and comes back as 400 with an error toast.

Verified by test: rendering the dialog with `availablePermissions={[]}` leaves the submit button enabled, and submitting sends `{organizationId, name, permissions: [], expiresAt: null}`.

Suggestion: keep (or re-derive) the empty-selection guard — e.g. disable when `[...grantedByDefault, ...selectedPermissions].length === 0`.

PR #1370, opened 2026-08-27
fix(api): default box visibility to private

G4614:
Fixes POL-205.

## Before

```
POST /v1/boxes {}  (no network.inbound specified)
  -> createBoxToCreateBox(dto)          [box-to-box.mapper.ts]
       createDto.public = undefined     (inbound.mode not specified)
  -> BoxService.create()                [box.service.ts:291]
       box.public = createBoxDto.public ?? true   ← BUG: silently public
  -> anonymous internet traffic reaches the box's preview URL, no auth
```

Same bug in `assignWarmPoolBox` (box.service.ts:355) for boxes served from
the warm pool.

## After

```
POST /v1/boxes {}
  -> createBoxToCreateBox(dto) -> createDto.public = undefined
  -> BoxService.create()
       box.public = createBoxDto.public ?? false   (both call sites)
  -> box is private by default; caller must opt in with
     network.inbound.mode="enabled" (or legacy public:true) to be reachable
```

This supersedes the Warning-header mitigation from the branch's prior commit —
once the default is private, a caller can no longer end up anonymously public
without asking, so the header (and its `isAnonymouslyPublicByDefault` helper)
served no further purpose and were removed.

**Note:** this is a breaking behavior change for any existing caller relying on
the old anonymous-public default. Flagging for release notes / staged rollout
per the issue's own migration plan.

## Test plan

- `box.service.spec.ts`'s existing `public defaults` tests pin both call
  sites; updated their expected values for the new default (undefined -> false,
  explicit true -> true) — 682 passed, 72 skipped, 0 failed on the full `api`
  suite
- Two-side verification: reverted the two `?? false` lines back to `?? true`,
  confirmed both pinning tests fail with the expected "expect false, got true"
  diff; restored, all green again
- `tsc --noEmit` clean

## Summary by CodeRabbit

* **New Features**
  * Newly created boxes now default to private when visibility is not specified.
  * Warm-pool assigned boxes also default to private.
  * Explicitly requested visibility settings continue to be honored.

* **Bug Fixes**
  * Tunnel requests to private boxes are now rejected.
  * Removed the warning response previously returned for boxes without an inbound network mode.

### DIFF
--- a/apps/api/src/box/services/box.service.spec.ts
+++ b/apps/api/src/box/services/box.service.spec.ts
@@ -304,8 +304,8 @@ describe('BoxService public defaults', () => {
   )
 
   it.each([
-    [undefined, true],
-    [false, false],
+    [undefined, false],
+    [true, true],
   ])('defaults a fresh box to public=%s', async (requestedPublic, expectedPublic) => {
     const { service, boxRepository } = makeCreateService()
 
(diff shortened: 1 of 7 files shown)


DefeatMan (approved):
LGTM

PR #1384, opened 2026-08-27
fix(activity): count preview and tunnel traffic as box activity

### LINKED ISSUE #1383: External preview/proxy traffic does not count as activity, so an actively-used box is idle-stopped
A box actively serving requests through its preview URL is idle-stopped at `last SDK call + autoStop`, as if the preview traffic never happened. Only SDK calls count as activity, contradicting the documented behaviour that "Preview URL traffic keeps a running box alive".

## Reproduction

Create a box with `autoStop=1` (60s), start a server on a port, then poll its external preview URL every 15s with **zero further SDK calls**:

```
T=15s   external GET=200  boxState=running
T=30s   external GET=200  boxState=running
T=45s   external GET=200  boxState=stopping   <- already stopping despite continuous 200s
T=60s   external GET=502  boxState=stopped
```

The box stops at exactly `last SDK call + 60s`. A zero-access control box with the same `autoStop` also stops after its window, so idle-stop itself works — the defect is that external traffic is not recorded as activity.

## Cause

The proxy does renew on every proxied request, but `POST /box/{id}/last-activity` sits on `BoxController`, whose class-level `OrganizationResourceActionGuard` runs before the route's own `OrGuard([BoxAccessGuard, ProxyGuard, RegionBoxAccessGuard])`. That guard exempts only a runner context, so a proxy context — which carries no organization by design — falls through to an organization check it can never satisfy, and every renewal returns 403.

Separately, CONNECT tunnels never renew activity at all, so a long-lived tunnel stream does not keep its box alive either.

## Expected

A running box reached through its preview URL or a CONNECT tunnel, with no SDK calls, is not idle-stopped while the traffic continues. A box with no access still stops after its `autoStop` window.

Linear: POL-326

DefeatMan:
## Summary

A box serving traffic through its preview URL was idle-stopped at `last SDK call + autoStop`, because every renewal the proxy sent was rejected with 403 — `POST /box/{id}/last-activity` sits on `BoxController`, whose class-level `OrganizationResourceActionGuard` runs ahead of the route's own `OrGuard` and turns away a machine context for carrying no organization. CONNECT tunnels never renewed at all.

Measuring that fix surfaced a second defect on the same path: the API accepts an `auto_stop` the renewal path can never honour. At `auto_stop: 30` on dev the box was running at t=31s and stopping at t=37s while its preview URL was polled, because the proxy renews on a fixed 50s poll. Windows shorter than one poll are now refused at the boundary instead of stored.

## Call graph

```text
Before
  GetProxyTarget          (Proxy · apps/proxy/pkg/proxy/get_box_target.go:35)
    └─ updateLastActivity (Proxy · apps/proxy/pkg/proxy/get_box_target.go:413)  — POSTs the renewal
         └─ canActivate   (OrganizationResourceActionGuard · apps/api/src/organization/guards/organization-resource-action.guard.ts:34)  ← BUG: exempts only isRunnerContext, so a proxy context falls through to an organization check it can never satisfy — 403
              └─ OrGuard([BoxAccessGuard, ProxyGuard, RegionBoxAccessGuard])  (· apps/api/src/box/controllers/box.controller.ts:446)  — never reached
                   └─ updateLastActivity  (BoxController · apps/api/src/box/controllers/box.controller.ts:447)  — never reached

  handleTunnelConnect     (Proxy · apps/proxy/pkg/proxy/tunnel.go:28)  ← BUG: renews nothing, so a live CONNECT stream never refreshes the idle timer
    └─ dialRunnerTunnel   (Proxy · apps/proxy/pkg/proxy/tunnel.go:107)

  createBox               (BoxliteBoxController · apps/api/src/boxlite-rest/boxlite-box.controller.ts:104)  — auto_stop: 30
    └─ createDto.autoStop (· apps/api/src/boxlite-rest/mappers/box-to-box.mapper.ts:55)
         └─ resolveLifecyclePolicy (BoxService · apps/api/src/box/services/box.service.ts:1449)  ← BUG: stores any non-negative value, including a window shorter than the proxy's 50s renewal poll (get_box_target.go:414)
              └─ autostopCheck (BoxManager · apps/api/src/box/managers/box.manager.ts:73)  — reaps the box at t≈31-37s with preview traffic flowing

After
  GetProxyTarget          (Proxy · apps/proxy/pkg/proxy/get_box_target.go:35)
    └─ updateLastActivity (Proxy · apps/proxy/pkg/proxy/get_box_target.go:413)  — POSTs the renewal
         └─ canActivate   (OrganizationResourceActionGuard · apps/api/src/organization/guards/organization-resource-action.guard.ts:50)  — admits a machine role when the handler declares that role's own guard
              └─ OrGuard([BoxAccessGuard, ProxyGuard, RegionBoxAccessGuard])  (· apps/api/src/box/controllers/box.controller.ts:446)
                   └─ updateLastActivity  (BoxController · apps/api/src/box/controllers/box.controller.ts:447)
                        └─ updateLastActivityAt  (BoxService · apps/api/src/box/services/box.service.ts:1080)

  handleTunnelConnect     (Proxy · apps/proxy/pkg/proxy/tunnel.go:28)  — renews for the life of the stream, ticker bound to the handler
    └─ updateLastActivity (Proxy · apps/proxy/pkg/proxy/get_box_target.go:413)
    └─ dialRunnerTunnel   (Proxy · apps/proxy/pkg/proxy/tunnel.go:117)

  createBox               (BoxliteBoxController · apps/api/src/boxlite-rest/boxlite-box.controller.ts:104)  — auto_stop: 30
    └─ createDto.autoStop (· apps/api/src/boxlite-rest/mappers/box-to-box.mapper.ts:55)
         └─ resolveLifecyclePolicy (BoxService · apps/api/src/box/services/box.service.ts:1461)  — 400: must be 0 (disabled) or at least 60 seconds
              └─ autostopCheck (BoxManager · apps/api/src/box/managers/box.manager.ts:73)  — never reached; no such box is created
```

Fixes #1383

Linear: POL-326

## Changes

- `OrganizationResourceActionGuard` generalises its runner-only escape hatch into a per-role table: a handler admits a machine role by declaring one of that role's own guards. The exemption stays opt-in and fail-closed — `ProxyGuard` alone speaks for the proxy, because `BoxAccessGuard` admits *any* proxy context for *any* box (`box-access.guard.ts:47`) and listing it would widen every route that uses it, not just the routes meant for the proxy.
- `handleTunnelConnect` renews activity for the life of the CONNECT stream, with the poll ticker bound to the handler. It renews ahead of the runner dial, matching the API-side proxy, which renews ahead of its own readiness gate (`boxlite-proxy.controller.ts:246-252`) so a box being actively reached for is not reaped while the path to it is still coming up.
- `resolveLifecyclePolicy` refuses a non-zero `auto_stop` below `MIN_AUTO_STOP_SECONDS` (60); `0` still disables AutoStop. Both surfaces converge on that resolver — the native camelCase `autoStop` and the snake_case `auto_stop` the boxlite-rest mapper feeds into it — so the floor is written once and covers create and warm-pool assignment alike. The minutes-based `POST /box/{id}/autostop/{interval}` route cannot express 1–59s at all and is untouched.
- `apps/api/.../box-last-activity-proxy-auth.spec.ts` exercises the real guard chain end to end for a proxy and a region-proxy context, and pins the negative case: a route declaring no proxy guard still returns 403. `box.service.lifecycle-policy.spec.ts` pins the floor, the `0` escape hatch and the 900s default.

Only one route was ever affected by the guard change. `ProxyGuard` and `RegionBoxAccessGuard` appear on three handlers; the two on `RunnerController` sit under a class that carries no `OrganizationResourceActionGuard` (`runner.controller.ts:57`), so they were never blocked.

## How to verify

```bash
cd apps/proxy && go test ./pkg/proxy/ -run TestTunnelConnectRenewsBoxActivityBeforeStreaming
cd apps/api  && ../node_modules/.bin/jest --testPathPatterns 'guard|box-last-activity|lifecycle-policy'
```

All three reproducers fail on `main`: the API pair gets 403 where it expects 201, the tunnel test times out waiting for a renewal that never arrives, and the floor's 1s/30s/59s cases report `Received function did not throw`. The pre-existing `organization-resource-action.guard.spec.ts` covers five runner-exemption cases and guards the refactor.

End to end: create a box with `autoStop=60`, start a server on a port, then poll its preview URL every 15s with no further SDK calls. The box stays `running` past the 60s window; a zero-access control box with the same `autoStop` still stops. `auto_stop: 30` now returns 400 rather than creating a box that gets reaped under traffic.

## Risks / rollout

- The guard change widens authorization, so it is worth a careful read. It is opt-in per handler and per role, and no route gains a role it did not already declare a guard for — but the blast radius of a mistake here is every route on a controller carrying `OrganizationResourceActionGuard`.
- The floor is a breaking change for any caller passing `auto_stop` between 1 and 59: those creates now 400 instead of quietly producing a box the renewal path cannot keep alive. Nothing in this repo passes such a value, and the window was unreachable before lifecycle intervals moved from minutes to seconds. Boxes already stored below the floor are left alone — they keep being swept on their stored value.
- Margin at the floor is thin: renewal dedups for 45s (`get_box_target.go:446`) against a 60s window, so a client polling every 15s still leaves the sweeper a narrow chance to read a stale timestamp. The API-side proxy solved this with `autoStop/2` (`boxlite-proxy.controller.ts:307-310`); doing the same here needs the proxy to know `autoStop`.
- `last_activity_at` / `last_activity_source` on the box resource are not part of this PR — the issue attributes that to POL-313 (`313-C7`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Active proxy tunnel connections now keep boxes from being incorrectly marked inactive while connections are established.
  * Proxy and regional proxy authentication can renew box activity where permitted.
  * Box auto-stop settings now reject nonzero intervals shorter than 60 seconds.

* **Documentation**
  * Clarified the 60-second minimum for applicable BoxLite Cloud auto-stop settings and documented server-specific behavior.

* **Tests**
  * Added coverage for activity renewal, proxy authentication, and auto-stop lifecycle rules.


apps/proxy/pkg/proxy/tunnel.go
```diff
@@ -44,6 +44,16 @@ func (p *Proxy) handleTunnelConnect(writer http.ResponseWriter, request *http.Re
...
 	}
+
+	// A CONNECT tunnel outlives any single HTTP request, so for tunnels the poll
+	// is the renewal rather than a fallback: bind its ticker to the stream, which
+	// ends when this handler returns. Renewing before the runner dial mirrors the
+	// API-side proxy, which renews ahead of its own readiness gate so a box being
+	// actively reached for is not reaped while the path to it is still coming up.
+	activityPoll := newActivityPollController()
+	defer activityPoll.stop()
+	go p.updateLastActivity(context.WithoutCancel(request.Context()), boxID, true, activityPoll.done)
```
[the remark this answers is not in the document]
DefeatMan: not a problem, verification fist

G4614:
Mechanism looks right, nothing to change here.

Measured on dev, `auto_stop: 30` (accepted verbatim — `@Min(0)`, `create-box.dto.ts:262`), one renewal at t=0 then left alone:

```
t=1s   running
t=31s  running
t=37s  stopping
t=49s  stopped
```

Reaping starts around t=31-37s, but the proxy's renewal cadence is a fixed 50s (`get_box_target.go:423`), so the next renewal never arrives in time — any box with `auto_stop` under the poll interval is reaped with traffic flowing. Worth a lower bound on `auto_stop` later.

For #1383's acceptance: its repro uses `autoStop=1` (60s), 10s of margin over the poll — if that case looks flaky after this lands, it is this, not the fix.

apps/proxy/pkg/proxy/tunnel.go
ltstriker, line 55: this poll method poll fixed 50sec. but current api have not such a limit.  make sure if there is a limit at first.

DefeatMan: You were right that there was no limit — there is one now, in 6f63877f. `resolveLifecyclePolicy` (`apps/api/src/box/services/box.service.ts:1461`) rejects a non-zero `auto_stop` below `MIN_AUTO_STOP_SECONDS` (60) with a 400, so the fixed 50s poll always has a window it can keep alive. `0` still disables AutoStop, and the minutes-based `POST /box/{id}/autostop/{interval}` route can only emit 0 or >= 60. Boxes already stored below the floor keep their value.

ltstriker (approved):

ltstriker (approved):

PR #1389, opened 2026-08-28
fix(guest): remove OwnershipFixer that overwrites image-USER-owned paths

G4614:
Fix `OwnershipFixer` chowning all rootfs inodes to root:root after every disk mount, breaking exec as the image USER (uid=1000).

Test plan:
- [ ] `make test:e2e:files` — dd write to /workspace must pass (was rc=1)
- [ ] `make test:e2e:exec` — default exec user should be uid=1000


src/guest/src/service/container.rs
```diff
@@ -244,6 +245,12 @@ impl ContainerService for GuestServer {
             }));
         }
 
+        // For disk-based rootfs: verify ownership matches the exec user.
+        // Defence-in-depth against ext4 build regressions; no-op on correct builds.
+        if matches!(&rootfs_init.strategy, Some(rootfs_init::Strategy::Disk(_))) {
+            check_rootfs_ownership(&shared_rootfs, &config.user);
```
ltstriker: seems not a good idea to chown all file to one owner. how about just not chown?

ltstriker (approved):

ltstriker (approved):

PR #1390, opened 2026-08-28
feat(api): report last activity in box metadata

DefeatMan:
## Summary

Box metadata reported `createdAt` and `updatedAt` but never when a box was last
*used* — the clock AutoStop measures idleness against. This adds
`lastActivityAt` (`last_activity_at` on the REST surface) to box metadata, read
from the same source the idle sweeper acts on, and surfaces it through the
generated clients, the SDKs, and `boxlite inspect`.

## Call graph

```text
Before
  toBoxDto              (BoxService · apps/api/src/box/services/box.service.ts:1090)
    └─ fromBox          (BoxDto · apps/api/src/box/dto/box.dto.ts:298)              — createdAt/updatedAt only
  toBoxDtos             (BoxService · apps/api/src/box/services/box.service.ts:1098)
    └─ fromBox          (BoxDto · apps/api/src/box/dto/box.dto.ts:298)              — per box, same fields
  …
  boxToBoxResponse      (mapper · apps/api/src/boxlite-rest/mappers/box-to-box.mapper.ts:18)
  …
  to_box_info           (BoxResponse · src/boxlite/src/rest/types.rs:320)           — no activity field to carry
    └─ InspectPresenter::from  (src/cli/src/commands/inspect.rs:65)

After
  toBoxDto              (BoxService · apps/api/src/box/services/box.service.ts:1090)
    ├─ getLastActivityAt      (BoxActivityService · apps/api/src/box/services/box-activity.service.ts:58)
    │                                                                              — delegates to the batch read
    └─ fromBox          (BoxDto · apps/api/src/box/dto/box.dto.ts:298)              — takes lastActivityAt as an argument
  toBoxDtos             (BoxService · apps/api/src/box/services/box.service.ts:1098)
    ├─ getLastActivityAtMany  (BoxActivityService · apps/api/src/box/services/box-activity.service.ts:71)
    │                                                                              — one pipelined ZSCORE + one query
    └─ fromBox          (BoxDto · apps/api/src/box/dto/box.dto.ts:298)              — per box, from the merged map
  …
  boxToBoxResponse      (mapper · apps/api/src/boxlite-rest/mappers/box-to-box.mapper.ts:18)  — emits last_activity_at
  …
  to_box_info           (BoxResponse · src/boxlite/src/rest/types.rs:320)           — parses it; unparseable ⇒ None
    └─ InspectPresenter::from  (src/cli/src/commands/inspect.rs:65)                 — renders State.LastActivityAt
```

## Changes

- `BoxActivityService.getLastActivityAtMany` reads many boxes in two round trips
  regardless of count: one pipelined `ZSCORE` over the Redis buffer, then a
  single `findBy(In(...))` for only the boxes the buffer did not answer. The
  single-box `getLastActivityAt` is now a thin wrapper over it, so both paths
  share one buffer-before-database precedence. A per-command pipeline error is
  raised rather than quietly answered from the database, matching what a
  single-box `ZSCORE` failure has always done.
- List responses take the batch path, so the new field does not put a read on
  the per-box N+1 path.
- The value is reported raw. The sweeper's fallback to `updatedAt` is a stop
  policy, not recorded activity, so a box that has never been used reports
  nothing rather than its creation time.
- Local runtimes have no idle sweeper and record no activity, so `BoxInfo`
  from a local box carries `None`.
- `BoxResponse::to_box_info` maps an absent *or unparseable* timestamp to `None`
  rather than `now()` (as `created_at`/`updated_at` do): dating activity to this
  instant would read as "just used", the exact opposite of the idleness this
  field describes. `#[serde(default)]` keeps responses from older servers
  readable.
- Regenerated clients (`api-client`, `api-client-go`), OpenAPI specs, and the
  Python/Node/Go/C SDK info surfaces; docs updated for CLI, Node, and Python.

## How to verify

```bash
make test:apps FILTER='BoxActivityService activity reads'
make test:unit:rust
```

- `boxlite inspect <box>` against a cloud box shows `State.LastActivityAt`; the
  same command against a local box shows `null`.

## Risks / rollout

- Additive and optional on every surface: a client that does not know the field
  ignores it, and a server that does not send it deserializes to `None`.
- List endpoints gain one pipelined Redis read plus at most one extra query per
  request — bounded by page size, not by box count.

## Summary by CodeRabbit

* **New Features**
  * Box information now includes the timestamp of its most recent recorded activity.
  * Activity timestamps are available through the REST API, OpenAPI schema, and C, Go, Node.js, and Python SDKs.
  * CLI inspection output displays the last activity time when available.
  * Local boxes report no recorded activity.
* **Documentation**
  * Updated API and SDK references with field formats and availability.
* **Tests**
  * Added coverage for buffered, stored, missing, and invalid timestamps.

### DIFF
--- a/apps/api/src/box/services/box-activity.service.ts
+++ b/apps/api/src/box/services/box-activity.service.ts
@@ -7,7 +7,7 @@ import { Injectable, Logger } from '@nestjs/common'
 import { InjectRedis } from '@nestjs-modules/ioredis'
 import Redis from 'ioredis'
 import { InjectDataSource } from '@nestjs/typeorm'
-import { DataSource } from 'typeorm'
+import { DataSource, In } from 'typeorm'
 import { Cron, CronExpression } from '@nestjs/schedule'
 import { RedisLockProvider } from '../common/redis-lock.provider'
 import { BoxLastActivity } from '../entities/box-last-activity.entity'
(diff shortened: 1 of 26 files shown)


apps/api/src/box/services/box-activity.service.ts
```diff
@@ -56,14 +56,60 @@ export class BoxActivityService {
...
+    // the same buffer in one round trip against any server version.
+    const bufferedScores = await pipeline.exec()
+
+    const unbufferedBoxIds: string[] = []
+    uniqueBoxIds.forEach((boxId, index) => {
+      const [error, score] = bufferedScores?.[index] ?? [null, null]
+      // A per-command failure is raised rather than answered from the database:
+      // a single-box ZSCORE failure has always surfaced to the caller.
+      if (error) {
+        throw error
```
G4614: Shall we give some warnings or err msgs to show the thrown error is caused by `lastactivity`, so that the downstream error can be easily located

G4614: also maybe a test case can be added

ltstriker (approved):

PR #1391, opened 2026-08-28
fix(proxy): gate box preview access on an active network-tunnel lease

G4614:
Box services must not be reachable from a guessed URL unless the owner explicitly opts in (`public: true`) or holds a `network tunnel` open.

Test plan:
- [x] `npx jest --testPathPatterns 'box.service.spec|boxlite-proxy.controller.spec'` — 48 tests pass
- [x] `go test ./proxy/...` — all proxy Go tests pass
- [x] `cargo check -p boxlite-cli` — CLI compiles cleanly
- [ ] e2e against dev not run

## Summary by CodeRabbit

- **New Features**
  - Network tunnels remain active while the CLI tunnel command runs, with automatic lease renewal.
  - Added support for explicitly closing network tunnels.
  - Local forwarding and URL-only tunnel modes continue until interrupted.
- **Bug Fixes**
  - Browser access automatically stops when the tunnel process exits or its lease expires.
  - Tunnel URLs use consistent casing and remain available during active forwarding.

### DIFF
--- a/apps/api/src/box/services/box.service.spec.ts
+++ b/apps/api/src/box/services/box.service.spec.ts
@@ -218,6 +218,7 @@ function makeNetworkTunnelService() {
     }),
   } as any
   const regionService = { findOne: jest.fn().mockResolvedValue(null) } as any
+  const redis = { setex: jest.fn().mockResolvedValue('OK'), del: jest.fn().mockResolvedValue(1) } as any
   const noop = {} as any
   const service = new BoxService(
     noop,
@@ -236,6 +237,7 @@ function makeNetworkTunnelService() {
     noop, // jobRepository
     noop, // jobService
   )
+  ;(service as any).redis = redis
   jest.spyOn(service, 'findOneByIdOrName').mockResolvedValue({
     id: 'MixedCaseBox',
     region: 'region-1',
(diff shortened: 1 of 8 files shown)


G4614:
Closing this — the liveness-lease approach is being replaced by a durable per-port exposure declaration. Recording why, since the problem it targets is real and still needs solving.

**The threat model does not hold as stated.** The commit message motivates this as "reachable at a guessable URL", but the preview hostname is `{port}-d-{hex(box uuid)}.{domain}` (`encodeDirectPreviewBoxId`) — 128 bits, not enumerable. And `box.public` already defaults to `false` on `main` (`box.entity.ts:133`), so a box is only reachable after the owner explicitly opts in. The real risk with an opted-in box is the URL being *shared or leaked*, and a liveness lease does not help there: whoever received the link uses it while the lease is live.

**What the lease costs:**

- Reachability is bound to a foreground process on the user's laptop. Close the terminal, sleep the machine, drop the SSH session — the URL dies within 15s. For "deploy a service and give someone the link", that is the core use case, not an edge case.
- Renewal re-calls `POST :boxId/network/tunnel` every 5s for its `SETEX` side effect. That endpoint is also the one that records activity and (in #1431) resumes a stopped box, so a foreground tunnel silently pins the box: `auto_stop` can never fire while it runs.
- The lease key carries no port, so opening 3000 makes 22 and 8080 reachable too — the opposite of what a per-port URL implies.
- `IsLive` returns `false` when `tunnelChecker` is nil, which is the no-Redis case, and the caller in `getBoxPublic` has no fallback branch. Any deployment without Redis (and any Redis blip, since errors are fail-closed) makes every public box unreachable.
- `getBoxPublic` caches `isPublic && IsLive` under the `boxPublicCache` key for 3s. So `DELETE .../network/tunnel` is not actually immediate — it can stay reachable for up to 3s — and symmetrically a freshly opened tunnel can 403 for up to 3s behind a stale `false`. The header comment accounts for the first direction but not the second.

**Direction instead:** make exposure a durable record per `(box, port)` with a visibility mode (`public` or credential-required), independent of box state. The credential half already exists — `box.authToken` accepted via header/query/cookie in `apps/proxy/pkg/proxy/auth.go` — which is the same shape as E2B's traffic access token. The fail-safe intent here (exposure that lapses if forgotten) is kept as a server-side expiry refreshed by real traffic, using the `last-activity` call the proxy already makes on every request, rather than a client heartbeat.

That also unblocks waking a stopped box from an incoming request, which is impossible while "exposed" means "a client process is currently renewing".

Branch stays if we want to revive any of it.

PR #1394, opened 2026-08-28
fix(core): map BoxOptions caller-input validation errors to 400

G4614:
Fixes POL-356.

**Depends on #1206 (POL-207)** — stacked on `network-info-inbound-outbound-split`
for the inbound/outbound NetworkSpec shape; diff will shrink once #1206 merges.

## Investigation

POL-356 as filed reads as "inbound.allow_net isn't rejected yet" — but it already
is, and has been on main, across every layer: core Rust (`options.rs`), Python/Node/Go
SDKs, the apps/api DTO, self-hosted `boxlite serve`, the CLI (flag not exposed),
and the OpenAPI spec. The actual bug is narrower and more interesting:

## Before

```
POST /v1/boxes {"network":{"inbound":{"mode":"enabled","allow_net":["10.0.0.0/8"]}}}
  -> build_box_options()                    [commands/serve/mod.rs]
  -> NetworkSpec::try_from(InboundNetworkConfig)  [runtime/options.rs]
       Err(BoxliteError::Config("inbound.allow_net is not supported yet..."))
  -> error_from_boxlite() -> err.http()     [shared/src/errors.rs:190]
       BoxliteError::Config(_) => (500, "ConfigError", "config_error")  ← BUG
  -> client sees: 500 Internal Server Error, for their own mistake
```

`Config` is used for two unrelated things: genuine caller-input mistakes (bad
`BoxOptions`) and real system/environment faults (jailer/sandbox setup, image
store). Mixing them meant every caller-input validation error in `options.rs`
— not just inbound.allow_net — surfaced as a 500 over `boxlite serve`.

## After

```
POST /v1/boxes {"network":{"inbound":{"mode":"enabled","allow_net":["10.0.0.0/8"]}}}
  -> NetworkSpec::try_from(InboundNetworkConfig)
       Err(BoxliteError::InvalidArgument("inbound.allow_net is not supported yet..."))
  -> err.http() -> (400, "InvalidArgumentError", "invalid_argument")
  -> client sees: 400 Bad Request with the same message
```

Retargets the ~12 `BoxOptions`/`NetworkSpec`/`PortSpec` validation call sites in
`options.rs` (inbound.allow_net, network mode parsing, ports-without-network,
remove-on-stop+detach, malformed volume mounts, invalid port host_ip) from
`Config` to `InvalidArgument`. Leaves every other `Config` use (jailer, image
store, custom kernel — genuine environment problems) untouched — that variant
still correctly maps to 500 there.

The REST client already reconstructs `InvalidArgument` symmetrically from the
wire (`rest/error.rs`'s `"invalid_argument"` dispatch arm predates this change),
so no client-side change was needed.

## Test plan

- Strengthened `test_network_config_rejects_unsupported_inbound_allow_net` and
  `test_sanitize_rejects_unsupported_inbound_allow_net` to pin the error variant
  (not just the message) — this is what `BoxliteError::http()` dispatches on
- Two-side verification: reverted only the 12 production call sites (kept the
  test assertions on `InvalidArgument`), both tests failed with
  `expected InvalidArgument (→ HTTP 400), got Config(...)`; restored, both pass
- Fixed `create_rejects_detached_remove_on_stop`, the one test that asserted
  the old `Config` variant for a caller-input error
- `cargo test -p boxlite --lib --features rest`: 1169 passed, 2 failed — both
  pre-existing and unrelated (confirmed identical failures on the unmodified
  tree): a missing vendored libkrun submodule in this sandbox, and a flaky
  watchdog-timing test
- `cargo check --workspace --features rest --tests --exclude boxlite-guest`: clean
  (`boxlite-guest` is Linux-only, unbuildable on this host regardless of this change)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation error handling for box, volume, network, and port configuration.
  * Invalid caller-provided settings are now consistently reported as invalid arguments, corresponding to HTTP 400 responses instead of server errors.
  * Added coverage for invalid network configuration scenarios.

### DIFF
--- a/src/boxlite/src/runtime/core.rs
+++ b/src/boxlite/src/runtime/core.rs
@@ -574,9 +574,12 @@ mod tests {
             .await
             .err()
             .expect("detached remove-on-stop must be rejected at create");
+        // POL-356: caller-input validation errors map to InvalidArgument
+        // (→ HTTP 400 over boxlite serve), not Config (→ 500) — this is the
+        // caller's mistake, not a server/environment fault.
         assert!(
-            matches!(err, BoxliteError::Config(_)),
-            "expected a Config error, got: {err:?}"
+            matches!(err, BoxliteError::InvalidArgument(_)),
+            "expected an InvalidArgument error, got: {err:?}"
         );
         assert!(
             err.to_string().contains("remove-on-stop is incompatible"),
--- a/src/boxlite/src/runtime/options.rs
+++ b/src/boxlite/src/runtime/options.rs
@@ -762,25 +762,25 @@ impl VolumeSpec {
     pub fn validate(&self) -> BoxliteResult<()> {
         let guest_path = &self.guest_path;
         match &self.managed_volume {
-            Some(volume) if !self.host_path.is_empty() => {
-                Err(boxlite_shared::errors::BoxliteError::InvalidArgument(format!(
+            Some(volume) if !self.host_path.is_empty() => Err(
+                boxlite_shared::errors::BoxliteError::InvalidArgument(format!(
                     "volume mount {guest_path:?} sets both managed_volume ({volume:?}) and \
                      host_path; use exactly one"
-                )))
-            }
-            Some(volume) if volume.trim().is_empty() => {
-                Err(boxlite_shared::errors::BoxliteError::InvalidArgument(format!(
+                )),
+            ),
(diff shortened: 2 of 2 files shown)


DefeatMan (approved):
LGTM

PR #1396, opened 2026-08-28
fix(rest): send legacy flat network shape when inbound is at default

G4614:
## Summary

Fixes backward compatibility for self-hosted users who upgrade their SDK/CLI before upgrading their server.

Old servers (pre-#1199) set `forbidNonWhitelisted: true` in NestJS, so the new nested `{outbound, inbound}` wire shape produces a 400 even when the user never configures inbound — the unknown `inbound` key is enough to reject the request.

Serialize as the legacy flat shape `{mode, allow_net}` when inbound is at its default (`Enabled`, empty allowlist). Emit the nested shape only when inbound is explicitly configured, which requires a server with #1199+ support anyway.

## Before/after

```
(before)
BoxOptions { network: Enabled, inbound_network: Enabled }   ← default
  -> POST /boxes { network: { outbound: {mode,allow_net}, inbound: {mode} } }
     ← 400 on old server (forbidNonWhitelisted rejects outbound/inbound keys)

(after)
BoxOptions { network: Enabled, inbound_network: Enabled }   ← default
  -> POST /boxes { network: { mode, allow_net } }           ← legacy flat shape
     ← 200 on any server

BoxOptions { ..., inbound_network: Disabled }               ← explicit inbound
  -> POST /boxes { network: { outbound: {...}, inbound: {...} } }
     ← requires #1199+ server
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network configuration compatibility by preserving the legacy format when default inbound networking settings are used.
  * Added support for the newer nested outbound and inbound network configuration format when inbound settings are explicitly customized.
  * Improved reliability by ensuring network settings are serialized in the correct format for each configuration.

### DIFF
--- a/src/boxlite/src/rest/types.rs
+++ b/src/boxlite/src/rest/types.rs
@@ -266,7 +266,11 @@ impl serde::Serialize for CreateBoxNetworkSpec {
     fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
         use serde::ser::SerializeMap;
         if self.legacy {
-            let len = if self.outbound.allow_net.is_empty() { 1 } else { 2 };
+            let len = if self.outbound.allow_net.is_empty() {
+                1
+            } else {
+                2
+            };
             let mut map = serializer.serialize_map(Some(len))?;
             map.serialize_entry("mode", &self.outbound.mode)?;
             if !self.outbound.allow_net.is_empty() {
--- a/src/boxlite/src/rest/types.rs
+++ b/src/boxlite/src/rest/types.rs
@@ -250,40 +250,30 @@ impl From<&crate::runtime::options::VolumeSpec> for CreateBoxVolumeSpec {
 
 /// Wire shape sent to the server when creating a box.
 ///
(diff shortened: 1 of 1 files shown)


src/boxlite/src/rest/types.rs
```diff
@@ -248,10 +248,38 @@ impl From<&crate::runtime::options::VolumeSpec> for CreateBoxVolumeSpec {
...
+/// Serializes as the legacy flat shape `{"mode","allow_net"}` when inbound is
+/// at its default (enabled, no allowlist), so servers that predate the
+/// inbound/outbound split (#1199) keep working. Serializes as the nested shape
+/// `{"outbound","inbound"}` when inbound is explicitly configured — that
+/// requires a server with #1199+ support.
+#[derive(Debug)]
 pub(crate) struct CreateBoxNetworkSpec {
     pub outbound: CreateBoxOutboundNetworkSpec,
     pub inbound: CreateBoxInboundNetworkSpec,
+    legacy: bool,
```
DorianZheng: what for?

DorianZheng (changes requested):

DorianZheng (approved):

PR #1403, opened 2026-08-28
fix(rest): verify advanced policy before box reuse

### LINKED ISSUE #1348: RestRuntime silently drops privileged mode; get_or_create skips capability-compatibility check over REST
Found while auditing PR #1296 (privileged-capabilities-decouple). Pre-existing on `main`, not introduced or fixed by that PR — confirmed via `git show --stat` that the affected code is byte-for-byte unchanged by #1296's diff.

## 1. `RestRuntime::create`/`get_or_create` silently drop `privileged`

`reject_remote_experimental_options` (`src/boxlite/src/rest/runtime.rs`) rejects `advanced.kernel` and `advanced.nested_virtualization` for remote runtimes with a clear `Unsupported` error, but never checks `advanced.privileged`. `CreateBoxRequest::from_options` builds the wire request purely from `advanced.capabilities()`; there is no `privileged` field anywhere in the client-side or server-side wire types (`CreateBoxRequest`/`CreateBoxAdvancedOptions` in both `src/boxlite/src/rest/types.rs` and `src/cli/src/commands/serve/types.rs`).

Repro: a Rust caller building `BoxOptions { advanced: AdvancedBoxOptions { privileged: true, .. }, .. }` and calling `RestRuntime::create`/`get_or_create` gets back `Ok(LiteBox)` — no error — and the server creates an ordinary, non-privileged box. Confirmed end-to-end against a real listener; captured wire request has no `advanced` key at all, let alone `privileged`.

Contrast: the same call with `nested_virtualization = true` against a dead port fails locally before any network attempt, exactly as designed. `privileged` gets no such guard.

## 2. `RestRuntime::get_or_create` performs zero compatibility check on reuse

Unlike the local `RuntimeImpl::create_inner`/`adopt_existing_box`, which calls `check_options_compatibility` (comparing `nested_virtualization`, `privileged` both directions, and effective capabilities) before reusing a box, `RestRuntime::get_or_create` does a plain `GET` by name and returns whatever it finds, with zero comparison against the request's `advanced` options.

This is structurally worse than a simple omission: `BoxResponse` (the `GET` response wire type) carries no `capabilities`/`privileged`/`advanced` field at all, so the comparison can't even be added at the client without a wire-protocol change first.

## Suggested fix

- Add `privileged` to the REST wire types (client + server), and either reject it in `reject_remote_experimental_options` (matching the local-only pattern for `kernel`/`nested_virtualization`) or wire it through properly if remote privileged boxes are meant to be supported.
- Extend `BoxResponse` to carry back enough of `advanced` (or add a dedicated inspect field) for `RestRuntime::get_or_create` to run an equivalent `check_options_compatibility` before reusing.

Found and reproduced (including full repro command output) by an automated correctness-review pass over PR #1296; not yet fixed anywhere.

ZenAlexa:
## Summary
Completes the named-box reuse half of #1348 after #1354 addressed privileged REST creation. `RestRuntime::get_or_create` now verifies the requested effective advanced policy against inspect-time metadata before adopting an existing box.

## Call graph

```text
Before
  get_or_create          (RestRuntime · src/boxlite/src/rest/runtime.rs:259) — fetches an existing named box
    └─ BoxResponse       (REST types · src/boxlite/src/rest/types.rs:351)    ← BUG: omits the policy needed for compatibility checks
         └─ LiteBox      (RestBox · src/boxlite/src/rest/runtime.rs:43)      — returned without the local runtime's reuse check

After
  box_info_to_response / box_info_to_dict / boxToBoxResponse               — report the effective advanced policy
    └─ BoxResponse       (REST types · src/boxlite/src/rest/types.rs:351)    — preserves optional policy metadata
         └─ AdvancedBoxInfo (core · src/boxlite/src/runtime/advanced_options.rs:592) — rejects incomplete response policies
              └─ get_or_create (RestRuntime · src/boxlite/src/rest/runtime.rs:259)
                   └─ adopt_existing_box (RestRuntime · src/boxlite/src/rest/runtime.rs:49)
                        └─ check_reuse_compatibility (AdvancedBoxOptions · src/boxlite/src/runtime/advanced_options.rs:924) — reuses only a compatible box
```

Fixes #1348

## Changes
- Adds inspect-time `AdvancedBoxInfo` metadata for effective capabilities, privileged mode, and nested virtualization; the CLI server, reference server, and hosted API produce the same response shape.
- Shares one compatibility check between local and REST reuse, including rejection of accidental privilege upgrades.
- Recovers the named-box lookup/create race with one re-read and validates the concurrent winner; when no winner exists, the original create error is preserved.
- Requires complete `add` and `drop` arrays in response metadata at both the OpenAPI and Rust decoder boundaries. Create-side capability defaults remain unchanged.
- Exposes the metadata through Python because the reference server serializes `BoxInfo` through that binding. C, Go, and Node read-side APIs remain unchanged.

## Validation

On `6c2c7797`, integrated with `main@1f68ee36`:

- Rust core and REST-feature unit executable: 1,218 passed; the WebSocket idle-watchdog timing test was excluded after reproducing its timeout twice. Its implementation, test, and dependency lockfile match current upstream. Running the exact test from current upstream reproduced the same timeout.
- Named-box reuse tests: 5 passed in the core suite and 15 passed in the REST suite.
- CLI serve tests: 65 passed. OpenAPI tests: 36 passed.
- Python SDK with the compiled native extension: 146 passed, 190 integration cases deselected. Rust binding tests: 13 passed.
- Node SDK with the compiled native extension: 98 passed.
- Restoring the pinned libkrun source resolved the missing-vendor fixture failure; its regression passed and is included in the complete Rust run.
- Focused formatting and diff checks passed.

The earlier API mapper run passed 13 tests. The current hosted workflows await repository approval.

## Rollout
Servers predating this response field remain inspectable. A REST client rejects named-box reuse until the server reports enough metadata to verify the policy. Upgrading the server restores reuse without recreating compatible boxes.


ZenAlexa:
The current head is based on `main`, closes both review threads, and passes CLA plus the repository review checks. The API drift, lint/format, test, and CodeQL workflows are `action_required`; please approve the current fork runs so full CI can start.

ZenAlexa:
I've added the class-level docs for inbound and outbound policy in both SDKs, including the allowlist behavior. Python's advanced-policy docs now explain the effective capabilities and creation modes.

I also switched the serve test to its existing fixture so it accepts the new `BoxInfo` field on main. Could a maintainer approve the workflows for this update?

ZenAlexa:
I integrated the current main branch and preserved the effective-policy checks across named-box reuse and the concurrent-create recovery path.

I also exercised the compiled Python and Node bindings; their SDK suites pass. The PR description now records the current validation and the isolated WebSocket timing timeout.

PR #1407, opened 2026-08-30
feat(lifecycle)!: enforce AutoStop/AutoDelete in serve, add CLI flags

DorianZheng:
The `auto_stop` / `auto_delete` / `auto_resume` fields already existed in the wire
contract, in `BoxOptions` and in every SDK, but nothing below the cloud control plane
ever read them back. `boxlite serve` now enforces them and the CLI can reach them.

Two deletion axes had been sharing one integer:

- `auto_remove` (bool) removes the box synchronously when it stops (`--rm`).
- `auto_delete` (seconds) is a deferred deadline a sweeper acts on later.

While fused, `auto_delete=1` was spent as the `--rm` sentinel, any deadline forced the
box non-detached, and `recover_boxes` destroyed pending deadlines at startup — so
"delete N seconds after stop" could not be expressed at all. The C ABI fused them too:
setting `auto_remove` cleared a caller's deadline.

## Call graph

Before

```
create_box                          (serve · src/cli/src/commands/serve/handlers/boxes.rs:16)  — accepts auto_stop/auto_delete
  └─ build_box_options              (serve · src/cli/src/commands/serve/mod.rs:789)            — copies the fields through
       └─ removes_on_stop           (BoxOptions · src/boxlite/src/runtime/options.rs:568)      ← BUG: auto_delete overrode auto_remove, so Some(1) was the --rm sentinel and no real deadline could coexist
  └─ reaper_loop                    (serve · src/cli/src/commands/serve/mod.rs:623)            ← BUG: reaped exited processes only; never consulted either deadline
recover_boxes                       (RuntimeImpl · src/boxlite/src/runtime/rt_impl.rs:1277)    ← BUG: destroyed pending deadlines at startup
```

After

```
create_box                          (serve · src/cli/src/commands/serve/handlers/boxes.rs:16)  — unchanged entry
  └─ build_box_options              (serve · src/cli/src/commands/serve/mod.rs:789)            — sets auto_remove:false, always detached
       └─ reject_local_unsupported_options (RuntimeImpl · src/boxlite/src/runtime/rt_impl.rs:1744) — refuses a deadline no sweeper will act on
            └─ with_lifecycle_sweeper (RuntimeBuilder · src/boxlite/src/experimental.rs:160)   — serve declares one; the embedded runtime does not
record_activity                     (middleware · src/cli/src/commands/serve/mod.rs:1141)      — stamps the per-box idle clock
  └─ record_box_activity            (AppState · src/cli/src/commands/serve/mod.rs:101)         — no-ops when this serve does not sweep
reaper_loop                         (serve · src/cli/src/commands/serve/mod.rs:623)            — same 30s tick, no new persisted state
  └─ run_lifecycle_once             (serve · src/cli/src/commands/serve/mod.rs:1200)           — returns false when another server owns the boxes
       └─ idle_or_seed              (serve · src/cli/src/commands/serve/mod.rs:1171)           — newest clock across {id, name}; an unseen box is seeded now
       └─ decide_lifecycle          (serve · src/cli/src/commands/serve/mod.rs:1076)           — pure: Stop | Delete | Leave
            └─ is_at_rest           (BoxStatus · src/boxlite/src/litebox/state.rs:73)          — one at-rest rule, shared with ls
       └─ retain_box_activity       (AppState · src/cli/src/commands/serve/mod.rs:130)         — drops clocks no box answers to
autoresume_allows                   (serve · src/cli/src/commands/serve/mod.rs:1373)           — pure gate; 409 only on implicit wake
SplitFusedDeletionAxes              (Migration · src/boxlite/src/db/migration/v10_to_v11.rs:44) — rewrites persisted boxes once
options_from_manifest               (import · src/boxlite/src/runtime/import.rs:76)            — same rewrite for archives below v6
  └─ archive_version_for_options    (archive · src/boxlite/src/litebox/archive.rs:78)          — stamps v6 when a pre-split importer would misread
```

**Key:** `serve` is the sole enforcer, and its exclusive `flock` on `$BOXLITE_HOME/.lock`
is what makes that sound — no second process can hold boxes it cannot see, so its
activity view is total. Idle comes only from the in-memory clock, so a restart grants
every box a full window instead of reading it as idle since boot. A `serve` that itself
proxies to another server leaves that server's boxes alone.

`AutoResume` gates only the operations that would wake a box implicitly — exec, files,
attach, metrics — returning 409 when disabled. Start, stop, snapshots, clone and export
resolve through an ungated path: they never need the VM up, and gating them answered a
metadata read with advice to start a box the caller had not asked to run.

CLI `run`/`create` gain `--auto-stop`, `--auto-delete` and `--no-auto-resume`, accepting
seconds or a suffixed duration, and `ls` shows both deadlines.

## Migrating the fused shape

The old meaning is stored in two places, and both are translated once, while the shape
can still be read as what it meant:

| Store | Mechanism |
|---|---|
| `box_config` (SQLite) | migration v10 → v11, `SplitFusedDeletionAxes` |
| archive manifest | `options_from_manifest` rewrite below archive **v6** |

Without the first, `recover_boxes` would delete every pre-upgrade box that was not
created with `--rm`, on the first command after upgrading — they carry `auto_delete:0`
beside the `auto_remove:true` default the old CLI never wrote. Without the second, an
archive of any created box becomes unimportable, because its `auto_remove:true` now
contradicts the `detach:true` beside it. An export carrying a deadline a pre-split
importer would misread is stamped v6 so that importer refuses it rather than acting on
the wrong policy.

## Breaking change

On the embedded runtime, `auto_delete>0` previously meant "remove the box the moment it
stops" and now means a deferred deadline, which that runtime refuses with `Unsupported`
because it runs no sweeper. `auto_delete=Some(0)` also no longer suppresses removal:
`auto_delete` used to take precedence over `auto_remove`, so `Some(0)` overrode it and
meant "keep the box after stop", while `removes_on_stop` now reads `auto_remove` alone,
which defaults to true. Callers wanting the old behaviour set `auto_remove` explicitly.

## Testing

| Suite | Result |
|---|---|
| Engine unit (`test:unit:rust`) | 1097/1097 |
| Shared | 57/57 |
| CLI unit | 230/230 |
| Reference server | 36/36 |
| clippy / fmt (`lint:fix`) | clean |

`lint:fix` covers the Rust and SDK sources; nothing lints `openapi/reference-server`
(ruff is scoped to `sdks/python`), so it is not evidence about that part of the diff.

The reference-server suite is not discovered by any make target (`make/test.mk` globs
`test_error*.py`, deliberately — nothing installs fastapi/pydantic), so it never runs in
CI. Concretely: an earlier commit on this branch had CI green while that suite was red on
a stale `BoxOptions(...)` expectation. It does run locally once its PEP 723 deps are
supplied:

```
uv run --with fastapi --with sse-starlette --with PyJWT --with python-multipart \
       --with python-dotenv --with uvicorn \
       python -m unittest discover -s openapi/reference-server/tests -p 'test_*.py'
```

**Never executed, by CI or locally:** `src/boxlite/tests/rest_integration.rs`, which this
PR modifies — it needs a hand-started reference server, and `BOXLITE_REST_URL` appears
nowhere in `.github/` or `make/`. Also unrun locally: the SDK binding suites and
`sdks/c/src/tests.rs`, which CI does cover.

The local pre-push matrix was skipped for this branch (`SKIP=full-test-matrix`) because
two tests fail for reasons unrelated to this diff, neither file being among the 54
changed:

- `ws_watchdog_fires_when_idle` (`src/boxlite/src/rest/litebox.rs:1710`) reproduces
  identically at the clean parent commit `446e9aece` with none of this diff applied
  (84 passed / 1 failed, 6.74s there vs 6.82s here), and still fails with all proxy
  environment variables cleared.
- `test_pull_nonexistent` (`src/cli/tests/pull.rs:31`) needs 18.4s of its own 30s budget
  when run alone and is killed mid-flight under the full 417-test parallel run.

## Known follow-ups

- Import bypasses the sweeper gate (`options_from_manifest` calls `sanitize()`, not
  `sanitize_local_options`), so an archived deadline can still reach a sweeperless
  runtime — the one hole in this PR's own "creating a box accepts a deadline only when
  the embedder declares a sweeper". Not closed here because `import.rs` deliberately
  *resolves* the fused legacy shape rather than rejecting it, so a refusal on that path
  needs its own design pass.
- Nothing installs the reference server's test dependencies, so
  `openapi/reference-server/tests/` is permanently CI-uncovered. That is how a stale
  `BoxOptions(...)` expectation in it went unnoticed until this PR. The fix is a setup
  target that installs the deps, not widening the discovery glob.
- `remove_box` / `stop_box` evict `state.boxes` by URL segment while every insert keys by
  canonical id, so a delete-by-name leaves the handle cached. Pre-existing on main.
- The idle clock reconciles the id and name spellings only; a unique id prefix or a
  percent-encoded name still stamps a key the sweep does not read. Out of contract —
  `box.openapi.yaml` tells clients to pass the identifier back verbatim.
- `openapi/box.openapi.yaml` documents `auto_stop` default 900 and `detach` default
  false; `serve` reads an omitted `auto_stop` as disabled and always detaches. Needs a
  product call rather than a patch.

### DIFF
--- a/.github/workflows/e2e-cloud.yml
+++ b/.github/workflows/e2e-cloud.yml
@@ -122,7 +122,7 @@ permissions:
 # A later run queues behind the in-progress one instead of cancelling it.
 # The `box` fixture's teardown is the only thing that deletes the boxes a
 # run creates: auto_remove is a no-op over REST (see
-# deprecated_auto_remove_does_not_change_rest_lifecycle_defaults) and the
+# auto_remove_is_never_transmitted_to_a_rest_runtime) and the
 # API defaults autoDelete to disabled, so a run killed mid-suite strands
 # every box it had open — they auto-stop after 15 min but are never
 # deleted. Note this bounds the damage rather than eliminating it:
--- a/docs/architecture/auto-stop-resume-design.md
+++ b/docs/architecture/auto-stop-resume-design.md
@@ -16,7 +16,9 @@ auto_delete: integer seconds, 0 disables
 auto_resume: boolean, default true; user operations resume an auto-stopped box when enabled
 ```
 
-The default `auto_stop` is `900` seconds, the default `auto_delete` is `0`, and the default `auto_resume` is `true`. Create and read APIs and the Rust, Python, Node.js, C, and Go SDK boundaries use these same modern lifecycle semantics. AutoResume is implemented by the cloud control plane; the embedded local runtime has no auto-stopped state to resume. SDKs continue to accept deprecated `auto_remove` for embedded remove-on-stop compatibility, and explicit `auto_delete` takes precedence there. The REST API does not expose `auto_remove`; leaving `auto_delete` unset preserves the remote server's default instead of translating the deprecated field to a timer.
+The default `auto_stop` is `900` seconds, the default `auto_delete` is `0`, and the default `auto_resume` is `true`. Create and read APIs and the Rust, Python, Node.js, C, and Go SDK boundaries use these same modern lifecycle semantics.
+
+`auto_remove` is a separate axis, not a deprecated spelling of `auto_delete`: it removes the box synchronously the moment it stops (docker `run --rm`) and needs no sweeper, while `auto_delete` is a deferred deadline something has to act on later. Setting one never clears the other, and the two are mutually exclusive at the CLI. The REST API does not expose `auto_remove`; leaving `auto_delete` unset preserves the remote server's default.
 
 The internal database columns are `autoStop`, `autoDelete`, `autoResume`, and `lastActivityAt`. Public names remain stable.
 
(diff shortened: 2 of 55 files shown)


DorianZheng:
Reverted at the author's request; the lifecycle work is not landing in this form.

PR #1413, opened 2026-08-31
feat(proxy): replace tunnel-live polling with WebSocket keepalive

G4614:
## Summary

Replace the 5s Redis TTL renewal loop with a persistent WebSocket connection for tunnel liveness.

**Before:** CLI polls `POST /network/tunnel` every 5s to renew a 15s Redis TTL. Browser access lingers up to 18s after CLI exits.

**After:** CLI holds a WebSocket to `/api/v1/boxes/{boxId}/network/tunnel/live`. API sets Redis key (no TTL) on connect, deletes it immediately on disconnect. Browser access lapses as soon as the TCP connection closes.

## Call graph

**Before**
```
CLI loop (5s)
  → POST /network/tunnel          BoxliteProxyController.proxyNetworkTunnel
    → BoxService.openNetworkTunnel  redis.setex(key, 15s)
Go proxy (per request)
  → NetworkTunnelChecker.IsLive   redis.EXISTS(key)
```

**After**
```
CLI
  → WS /api/v1/boxes/{id}/network/tunnel/live   TunnelLiveWsService.upgrade
      onConnect → redis.SET(key)                (no TTL)
      onClose   → redis.DEL(key)               (immediate)
Go proxy (per request)
  → NetworkTunnelChecker.IsLive   redis.EXISTS(key)   (unchanged)
```

## Test plan

- [ ] e2e: open tunnel, verify browser access, Ctrl-C, verify 403 within ~1s
- [ ] e2e: network drop (kill WS), verify Redis key deleted, browser gets 403
- Unit tests: 49/49 pass

### DIFF
--- a/.zcode/plans/plan-sess_790c53a9-7b87-4786-a753-5694e5e1b3c4.md
+++ b/.zcode/plans/plan-sess_790c53a9-7b87-4786-a753-5694e5e1b3c4.md
@@ -1,0 +1,12 @@
+继续对 PR #1360 的实测验证（前期准备已完成：worktree 在 /tmp/pr1360，依赖已软链就位）：
+
+1. **跑 PR 自带测试**：`./node_modules/.bin/jest --config api/jest.config.ts --rootDir api`（API 侧）和 vitest（dashboard 侧），核实 PR 描述的通过数（608 + 158）。
+2. **写探针测试验证四个疑点**（写在 worktree 内）：
+   - `effectivePermissions`：`role: ADMIN` + `apiKey` 并存时返回 null（提权验证）
+   - `RestCreateApiKeyDto`：`expires_at: "not-a-date"` 经真实 ValidationPipe 是 400 还是穿透
+   - `BoxliteApiKeyController.remove()`：窄 key 吊销宽 key 是否无检查
+   - `CreateApiKeyDialog`：`availablePermissions` 为空时 payload 是否为 `permissions: []`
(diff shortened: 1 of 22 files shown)


G4614:
Closing: this builds on the liveness-lease model that was dropped with #1391, and merging it would now break a shipped behaviour rather than just carry a dead mechanism.

**Why the premise is gone.** The lease makes reachability mean "a client process is currently holding it open" — this PR makes that stricter, not looser, by tying it to a live TCP connection instead of a 15s TTL. The threat it was introduced for does not hold: the preview hostname is `{port}-d-{hex(box uuid)}`, 128 bits, and `box.public` already defaults to false on `main`, so an exposed box is an explicit opt-in. A leaked URL is leaked while the lease is live, so a lease does not address the real risk. Credentialed access for private boxes already exists in `apps/proxy/pkg/proxy/auth.go`.

**Why it now actively conflicts.** `getBoxPublic` treating a lease-less box as private runs *before* the resume added in #1467, so a request arriving at a preview URL would be rejected as private before it could wake a stopped box. That lands the opposite of what auto_resume promises.

**Also:** the branch is CONFLICTING against `main` (opened Aug 31; #1431 has since changed the same handler), and it carries `.zcode/plans/*.md`, `target` and `Cargo.lock` by accident.

**Worth rescuing separately, not lost on purpose:** the five POL-257 e2e cases here (`test_pol257_admin_fail_closed`, `cross_org_isolation`, `network_disabled`, `public_default`, `secrets_rest_path`) are unrelated to tunnel liveness, are not on `main`, and are not carried by any other open PR. `test_volume_crud_and_mount.py` is already covered by #1232. The branch stays, so these can be cherry-picked.

PR #1417, opened 2026-08-31
feat(api): enforce subscription limits on hosted box creation

mamamiyear:
## Call graph

```text
Before
createBox (BoxliteBoxController · apps/api/src/boxlite-rest/boxlite-box.controller.ts:112)
└─ create (BoxService · apps/api/src/box/services/box.service.ts:187)
   ├─ fresh Box → select Runner → insert (BoxRepository:77)
   │  └─ persist without organization subscription admission
   └─ warm-pool Box → select candidate → update ownership (BoxRepository:118)
      └─ persist without organization subscription admission

After
createBox (BoxliteBoxController · apps/api/src/boxlite-rest/boxlite-box.controller.ts:112)
├─ resolveMaxCreatedBoxes (CommerceBoxLimitService:52)
│  ├─ BILLING_API_URL unset → unlimited
│  ├─ inFlightResolutions → coalesce same-key work within this Node process
│  ├─ Redis GET commerce:box-limit:v1:<organizationId>
│  │  └─ hit → return shared plan resolution cached for 30 seconds
│  └─ miss → fetchMaxCreatedBoxes (CommerceBoxLimitService:135)
│     ├─ Promise.all
│     │  ├─ GET /plan
│     │  └─ GET /organization/{organizationId}/plan
│     ├─ valid matched/default plan → Redis SET EX 30 → return limit or unlimited
│     ├─ unmatched active plan, timeout, transport error, or 5xx
│     │  └─ warning → unlimited for this request → do not cache
│     └─ 4xx or malformed successful response → 503 upstream_unavailable
└─ create (..., { maxCreatedBoxes }) (BoxService:187)
   ├─ existing validation and resource preparation
   ├─ fresh Box → select Runner → insert (BoxRepository:77)
   ├─ warm-pool Box → select candidate → update ownership (BoxRepository:118)
   └─ both persistence paths → withCreationAdmission (BoxRepository:169)
      └─ PostgreSQL SERIALIZABLE transaction
         ├─ COUNT organization Boxes, excluding ERROR, DESTROYING, DESTROYED,
         │  ARCHIVING, and ARCHIVED (all UNKNOWN Boxes count)
         ├─ count >= limit → rollback → 429 resource_exhausted
         ├─ count < limit → INSERT or UPDATE plus last activity → commit
         └─ SQLSTATE 40001 → retry complete decision, at most 5 attempts
            └─ exhausted → 503 upstream_unavailable + Retry-After: 5
```

## Summary

Enforce the Commerce subscription limit on hosted CREATE BOX requests used by
Dashboard, CLI REST, and SDK REST clients. The Commerce result is resolved before
creation, while PostgreSQL remains the source of truth for the counted Box total
and makes the final concurrent admission decision.

Successful per-organization plan resolutions are shared through Redis for 30
seconds. Concurrent misses are coalesced within each Node process. Redis stores
neither Box counts nor quota reservations and is not an admission lock.

Local SDK creation, START, RECOVER, and warm-pool provisioning remain outside
this change.

Supersedes #1400

Refs boxlite-ai/boxlite-commerce#68

## Changes

- Resolve the effective limit from Commerce with parallel, two-second
  `GET /plan` and `GET /organization/{organizationId}/plan` requests using
  `BILLING_API_URL` and bearer `USAGE_EXPORT_TOKEN`.
- Validate the Commerce URL and shared token at API startup when
  `BILLING_API_URL` is configured. An unset URL preserves unlimited
  self-hosted and private-deployment behavior.
- Use the first ordered public plan for unsubscribed or suspended organizations,
  match active public plan ids, and treat `concurrencyLimit: null` as a real
  unlimited result.
- Cache successful finite and unlimited resolutions in shared Redis for 30
  seconds, and coalesce same-key work with a process-local in-flight promise.
  Redis read or write failures do not block the request.
- On a timeout, transport failure, or Commerce 5xx, log a warning and allow the
  current request without caching that fallback. Apply the same warning,
  unlimited, and no-cache behavior when an active plan id is absent from the
  public catalog, so the next request rechecks Commerce.
- Return `503 upstream_unavailable` for Commerce 4xx responses or malformed
  successful responses.
- Pass the optional limit only through hosted CREATE. For both fresh inserts and
  warm-pool ownership updates, count and persist in one PostgreSQL
  `SERIALIZABLE` transaction.
- Count the organization non-finalized Box total while excluding `ERROR`,
  `DESTROYING`, `DESTROYED`, `ARCHIVING`, and `ARCHIVED`. `UNKNOWN`, `STOPPED`,
  and every other state remain counted.
- Return `429 resource_exhausted` when `count >= limit`. Retry only SQLSTATE
  `40001` by rerunning the complete transaction up to five times; exhaustion
  returns `503 upstream_unavailable` with `Retry-After: 5`.
- Keep lookup-cache invalidation and events after commit, extend portable
  OpenAPI responses, document the Commerce contract, and exercise the current
  Commerce controllers through a cross-repository contract test.

## How to verify

- `make test:apps FILTER="creation limit|CommerceBoxLimit"`
- `make test:apps:infra`
- `make test:unit:openapi`
- `make lint:apps`
- `git diff --check`
- The API CI job provisions Redis 7 and PostgreSQL 16, checks out the current
  boxlite-commerce repository, and runs the API and cross-repository contract
  suites.

## Risks / rollout

- Deploy boxlite-ai/boxlite-commerce#68 before this change and configure matching
  `USAGE_EXPORT_TOKEN` and Commerce `USAGE_INGEST_TOKEN` values.
- A successful Commerce change can take up to 30 seconds to propagate through
  the Redis cache.
- In-flight request coalescing is process-local. Redis shares completed
  resolutions, but separate API instances may refresh the same expired key.
- Temporary Commerce transport or 5xx failures and unmatched active plan ids
  intentionally fail open for the current request and are never cached.
- `SERIALIZABLE` contention can trigger five short complete-transaction
  attempts. Persistent contention returns 503 with a five-second retry hint.
- No database migration, schema or index change, PostgreSQL advisory lock, Redis
  Box-count cache, Redis quota reservation, START or RECOVER restriction,
  Dashboard UI change, or SDK implementation change is included.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -422,6 +422,12 @@ jobs:
         with: # Native workspace dependencies are submodules.
           submodules: "recursive"
           persist-credentials: ${{ false }}
+      - name: Checkout Commerce contract provider
+        uses: actions/checkout@v5
+        with:
+          repository: boxlite-ai/boxlite-commerce
+          path: .contract/boxlite-commerce
+          persist-credentials: ${{ false }}
       - name: Enable Corepack
         run: "corepack enable"
       - uses: actions/setup-node@v4
@@ -431,7 +437,12 @@ jobs:
           cache-dependency-path: "apps/yarn.lock"
       - name: Install app dependencies
         run: "yarn install --immutable"
+      - name: Install Commerce contract dependencies
+        working-directory: .contract/boxlite-commerce
+        run: "npm ci --ignore-scripts --no-audit --no-fund"
       - name: Run API unit tests
+        env:
+          BOXLITE_COMMERCE_REPOSITORY: ${{ github.workspace }}/.contract/boxlite-commerce
         run: "yarn nx run api:test"
 
   # End API test job.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -422,11 +422,19 @@ jobs:
         with: # Native workspace dependencies are submodules.
           submodules: "recursive"
           persist-credentials: ${{ false }}
+      # Fork and Dependabot pull_request runs cannot read repository secrets.
+      # The merge_group run still enforces this contract before merge.
       - name: Checkout Commerce contract provider
+        id: commerce_checkout
+        if: >-
+          github.event_name != 'pull_request' ||
+          (github.event.pull_request.head.repo.full_name == github.repository &&
+          github.actor != 'dependabot[bot]')
         uses: actions/checkout@v5
         with:
           repository: boxlite-ai/boxlite-commerce
           path: .contract/boxlite-commerce
+          token: ${{ secrets.GH_PAT }}
           persist-credentials: ${{ false }}
       - name: Enable Corepack
         run: "corepack enable"
@@ -438,11 +446,12 @@ jobs:
       - name: Install app dependencies
         run: "yarn install --immutable"
       - name: Install Commerce contract dependencies
+        if: steps.commerce_checkout.outcome == 'success'
         working-directory: .contract/boxlite-commerce
         run: "npm ci --ignore-scripts --no-audit --no-fund"
       - name: Run API unit tests
         env:
-          BOXLITE_COMMERCE_REPOSITORY: ${{ github.workspace }}/.contract/boxlite-commerce
+          BOXLITE_COMMERCE_REPOSITORY: ${{ steps.commerce_checkout.outcome == 'success' && format('{0}/.contract/boxlite-commerce', github.workspace) || '' }}
         run: "yarn nx run api:test"
 
   # End API test job.
--- a/apps/API.md
+++ b/apps/API.md
@@ -689,12 +689,11 @@ that widen [organization suspend/unsuspend](#control-plane-api) exist for
 Commerce, which does not currently call them.
 
 Admission applies the plan value to the organization's non-finalized Box total.
-`STOPPED` Boxes and `UNKNOWN` Boxes with `pending=true` count, while `ERROR`,
-settled `UNKNOWN`, `DESTROYING`, `DESTROYED`, `ARCHIVING`, and `ARCHIVED` do
-not. The two admission reads have a cross-repository contract test that runs
-the current Commerce controllers and passes their responses through the
-BoxLite consumer. The remaining billing client is hand-written rather than
-generated, so neither side detects its
+`STOPPED` and `UNKNOWN` Boxes count, while `ERROR`, `DESTROYING`, `DESTROYED`,
+`ARCHIVING`, and `ARCHIVED` do not. The two admission reads have a
+cross-repository contract test that runs the current Commerce controllers and
+passes their responses through the BoxLite consumer. The remaining billing
+client is hand-written rather than generated, so neither side detects its
 divergence. Five of its 20 routes — the billing-email group — have no
 implementation in Commerce, of which only `email/verify` has a live caller here
 ([`EmailVerify.tsx`](./dashboard/src/pages/EmailVerify.tsx)); Commerce in turn
--- a/apps/API.md
+++ b/apps/API.md
@@ -667,8 +667,13 @@ The Billing API, Box admission, and Usage export rows are the same Commerce
 service reached three ways, on URLs that normally differ by path rather than
 host. The browser API and Box admission reads live under `/api/billing`, which
 `BILLING_API_URL` already includes; admission authenticates with the shared
-service token. The internal usage routes are served off the bare origin. The API
-itself requires `USAGE_EXPORT_URL` whenever export is on and never derives it
+service token. The API validates that URL and token presence at startup. A
+request-time transport or 5xx failure is logged and temporarily leaves creation
+unlimited without caching the fallback; rejected credentials and malformed
+successful responses still return 503. Successful per-organization resolutions
+are cached in Redis for 30 seconds. The internal usage routes are served off the
+bare origin. The API itself requires `USAGE_EXPORT_URL` whenever export is on
+and never derives it
 ([`configuration.ts`](./api/src/config/configuration.ts)); it is the SST stack
 that supplies a default of `BILLING_API_URL`'s origin
 ([`api.ts`](./infra/stack/api.ts)), so a deployment outside that stack must set
@@ -682,8 +687,13 @@ plane using the caller's own token. The `billing` API role and `BILLING_API_KEY`
 that widen [organization suspend/unsuspend](#control-plane-api) exist for
 Commerce, which does not currently call them.
 
-Because the billing client is hand-written rather than generated, neither side
-detects divergence. Five of its 20 routes — the billing-email group — have no
+Admission applies the plan value to the organization's non-finalized Box total:
+`STOPPED` Boxes count, while `ERROR`, `UNKNOWN`, `DESTROYING`, `DESTROYED`,
+`ARCHIVING`, and `ARCHIVED` do not. The two admission reads have a
+cross-repository contract test that runs the current Commerce controllers and
+passes their responses through the BoxLite consumer. The remaining billing
+client is hand-written rather than generated, so neither side detects its
+divergence. Five of its 20 routes — the billing-email group — have no
 implementation in Commerce, of which only `email/verify` has a live caller here
 ([`EmailVerify.tsx`](./dashboard/src/pages/EmailVerify.tsx)); Commerce in turn
 serves card default/delete, `plan/pay-open-charge`, and a credit admission API
--- a/apps/API.md
+++ b/apps/API.md
@@ -688,12 +688,13 @@ plane using the caller's own token. The `billing` API role and `BILLING_API_KEY`
 that widen [organization suspend/unsuspend](#control-plane-api) exist for
 Commerce, which does not currently call them.
 
-Admission applies the plan value to the organization's non-finalized Box total:
-`STOPPED` Boxes count, while `ERROR`, `UNKNOWN`, `DESTROYING`, `DESTROYED`,
-`ARCHIVING`, and `ARCHIVED` do not. The two admission reads have a
-cross-repository contract test that runs the current Commerce controllers and
-passes their responses through the BoxLite consumer. The remaining billing
-client is hand-written rather than generated, so neither side detects its
+Admission applies the plan value to the organization's non-finalized Box total.
+`STOPPED` Boxes and `UNKNOWN` Boxes with `pending=true` count, while `ERROR`,
+settled `UNKNOWN`, `DESTROYING`, `DESTROYED`, `ARCHIVING`, and `ARCHIVED` do
+not. The two admission reads have a cross-repository contract test that runs
+the current Commerce controllers and passes their responses through the
+BoxLite consumer. The remaining billing client is hand-written rather than
+generated, so neither side detects its
 divergence. Five of its 20 routes — the billing-email group — have no
 implementation in Commerce, of which only `email/verify` has a live caller here
 ([`EmailVerify.tsx`](./dashboard/src/pages/EmailVerify.tsx)); Commerce in turn
--- a/apps/API.md
+++ b/apps/API.md
@@ -645,12 +645,13 @@ their base URL from `GET /api/config` and disable the corresponding UI when it
 is unset; the same document carries PostHog's key and host to the browser.
 
 <details>
-<summary><b>External interfaces</b> · 3 APIs, 2 SaaS integrations</summary>
+<summary><b>External interfaces</b> · 4 APIs, 2 SaaS integrations</summary>
 
 | Interface     | Client                                                                                 | Base URL source                  | What it covers                                                                                                                                                                                             |
 | ------------- | -------------------------------------------------------------------------------------- | -------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | Analytics API | [`libs/analytics-api-client`](./libs/analytics-api-client/), generated                 | `analyticsApiUrl`                | 8 routes: per-box and organization usage, aggregates, and charts, plus box logs, metrics, and traces.                                                                                                      |
 | Billing API   | [`billingApiClient.ts`](./dashboard/src/billing-api/billingApiClient.ts), hand-written | `billingApiUrl`                  | 20 routes: wallet and top-ups, plans, payment methods, invoices, usage series and prices, coupons, billing emails, portal and checkout URLs.                                                               |
+| Box admission | [`commerce-box-limit.service.ts`](./api/src/boxlite-rest/commerce-box-limit.service.ts) | `billingApiUrl`                  | 2 server-side reads on Commerce: `GET /plan` and `GET /organization/{organizationId}/plan`, bearer `USAGE_EXPORT_TOKEN`, used by hosted CREATE BOX.                                                        |
 | Usage export  | [`api/src/usage/services`](./api/src/usage/services/), hand-written                    | `usageExport.url`                | 2 routes on the same Commerce service, server-side: `POST /internal/usage-events` and `POST /internal/allocation-snapshot`, bearer `USAGE_EXPORT_TOKEN`, which must equal Commerce's `USAGE_INGEST_TOKEN`. |
 | PostHog       | `posthog-js` in the dashboard; `posthog-node` in the API                               | `posthog.apiKey`, `posthog.host` | Browser product analytics; server-side `api_*` operation events from the global metrics interceptor, two `groupIdentify` calls, and feature-flag evaluation.                                               |
 | Pylon         | [`App.tsx`](./dashboard/src/App.tsx) widget, production builds only                    | `pylonAppId`                     | Outbound support-chat widget; it registers no route here.                                                                                                                                                  |
@@ -662,12 +663,12 @@ routes, and the dashboard's box telemetry views call the analytics client only �
 they disable themselves when `analyticsApiUrl` is unset rather than falling back.
 Nothing in this directory registers the analytics or billing paths.
 
-The Billing API and Usage export rows are the same Commerce service reached two
-ways, on URLs that normally differ by path rather than host: the browser API
-lives under `/api/billing`, which `BILLING_API_URL` already includes, while the
-internal routes are served off the bare origin because they authenticate a
-service rather than a user. The API itself requires `USAGE_EXPORT_URL` whenever
-export is on and never derives it
+The Billing API, Box admission, and Usage export rows are the same Commerce
+service reached three ways, on URLs that normally differ by path rather than
+host. The browser API and Box admission reads live under `/api/billing`, which
+`BILLING_API_URL` already includes; admission authenticates with the shared
+service token. The internal usage routes are served off the bare origin. The API
+itself requires `USAGE_EXPORT_URL` whenever export is on and never derives it
 ([`configuration.ts`](./api/src/config/configuration.ts)); it is the SST stack
 that supplies a default of `BILLING_API_URL`'s origin
 ([`api.ts`](./infra/stack/api.ts)), so a deployment outside that stack must set
--- a/apps/API.md
+++ b/apps/API.md
@@ -667,12 +667,13 @@ The Billing API, Box admission, and Usage export rows are the same Commerce
 service reached three ways, on URLs that normally differ by path rather than
 host. The browser API and Box admission reads live under `/api/billing`, which
 `BILLING_API_URL` already includes; admission authenticates with the shared
-service token. The API validates that URL and token presence at startup. A
-request-time transport or 5xx failure is logged and temporarily leaves creation
-unlimited without caching the fallback; rejected credentials and malformed
-successful responses still return 503. Successful per-organization resolutions
-are cached in Redis for 30 seconds. The internal usage routes are served off the
-bare origin. The API itself requires `USAGE_EXPORT_URL` whenever export is on
+service token. When `BILLING_API_URL` is configured, the API validates that URL
+and the shared token at startup. A request-time transport or 5xx failure is
+logged and temporarily leaves creation unlimited without caching the fallback;
+rejected credentials and malformed successful responses still return 503.
+Successful per-organization resolutions are cached in Redis for 30 seconds. The
+internal usage routes are served off the bare origin. The API itself requires
+`USAGE_EXPORT_URL` whenever export is on
 and never derives it
 ([`configuration.ts`](./api/src/config/configuration.ts)); it is the SST stack
 that supplies a default of `BILLING_API_URL`'s origin
--- a/apps/api/src/box/errors/box-creation-limit.error.ts
+++ b/apps/api/src/box/errors/box-creation-limit.error.ts
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2026 BoxLite AI
+ * SPDX-License-Identifier: AGPL-3.0
+ */
+
+import { HttpException, HttpStatus } from '@nestjs/common'
+
+export const BOX_CREATION_LIMIT_EXCEEDED_CODE = 'resource_exhausted'
+export const BOX_CREATION_ADMISSION_UNAVAILABLE_CODE = 'upstream_unavailable'
+
+export class BoxCreationLimitExceededError extends HttpException {
+  constructor(currentCount: number, limit: number) {
+    super(
+      {
+        message: `Organization box creation limit reached: ${currentCount} of ${limit} boxes are already counted`,
+        code: BOX_CREATION_LIMIT_EXCEEDED_CODE,
+      },
+      HttpStatus.TOO_MANY_REQUESTS,
+    )
+    this.name = 'BoxCreationLimitExceededError'
+  }
+}
+
+export class BoxCreationAdmissionUnavailableError extends HttpException {
+  constructor(message = 'Box creation admission is temporarily unavailable') {
+    super({ message, code: BOX_CREATION_ADMISSION_UNAVAILABLE_CODE }, HttpStatus.SERVICE_UNAVAILABLE)
+    this.name = 'BoxCreationAdmissionUnavailableError'
+  }
+}
--- a/apps/api/src/box/errors/box-creation-limit.error.ts
+++ b/apps/api/src/box/errors/box-creation-limit.error.ts
@@ -22,7 +22,10 @@ export class BoxCreationLimitExceededError extends HttpException {
 }
 
 export class BoxCreationAdmissionUnavailableError extends HttpException {
-  constructor(message = 'Box creation admission is temporarily unavailable') {
+  constructor(
+    message = 'Box creation admission is temporarily unavailable',
+    readonly retryAfterSeconds?: number,
+  ) {
     super({ message, code: BOX_CREATION_ADMISSION_UNAVAILABLE_CODE }, HttpStatus.SERVICE_UNAVAILABLE)
     this.name = 'BoxCreationAdmissionUnavailableError'
   }
--- a/apps/api/src/box/repositories/box.repository.ts
+++ b/apps/api/src/box/repositories/box.repository.ts
@@ -50,8 +50,8 @@ const CREATION_ADMISSION_RETRY_AFTER_SECONDS = 5
 
 const BOX_CREATION_LIMIT_EXCLUDED_STATES = [
   BoxState.ERROR,
-  // UNKNOWN is treated like ERROR for admission: it is not a usable Box the
-  // organization should have to destroy before creating a replacement.
+  // A settled UNKNOWN Box is treated like ERROR for admission: it is not
+  // usable, while a pending UNKNOWN Box still reserves its creation slot.
   BoxState.UNKNOWN,
   BoxState.DESTROYING,
   BoxState.DESTROYED,
@@ -179,10 +179,17 @@ export class BoxRepository extends BaseRepository<Box> {
       try {
         return await this.dataSource.transaction('SERIALIZABLE', async (entityManager) => {
           const currentCount = await entityManager.count(Box, {
-            where: {
-              organizationId,
-              state: Not(In(BOX_CREATION_LIMIT_EXCLUDED_STATES)),
-            },
+            where: [
+              {
+                organizationId,
+                state: Not(In(BOX_CREATION_LIMIT_EXCLUDED_STATES)),
+              },
+              {
+                organizationId,
+                state: BoxState.UNKNOWN,
+                pending: true,
+              },
+            ],
           })
           if (currentCount >= maxCreatedBoxes) {
             throw new BoxCreationLimitExceededError(currentCount, maxCreatedBoxes)
--- a/apps/api/src/box/repositories/box.repository.ts
+++ b/apps/api/src/box/repositories/box.repository.ts
@@ -22,10 +22,7 @@ import { BoxDesiredStateUpdatedEvent } from '../events/box-desired-state-updated
 import { BoxPublicStatusUpdatedEvent } from '../events/box-public-status-updated.event'
 import { BoxOrganizationUpdatedEvent } from '../events/box-organization-updated.event'
 import { BoxLookupCacheInvalidationService } from '../services/box-lookup-cache-invalidation.service'
-import {
-  BoxCreationAdmissionUnavailableError,
-  BoxCreationLimitExceededError,
-} from '../errors/box-creation-limit.error'
+import { BoxCreationAdmissionUnavailableError, BoxCreationLimitExceededError } from '../errors/box-creation-limit.error'
 
 // SQLSTATE lock_not_available — a statement waited longer than lock_timeout to
 // acquire a lock.
@@ -49,9 +46,13 @@ const CREATION_ADMISSION_MAX_ATTEMPTS = 5
 const CREATION_ADMISSION_BASE_BACKOFF_MS = 5
 const CREATION_ADMISSION_MAX_BACKOFF_MS = 40
 const CREATION_ADMISSION_JITTER_MS = 5
+const CREATION_ADMISSION_RETRY_AFTER_SECONDS = 5
 
 const BOX_CREATION_LIMIT_EXCLUDED_STATES = [
   BoxState.ERROR,
+  // UNKNOWN is treated like ERROR for admission: it is not a usable Box the
+  // organization should have to destroy before creating a replacement.
+  BoxState.UNKNOWN,
   BoxState.DESTROYING,
   BoxState.DESTROYED,
   BoxState.ARCHIVING,
@@ -194,14 +195,20 @@ export class BoxRepository extends BaseRepository<Box> {
           throw error
         }
         if (attempt === CREATION_ADMISSION_MAX_ATTEMPTS - 1) {
-          throw new BoxCreationAdmissionUnavailableError('Box creation admission remained contended; retry the request')
+          throw new BoxCreationAdmissionUnavailableError(
+            'Box creation admission remained contended; retry the request',
+            CREATION_ADMISSION_RETRY_AFTER_SECONDS,
+          )
         }
 
         await this.waitBeforeAdmissionRetry(attempt)
       }
     }
 
-    throw new BoxCreationAdmissionUnavailableError()
+    throw new BoxCreationAdmissionUnavailableError(
+      'Box creation admission remained contended; retry the request',
+      CREATION_ADMISSION_RETRY_AFTER_SECONDS,
+    )
   }
 
   private isSerializationFailure(error: unknown): boolean {
--- a/apps/api/src/box/repositories/box.repository.ts
+++ b/apps/api/src/box/repositories/box.repository.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: AGPL-3.0
  */
 
-import { DataSource, EntityManager, FindOptionsWhere } from 'typeorm'
+import { DataSource, EntityManager, FindOptionsWhere, In, Not } from 'typeorm'
 import { Box } from '../entities/box.entity'
 import { BoxLastActivity } from '../entities/box-last-activity.entity'
 import { BoxMigration } from '../entities/box-migration.entity'
@@ -22,6 +22,10 @@ import { BoxDesiredStateUpdatedEvent } from '../events/box-desired-state-updated
 import { BoxPublicStatusUpdatedEvent } from '../events/box-public-status-updated.event'
 import { BoxOrganizationUpdatedEvent } from '../events/box-organization-updated.event'
 import { BoxLookupCacheInvalidationService } from '../services/box-lookup-cache-invalidation.service'
+import {
+  BoxCreationAdmissionUnavailableError,
+  BoxCreationLimitExceededError,
+} from '../errors/box-creation-limit.error'
 
 // SQLSTATE lock_not_available — a statement waited longer than lock_timeout to
 // acquire a lock.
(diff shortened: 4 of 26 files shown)


apps/api/src/boxlite-rest/commerce-box-limit.service.ts
```diff
@@ -0,0 +1,141 @@
...
+    } catch {
+      throw new BoxCreationAdmissionUnavailableError('Commerce plan service is unavailable')
+    }
+
+    const [catalogResponse, organizationPlanResponse] = responses
+    const catalog = this.parseCatalog(catalogResponse.data)
+    const organizationPlan = this.parseOrganizationPlan(organizationPlanResponse.data)
+    const selectedPlan =
+      !organizationPlan || organizationPlan.entitlements === 'suspended'
+        ? catalog[0]
```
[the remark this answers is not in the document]
mamamiyear: Thanks. The Commerce side currently returns the public plan catalog from lower to higher rank, so catalog[0] intentionally represents the fallback tier. We are not adding a separate hard-coded plan identifier in this PR.

apps/api/src/box/errors/box-creation-limit.error.ts
```diff
@@ -0,0 +1,29 @@
...
+export const BOX_CREATION_ADMISSION_UNAVAILABLE_CODE = 'upstream_unavailable'
+
+export class BoxCreationLimitExceededError extends HttpException {
+  constructor(currentCount: number, limit: number) {
+    super(
+      {
+        message: `Organization box creation limit reached: ${currentCount} of ${limit} boxes are already counted`,
+        code: BOX_CREATION_LIMIT_EXCEEDED_CODE,
+      },
+      HttpStatus.TOO_MANY_REQUESTS,
```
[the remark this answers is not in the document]
mamamiyear: Thanks for flagging this. The API currently follows the repository-wide flat error response produced by AllExceptionsFilter. To stay consistent with the existing repository convention, we are not changing the response envelope in this PR.

apps/api/src/boxlite-rest/commerce-box-limit.service.ts
```diff
@@ -0,0 +1,141 @@
...
+  private normalizeBaseUrl(rawBaseUrl: string): string {
+    let parsed: URL
+    try {
+      parsed = new URL(rawBaseUrl)
+    } catch {
+      throw new BoxCreationAdmissionUnavailableError('Commerce API URL is invalid')
+    }
+
+    if (
+      !['http:', 'https:'].includes(parsed.protocol) ||
```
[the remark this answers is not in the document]
mamamiyear: Thanks. All production Commerce endpoints use HTTPS. HTTP support is intentionally retained for local test environments, so we are not enforcing HTTPS-only in this PR.

apps/api/src/boxlite-rest/commerce-box-limit.service.ts
```diff
@@ -0,0 +1,227 @@
...
+
+    return rawBaseUrl.replace(/\/+$/, '')
+  }
+
+  private parseCatalog(value: unknown): PublicPlan[] {
+    if (!Array.isArray(value) || value.length === 0) {
+      throw new BoxCreationAdmissionUnavailableError('Commerce returned an empty or invalid plan catalog')
+    }
+
+    return value.map((candidate) => {
```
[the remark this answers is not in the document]
mamamiyear: This is indeed guaranteed by the commerce API.

DefeatMan (commented):
Reviewed the branch at `b868b19` against `main`, from the code rather than the linked design doc.

Two things I checked that came back clean, so nobody needs to re-verify them: the API Tests job runs a real Postgres 16 service with `DB_HOST=localhost` (`.github/workflows/test.yml:398-415`), so the new `describeIfDatabase` concurrency suite really does execute in CI even though it was skipped locally; and gating at the controller is genuinely complete — `apps/api/src/boxlite-rest/boxlite-box.controller.ts:120` is the only caller of `boxService.create` in the tree, and `BoxController` exposes no create route. The `SERIALIZABLE` count-and-write shape itself looks right to me.

One point that doesn't belong on a line: **the PR description no longer describes this branch.** It says the change lands "without adding Redis quota state," that it "replaces the broader Redis reservation and cache design in #1400 with the minimum database-backed admission path," and under Risks that "No database migration, Redis quota cache, START/RECOVER limit, Dashboard UI, or SDK implementation change is included." But the head commit is `perf(api): cache Commerce box creation limits`, and `commerce-box-limit.service.ts` now carries a 30s Redis cache plus an in-process in-flight dedup map. The "After" call graph shows neither. The description is doing real explanatory work on this PR, so it's worth refreshing — otherwise reviewers evaluate a design the branch no longer has.

The inline notes below are the ones I think need a decision before merge. None of them are objections to the core admission logic.

apps/api/src/boxlite-rest/commerce-box-limit.service.ts
```diff
@@ -0,0 +1,227 @@
...
+    const catalog = this.parseCatalog(catalogResponse.data)
+    const organizationPlan = this.parseOrganizationPlan(organizationPlanResponse.data)
+    const selectedPlan =
+      !organizationPlan || organizationPlan.entitlements === 'suspended'
+        ? catalog[0]
+        : catalog.find((plan) => plan.id === organizationPlan.planId)
+
+    // A plan absent from the public catalog may be a negotiated subscription;
+    // leave it unlimited until Commerce exposes its effective limit.
+    return selectedPlan?.concurrencyLimit ?? undefined
```
DefeatMan: This is the finding I'd most want resolved before merge.

Every other failure mode in this service fails closed with 503 — bad URL, missing token, unreachable Commerce, malformed catalog, malformed org plan. But when the organization's `planId` simply isn't present in the public catalog, `catalog.find(...)` returns `undefined`, `undefined?.concurrencyLimit ?? undefined` collapses to `undefined`, and the caller reads that as **unlimited**. It is then written to Redis as `unlimited` and reused for the next 30s.

I follow the intent from the comment above it — a negotiated subscription shouldn't be capped at a public tier. The problem is that the same branch also swallows the failure cases: a plan id renamed on the Commerce side, a plan de-listed from the public catalog, or a catalog response that is well-formed but incomplete. In each of those, enforcement silently switches off for every org on that plan, and nothing logs it.

At minimum, could this `this.logger.warn(...)` with the unmatched `planId`? A counter would be better given this gates billing. And if the negotiated-plan case is expected to be common, an explicit signal from Commerce (say `entitlements: 'custom'`) would separate "deliberately uncapped" from "we lost track of this plan," so the fail-open is reachable only on purpose.

apps/api/src/boxlite-rest/commerce-box-limit.service.ts
```diff
@@ -0,0 +1,227 @@
...
+
+    if (!Object.prototype.hasOwnProperty.call(value, 'plan')) {
+      if (Object.keys(value).length === 0) {
+        return undefined
+      }
+      throw new BoxCreationAdmissionUnavailableError('Commerce returned an invalid organization plan')
+    }
+
+    const plan = value.plan
+    if (!isRecord(plan) || typeof plan.planId !== 'string' || plan.planId.length === 0) {
```
DefeatMan, line 260: `parseOrganizationPlan` accepts "no subscription" only when the body is a *literally empty* object. Two shapes Commerce could reasonably return instead both throw 503 and block all box creation for that org:

- `{"plan": null}` — `hasOwnProperty('plan')` is true, so it reaches this line, `isRecord(null)` is false, 503.
- `{"organizationId": "...", "plan": null}`, or any envelope carrying metadata beside a null plan — non-empty with no `plan` key resolves to the 503 on line 210.

`{"plan": null}` is the more idiomatic encoding of "this org has no plan," and since the producing contract lives in boxlite-commerce#68 and isn't merged, nothing currently pins which shape ships. Treating `plan == null` as the unsubscribed case alongside `{}` would make this robust to either.

Given the cross-repo coupling, a contract test (or generating the client from the Commerce spec) would be worth more than the parser hardening on its own — it would turn a shape change into a CI failure here rather than a production 503.

apps/api/src/boxlite-rest/commerce-box-limit.service.ts
```diff
@@ -0,0 +1,227 @@
...
+    @InjectRedis() private readonly redis: Redis,
+  ) {}
+
+  async resolveMaxCreatedBoxes(organizationId: string): Promise<number | undefined> {
+    const rawBaseUrl = this.configService.get('billingApiUrl')?.trim()
+    if (!rawBaseUrl) {
+      return undefined
+    }
+
+    const token = this.configService.get('usageExport.token')?.trim()
```
DefeatMan, line 58: Both of these are deploy-time misconfigurations that only surface when a customer calls CREATE:

- `BILLING_API_URL` set while `USAGE_EXPORT_TOKEN` is empty → 503 from this line on every create.
- `BILLING_API_URL` malformed, or carrying userinfo / query / fragment → 503 from `normalizeBaseUrl` (lines 159-186) on every create.

This repo already has the opposite convention for exactly this class of error. `usageExportConfig` throws at boot when export is armed without a URL or token (`apps/api/src/config/configuration.ts:152-165`), and the comment there spells out the reasoning — a stage pointed at a billing service but never given the shared secret should crash-loop on deploy rather than fail quietly. Meanwhile `configuration.ts:484` currently takes `BILLING_API_URL` raw, with no validation at all.

Moving both checks into `configuration.ts` would let a bad deploy fail fast instead of becoming a customer-visible CREATE outage discovered by the first user who tries. The per-request checks can stay as cheap assertions.

apps/api/src/boxlite-rest/commerce-box-limit.service.ts
```diff
@@ -0,0 +1,227 @@
...
+import { InjectRedis } from '@nestjs-modules/ioredis'
+import { Injectable, Logger } from '@nestjs/common'
+import axios from 'axios'
+import { Redis } from 'ioredis'
+import { TypedConfigService } from '../config/typed-config.service'
+import { BoxCreationAdmissionUnavailableError } from '../box/errors/box-creation-limit.error'
+
+const COMMERCE_REQUEST_TIMEOUT_MS = 2_000
+const COMMERCE_BOX_LIMIT_CACHE_KEY_PREFIX = 'commerce:box-limit:v1:'
+const COMMERCE_BOX_LIMIT_CACHE_TTL_SECONDS = 30
```
DefeatMan, line 15: Worth making this an explicit, documented decision rather than an emergent one: with the cache cold, a Commerce outage takes down hosted CREATE entirely — two parallel reads at a 2s timeout, then `BoxCreationAdmissionUnavailableError`. The 30s TTL narrows the exposure to roughly one request per org per 30s, but every expiry re-opens it.

Since the cached value is already the last known good limit, serving it past its TTL when Commerce is unreachable (stale-while-revalidate) would cap the blast radius at "limits are up to N seconds stale" instead of "nobody can create a box." The limit only tightens on a plan downgrade, so stale-permissive looks like the cheaper failure mode of the two.

If failing closed is the deliberate call — reasonable, since this is billing enforcement — it'd be good to state that in Risks / rollout, along with the fact that Commerce availability is now a hard dependency of hosted CREATE.

apps/api/src/box/repositories/box.repository.ts
```diff
@@ -37,6 +41,23 @@ const PG_LOCK_TIMEOUT_CODE = '55P03'
...
+const CREATION_ADMISSION_BASE_BACKOFF_MS = 5
+const CREATION_ADMISSION_MAX_BACKOFF_MS = 40
+const CREATION_ADMISSION_JITTER_MS = 5
+
+const BOX_CREATION_LIMIT_EXCLUDED_STATES = [
+  BoxState.ERROR,
+  BoxState.DESTROYING,
+  BoxState.DESTROYED,
+  BoxState.ARCHIVING,
+  BoxState.ARCHIVED,
```
DefeatMan, line 56: Two observations about this exclusion list.

**`UNKNOWN` counts, `ERROR` doesn't.** A box that has drifted into `UNKNOWN` consumes a quota slot indefinitely, and the user has no way to see why they're at their cap or to clear it. That asymmetry reads as unintentional — if `ERROR` is excluded because a dead box shouldn't count against the plan, `UNKNOWN` seems to qualify on the same grounds.

**The counted set doesn't match what the Commerce field is called.** `STOPPED` counts, so `concurrencyLimit` is being enforced as "total boxes not yet destroyed." Those diverge in both directions: an org whose boxes are all stopped consumes zero concurrency yet is blocked from creating; and because START and RECOVER are out of scope here, an org sitting at the cap can start all N boxes at once, so actual concurrency isn't bounded by anything in this PR.

That's a fine first increment — I'm not asking for START gating here. But it may be worth naming the API-side concept for what it enforces (`maxCreatedBoxes` already reads that way) and calling the concurrency gap out as explicit follow-up, so the Commerce field name doesn't imply a guarantee nothing currently makes.

apps/api/src/box/repositories/box.repository.ts
```diff
@@ -136,6 +165,63 @@ export class BoxRepository extends BaseRepository<Box> {
...
+          }
+
+          return persist(entityManager)
+        })
+      } catch (error) {
+        if (!this.isSerializationFailure(error)) {
+          throw error
+        }
+        if (attempt === CREATION_ADMISSION_MAX_ATTEMPTS - 1) {
+          throw new BoxCreationAdmissionUnavailableError('Box creation admission remained contended; retry the request')
```
DefeatMan: Two suggestions on the exhausted-retries path.

**503 may be the wrong code here.** The total retry budget is small — 5 attempts with backoff capped at 40ms plus up to 5ms jitter, so on the order of 150ms of wall clock before giving up. An organization scripting parallel creates can reach that, and the client then sees a 503, which reads as "the service is broken" rather than "you're contending, back off." A 429 is closer to the truth for this case and would let clients reuse the same backoff path they already need for the limit-exceeded 429.

**Neither response carries `Retry-After`.** This 503 and `BoxCreationLimitExceededError` are both now documented responses on the CREATE contract, and `Retry-After` is what the SDK and CLI need in order to back off instead of hammering. For the limit case there may be no meaningful hint to give, but for contention the retry window is known and cheap to surface.

apps/api/src/box/services/box.service.ts
```diff
@@ -332,12 +340,20 @@ export class BoxService {
     }
   }
 
+  private insertWithCreationAdmission(box: Box, maxCreatedBoxes?: number): Promise<Box> {
```
DefeatMan: `insertWithCreationAdmission` doesn't change behavior. `this.boxRepository.insert(box, undefined)` and `this.boxRepository.insert(box)` are the same call — the parameter is optional, and `withCreationAdmission` branches on `maxCreatedBoxes === undefined` in both cases. The same holds for `const creationAdmission = maxCreatedBoxes === undefined ? {} : { maxCreatedBoxes }` on line 356: spreading `{ maxCreatedBoxes: undefined }` reaches the identical branch.

I checked whether `exactOptionalPropertyTypes` would make the distinction meaningful — it isn't set in `apps/tsconfig.base.json` or any tsconfig in the workspace, so this really is a no-op.

All three sites can pass the value straight through and the helper can go, which also removes the impression that `undefined` is being handled specially somewhere downstream.

openapi/box.openapi.yaml
```diff
@@ -1323,6 +1327,19 @@ components:
...
+    ResourceExhaustedError:
+      description: The organization has reached its active Box creation limit
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ErrorResponse"
+          example:
+            error:
+              message: "Organization box creation limit reached: 2 of 2 boxes are already counted"
+              type: ResourceExhaustedError
```
DefeatMan: This example can't be produced by the server. `ErrorModel` requires a nested `error` envelope with `message`, `type`, and `code`, but `AllExceptionsFilter` — the only exception filter registered (`apps/api/src/main.ts:85`; there's no boxlite-rest-specific filter) — emits a flat body of `{ statusCode, error: <status text>, message, code }` (`apps/api/src/filters/all-exceptions.filter.ts:80-86`). No `type`, no `request_id`, and an undocumented `statusCode`.

From your reply to CodeRabbit I understand the flat shape is the intentional repo-wide convention and that changing the envelope is out of scope — agreed, and I'm not asking for that here. The narrower point is that this PR adds a *new* hand-written example asserting three fields the server never returns, on a surface whose own header comment calls `box.openapi.yaml` the contract.

Dropping the `example:` block would be enough — the `$ref` alone still documents the 429 — or, if the example is valuable, matching it to what `AllExceptionsFilter` actually emits. Either way it'd be good to have the spec-vs-filter reconciliation tracked somewhere so it stops accreting new instances.

apps/api/src/boxlite-rest/commerce-box-limit.service.ts
```diff
@@ -0,0 +1,227 @@
...
+      if (/^(0|[1-9]\d*)$/.test(rawLimit) && Number.isSafeInteger(limit)) {
+        return { hit: true, maxCreatedBoxes: limit }
+      }
+    }
+
+    this.logger.warn(`Ignoring invalid Commerce box limit cache value for ${cacheKey}`)
+    return { hit: false }
+  }
+
+  private async cacheResolvedLimit(cacheKey: string, maxCreatedBoxes: number | undefined): Promise<void> {
```
DefeatMan, line 124: Follow-up to the thread on line 156 — raising it separately because the fix belongs here rather than there.

`cacheResolvedLimit` persists `undefined` as the literal string `unlimited` on the same 30s TTL as a real limit. But `undefined` reaches this function from two very different places:

- the organization is genuinely on an uncapped plan (`concurrencyLimit: null`) — a real answer, safe to cache;
- `catalog.find(...)` failed to match the org's `planId` (lines 151-156) — not an answer at all, just a lookup miss coerced into "unlimited".

The second case is the problem. A single transient Commerce response with an incomplete catalog, or a plan id renamed upstream, gets latched into Redis and every subsequent create for that org is unrestricted for the next 30 seconds — including creates served by *other* API instances, since the cache is shared. One bad read amplifies into a 30-second enforcement outage for that tenant, and nothing in logs or metrics marks it.

Either of these would close it:

- don't cache the unmatched-plan result — return it for the current request but skip `cacheResolvedLimit`, so the next request re-reads Commerce;
- or distinguish the two sources in the cached value (`unlimited` vs something like `unmatched`) and give the second a much shorter TTL.

Together with the logging suggestion on line 156, that would make this case both visible and short-lived rather than silent and sticky.

DefeatMan (commented):
Two follow-ups on the Redis caching layer, both about how the cache interacts with the Commerce lookup rather than about the admission transaction itself.

apps/api/src/boxlite-rest/commerce-box-limit.service.ts
```diff
@@ -0,0 +1,227 @@
...
+type CachedBoxLimit = { hit: true; maxCreatedBoxes: number | undefined } | { hit: false }
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && value !== null && !Array.isArray(value)
+}
+
+@Injectable()
+export class CommerceBoxLimitService {
+  private readonly logger = new Logger(CommerceBoxLimitService.name)
+  private readonly inFlightResolutions = new Map<string, Promise<number | undefined>>()
```
DefeatMan, line 45: `inFlightResolutions` deduplicates concurrent lookups within a single Node process, but the API runs multiple instances against one shared Redis. Each instance holds its own `Map`, so this is per-process coalescing rather than a global singleflight.

The effect shows up at TTL expiry: the Redis key drops for an organization, and every instance currently handling a create for that org misses the cache and issues its own pair of Commerce calls. With N instances that's up to 2N requests to `GET /plan` and `GET /organization/{id}/plan` in the same instant, recurring every 30 seconds for any org creating boxes steadily. `GET /plan` is the wasteful half — it returns the same catalog for every organization, so nearly all of that traffic is redundant.

At current scale this is probably fine and I wouldn't block on it. Two things worth weighing:

- `GET /plan` is organization-independent. Caching the catalog under its own key with a longer TTL would collapse most of the burst; only the per-org read genuinely needs the 30s refresh.
- If Commerce ever rate-limits or degrades under that burst, this PR's fail-closed behaviour converts it into simultaneous 503s on CREATE across every instance.

At minimum, a short comment noting the dedup is per-process would help — otherwise someone later reads this `Map` as a guarantee that Commerce sees one request per org per TTL, which it isn't.

apps/api/src/box/repositories/box.repository.ts
```diff
@@ -136,6 +165,63 @@ export class BoxRepository extends BaseRepository<Box> {
...
+    maxCreatedBoxes: number | undefined,
+    persist: (entityManager: EntityManager) => Promise<T>,
+  ): Promise<T> {
+    if (maxCreatedBoxes === undefined) {
+      return this.dataSource.transaction(persist)
+    }
+
+    for (let attempt = 0; attempt < CREATION_ADMISSION_MAX_ATTEMPTS; attempt++) {
+      try {
+        return await this.dataSource.transaction('SERIALIZABLE', async (entityManager) => {
```
DefeatMan, line 177: The Changes section says this lands "without adding Redis quota state, advisory locks, schema, or indexes." Redis state and schema I follow — both add real operational surface. Advisory locks I'd like to push back on, because I think they're strictly better than `SERIALIZABLE` for this specific decision, and they'd let you delete code rather than add it.

### Why SERIALIZABLE is an awkward fit here

The count is `WHERE organizationId = $1 AND state NOT IN (...)`. SSI can't record that predicate, so it takes predicate locks on the *physical* objects the scan touched — index pages and heap tuples. Two consequences:

**Predicate locks aren't isolated per organization.** `organizationId` is a random UUID (`box.entity.ts`), so organizations are fully interleaved across `box_organizationid_idx` b-tree pages — one page holds entries for many unrelated orgs. Org B's insert landing on a page Org A's count locked produces a `40001` that has nothing to do with either org's quota.

**Escalation turns that from occasional into systemic.** `box_organizationid_idx` covers `organizationId` only, so evaluating `state NOT IN (...)` requires heap fetches, which leave tuple-level predicate locks behind. Past `max_pred_locks_per_page` (default 2) those become page locks; past `max_pred_locks_per_relation` (default `max_pred_locks_per_transaction`/2 = 32) they escalate to a **relation-level lock on `box`**. An organization with a few hundred boxes can cross that just by counting them — and once it does, every concurrent hosted CREATE from every other tenant conflicts with it. A seq scan (stale stats, small table) reaches the same state immediately.

The retry budget can't absorb that: 5 attempts with backoff capped at 40ms + ≤5ms jitter is ~150ms total, and each attempt holds a pooled connection.

### What an advisory lock gives instead

`pg_advisory_xact_lock` keys on a number, not on storage. Different organizations produce different `LOCKTAG`s, so there's no false sharing and no escalation path — it's one entry in the same shared lock table that regular row locks already use, and it releases automatically at COMMIT or ROLLBACK, so it can't leak back into the pool.

```ts
private async withCreationAdmission<T>(
  organizationId: string,
  maxCreatedBoxes: number | undefined,
  persist: (entityManager: EntityManager) => Promise<T>,
): Promise<T> {
  if (maxCreatedBoxes === undefined) {
    return this.dataSource.transaction(persist)
  }

  return this.dataSource.transaction(async (entityManager) => {
    // First statement in the transaction: a consistent lock order against the
    // row locks taken later, so admission can't deadlock with updateWhere().
    await entityManager.query(`SET LOCAL lock_timeout = '${ADMISSION_LOCK_TIMEOUT_MS}ms'`)
    await entityManager.query('SELECT pg_advisory_xact_lock($1, hashtext($2))', [
      BOX_ADMISSION_LOCK_NAMESPACE,
      organizationId,
    ])

    const currentCount = await entityManager.count(Box, {
      where: { organizationId, state: Not(In(BOX_CREATION_LIMIT_EXCLUDED_STATES)) },
    })
    if (currentCount >= maxCreatedBoxes) {
      throw new BoxCreationLimitExceededError(currentCount, maxCreatedBoxes)
    }
    return persist(entityManager)
  })
}
```

Contention is then scoped to one organization, and it costs a short wait instead of an abort. What comes out: the retry loop, `CREATION_ADMISSION_MAX_ATTEMPTS`, the three backoff constants, `waitBeforeAdmissionRetry`, `isSerializationFailure`, `PG_SERIALIZATION_FAILURE_CODE`, and the contention branch of `BoxCreationAdmissionUnavailableError` (the Commerce branch still needs it). Net deletion.

Use the two-argument form so these locks sit in their own namespace and can't collide with advisory locks taken elsewhere in the database. `hashtext` collisions only cause two unrelated orgs to serialize against each other briefly — no correctness impact.

### The honest tradeoff

An advisory lock blocks where SSI aborts, so concurrent same-org creates queue on a pooled connection instead of failing fast. That needs a bound — and **this file already has the pattern**: line 295 sets `SET LOCAL lock_timeout` on the proxy auto-start UPDATE, line 336 treats SQLSTATE `55P03` as a lost race, and the comment at lines 34-41 lays out exactly this reasoning ("there is no global statement/lock timeout configured ... without this bound a contended row could pin a pooled connection indefinitely"). Admission can reuse it directly: on `55P03`, return the same 429 rather than a 503, which is also more accurate than what the current contention path returns.

If queuing isn't wanted at all, `pg_try_advisory_xact_lock` fails immediately and turns this into a one-line difference.

### The question

Was the advisory-lock option ruled out on operational grounds, or was it primarily about keeping the diff minimal? If the latter, this version is smaller than what's here now. If there's a constraint I'm not seeing — something about the pooler, or a policy against advisory locks — that would be useful to record in the PR body, since "no advisory locks" currently reads as a deliberate choice without a stated reason.

DefeatMan (commented):
One design-level suggestion on the choice of `SERIALIZABLE` as the serialization primitive. Filed as a question rather than a request for change — the Changes section rules out advisory locks explicitly, so I assume there was a reason, and I'd like to understand it.

apps/api/src/config/configuration.ts
```diff
@@ -78,6 +78,30 @@ function requiredHttpUrl(value: string, name: string): string {
...
+function billingApiUrlConfig(env: NodeJS.ProcessEnv = process.env): string | undefined {
+  const rawUrl = env.BILLING_API_URL?.trim()
+  if (!rawUrl) {
+    return undefined
+  }
+  if (!env.USAGE_EXPORT_TOKEN?.trim()) {
+    throw new Error('USAGE_EXPORT_TOKEN is required when BILLING_API_URL is set')
+  }
+
+  const url = requiredHttpUrl(rawUrl, 'BILLING_API_URL')
```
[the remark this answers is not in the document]
mamamiyear: Thanks. Please see the earlier discussion: https://github.com/boxlite-ai/boxlite/pull/1417#discussion_r3896008893. Production Commerce endpoints use HTTPS, while HTTP support is intentionally retained for local test environments. We are keeping that behavior in this PR.

apps/api/src/boxlite-rest/commerce-box-limit.service.ts
```diff
@@ -0,0 +1,272 @@
...
+      responses = await Promise.all([
+        axios.get<unknown>(`${baseUrl}/plan`, requestConfig),
+        axios.get<unknown>(`${baseUrl}/organization/${encodeURIComponent(organizationId)}/plan`, requestConfig),
+      ])
+    } catch (error) {
+      if (this.isTemporarilyUnavailable(error)) {
+        this.logger.warn(
+          `Commerce box limit lookup for organization ${organizationId} is unavailable (${this.unavailabilityReason(error)}); temporarily allowing Box creation`,
+        )
+        return { maxCreatedBoxes: undefined, isCacheable: false }
```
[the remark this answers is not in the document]
mamamiyear: Thanks for calling out the fail-open trade-off. This is intentional under the current product policy: for temporary transport failures and Commerce 5xx responses, hosted CREATE remains available, logs a warning, and does not cache the unlimited fallback, so the next request retries Commerce. Commerce 4xx responses and malformed successful responses still fail closed with 503. We will keep this behavior for this PR.

apps/API.md
```diff
@@ -662,12 +663,17 @@ routes, and the dashboard's box telemetry views call the analytics client only 
...
-service rather than a user. The API itself requires `USAGE_EXPORT_URL` whenever
-export is on and never derives it
+The Billing API, Box admission, and Usage export rows are the same Commerce
+service reached three ways, on URLs that normally differ by path rather than
+host. The browser API and Box admission reads live under `/api/billing`, which
+`BILLING_API_URL` already includes; admission authenticates with the shared
+service token. The API validates that URL and token presence at startup. A
+request-time transport or 5xx failure is logged and temporarily leaves creation
+unlimited without caching the fallback; rejected credentials and malformed
+successful responses still return 503. Successful per-organization resolutions
```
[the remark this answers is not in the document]
mamamiyear: Thanks. I clarified the wording in 3416d728: startup validation of the URL and shared token applies when `BILLING_API_URL` is configured. The documented request-time behavior remains unchanged and matches the current product policy: transport failures and 5xx responses log a warning, temporarily allow creation, and are not cached; rejected credentials and malformed successful responses still return 503.

.github/workflows/test.yml
```diff
@@ -422,6 +422,12 @@ jobs:
         with: # Native workspace dependencies are submodules.
           submodules: "recursive"
           persist-credentials: ${{ false }}
+      - name: Checkout Commerce contract provider
+        uses: actions/checkout@v5
+        with:
+          repository: boxlite-ai/boxlite-commerce
```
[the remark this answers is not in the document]
mamamiyear: Thanks for raising this. Omitting `ref` is intentional: this cross-repository contract test is meant to continuously validate BoxLite against the current `boxlite-commerce` default branch and surface contract drift as soon as it lands. Pinning a SHA would make individual runs reproducible, but it would also stop testing new Commerce contract changes until the pin is manually advanced. We are accepting that trade-off here and will keep this checkout unpinned.

apps/api/src/boxlite-rest/commerce-box-limit.service.ts
mamamiyear: Thanks for raising this. The current Commerce plan contract does not yet expose an explicit field that distinguishes negotiated plans, so this PR intentionally keeps the current-request fail-open behavior for an unmatched `planId`. The lookup now emits a warning and marks that resolution as non-cacheable, so it is not written to Redis and the next request consults Commerce again. We will adapt to the explicit field once Commerce extends the plan contract.

apps/api/src/boxlite-rest/commerce-box-limit.service.ts
mamamiyear: Thanks for the suggestion. The current Commerce contract represents no subscription as `{}` and does not emit `{"plan": null}`, so the consumer remains strict instead of accepting a response shape that does not exist today. We added cross-repository contract coverage that passes the current Commerce controller responses through the BoxLite consumer. Fork and Dependabot pull requests skip the provider checkout because repository secrets are unavailable, while the `merge_group` run still enforces the contract before merge.

apps/api/src/boxlite-rest/commerce-box-limit.service.ts
mamamiyear: Agreed. These deployment errors are now validated during configuration loading: when `BILLING_API_URL` is set, `USAGE_EXPORT_TOKEN` is required, and the URL is validated before the API starts. Invalid URLs, credentials in userinfo, queries, fragments, or a missing token therefore fail startup. The request-time checks remain as defensive validation.

apps/api/src/boxlite-rest/commerce-box-limit.service.ts
mamamiyear: Thanks for calling out the availability trade-off. Under the current product policy, a timeout, transport failure, or Commerce 5xx response must not block CREATE: the API logs a warning, allows only the current request, and does not cache the unlimited fallback, so the next request retries Commerce. Commerce 4xx responses and malformed successful responses still return 503. This behavior is now documented explicitly.

apps/api/src/box/repositories/box.repository.ts
mamamiyear: Agreed. A settled `UNKNOWN` Box is now treated like `ERROR` and excluded, while an `UNKNOWN` Box with `pending=true` still counts because a fresh Box is initially persisted in that state and must reserve a slot to preserve the concurrent admission invariant. `STOPPED` continues to count as required, so the API-side concept is the organization's creation-admission total rather than runtime concurrency; START and RECOVER remain outside this PR. The repository documentation was updated in `8f776cd4`, and the PR description was updated as well to make this distinction explicit.

apps/api/src/box/repositories/box.repository.ts
mamamiyear: Thanks for the suggestions. Exhausting serialization retries remains a `503 upstream_unavailable` because it means the service could not complete a consistent admission decision; it does not establish that the organization has exhausted its quota. That response now includes `Retry-After: 5` so clients can back off. A confirmed `count >= limit` still returns 429; it has no retry hint because the service cannot predict when a Box slot will be released.

apps/api/src/box/services/box.service.ts
mamamiyear: Agreed. Those branches did not change behavior and only added indirection. `insertWithCreationAdmission` and the conditional object wrapping have been removed. Both the fresh-Box and warm-pool paths now pass `maxCreatedBoxes` directly, and the repository handles `undefined` in one place.

openapi/box.openapi.yaml
mamamiyear: Agreed. Reworking the repository-wide error envelope remains outside this PR, but the new example must reflect the server's current behavior. The 429 example now uses the flat shape emitted by `AllExceptionsFilter` and no longer includes the nested error object, `type`, or `request_id` fields that the server does not return.

apps/api/src/boxlite-rest/commerce-box-limit.service.ts
mamamiyear: Agreed. A real `concurrencyLimit: null` remains a valid, cacheable unlimited result. An unmatched `planId` now logs a warning and returns a resolution with `isCacheable: false`, so it applies only to the current request, is not written to Redis, and the next request reads Commerce again.

apps/api/src/boxlite-rest/commerce-box-limit.service.ts
mamamiyear: Agreed. The `Map` only coalesces concurrent misses within one Node process; it is not a global singleflight. The code comment and PR risks now state that Redis shares completed resolutions, while separate API instances may independently refresh the same expired key. We accept that trade-off at the current scale and are not adding a distributed lock or a separate catalog cache in this PR.

apps/api/src/box/repositories/box.repository.ts
mamamiyear: Thanks for the detailed analysis. Keeping `SERIALIZABLE` here is intentional: this requirement calls for optimistic concurrency control. PostgreSQL `SERIALIZABLE` is based on SSI and does not queue all requests for serial execution up front; transactions still execute concurrently against their own snapshots, while PostgreSQL tracks read/write dependencies using non-blocking predicate locks. When it detects a dependency pattern that could make the committed result inconsistent with every possible serial order, it aborts one transaction with SQLSTATE `40001`, and this service reruns the complete count plus INSERT/UPDATE. Requests for different organizations can usually proceed concurrently, and requests for the same organization do not need to acquire a mutex before doing the calculation.

By contrast, taking a per-organization `pg_advisory_xact_lock` would turn same-organization requests into blocking pessimistic serialization and introduce new runtime semantics around lock-key derivation, wait timeouts, and error mapping. The admission critical section currently contains only the count and one INSERT/UPDATE; only SQLSTATE `40001` retries the complete transaction, for at most five attempts. We have no production evidence that predicate-lock false sharing is currently a material bottleneck, so to keep this change minimal we are not switching to advisory locks in this PR. If monitoring later shows cross-organization conflicts or connection-pool pressure, we can evaluate that as a separate optimization.

apps/api/src/box/repositories/box.repository.ts
```diff
@@ -136,6 +166,76 @@ export class BoxRepository extends BaseRepository<Box> {
...
+            where: [
+              {
+                organizationId,
+                state: Not(In(BOX_CREATION_LIMIT_EXCLUDED_STATES)),
+              },
+              {
+                organizationId,
+                state: BoxState.UNKNOWN,
+                pending: true,
+              },
```
DefeatMan: Follow-up to the thread on line 59: the exclusion this branch pairs with is unreachable, so the behaviour the comment above it describes does not exist.

`BoxDesiredState` has only `destroyed | started | stopped | resized` — there is no `unknown` member. So whenever `state === UNKNOWN`, `getInvariantChanges()` always takes `!pending && String(state) !== String(desiredState)` (`box.entity.ts:290`) and can never take the clearing rule on line 293, which forces `pending = true`. Every write path runs it: `insert` (line 87), `update` (line 139) and `updateWhere` (line 280). The only production writer of `state = UNKNOWN` is `box-start.action.ts:157`, inside `handleRunnerBoxStartedStateCheck`, and it goes through `updateBoxState` -> `boxRepository.update`. The one migration that writes `pending = false` (`migrations/pre-deploy/1786300000000-retire-unused-job-types-migration.ts`) only touches `state = 'resizing'` rows.

So `state = 'unknown' AND pending = false` cannot occur. Excluding `UNKNOWN` on line 55 and adding `UNKNOWN AND pending = true` back here cancels out exactly — this two-branch predicate is equivalent to simply dropping `BoxState.UNKNOWN` from `BOX_CREATION_LIMIT_EXCLUDED_STATES`.

Two consequences.

**The documented refinement is not there.** Lines 53-54, the PR description, and `apps/API.md:692-693` all state that a settled `UNKNOWN` Box is excluded like `ERROR`. Nothing can produce one. My original point on line 59 therefore stands unchanged: a Box that drifts into `UNKNOWN` holds a creation slot indefinitely, and the user can neither see why they are at their cap nor clear it. `handleRunnerBoxStartedStateCheck` gives `STARTING`, `RESTORING` and `CREATING` a `checkTimeoutError`, but its `UNKNOWN` case (lines 156-159) writes `UNKNOWN` back and keeps syncing, so no timeout ever converts it to `ERROR` and releases the slot.

**The test that covers it asserts an impossible row.** `box.repository.creation-limit.integration.spec.ts:165-179` passes only because it inserts through the bare `Repository` with `existing.pending = false`, which bypasses `enforceInvariants()`. For `state = UNKNOWN` it writes a row shape `BoxRepository` cannot write.

To be clear about what `09d45ee1` got right: with `UNKNOWN` fully excluded, a freshly inserted Box — `state = UNKNOWN` from the entity default, `pending = true` from `box.service.ts:310` — did not count itself, so the limit was not applied to the pending set at all. That was a real bug and fixing it was correct. It is only the settled/pending distinction layered on top that cannot be reached.

Suggestion: drop `BoxState.UNKNOWN` from the exclusion list and delete this second `where` branch. That is the same predicate expressed as one condition instead of an OR of two, which also leaves fewer predicate locks behind for the `SERIALIZABLE` decision. Then either give a stuck `UNKNOWN` Box a timeout into `ERROR` so its slot is released, or change lines 53-54, the PR description and `apps/API.md` to say plainly that `UNKNOWN` counts.

mamamiyear: Thanks for pointing this out.

This issue came from an incorrect interpretation of the state semantics while addressing the earlier review feedback. I mistakenly treated `state=UNKNOWN && pending=true` as exclusively representing a newly created Box that was still being processed, and assumed that there was a `state=UNKNOWN && pending=false` case that should be excluded. This introduced an unnecessary special-case branch.

The implementation has now been adjusted so that admission counts all `UNKNOWN` Boxes. The `pending`-dependent query branch has been removed, and the documentation and tests have been updated accordingly.

Under the current product policy, an `UNKNOWN` Box continues to consume quota. When this state occurs, the Box should be deleted by the user or an administrator, and the corresponding quota is released once the Box enters a state excluded from admission counting.

DefeatMan (approved):

PR #1424, opened 2026-09-01
feat(dashboard): add network section to box detail page

Mandalorian-Wang:
## Summary

The box detail page showed nothing about networking. A user who started a server inside a box had no way to learn its preview URL, or whether the box was reachable from outside the organization, without leaving the console for the API.

This adds a `network` section to the detail sheet with two things: what `access` the preview URLs require, and a way to get the URL for a port.

```
▪ NETWORK ─────────────────────── GET URL
access ·········· your organization  CHANGE
```

`GET URL` opens a dialog that asks for a port and returns that port's preview URL, with copy, open, who-can-open-it and the bind hint.

## Screenshots

**The section on the detail sheet**

network section

Two rows: what `access` the preview URLs require, and `get url`. It sits between `volumes` and `activity` in the same spec-sheet idiom as the rest of the column.

**`GET URL` asks for a port**

get url dialog

`Open ↗` stays disabled until a URL exists.

**…and returns that port's URL**

resolved url

With copy, who can open it, and the bind hint — stated here because the URL opens in a new tab, where a failure is invisible to the console.

**`CHANGE` confirms before opening access**

confirmation

The dialog says what opens *and* what does not: the web terminal and the management API are unaffected.

**Once open, the value carries the risk**

public

`anyone` is amber — hue marks the risk, while the key/value greyscale convention the rest of the sheet uses is left intact. The confirmation is symmetric, so the state can be closed again from the same control.

### Before

```
BoxDetails (apps/dashboard/src/components/boxes/BoxDetails.tsx)
├── SectionHeader "general"     → box id · image · region
├── SectionHeader "resources"   → cpu · memory · disk        ← two headers, 6 rows
├── SectionHeader "volumes"     → mounts (conditional)
├── SectionHeader "timestamps"  → created · last event       ← names the value type
└── (no networking anywhere)                                 ← BUG: preview URL and
                                                               exposure unreachable
                                                               from the console
```

### After

```
BoxDetails
├── SectionHeader "general"     → box id · image · region · cpu · memory · disk
├── SectionHeader "volumes"     → mounts (conditional)
├── BoxNetworkSection ────────────────────────────────────────────────────┐
│     useUpdateBoxPublicStatusMutation                                    │
│       └── boxApi.updatePublicStatus → invalidate boxes.detail           │
│     useBoxPreviewUrlMutation                                            │
│       └── boxApi.getPortPreviewUrl(box, port) → url                     │
│                                                                         │
│     ├── BoxNetworkPanel      (presentational)                           │
│     │     SectionHeader "network" → get url                             │
│     │     access row → your organization | anyone  ·  change            │
│     │     AlertDialog → scoped confirmation                             │
│     └── BoxPreviewUrlDialog  (presentational)                           │
│           port input → get url → url · copy · open ↗                    │
└─────────────────────────────────────────────────────────────────────────┘
└── SectionHeader "activity"    → created · last event
```

No component issues HTTP: both server calls sit in the hooks layer.

## Design notes

**The port is asked for, not discovered.** The console cannot enumerate what is listening inside a guest, and every route was checked: `POST /exec` returns only an `execution_id`; `GET /executions/{id}` returns only status and exit code; output is reachable solely over `GET /executions/{id}/attach`, a WebSocket upgrade whose `Authorization` header a browser cannot set (`boxlite-ws-proxy.service.ts:176`); `GET /box/{id}` and `GET /metrics` carry no port data; and file transfer refuses `/proc` outright. Offering a list of common ports would have dressed a guess up as discovery, so the dialog asks instead. A `GET /boxes/{id}/listening-ports` endpoint would settle this — see the follow-up below.

**The confirmation is scoped to preview traffic.** `public` makes the proxy skip authentication for preview requests (`get_box_target.go:87`) and raw tunnel CONNECTs (`tunnel.go:43`). It does not touch the web terminal — the same line exempts `TERMINAL_PORT` — nor the management API. The dialog says both what opens and what does not, because an unscoped "make this box public" reads as handing over the whole box.

**`access: your organization`** describes the URL the dialog hands out, which carries no credential: `getPortPreviewUrl` returns `url` and `token` as separate fields and only `url` is shown. A non-member opening it is bounced through OIDC and refused by the membership check in `preview.controller.ts:142`. Access is credential-based, not network-based — there is no inbound IP allowlist (`openapi/box.openapi.yaml`: "no layer enforces an inbound allowlist today").

**The preview URL is a mutation, not a query.** It runs on click for a port the user names, and its response also carries the box auth token, which should not sit in a shared cache where devtools, a HAR capture or error reporting would pick it up.

**Two section renames come along.** `resources` folds into `general` — six rows did not need two headers — and `timestamps` becomes `activity`, which names the subject rather than the data type and is the platform's own term for it (`POST /:boxId/last-activity`), as well as what auto-stop is measured against.

## Tests

```
✓ src/components/boxes/BoxDetails.test.tsx (3 tests)

Test Files  1 passed (1)
     Tests  3 passed (3)
```

`BoxNetworkPanel` and `BoxPreviewUrlDialog` are presentational and driven by props; their states — restricted, public, saving, viewer-only, fetching, resolved, failed — are covered by stories rather than assertions, since there is no logic to assert beyond what renders.

## Review feedback addressed

- **Network I/O moved out of the component.** Both calls are hooks now (`useUpdateBoxPublicStatusMutation`, `useBoxPreviewUrlMutation`), cache invalidation is defined once per resource, and no raw Axios remains — the earlier hand-written `POST /exec` request is gone with the port probe it served.
- **The exec response shape.** The review was right that `POST /exec` returns `execution_id` rather than `stdout`, which meant the port probe would have silently parsed an empty string in production while the mock hid it. Verified against dev, and the probe was removed rather than patched, since its output is only reachable over a WebSocket the browser cannot authenticate.

## Follow-up, not in this PR

A `GET /boxes/{id}/listening-ports` endpoint — the runner can read `/proc/net/tcp` server-side without the browser's WebSocket constraint. With it, this section can list the reachable URLs directly and flag a service bound to `127.0.0.1` before the user opens a tab that fails. Only `BoxNetworkSection` would change.

### DIFF
--- a/apps/dashboard/src/components/boxes/BoxPreviewUrlDialog.stories.tsx
+++ b/apps/dashboard/src/components/boxes/BoxPreviewUrlDialog.stories.tsx
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2026 BoxLite AI
+ * SPDX-License-Identifier: AGPL-3.0
+ */
+
+import type { Meta, StoryObj } from '@storybook/react'
+import { useState } from 'react'
+import { BoxPreviewUrlDialog } from './BoxPreviewUrlDialog'
+
+const RESTRICTED_URL = 'https://3000-d-3874534e7a76373849495666.proxy.dev.boxlite.ai'
+
+const meta: Meta<typeof BoxPreviewUrlDialog> = {
+  title: 'Boxes/BoxPreviewUrlDialog',
+  component: BoxPreviewUrlDialog,
+}
+
+export default meta
+type Story = StoryObj<typeof BoxPreviewUrlDialog>
+
+const noop = () => {}
+
+/** Opened, nothing asked for yet. */
+export const Empty: Story = {
+  name: '1 · Asking for a port',
+  args: { open: true, onOpenChange: noop, isPublic: false, onFetchUrl: noop },
+}
+
+export const Fetching: Story = {
+  name: '2 · Fetching',
+  args: { open: true, onOpenChange: noop, isPublic: false, isFetching: true, onFetchUrl: noop },
+}
+
+/** Restricted box: the URL only opens for organization members. */
+export const RestrictedResult: Story = {
+  name: '3 · URL on a restricted box',
+  args: { open: true, onOpenChange: noop, isPublic: false, url: RESTRICTED_URL, onFetchUrl: noop },
+}
+
+/** Public box: the wording has to make the anonymous reach unmissable. */
+export const PublicResult: Story = {
+  name: '4 · URL on a public box',
+  args: { open: true, onOpenChange: noop, isPublic: true, url: RESTRICTED_URL, onFetchUrl: noop },
+}
+
+export const Failed: Story = {
+  name: '5 · Request failed',
+  args: {
+    open: true,
+    onOpenChange: noop,
+    isPublic: false,
+    error: 'Could not get a URL for that port.',
+    onFetchUrl: noop,
+  },
+}
+
+/** Type a port and watch the URL resolve, without a backend. */
+export const Interactive: Story = {
+  name: '6 · Interactive (type a port)',
+  render: () => {
+    const [url, setUrl] = useState<string | undefined>()
+    const [isFetching, setIsFetching] = useState(false)
+
+    return (
+      <BoxPreviewUrlDialog
+        open
+        onOpenChange={noop}
+        isPublic={false}
+        url={url}
+        isFetching={isFetching}
+        onFetchUrl={(port) => {
+          setIsFetching(true)
+          setUrl(undefined)
+          setTimeout(() => {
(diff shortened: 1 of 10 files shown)


DorianZheng:
will be merged after ci succeed and coderabiit issues resolved

apps/dashboard/src/components/boxes/BoxPreviewUrlDialog.stories.tsx
```diff
@@ -0,0 +1,81 @@
+/*
+ * Copyright 2026 BoxLite AI
+ * SPDX-License-Identifier: AGPL-3.0
+ */
+
+import type { Meta, StoryObj } from '@storybook/react'
+import { useState } from 'react'
+import { BoxPreviewUrlDialog } from './BoxPreviewUrlDialog'
```
DorianZheng, line 8: should be `BoxTunnelUriDialog`

PR #1427, opened 2026-09-01
fix(api): destroy and reap boxes whose creation never finished

G4614:
Destroys the box record when creation fails to start, and reaps the boxes that path cannot reach — a caller timeout (box still CREATING and pending, which `destroy()` refuses) or the API process dying mid-create.

`cleanupStaleErrorBoxes` never covered these: it only matches boxes already at `desiredState = DESTROYED`, while a box that never came up sits at `STARTED`, so it matched nothing and stayed forever. The new `reapFailedBoxStartups` cron decides from persisted state alone, and is deliberately conservative about what it condemns:

- an open `CREATE_BOX` job means hands off — declaring a create dead belongs to `job.service`'s stale-job timeout, not to this sweep;
- `CREATING` is reaped only after `boxSync.failedStartupReapSeconds` (30m default, well past that timeout);
- errored boxes keep their existing 7 day tombstone window, so `errorReason` and `recoverable` stay readable;
- a completed `CREATE_BOX` job exempts the box — one that came up and failed later stays the user's to delete;
- each sweep is capped at 100, oldest first, and holds the same state-change lock as the sync loop.

It soft-deletes through `desiredState = DESTROYED` rather than deleting rows, so the destroy action still releases whatever the runner allocated.

Test plan:
- [x] `jest src/box src/boxlite-rest src/config` — 389 passed, 0 failed (10 new cases for `reapFailedBoxStartups`)
- [x] `tsc --noEmit` on `tsconfig.app.json` and `tsconfig.spec.json` — clean
- [x] `eslint` + `prettier --check` on the changed files
- [ ] Not run: staging soak against a real stalled create (needs a runner that hangs mid-create)

## Summary by CodeRabbit

- **Bug Fixes**
  - Failed box startups are now cleaned up automatically, preventing orphaned boxes from remaining in the box list.
  - Stalled or errored box startups are identified and marked for destruction.
  - Box startup cleanup continues even if an individual cleanup operation encounters an error.
  - Boxes with active or progressing startup jobs are left unchanged.
  - Errors from failed startups are still reported after cleanup is attempted.
  - Cleanup safely handles concurrent state changes and limits each sweep to the oldest affected boxes.

### DIFF
--- a/apps/api/src/box/services/box.service.spec.ts
+++ b/apps/api/src/box/services/box.service.spec.ts
@@ -372,8 +372,9 @@ describe('BoxService public defaults', () => {
 })
 
 // The reaper only touches boxRepository + jobRepository + configService +
-// eventEmitter, and decides purely from persisted state — which is what makes
-// it able to clean up after a request, or a whole API process, that is gone.
+// eventEmitter + redisLockProvider, and decides purely from persisted state —
+// which is what makes it able to clean up after a request, or a whole API
+// process, that is gone.
 function makeReaperService() {
   const boxRepository = {
     find: jest.fn().mockResolvedValue([]),
@@ -381,15 +382,15 @@ function makeReaperService() {
   } as any
   const jobRepository = {
     count: jest.fn().mockResolvedValue(0),
-    find: jest.fn().mockResolvedValue([]),
   } as any
   const configService = {
     getOrThrow: jest.fn((key: string) => {
-      if (key === 'boxSync.startConfirmationStallSeconds') return 60
+      if (key === 'boxSync.failedStartupReapSeconds') return 1800
       throw new Error(`unexpected config key ${key}`)
     }),
   } as any
   const eventEmitter = { emit: jest.fn(), emitAsync: jest.fn() } as any
+  const redisLockProvider = { lock: jest.fn().mockResolvedValue(true), unlock: jest.fn() } as any
   const noop = {} as any
   const service = new BoxService(
     boxRepository, // boxRepository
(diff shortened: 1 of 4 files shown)


DorianZheng:
what if the api crash, there are still orphaned boxes

G4614:
Good catch — fixed in 7a9c7a34, and it turned out the try/catch was weaker than I thought.

Two things the request-path cleanup cannot do:
- **API crashes mid-create** — your case; nothing runs the catch.
- **Caller times out** — the box is still `CREATING` with `pending = true` (set by `create()`), and `destroy()` rejects a pending box, so the catch only logged a warning.

Neither was covered by `cleanupStaleErrorBoxes`: it requires `desiredState = DESTROYED`, but `box.manager` sets `state = ERROR` without touching `desiredState`, so a failed-to-start box sits at `STARTED` and matched nothing — it was leaking permanently, not for 7 days.

Added `reapFailedBoxStartups` (every minute), which decides from persisted state only:
- `CREATING` older than `startConfirmationStallSeconds` with no `CREATE_BOX` job that could still move it (unclaimed or recently claimed jobs are left alone);
- `ERROR` with no completed `CREATE_BOX` job — so a box that came up and errored later stays visible and yours to delete.

It soft-deletes through `desiredState = DESTROYED` rather than deleting the row, so the destroy action still releases whatever the runner allocated.

Tested: `jest src/box src/boxlite-rest` — 314 passed, 7 new cases. Not run: a staging soak against a genuinely stalled create.

PR #1428, opened 2026-09-01
test(e2e): verify cross-org resource isolation

G4614:
Adds `apps/e2e/cases/test_cross_org_isolation.py` — end-to-end coverage proving one organisation's credentials cannot observe or manipulate another's — and wires the second-org key into `e2e-cloud.yml` so the module actually runs.

Test-only change; no production code is touched.

Two vectors are asserted separately because they fail differently: a request into org-A's namespace is refused by `OrganizationAccessGuard` (401/403), while a request into org-B's own namespace carrying org-A's box id passes the guard and misses the org-scoped lookup (404). Only the first catches a regression exposing org-A's resources inside org-A's URL space; only the second proves the API does not leak their existence.

**This PR needs the repository secret `BOXLITE_E2E_CROSS_ORG_API_KEY` added before it verifies anything in CI.** Without it the module skips — announced on stderr locally and as a `::warning::` in the workflow, so lost coverage does not read as a green run.

Test plan:
- [x] `BOXLITE_E2E_CROSS_ORG_API_KEY=<key> BOXLITE_E2E_PROFILE=p1 BOXLITE_E2E_SKIP_PATH_VERIFY=1 python3 -m pytest apps/e2e/cases/test_cross_org_isolation.py -v` against `api.dev.boxlite.ai` with two real orgs — 8 passed
- [x] Mutation check on the positive control: binding the marker service to a port the tunnel does not read makes `test_cross_org_cannot_tunnel_into_org_a_box` fail loudly (`Response: b''`) rather than pass on empty output
- [x] Key unset — 8 skipped, exit 0, reason on stderr
- [x] Key unset plus `BOXLITE_E2E_REQUIRE_CROSS_ORG=1` — collection error, exit 2
- [ ] CI run of `e2e-cloud.yml` — blocked until the repository secret exists

`BOXLITE_E2E_SKIP_PATH_VERIFY=1` was needed locally only because the runner journal that `verify_runner_saw_all_boxes` reads is not reachable from a laptop against a remote deployment; CI does not set it.

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for cross-organization isolation.
  * Verified that credentials, boxes, deletions, and network tunnels cannot be used across organization boundaries.
  * Improved tunnel readiness checks for more reliable test results.
  * Added configuration validation so missing cross-organization test credentials are reported clearly when required.

### DIFF
--- a/apps/e2e/cases/test_cross_org_isolation.py
+++ b/apps/e2e/cases/test_cross_org_isolation.py
@@ -1,8 +1,7 @@
-"""POL-257 cross-org isolation tests (automated).
+"""Cross-organisation resource isolation tests.
 
 Verifies that one organisation's credentials cannot observe or manipulate
-another organisation's resources — the launchgate security requirement tracked
-in POL-257 (cases 257-1c, 257-3, 257-3a, 257-3b, 257-4).
+another organisation's resources.
 
 ## Setup
 
@@ -20,15 +19,15 @@ secret injection; locally use a second boxlite account.
 
 ## What is tested
 
-* 257-1c  A valid key from org-B is rejected (401) for org-A's namespace.
-* 257-3   GET  /v1/{prefix-A}/boxes/{box-A-id}  with org-B key → 404.
-* 257-3a  DELETE /v1/{prefix-A}/boxes/{box-A-id} with org-B key → 404.
-* 257-3b  POST /v1/{prefix-A}/boxes/{box-A-id}/exec with org-B key → 404.
-* 257-4   A box launched by org-B cannot reach org-A's box on its internal
-          address (bidirectional network isolation).
+* A valid key from org-B is rejected (401) for org-A's namespace.
+* GET  /v1/{prefix-A}/boxes/{box-A-id}  with org-B key → 404.
+* DELETE /v1/{prefix-A}/boxes/{box-A-id} with org-B key → 404; box survives.
+* POST /v1/{prefix-A}/boxes/{box-A-id}/exec with org-B key → 404.
+* A box launched by org-B cannot reach org-A's box on its internal address
+  (bidirectional network isolation).
 
-The 404 (not 403/401) requirement for 257-3/3a/3b is deliberate: the API must
-not leak the existence of resources belonging to another org.
+The 404 (not 403/401) requirement for cross-org resource access is deliberate:
+the API must not leak the existence of resources belonging to another org.
 """
 from __future__ import annotations
 
@@ -184,16 +183,16 @@ async def org_a_box(org_a_rt, image):
 
 
 # ---------------------------------------------------------------------------
-# 257-1c  A valid org-B key is rejected for org-A's namespace
+# Tests
 # ---------------------------------------------------------------------------
 
 @pytest.mark.asyncio
-async def test_257_1c_cross_org_key_rejected_for_other_org_namespace(
+async def test_cross_org_key_rejected_for_other_org_namespace(
     org_a_ctx: E2EAuthContext,
     org_b_token: str,
     org_a_box,
 ):
-    """org-B's API key must not be accepted in org-A's URL namespace (257-1c).
+    """org-B's API key must not be accepted in org-A's URL namespace.
 
     The endpoint under test is org-A's box list.  A different-org key must get
     401 — not 200 — because the key resolves to a different tenant scope.
@@ -206,25 +205,19 @@ async def test_257_1c_cross_org_key_rejected_for_other_org_namespace(
     )
 
 
-# ---------------------------------------------------------------------------
-# 257-3  Cross-org GET on another org's box → 404
-# ---------------------------------------------------------------------------
-
 @pytest.mark.asyncio
-async def test_257_3_cross_org_get_box_returns_404(
+async def test_cross_org_get_box_returns_404(
     org_a_ctx: E2EAuthContext,
     org_b_token: str,
     org_b_prefix: str,
     org_a_box,
 ):
-    """GET /v1/{prefix-B}/boxes/{box-A-id} with org-B token → 404 (257-3).
+    """GET /v1/{prefix-B}/boxes/{box-A-id} with org-B token → 404.
 
     The response must be 404, not 403/401, to avoid leaking the existence of
     org-A's resources to org-B.
     """
     box_id = org_a_box.id
-    # Use org-B's prefix in the URL so the key is valid for the route, but
-    # request a box that belongs to org-A.
     if org_b_prefix:
         url = org_a_ctx.url_for(f"/v1/{org_b_prefix}/boxes/{box_id}")
     else:
@@ -236,19 +229,15 @@ async def test_257_3_cross_org_get_box_returns_404(
     )
 
 
-# ---------------------------------------------------------------------------
-# 257-3a  Cross-org DELETE on another org's box → 404
-# ---------------------------------------------------------------------------
-
 @pytest.mark.asyncio
-async def test_257_3a_cross_org_delete_box_returns_404(
+async def test_cross_org_delete_box_returns_404(
     org_a_ctx: E2EAuthContext,
     org_b_token: str,
     org_b_prefix: str,
     org_a_box,
     org_a_rt,
 ):
-    """DELETE /v1/{prefix-B}/boxes/{box-A-id} with org-B token → 404 (257-3a).
+    """DELETE /v1/{prefix-B}/boxes/{box-A-id} with org-B token → 404.
 
     The box must still be alive in org-A after the attempt.
     """
@@ -267,18 +256,14 @@ async def test_257_3a_cross_org_delete_box_returns_404(
     assert info is not None, "org-A's box disappeared after a cross-org DELETE attempt"
 
 
-# ---------------------------------------------------------------------------
-# 257-3b  Cross-org exec on another org's box → 404
-# ---------------------------------------------------------------------------
-
 @pytest.mark.asyncio
-async def test_257_3b_cross_org_exec_returns_404(
+async def test_cross_org_exec_returns_404(
     org_a_ctx: E2EAuthContext,
     org_b_token: str,
     org_b_prefix: str,
     org_a_box,
 ):
-    """POST …/exec on org-A's box with org-B token → 404 (257-3b)."""
+    """POST …/exec on org-A's box with org-B token → 404."""
     box_id = org_a_box.id
     if org_b_prefix:
         url = org_a_ctx.url_for(f"/v1/{org_b_prefix}/boxes/{box_id}/exec")
@@ -293,24 +278,18 @@ async def test_257_3b_cross_org_exec_returns_404(
     )
 
 
-# ---------------------------------------------------------------------------
-# 257-4  Network isolation: org-B box cannot reach org-A box internally
-# ---------------------------------------------------------------------------
-
 @pytest.mark.asyncio
-async def test_257_4_cross_org_network_isolation(
+async def test_cross_org_network_isolation(
     image,
     org_a_rt,
     org_b_rt,
 ):
-    """Boxes in different orgs cannot reach each other over the internal
-    network (257-4).
+    """Boxes in different orgs cannot reach each other over the internal network.
 
     Start a simple HTTP server in an org-A box; attempt to connect to it from
-    an org-B box using the box-id-based DNS pattern and common private
-    addresses. All attempts must time out or fail.
+    an org-B box using the box-id-based DNS pattern. All attempts must fail.
     """
-    PORT = 9527  # arbitrary; unlikely to conflict
+    PORT = 9527
 
     box_a = box_b = None
     try:
@@ -321,7 +300,6 @@ async def test_257_4_cross_org_network_isolation(
             boxlite.BoxOptions(image=image, auto_delete=300, auto_stop=120),
         )
 
-        # Start a trivial HTTP server in org-A's box.
         await box_a.exec(
             "/bin/sh",
             ["-c", f"printf 'HTTP/1.0 200 OK\\r\\n\\r\\nORG_A_DATA' | nc -l -p {PORT} &>/dev/null &"],
@@ -329,8 +307,6 @@ async def test_257_4_cross_org_network_isolation(
         )
         time.sleep(1)
 
-        # Try to reach it from org-B's box via several address patterns.
-        # All should fail (DNS resolution failure or connection refused).
         reach_attempts = [
             f"http://{box_a.id}.box:{PORT}/",
             f"http://{box_a.id}.box.internal:{PORT}/",
@@ -338,12 +314,7 @@ async def test_257_4_cross_org_network_isolation(
         for target_url in reach_attempts:
             result = await box_b.exec(
                 "/bin/sh",
-                [
-                    "-c",
-                    (
-                        f"curl -s --max-time 3 {target_url!r} 2>&1 || echo 'BLOCKED'"
-                    ),
-                ],
+                ["-c", f"curl -s --max-time 3 {target_url!r} 2>&1 || echo 'BLOCKED'"],
                 None,
             )
             import asyncio
@@ -352,8 +323,6 @@ async def test_257_4_cross_org_network_isolation(
                 out_chunks.append(chunk)
             out = b"".join(out_chunks).decode("utf-8", "replace")
 
-            # Either the connection fails (BLOCKED / curl exit non-zero) or
-            # org-A's distinct marker never appears.
             assert "ORG_A_DATA" not in out, (
                 f"org-B box reached org-A box at {target_url}: "
                 f"network isolation breach. Output: {out!r}"
--- a/apps/e2e/cases/test_cross_org_isolation.py
+++ b/apps/e2e/cases/test_cross_org_isolation.py
@@ -169,17 +169,30 @@ async def org_b_rt(org_a_ctx: E2EAuthContext, org_b_token: str, org_b_prefix: st
             pass
 
 
-@pytest_asyncio.fixture(scope="module")
-async def org_a_box(org_a_rt, image):
-    """A running box in org-A, shared across this module's tests."""
-    box = await org_a_rt.create(
-        boxlite.BoxOptions(image=image, auto_delete=600, auto_stop=300),
+@pytest.fixture(scope="module")
+def org_a_box(org_a_ctx: E2EAuthContext, image: str):
+    """A running box in org-A, shared across this module's tests.
+
+    Created via raw HTTP so this fixture is independent of SDK version.
+    Returns a simple namespace with an ``.id`` attribute.
+    """
+    import types
+
+    # Create
+    url = org_a_ctx.url_for(org_a_ctx.v1("boxes"))
+    status, body = _request(
+        "POST", url, token=org_a_ctx.token,
+        body={"image": image, "auto_delete": 600, "auto_stop": 300},
     )
+    assert status == 201, f"Failed to create org-A box: {status} {body}"
+    box_id = body["box_id"]
+
+    box = types.SimpleNamespace(id=box_id)
     yield box
-    try:
-        await org_a_rt.remove(box.id, force=True)
-    except Exception:
-        pass
+
+    # Cleanup
+    del_url = org_a_ctx.url_for(org_a_ctx.v1(f"boxes/{box_id}"))
+    _request("DELETE", del_url, token=org_a_ctx.token)
 
 
 # ---------------------------------------------------------------------------
@@ -199,9 +212,9 @@ async def test_cross_org_key_rejected_for_other_org_namespace(
     """
     url = org_a_ctx.url_for(org_a_ctx.v1("boxes"))
     status, _ = _request("GET", url, token=org_b_token)
-    assert status == 401, (
-        f"Expected 401 when using org-B key in org-A's namespace, got {status}. "
-        "A valid key from another org must not be accepted in a foreign namespace."
+    assert status in (401, 403), (
+        f"Expected 401 or 403 when using org-B key in org-A's namespace, got {status}. "
+        "A valid key from another org must be rejected in a foreign namespace."
     )
 
 
@@ -235,7 +248,6 @@ async def test_cross_org_delete_box_returns_404(
     org_b_token: str,
     org_b_prefix: str,
     org_a_box,
-    org_a_rt,
 ):
     """DELETE /v1/{prefix-B}/boxes/{box-A-id} with org-B token → 404.
 
@@ -252,8 +264,9 @@ async def test_cross_org_delete_box_returns_404(
     )
 
     # Box must still be accessible by org-A (org-B's DELETE had no effect).
-    info = await org_a_rt.get_info(box_id)
-    assert info is not None, "org-A's box disappeared after a cross-org DELETE attempt"
+    get_url = org_a_ctx.url_for(org_a_ctx.v1(f"boxes/{box_id}"))
+    get_status, _ = _request("GET", get_url, token=org_a_ctx.token)
+    assert get_status == 200, "org-A's box disappeared after a cross-org DELETE attempt"
 
 
 @pytest.mark.asyncio
@@ -294,10 +307,10 @@ async def test_cross_org_network_isolation(
     box_a = box_b = None
     try:
         box_a = await org_a_rt.create(
-            boxlite.BoxOptions(image=image, auto_delete=300, auto_stop=120),
+            boxlite.BoxOptions(image=image, auto_remove=True),
         )
         box_b = await org_b_rt.create(
-            boxlite.BoxOptions(image=image, auto_delete=300, auto_stop=120),
+            boxlite.BoxOptions(image=image, auto_remove=True),
         )
 
         await box_a.exec(
--- a/apps/e2e/cases/test_cross_org_isolation.py
+++ b/apps/e2e/cases/test_cross_org_isolation.py
@@ -307,6 +307,41 @@ async def _fetch_marker_over_tunnel(box, port: int) -> bytes:
         await connection.close()
 
 
+async def _await_marker_over_tunnel(box, port: int, *, timeout: float = 30.0) -> bytes:
+    """Poll `_fetch_marker_over_tunnel` until the marker shows up.
+
+    `_start_marker_service` returns as soon as `sh` reports the background
+    pid, which is before `python3` has bound *port*.  A single fetch can
+    therefore connect to a closed port and read nothing, failing the positive
+    control for a timing reason rather than a reachability one.
+
+    The last response and the last error are both carried out so a genuine
+    failure (tunnel refused, service crashed) stays diagnosable instead of
+    surfacing as an empty read.
+    """
+    deadline = asyncio.get_running_loop().time() + timeout
+    response = b""
+    last_error: BaseException | None = None
+    while True:
+        try:
+            response = await _fetch_marker_over_tunnel(box, port)
+            last_error = None
+            if _TUNNEL_MARKER in response:
+                return response
+        except (OSError, asyncio.TimeoutError, RuntimeError) as exc:
+            last_error = exc
+            response = b""
+        if asyncio.get_running_loop().time() >= deadline:
+            break
+        await asyncio.sleep(0.5)
+    if last_error is not None:
+        raise AssertionError(
+            f"marker service in box {box.id} never became reachable over its "
+            f"own tunnel within {timeout}s; last error: {last_error!r}"
+        ) from last_error
+    return response
+
+
 # ---------------------------------------------------------------------------
 # Tests — foreign namespace (403)
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


apps/e2e/cases/test_cross_org_isolation.py
```diff
@@ -0,0 +1,349 @@
...
+import pytest_asyncio
+
+sys.path.insert(0, str(Path(__file__).parent.parent / "lib"))
+from e2e_auth import auth_context, E2EAuthContext
+
+# ---------------------------------------------------------------------------
+# Skip the whole module when cross-org credentials are absent
+# ---------------------------------------------------------------------------
+
+_CROSS_ORG_KEY = os.environ.get("BOXLITE_E2E_CROSS_ORG_API_KEY", "").strip()
```
DefeatMan, line 78: **This module never runs anywhere.**

`BOXLITE_E2E_CROSS_ORG_API_KEY` is set nowhere in the repository — grepping `.github/`, `apps/e2e/run.sh`, `bootstrap.sh`, and `fixture_setup.py` for `CROSS_ORG` returns nothing. The `skipif` on this constant fires unconditionally, and CI on this PR shows all five tests SKIPPED.

The PR therefore adds zero executed coverage while reporting green, and the skip is silent — which is why none of the other issues in this file were caught.

Please wire the key into the e2e workflow, and consider making the absence loud (fail the job, or gate on an explicit `--run-cross-org` flag) so this can't regress into a silent skip again.

apps/e2e/cases/test_cross_org_isolation.py
```diff
@@ -0,0 +1,349 @@
...
+def _discover_prefix(url: str, token: str) -> str:
+    """Return path_prefix for *token* by calling /v1/me, or '' on failure."""
+    req = urllib.request.Request(
+        f"{url.rstrip('/')}/v1/me",
+        method="GET",
+        headers={"Authorization": f"Bearer {token}"},
+    )
+    try:
+        with urllib.request.urlopen(req, timeout=15) as resp:
+            return (json.loads(resp.read() or "null") or {}).get("path_prefix", "")
```
DefeatMan: `.get("path_prefix", "")` returns `None`, not `""`, when the server sends an explicit `"path_prefix": null` — which the OpenAPI prefix parameter allows. That yields URLs like `/v1/None/boxes/{id}`.

Separately, the `except Exception: return ""` on L78-79 turns an expired org-B key, a 500, or a DNS failure into a silent empty prefix, after which every test in this module quietly falls back to `/v1/boxes/{id}` and stops testing anything.

The shared helper `e2e_auth.discover_path_prefix` already does the right thing here: it raises `RuntimeError(f"GET /v1/me failed with HTTP {exc.code}")` and uses `or ""` rather than a default. Suggest reusing it instead of this local copy.

apps/e2e/cases/test_cross_org_isolation.py
```diff
@@ -0,0 +1,349 @@
...
+    data: bytes | None = None
+    if body is not None:
+        headers["Content-Type"] = "application/json"
+        data = json.dumps(body).encode()
+    req = urllib.request.Request(url, method=method, headers=headers, data=data)
+    try:
+        with urllib.request.urlopen(req, timeout=timeout) as resp:
+            raw = resp.read()
+            return resp.status, (json.loads(raw) if raw else None)
+    except urllib.error.HTTPError as exc:
```
DefeatMan, line 143: `_request` catches only `urllib.error.HTTPError`. A `URLError` or socket timeout escapes uncaught.

Both call sites are inside the module-scoped `org_a_box` fixture, so a single transient reset or a >30s stall on `POST /v1/{prefixA}/boxes` raises out of the fixture *before* `yield` (L191): pytest reports four fixture ERRORs with a raw urllib traceback, and because post-yield cleanup never runs, an already-created box is orphaned. The same escape on the teardown call (L195) turns module teardown into an ERROR.

Suggest catching `urllib.error.URLError` / `TimeoutError` as well, and moving box creation into a `try/finally` that can clean up a partially-created box.

apps/e2e/cases/test_cross_org_isolation.py
```diff
@@ -0,0 +1,349 @@
...
+
+@pytest.fixture(scope="module")
+def org_b_prefix(org_a_ctx: E2EAuthContext, org_b_token: str) -> str:
+    explicit = os.environ.get("BOXLITE_E2E_CROSS_ORG_PREFIX", "").strip()
+    if explicit:
+        return explicit
+    return _discover_prefix(org_a_ctx.url, org_b_token)
+
+
+@pytest_asyncio.fixture(scope="module")
```
DefeatMan: This fixture (and `org_b_rt` at L152, and the raw-HTTP creation at L183) calls `boxlite.Boxlite.rest(opts)` directly, bypassing conftest's `_TrackingRuntime` wrapper (`conftest.py:33-55`).

That wrapper exists so the autouse `verify_runner_saw_all_boxes` fixture can prove each box actually reached the runner, failing the test with a path-bypass error if the SDK silently degrades to local FFI. With the raw runtime, `rt._created` stays empty and `conftest.py:113-114` returns early — so boxes created by this module are never path-verified.

Suggest going through the same tracking wrapper the rest of the suite uses.

apps/e2e/cases/test_cross_org_isolation.py
```diff
@@ -0,0 +1,349 @@
...
+    org_a_ctx: E2EAuthContext,
+    org_b_token: str,
+    org_a_box,
+):
+    """org-B's API key must not be accepted in org-A's URL namespace.
+
+    The endpoint under test is org-A's box list.  A different-org key must get
+    401 — not 200 — because the key resolves to a different tenant scope.
+    """
+    url = org_a_ctx.url_for(org_a_ctx.v1("boxes"))
```
DefeatMan: This test has no guard that org-A actually has a non-empty path prefix.

When `path_prefix` is empty, `org_a_ctx.v1("boxes")` returns `/v1/boxes` (`e2e_auth.py:68-72`). With no `:prefix` route param, `OrganizationAccessGuard`'s mismatch branch never fires — `organizationIdParam` is undefined, so it falls through to `authContext.organizationId` (org-B) and returns org-B's own box list with 200, failing `status in (401, 403)`.

Nothing in the module asserts that this is a multi-tenant deployment. Suggest a `pytest.skip` when `not org_a_ctx.path_prefix`, so the test is explicit about the deployment shape it requires.

apps/e2e/cases/test_cross_org_isolation.py
```diff
@@ -0,0 +1,349 @@
...
+    org_a_box,
+):
+    """GET /v1/{prefix-B}/boxes/{box-A-id} with org-B token → 404.
+
+    The response must be 404, not 403/401, to avoid leaking the existence of
+    org-A's resources to org-B.
+    """
+    box_id = org_a_box.id
+    if org_b_prefix:
+        url = org_a_ctx.url_for(f"/v1/{org_b_prefix}/boxes/{box_id}")
```
DefeatMan: **This does not exercise the cross-org vector the PR describes.**

The URL is built from `org_b_prefix` — org-B's key pointed at org-B's *own* namespace. `boxlite-box.controller.ts:145` scopes lookups by `authContext.organizationId`, so a foreign box id is simply not found; this 404 is the same 404 any random string would produce.

The PR body and the module docstring (L29-31) both state the tests hit `/v1/{prefix-A}/boxes/{box-A-id}`, which *is* the interesting vector — a regression exposing org-A's boxes inside org-A's namespace would not be caught by what's written here.

Same issue at L258 and L282. Suggest `org_a_ctx.path_prefix` in the URL with org-B's credentials on the request.

apps/e2e/cases/test_cross_org_isolation.py
```diff
@@ -0,0 +1,349 @@
...
+        url = org_a_ctx.url_for(f"/v1/boxes/{box_id}")
+    status, _ = _request("DELETE", url, token=org_b_token)
+    assert status == 404, (
+        f"Expected 404 when org-B deletes org-A's box {box_id}, got {status}."
+    )
+
+    # Box must still be accessible by org-A (org-B's DELETE had no effect).
+    get_url = org_a_ctx.url_for(org_a_ctx.v1(f"boxes/{box_id}"))
+    get_status, _ = _request("GET", get_url, token=org_a_ctx.token)
+    assert get_status == 200, "org-A's box disappeared after a cross-org DELETE attempt"
```
DefeatMan: **This assertion is vacuous — it passes whether or not the cross-org DELETE was honoured.**

If a regression lets org-B's DELETE actually destroy org-A's box, `boxService.destroy()` applies `Box.getSoftDeleteUpdate(box)` = `{pending: true, desiredState: DESTROYED, name: 'DESTROYED_...'}` (`box.entity.ts:209-215`) — it never sets `state`. The GET here runs `findOneByIdOrName`, whose filter is `state: Not(BoxState.DESTROYED)` (`box.service.ts:654`). The row's state is still `started`, so it returns 200 and this assertion passes.

So the single assertion guarding against a destructive cross-org breach reports "isolation holds" on exactly the input where org-A's box was just destroyed.

There's a second, compounding problem: if the DELETE does succeed, the `assert status == 404` at L262 fires first and this check never runs at all — and the now-destroyed module-scoped box makes the later exec test 404 for the wrong reason.

Suggest asserting on `pending` / `desiredState` / `name` from the GET response body instead of on the status code.

apps/e2e/cases/test_cross_org_isolation.py
```diff
@@ -0,0 +1,349 @@
...
+):
+    """POST …/exec on org-A's box with org-B token → 404."""
+    box_id = org_a_box.id
+    if org_b_prefix:
+        url = org_a_ctx.url_for(f"/v1/{org_b_prefix}/boxes/{box_id}/exec")
+    else:
+        url = org_a_ctx.url_for(f"/v1/boxes/{box_id}/exec")
+    status, _ = _request(
+        "POST", url, token=org_b_token,
+        body={"cmd": ["/bin/sh", "-c", "id"]},
```
DefeatMan: `{"cmd": [...]}` is not a valid `ExecRequest`. `openapi/box.openapi.yaml` declares `required: [command]` with `command: string` and `args: array` — there is no `cmd` field — and `src/cli/src/commands/serve/types.rs` mirrors it with `pub command: String`.

Against `boxlite serve`, the axum `Json` extractor rejects this body with 422 before any authorization check runs, so `assert status == 404` fails. Against the NestJS target, the proxy controller has no `ValidationPipe` and 404s at the ownership lookup first — so the test passes, but the malformed body proves the request never reached exec at all.

Either way this test is not validating exec authorization. Suggest `{"command": "/bin/sh", "args": ["-c", "id"]}`.

apps/e2e/cases/test_cross_org_isolation.py
```diff
@@ -0,0 +1,349 @@
...
+    """Boxes in different orgs cannot reach each other over the internal network.
+
+    Start a simple HTTP server in an org-A box; attempt to connect to it from
+    an org-B box using the box-id-based DNS pattern. All attempts must fail.
+    """
+    PORT = 9527
+
+    box_a = box_b = None
+    try:
+        box_a = await org_a_rt.create(
```
DefeatMan: The two boxes are never pinned to the same runner host.

The scheduler assigns `box.runnerId` per box (`box.manager.ts` selects from `runnerService.findAllReady()`), so on a multi-runner deployment org-A's and org-B's boxes land on different physical machines. Whatever the guest network policy is, `box_b` cannot reach `box_a`, the probe fails, and the assertion below passes — for reasons that have nothing to do with isolation.

The interesting case is precisely the one this can't reach: both orgs co-tenant on one runner sharing a host bridge.

Suggest pinning both boxes to the same runner, or at minimum reading `runnerId` from both and `pytest.skip`-ing with an explicit message when they differ, so a pass can't be mistaken for coverage.

apps/e2e/cases/test_cross_org_isolation.py
```diff
@@ -0,0 +1,349 @@
...
+            boxlite.BoxOptions(image=image, auto_remove=True),
```
DefeatMan: `BoxOptions(image=image, auto_remove=True)` sends neither `auto_stop` nor `auto_delete`, so the API applies `DEFAULT_AUTO_STOP_SECONDS = 900` and `AUTO_DELETE_DISABLED = 0` — server-side reaping is off.

That matters because `POST /boxes` persists the row *before* `createBox` calls `boxStateWaiter.waitForStarted(box.id, org.id, 30)`. A cold image pull exceeding 30s raises `RequestTimeoutException`; `create()` then raises, `box_a` stays `None`, the `finally` skips it (`if box is not None`), and the box is left running. It idles, stops, and is never reaped.

The module's own `org_a_box` fixture sets `auto_delete: 600` for exactly this reason (L185), so this call site contradicts it. Suggest passing `auto_delete` here too.

apps/e2e/cases/test_cross_org_isolation.py
```diff
@@ -0,0 +1,349 @@
...
+        )
+        box_b = await org_b_rt.create(
+            boxlite.BoxOptions(image=image, auto_remove=True),
+        )
+
+        await box_a.exec(
+            "/bin/sh",
+            ["-c", f"printf 'HTTP/1.0 200 OK\\r\\n\\r\\nORG_A_DATA' | nc -l -p {PORT} &>/dev/null &"],
```
DefeatMan: This listener command is broken in two independent ways, so it never starts.

1. `nc` is not in the image. `apps/box-images/base.Dockerfile` is `debian:bookworm-slim` and installs bash, ca-certificates, curl, git, jq, less, openssh-client, procps, python3, sudo, tzdata, unzip, wget, zip — no netcat, no busybox. The command fails with `sh: 1: nc: not found`. (`curl`, used on L324, *is* present.)
2. `&>` is a bashism, and the program here is `/bin/sh` — dash on Debian. Dash does not parse `cmd &>/dev/null &` as redirect-all; it parses `&` as backgrounding, then `>/dev/null &` separately.

`python3` is available in the image, so a one-liner `python3 -m http.server` or a small socket script would be more robust than netcat. And `>/dev/null 2>&1 &` for the redirect.

apps/e2e/cases/test_cross_org_isolation.py
```diff
@@ -0,0 +1,349 @@
...
+            None,
+        )
+        time.sleep(1)
+
+        reach_attempts = [
+            f"http://{box_a.id}.box:{PORT}/",
```
DefeatMan: `<box-id>.box` and `<box-id>.box.internal` are not real hostnames — a repo-wide grep for them returns only this file.

`src/guest/src/container/start.rs:71` writes a constant hostname `boxlite` and an `/etc/hosts` with no per-box entries, and `src/boxlite/src/net/constants.rs` pins `GUEST_IP` to 192.168.127.2 for every box. Both URLs fail DNS resolution unconditionally, so `curl` returns nothing regardless of whether cross-org networking is actually isolated — which the assertion below then reads as success.

This needs a genuinely routable target address for the probe to mean anything.

DefeatMan (commented):
## Review: cross-org isolation e2e tests

Thanks for taking on the cross-org vector — this is exactly the kind of coverage the repo needs. The problem is that as written, this suite provides no security assurance while reporting green, so I'd like to work through the issues below before it lands.

**Root cause.** `BOXLITE_E2E_CROSS_ORG_API_KEY` is not set anywhere in the repository — not in `.github/`, `apps/e2e/run.sh`, `bootstrap.sh`, or `fixture_setup.py`. The module-level `pytestmark = pytest.mark.skipif(...)` therefore always fires, and CI on this PR shows all five tests SKIPPED. The module has never executed, which is why none of the defects below surfaced.

The findings fall into three groups:

**1. Would error immediately once the key is wired up** — `result.stdout` accessed as a property when the SDK exposes it as a method; `b"".join()` over a `str` stream; `nc` not present in the test image; `&>` is a bashism under dash; `<box-id>.box` hostnames that exist nowhere in the codebase; `{"cmd": [...]}` is not a valid `ExecRequest`.

**2. Would pass even when a real vulnerability exists** — three independent mechanisms:
- The delete test's survival check is vacuous, because the soft delete never writes `state` (details inline at L269).
- The network test has no positive control, so empty output satisfies the assertion.
- The two boxes are never pinned to the same runner, so on multi-runner deployments the test passes for the wrong reason.

**3. Does not exercise the advertised vector** — all three 404 tests build URLs from org-B's *own* prefix, so they only prove "an unknown id in my own namespace 404s". The PR description and the module docstring (L29-31) both say the tests hit `/v1/{prefix-A}/boxes/{box-A-id}`; the code does not. Either the code or the description needs to change — I'd suggest the code, since the description names the actual attack vector.

### Suggested order of work

1. Wire the key into CI, and make the skip loud (or make the job required) — a silently skipped security test is worse than no test.
2. Fix the URL construction to org-B key + org-A prefix + org-A box id.
3. Replace the delete test's `GET == 200` liveness check with assertions on `pending` / `desiredState` / `name`.
4. Add a positive control to the network test, and pin or assert co-tenancy on one runner.
5. Reuse the existing helpers rather than the hand-rolled ones: `e2e_auth.discover_path_prefix`, conftest's `_TrackingRuntime`, `collect_stream`.

### One meta-suggestion

The only property that matters for a security test is that **it goes red when the vulnerability is present**. Before merging, please run a mutation check: temporarily disable `OrganizationAccessGuard` and confirm all five tests fail. As it stands they stay green with the guard removed — that check is worth more than any individual fix above.

Separately, and outside the scope of this PR: `Box.getSoftDeleteUpdate` (`box.entity.ts:209-215`) sets `pending` / `desiredState` / `name` but never `state`, while `findOneByIdOrName` filters on `state: Not(BoxState.DESTROYED)` (`box.service.ts:654`). A soft-deleted box therefore still returns 200 from GET. That may well be intentional, but it's worth a second look since it's what makes the test above vacuous.

apps/e2e/cases/test_cross_org_isolation.py
```diff
@@ -0,0 +1,349 @@
...
+        ]
+        for target_url in reach_attempts:
+            result = await box_b.exec(
+                "/bin/sh",
+                ["-c", f"curl -s --max-time 3 {target_url!r} 2>&1 || echo 'BLOCKED'"],
+                None,
+            )
+            import asyncio
+            out_chunks: list[bytes] = []
+            async for chunk in result.stdout:
```
DefeatMan: `result.stdout` is a method, not a property. `sdks/python/src/exec.rs` defines `fn stdout(&self)` under `#[pymethods]` with no `#[getter]` — contrast `PyBox::id`, which does have one.

So `async for chunk in result.stdout:` iterates a bound method and raises `TypeError: 'builtin_function_or_method' object is not async iterable` before any assertion runs.

Every other e2e test calls it correctly — e.g. `conftest.py:173` uses `collect_stream(ex.stdout())`.

apps/e2e/cases/test_cross_org_isolation.py
```diff
@@ -0,0 +1,349 @@
...
+                out_chunks.append(chunk)
+            out = b"".join(out_chunks).decode("utf-8", "replace")
```
DefeatMan: `b"".join(...)` assumes bytes, but the stdout stream yields `str`: `src/boxlite/src/litebox/exec.rs:416-417` declares `impl Stream for ExecStdout { type Item = String; }`, surfaced by pyo3 as Python `str`. This raises `TypeError: sequence item 0: expected a bytes-like object, str found` — a second error that surfaces as soon as L335 is fixed.

The `out_chunks: list[bytes]` annotation on L334 is wrong for the same reason.

`conftest.py:166` already handles both cases with an `isinstance(ch, bytes)` guard; reusing `collect_stream` would avoid this entirely.

apps/e2e/cases/test_cross_org_isolation.py
```diff
@@ -0,0 +1,349 @@
...
+
+            assert "ORG_A_DATA" not in out, (
```
DefeatMan: **This assertion has no positive control, so it passes when nothing works at all.**

`assert "ORG_A_DATA" not in out` is satisfied by empty output. Nothing in this test verifies that the org-A listener ever started, or that org-A itself can reach it. Given the issues flagged above — `nc` missing, the dash mis-parse, unresolvable hostnames, and the two `TypeError`s — `out` will be empty or `BLOCKED` in every real run, and the test will report success.

A genuine cross-org breach and a completely broken setup are indistinguishable here, and today it's the second one.

Suggest a two-step shape: first assert org-A *can* reach its own listener (fail loudly if not — that's the control), then assert org-B cannot.

apps/e2e/cases/test_cross_org_isolation.py
G4614: Fixed. `BOXLITE_E2E_CROSS_ORG_API_KEY` is now wired into the `Run E2E suite` step of `.github/workflows/e2e-cloud.yml`, and the skip is no longer silent on either side:

- A new `Check cross-org secret` step emits a `::warning::` annotation when the secret is unset, so the lost coverage shows up in the run summary instead of as five green SKIPs.
- `BOXLITE_E2E_REQUIRE_CROSS_ORG` turns the skip into a **collection error** (`RuntimeError` at import). The workflow sets it whenever the secret is present, so once configured a later regression that unsets it fails the job.
- When the module does skip, it prints the reason to stderr.

I did not make a missing secret an unconditional hard failure: only an org admin can add it, so that would red-fail every run until then. `BOXLITE_E2E_REQUIRE_CROSS_ORG=1` set unconditionally is the one-line change to make it mandatory, and the workflow comment says so.

apps/e2e/cases/test_cross_org_isolation.py
G4614: Fixed — the local copy is gone; `org_b_prefix` now calls `e2e_auth.discover_path_prefix`. That closes both holes you named: it uses `or ""` rather than `.get(..., "")` so an explicit `"path_prefix": null` cannot become `None`, and it raises `RuntimeError` on a non-200 `/v1/me` rather than degrading into an empty prefix that silently unroutes every test in the module.

apps/e2e/cases/test_cross_org_isolation.py
G4614: Fixed. `_request` now catches `urllib.error.URLError`, `TimeoutError` and `OSError` and re-raises them as a typed `TransportError`, so a transport fault is distinguishable from "the server rejected us" — which is the whole point of these assertions — instead of escaping as a raw urllib traceback.

`org_a_box` is now `try/finally` around the `yield`, with cleanup guarded on `box_id is not None`. The one path that genuinely cannot self-clean is a POST that times out *after* the API persisted the row: we never learn the id. That's why the fixture keeps `auto_delete: 600` — noted in its docstring now, so the value doesn't look incidental.

apps/e2e/cases/test_cross_org_isolation.py
G4614: Partly fixed, and I want to be precise about the part that isn't.

org-A's box in the network test now comes from the session `rt` fixture, so it goes through `_TrackingRuntime` and the autouse `verify_runner_saw_all_boxes` guard does cover it.

`org_a_rt` is deleted outright. `org_b_rt` is wrapped in `_TrackingRuntime` for consistency, but wrapping alone would not have bought verification: the autouse fixture only resets and reads `rt._created` (`conftest.py:105,112`), so a wrapped non-session runtime records ids that nothing checks. Since org-B never creates a box in the reworked tests, there is nothing to verify there — the fixture docstring states this rather than implying coverage it doesn't have.

apps/e2e/cases/test_cross_org_isolation.py
G4614: Fixed. New module-scoped `org_a_prefix` fixture calls `pytest.skip` when `org_a_ctx.path_prefix` is empty, with a message naming the deployment shape ("only observable on a prefix-routed (multi-tenant) deployment"). Every foreign-namespace test depends on it, so the fall-through-to-200 case you describe can no longer be read as a pass.

apps/e2e/cases/test_cross_org_isolation.py
G4614: Agreed, and fixed — though the conclusion is that *both* vectors are worth asserting, with different expected answers, so I kept both rather than just switching.

Reading `OrganizationAccessGuard` (`organization-access.guard.ts:54-64`), the two are genuinely distinct failure modes:

- `/v1/{prefix-A}/…` + org-B key → the mismatch branch returns false → **403**. This is the vector the PR body claimed and the only one that catches a regression exposing org-A's resources inside org-A's own namespace.
- `/v1/{prefix-B}/boxes/{box-A-id}` → guard passes, lookup is scoped by `authContext.organizationId` → **404**. Weak on its own, as you say, but it is the no-existence-leak property, and 404-not-403 is the assertion.

So the tests are now two parametrized groups — `test_org_b_key_rejected_in_org_a_namespace` (list/get/delete/exec → 401/403) and `test_org_a_box_is_not_found_in_org_b_namespace` (get/exec → 404) — and the module docstring explains why the same operation has two different correct answers.

apps/e2e/cases/test_cross_org_isolation.py
G4614: Fixed, both halves.

The assertion is now on the body, not the status code. `boxToBoxResponse` (`box-to-box.mapper.ts:18-33`) doesn't surface `pending`/`desiredState`, but it does surface `name`, and `getSoftDeleteUpdate` rewrites it to `DESTROYED_<name>_<ts>` — so the test asserts the name does not start with `DESTROYED_` and, when the create response carried one, that it equals the baseline captured at fixture creation. That is the signal that survives the `state: Not(DESTROYED)` read filter.

On the ordering problem: both DELETE vectors are now attempted first, their statuses collected into a dict, and the box inspected unconditionally afterwards. The status-code assertions come last and quote the collected outcomes, so a surprising status can no longer preempt the check that actually matters.

apps/e2e/cases/test_cross_org_isolation.py
G4614: Fixed — the body is now `{"command": "/bin/sh", "args": ["-c", "id"]}` everywhere exec appears (both the 403 and the 404 parametrizations), matching `ExecRequest`'s `required: [command]` at `openapi/box.openapi.yaml:2140`. The test docstring notes why: a rejection has to be an authorization decision, not a 422 from body validation.

apps/e2e/cases/test_cross_org_isolation.py
G4614: Dropped the in-guest probe entirely rather than pinning runners — pinning would have made the test schedulable but still meaningless.

There is no box-to-box network to isolate: `net/constants.rs` gives every box the same `GUEST_IP` 192.168.127.2 in its own netstack, and `start.rs:71` writes a constant `boxlite` hostname. Co-tenancy on one runner doesn't create mutual reachability, so a same-runner probe would have been a second way to pass for the wrong reason.

The surface that does exist between an org and a box is the control-plane tunnel (`POST /{prefix}/boxes/{box_id}/network/tunnel`), so the test now asserts on that: org-A opens a tunnel to its own box and reads a marker (positive control), then org-B is denied — `org_b_rt.get(box_a.id)` returns `None`, and both raw tunnel vectors are refused 404/403.

apps/e2e/cases/test_cross_org_isolation.py
G4614: Resolved by the rework — the org-B box is gone entirely. The network test now creates a single box, via the session `rt` fixture, and probes the control-plane tunnel instead of in-guest networking, so there is no second `BoxOptions` call site to contradict the fixture.

The underlying concern you identified still applies to `org_a_box`, and its `auto_delete: 600` is now documented as deliberate: a POST that times out after the API persisted the row never yields an id we could clean up, so server-side reaping is the only backstop.

apps/e2e/cases/test_cross_org_isolation.py
G4614: Fixed by replacing the listener wholesale. The test now runs `apps/e2e/fixtures/service_in_box_server.py` — the same base64+`python3 -u -c` pattern `test_sdk_tunnel.py` uses — so no netcat, and the redirect is `>/tmp/… 2>&1 &`, not the dash-hostile `&>`. `python3` is in `base.Dockerfile`, as you note.

It also gives the test a real HTTP server with a known marker, which is what the positive control needs.

apps/e2e/cases/test_cross_org_isolation.py
G4614: Fixed — both URLs are gone. Confirmed your reading: `start.rs:71` writes a constant `boxlite` hostname with no per-box `/etc/hosts` entries, and `GUEST_IP` is 192.168.127.2 for every box, so there was never an address to dial.

The probe now targets the control-plane tunnel, which is a genuinely routable surface, and the module docstring records why in-guest addressing was the wrong target rather than leaving it as a fixed typo.

apps/e2e/cases/test_cross_org_isolation.py
G4614: Fixed — the hand-rolled stream loop is gone. The test uses `conftest.drain(ex)` (which calls `ex.stdout()` / `ex.stderr()` correctly and drains them concurrently) for the service startup, and reads the tunnel response over the connection API. No `result.stdout` attribute access remains.

apps/e2e/cases/test_cross_org_isolation.py
G4614: Fixed with the same change. `collect_stream` already normalises `str` and `bytes` via its `isinstance` guard (`conftest.py:166`), so reusing `drain` removes both the `b"".join` and the wrong `list[bytes]` annotation. The only bytes left in the module is the tunnel marker, which is compared against the raw socket response — bytes on both sides there.

apps/e2e/cases/test_cross_org_isolation.py
G4614: Fixed — this was the most important one, and it drove the shape of the rewrite.

The test is now control-then-denial:

1. org-A starts the marker service in its own box and fetches it over its own tunnel. If that fails, the test fails with "positive control failed: … so the denial below would prove nothing". Empty output can no longer read as success.
2. Only then is org-B denied — `org_b_rt.get(box_a.id) is None`, plus 404 on the tunnel endpoint from org-B's namespace and 401/403 from org-A's.

A broken setup and a real breach are now distinguishable in both directions.

apps/e2e/cases/test_cross_org_isolation.py
```diff
@@ -0,0 +1,523 @@
...
+    try:
+        await _start_marker_service(box_a)
+
+        # 1. Positive control: the service is up and reachable by its owner.
+        response = await _fetch_marker_over_tunnel(box_a, _TUNNEL_PORT)
+        assert _TUNNEL_MARKER in response, (
+            "positive control failed: org-A could not reach its own service in "
+            f"box {box_a.id} over its own tunnel, so the denial below would "
+            f"prove nothing. Response: {response[:512]!r}"
+        )
```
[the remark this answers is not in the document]
G4614: Valid, fixed in d5270719b — though the failure mode was worse than a flake: running the module against a second real org, this timing window was one of three reasons the tunnel test wasn't testing what it claimed.

Fixed as `_await_marker_over_tunnel`, a helper rather than an inline loop. I didn't take the proposed diff as written: its `except Exception: response = b""` swallows a genuine failure (tunnel refused, service crashed, SDK error) and leaves only an empty `response` at the deadline, which reads identically to "service never bound". The helper keeps the last exception and raises with it, so the two stay distinguishable.

Two further defects the same run surfaced, both of which meant a denial vector never reached an authorization check:

- `assert await org_b_rt.get(box_a.id) is None` failed on *correct* behaviour — the REST runtime raises `RuntimeError("Box with ID or name … not found")` rather than returning None. Now accepts either shape, asserting on the not-found reason.
- The tunnel `port` was sent in the JSON body, but it's `@Query('port', ParseIntPipe)` (`boxlite-proxy.controller.ts:208`, `openapi/box.openapi.yaml:952`). `ParseIntPipe` rejected the missing query param with 400 *before* `findOneByIdOrName` ran, so both ownership vectors 400'd without the ownership check ever executing.

Verified against `api.dev.boxlite.ai` with two real orgs: 8 passed. Mutation check on the control itself — binding the service to a port the tunnel doesn't read makes it fail loudly (`Response: b''` plus the "would prove nothing" message) instead of passing on empty output.

PR #1430, opened 2026-09-01
fix(rest): classify codeless error bodies by HTTP status

Monsteer714:
## Summary

A structured error body that names no `error.code` was reported to callers as
`BoxliteError::Internal`, blaming the server for refusals the caller caused. A
recognized `error.code` still names the variant; otherwise the HTTP status now
decides, through a baseline table of the statuses the API answers codelessly —
so a codeless refusal keeps the class its status states, and a failure the
server answered for itself is never attributed to an intermediary.

## Call graph

Before
```
create                    (CreateCmd · src/cli/src/commands/create.rs:59)
  └─ handle_error         (RestClient · src/boxlite/src/rest/client.rs)
       └─ map_http_body   (rest::error)
            └─ into_error_model (FlatErrorResponse · src/boxlite/src/rest/types.rs)
                 │                        ← BUG: a body with no `code` is given
                 │                          the fabricated code "internal"
                 └─ map_http_error  (rest::error)   — dispatches on it ⇒ Internal
                      └─ map_http_status (rest::error) — unreachable behind it
```

After
```
create                    (CreateCmd · src/cli/src/commands/create.rs:59)
  └─ handle_error         (RestClient · src/boxlite/src/rest/client.rs:239)
       └─ map_http_body   (rest::error · src/boxlite/src/rest/error.rs:28)
            ├─ into_error_model (FlatErrorResponse · types.rs:37) — absent code stays empty
            ├─ map_enveloped      (error.rs:48)
            │    ├─ refine_by_code   (error.rs:73)  — a known code wins
            │    ├─ refine_by_status (error.rs:111) — 400 ⇒ InvalidArgument,
            │    │                     404/408/409/428/429 likewise
            │    └─ else ⇒ Internal — codeless or unknown-code alike: the
            │                server answered, no intermediary is implicated
            └─ else: envelope_message (error.rs:138) — gin `{"error": text}`
                     lends its sentence; class from map_status (error.rs:180)

connect_box_network_tunnel (RestClient · client.rs)
  └─ read_connect_refusal  (client.rs:37)   — proxy's sentence; bounded read,
       │                     2s dedicated timeout (client.rs:34)
       └─ map_plain_reply  (error.rs:166)   — never blames an intermediary

attach (WS upgrade)
  └─ map_ws_error          (client.rs:616)  — attach arms only (404/409/410 +
                             auth); envelope or gin sentence surfaced; all
                             else, 5xx included, reads like a REST reply
```

Fixes #<issue>

## Changes

- `FlatErrorResponse::into_error_model` no longer invents `code: "internal"`;
  a missing code decodes to the empty string. `ErrorModel` is a tolerant
  reader — `code` and `type` both default — so an envelope missing either
  still decodes instead of handing the caller raw JSON.
- Status baseline holds one arm per status the API answers codelessly (400,
  404, 408, 409, 428, 429, plus the 401/403 auth arms). Statuses only ever
  answered with a code (410, 422) deliberately get no arm.
- Any enveloped body that resolves no further — codeless, or naming a code
  this client has not learned yet — stays `Internal`: the server answered
  for itself. The bare-5xx "likely a proxy" attribution fires only when
  nothing decoded at all.
- The runner's bare `{"error": "<text>"}` rejections lend their sentence to
  the message; the class stays status-derived.
- CONNECT tunnel: the proxy's own sentence is read off the reply (limit
  enforced while reading, under a dedicated 2s timeout) and a refusal the
  proxy wrote is never attributed to an intermediary.
- WS upgrade: only the attach contract's arms remain (the reattach loop pins
  404/409/410); the envelope's or gin's sentence is surfaced instead of raw
  JSON, and every other status — 5xx included — is read like any other REST
  reply, so a 503 named `engine_unavailable` arrives as `Engine`, not as a
  proxy fault.

## How to verify

```
make test:unit:rust FILTER=rest::error
```

Against a running control plane, a box create above the org's per-box CPU
ceiling is refused as `invalid argument` rather than `internal error`:

```
boxlite --profile <p> create --cpus 200 <image>
```

## Risks / rollout

Client-side error classification only; no wire or server change. Callers that
matched on `Internal` for these refusals now see the status-derived variant
(`InvalidArgument`, `NotFound`, `InvalidState`, `ResourceExhausted`); a
framework route-miss 404 (stale `path_prefix`) is now `NotFound` as well,
which resource-probing callers treat as absence. An unknown-code 5xx moves
from the proxy-attributed `Network` text to `Internal` with the server's own
sentence.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of connection refusals by preserving proxy messages with size and timeout safeguards.
  - Improved WebSocket rejection handling with clearer messages and consistent HTTP error mapping.
  - Refined error classification when response codes or error details are missing, including clearer handling of server and network errors.
  - Preserved meaningful messages from plain-text and common server error responses.
  - Improved tolerance for incomplete error responses while retaining diagnostic details.

- **Documentation**
  - Clarified error-response behavior and fallback classification in developer-facing documentation and tests.

### DIFF
--- a/src/boxlite/src/rest/client.rs
+++ b/src/boxlite/src/rest/client.rs
@@ -11,7 +11,7 @@ use tokio::sync::RwLock;
 use boxlite_shared::errors::{BoxliteError, BoxliteResult};
 
 use super::credential::{AccessToken, Credential};
-use super::error::{map_http_body, map_http_status};
+use super::error::{map_http_body, map_plain_reply};
 use super::options::BoxliteRestOptions;
 use super::types::ServerConfig;
 use crate::runtime::auth::Principal;
@@ -21,6 +21,37 @@ const REFRESH_LEEWAY: Duration = Duration::from_secs(60);
 /// File transfers cannot outlive the runner's 24-hour box lifetime cap.
 const FILE_REQUEST_TIMEOUT: Duration = Duration::from_secs(24 * 60 * 60);
 const TUNNEL_SETUP_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// The proxy states a refused tunnel in one short line (Go's `http.Error`).
+/// The limit is enforced while reading, not after: a peer that answers a
+/// handshake with megabytes is not stating a reason, and buffering it first
+/// would be trusting it for length. Past the limit the status speaks alone.
+const CONNECT_REFUSAL_MAX_BYTES: usize = 512;
+
+/// Reading the sentence off an already-arrived refusal is instant against
+/// `apps/proxy` (`http.Error` flushes the body with the headers); this bound
+/// only covers a peer that stalls mid-body, and a refusal should fail fast.
+const CONNECT_REFUSAL_READ_TIMEOUT: Duration = Duration::from_secs(2);
+
+/// Read the sentence a peer sent with its non-2xx CONNECT reply.
+async fn read_connect_refusal(response: hyper::Response<hyper::body::Incoming>) -> String {
+    use http_body_util::{BodyExt, Limited};
+
+    const FALLBACK: &str = "CONNECT proxy rejected tunnel";
+    let body = Limited::new(response.into_body(), CONNECT_REFUSAL_MAX_BYTES);
+    let Ok(Ok(collected)) =
+        tokio::time::timeout(CONNECT_REFUSAL_READ_TIMEOUT, body.collect()).await
+    else {
+        return FALLBACK.to_string();
+    };
+    let bytes = collected.to_bytes();
+    let text = String::from_utf8_lossy(&bytes);
+    let text = text.trim();
+    if text.is_empty() {
+        return FALLBACK.to_string();
+    }
+    text.to_string()
+}
 
 type TunnelConnector =
     hyper_rustls::HttpsConnector<hyper_util::client::legacy::connect::HttpConnector>;
@@ -390,7 +421,12 @@ impl ApiClient {
             .map_err(|e| BoxliteError::Network(format!("CONNECT request failed: {e}")))?;
         let status = response.status();
         if !status.is_success() {
-            return Err(map_http_status(status, "CONNECT proxy rejected tunnel"));
+            // Our own proxy answers this handshake with a verdict of its own,
+            // in plain text rather than our envelope. Carry its sentence and
+            // read the class off the status; a refusal it wrote is never an
+            // intermediary's fault.
+            let detail = read_connect_refusal(response).await;
+            return Err(map_plain_reply(status, &detail));
         }
         let upgraded = hyper::upgrade::on(response)
             .await
@@ -564,19 +600,31 @@ pub(super) fn transport_error(err: reqwest::Error) -> BoxliteError {
     BoxliteError::Network(detail)
 }
 
-/// Map a tungstenite connect error to a typed `BoxliteError`. The WS
-/// upgrade returns HTTP status codes for rejections (404 for a missing
-/// session, 409 for an already-attached one, 410 once an exec has been
-/// reaped). Symmetric with the REST mapper in [`super::error`].
+/// Map a tungstenite connect error to a typed `BoxliteError`.
+///
+/// The explicit arms below are the WS attach contract, not a copy of the
+/// REST baseline: on this handshake 404 means the session is gone, 409 that
+/// the single attach slot is taken, 410 that the reaper got there first —
+/// and the reattach loop in `litebox` branches on exactly those classes
+/// (`AlreadyExists` to back off, `NotFound`/`SessionReaped` to stop), with
+/// its tests pinning them. The backends also do not agree on a body shape
+/// here (`serve` rejects with the wire envelope, the runner with gin's
+/// `{"error": text}`), so the classes stay keyed on the status; the envelope,
+/// when present, contributes its sentence rather than raw JSON. Statuses
+/// with no meaning of their own on this handshake are read like any other
+/// REST reply.
 fn map_ws_error(err: tokio_tungstenite::tungstenite::Error) -> BoxliteError {
     use tokio_tungstenite::tungstenite::Error as TgErr;
     if let TgErr::Http(resp) = &err {
         let status = resp.status();
-        let body = resp
+        let raw = resp
             .body()
             .as_ref()
             .map(|b| String::from_utf8_lossy(b).into_owned())
             .unwrap_or_default();
+        // `serve` rejects the upgrade with the same envelope every REST
+        // route answers with; carry its sentence, never the JSON itself.
+        let body = super::error::envelope_message(&raw).unwrap_or_else(|| raw.clone());
         return match status.as_u16() {
             404 => BoxliteError::NotFound(if body.is_empty() {
                 "session not found".to_string()
@@ -594,12 +642,12 @@ fn map_ws_error(err: tokio_tungstenite::tungstenite::Error) -> BoxliteError {
                 body
             }),
             401 | 403 => BoxliteError::Config(format!("WS auth rejected ({}): {}", status, body)),
-            502..=504 => BoxliteError::Network(format!(
-                "WS upstream returned HTTP {} (proxy or load balancer): {}",
-                status,
-                if body.is_empty() { "<empty>" } else { &body }
-            )),
-            _ => BoxliteError::Internal(format!("WS upgrade failed (HTTP {}): {}", status, body)),
+            // No WS-specific meaning — 5xx included: read it like any other
+            // REST reply, envelope, flat body, or bare status alike. A
+            // refusal the server names (`invalid_argument` on a 400,
+            // `engine_unavailable` on a 503) keeps its class instead of
+            // arriving as a server fault or an intermediary's.
+            _ => map_http_body(status, &raw),
         };
     }
     BoxliteError::Network(format!("WS connect failed: {}", err))
@@ -624,8 +672,7 @@ mod tests {
     use super::ensure_capability;
     use super::*;
     use crate::rest::credential::{AccessToken, Credential};
-    use crate::rest::error::map_http_error;
-    use crate::rest::types::FlatErrorResponse;
+    use crate::rest::error::map_http_body;
     use async_trait::async_trait;
     use boxlite_shared::errors::BoxliteError;
     use std::sync::atomic::{AtomicUsize, Ordering};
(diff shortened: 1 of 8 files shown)


src/boxlite/src/rest/client.rs
```diff
@@ -599,7 +640,11 @@ fn map_ws_error(err: tokio_tungstenite::tungstenite::Error) -> BoxliteError {
                 status,
                 if body.is_empty() { "<empty>" } else { &body }
             )),
-            _ => BoxliteError::Internal(format!("WS upgrade failed (HTTP {}): {}", status, body)),
+            // No WS-specific meaning: read it like any other REST reply —
+            // envelope, flat body, or bare status alike — so a refusal the
+            // server names (an `invalid_argument` 400) keeps its class
+            // instead of arriving as a server fault.
+            _ => map_http_body(status, &raw),
```
DefeatMan, line 650: The `502..=504` arm just above this fallback is the one arm in `map_ws_error` that isn't part of the attach contract the new doc comment defends. 404/409/410 are pinned by the reattach loop; 502–504 is a copy of the REST bare-5xx heuristic — the exact attribution this PR removes one layer down.

Today, when `serve` rejects the upgrade with its own envelope on a 503:

```json
{"error":{"message":"engine is down","type":"EngineError","code":"engine_unavailable"}}
```

the caller gets `Network("WS upstream returned HTTP 503 (proxy or load balancer): engine is down")` — a fault the server named for itself, attributed to an intermediary. That is the same mistake `map_enveloped` now refuses to make.

Letting 502–504 fall through to this `_` arm gives `Engine(_)` when an envelope decoded, and a genuinely bare 5xx still reaches the "no error envelope; likely a proxy" wording, because that arm lives in `map_status` — nothing is lost. Nothing in the reattach loop branches on 5xx, so the contract is unaffected.

src/boxlite/src/rest/error.rs
```diff
@@ -21,23 +27,47 @@ use super::types::{ErrorModel, ErrorResponse, FlatErrorResponse};
...
+        // used to render — and never an intermediary's: the API answers
+        // codelessly at 503 while under maintenance or missing object
+        // storage, and calling that a proxy failure is this bug in mirror.
+        return BoxliteError::Internal(msg);
+    }
+    // A code this client has not learned yet: the server named a class, this
+    // client just cannot read it. Such a body keeps the status-only path,
+    // intermediary attribution included, rather than the `Internal` above —
+    // the server did say something more specific than "no code at all".
+    map_status(status, &msg)
```
DefeatMan: This arm can produce the exact message the rest of the PR exists to prevent. `map_status` falls through to its `502..=504` arm, so an envelope that *did* decode but names a code this client hasn't learned yet renders as:

> upstream returned HTTP 503 (no error envelope; likely a proxy or load balancer in front of the server). Body: …

which is untrue on its face — an envelope decoded — and the CLI turns `Network` into "check your HTTP_PROXY". The reasoning three lines up argues a codeless 5xx must stay `Internal` because "calling that a proxy failure is this bug in mirror"; a body that carries a code is *stronger* evidence it came from our server, yet it gets the proxy attribution.

Every code in `box.openapi.yaml`'s enum has an arm in `refine_by_code` today, so this is forward-compat only — but it is precisely the forward-compat case the module doc invokes: a server adds `code: "region_draining"` on a 503, an older client blames the user's proxy. Suggest the unknown-code tail reuse the codeless tail:

```rust
    // A code this client has not learned yet: the server named a class, this
    // client just cannot read it — but it did answer, so no intermediary is
    // implicated. The status baseline above already had its say.
    BoxliteError::Internal(msg)
```

That also drops the second `refine_by_status` call `map_status` makes after `map_enveloped` already got `None` from it.

If keeping the status-only path for unknown codes is deliberate, `statuses_without_a_codeless_producer_get_no_baseline` would be a good place to pin an unknown-code 503 either way — it's the one branch here no test covers.

src/boxlite/src/rest/error.rs
```diff
@@ -58,29 +88,82 @@ pub(crate) fn map_http_error(status: StatusCode, body: &ErrorModel) -> BoxliteEr
...
+        428 => BoxliteError::InvalidState(msg),
+        429 => BoxliteError::ResourceExhausted(msg),
+        _ => return None,
+    })
+}
+
+/// The human sentence inside a body, whichever of our envelope shapes carries
+/// it — for a caller that keeps status semantics of its own (the WS upgrade)
+/// but must not surface raw JSON as the error text.
+pub(crate) fn envelope_message(text: &str) -> Option<String> {
```
DefeatMan, line 138: `envelope_message` decodes the two envelope shapes but not the third one the `map_ws_error` doc comment names — gin's bare `{"error": "<text>"}`. `ErrorResponse` rejects it (`error` is a string, not an object) and `FlatErrorResponse` rejects it (no top-level `message`), so it returns `None` and the caller falls back to `raw`.

That is the runner's *only* WS attach rejection shape, and it covers both statuses the reattach loop keys on (`boxlite_exec_attach.go:140` and `:146`):

```go
ctx.JSON(http.StatusNotFound, gin.H{"error": fmt.Sprintf("execution %s not found", execId)})
ctx.JSON(http.StatusConflict, gin.H{"error": fmt.Sprintf("execution %s already attached", execId)})
```

So against a runner the user sees `NotFound("{\"error\":\"execution e1 not found\"}")` — the raw JSON `ws_upgrade_rejections_read_the_envelope` asserts must not leak, in the most common rejection there is. A third arm covers it:

```rust
    #[derive(Deserialize)]
    struct BareError {
        error: String,
    }
    serde_json::from_str::<BareError>(text).ok().map(|resp| resp.error)
```

Worth having `map_http_body` reach for the same thing on its `else` branch: every runner exec route answers `gin.H{"error": err.Error()}` (`boxlite_exec.go:42, 136, 141, 156, 161, 228`, `boxlite_metrics.go:29,37`), and all of those currently arrive with raw JSON as the message. The class stays status-derived either way — this is only about the sentence.

src/boxlite/src/rest/types.rs
```diff
@@ -45,18 +50,25 @@ impl FlatErrorResponse {
...
 #[derive(Debug, Deserialize)]
 pub(crate) struct ErrorModel {
     pub message: String,
-    /// Preserved for diagnostics / log enrichment; dispatch happens on
-    /// `code` (the snake string).
+    /// Preserved for diagnostics / log enrichment; the variant is chosen
+    /// from the status and `code`, never from this field.
     #[serde(rename = "type")]
     #[allow(dead_code)]
     pub error_type: String,
```
DefeatMan, line 65: Nit, and the same trap one field over: `code` just got `#[serde(default)]` because "as a required field it made the whole body unparseable and the caller was handed the raw JSON instead". `error_type` is still required — and it is `#[allow(dead_code)]`, with a doc line two lines up saying the variant is "chosen from the status and `code`, never from this field". So a field nothing dispatches on decides whether the body decodes at all.

`{"error":{"message":"cpu 200 exceeds the limit"}}` fails `ErrorResponse` on the missing `type`, then fails `FlatErrorResponse` on the missing top-level `message`, and lands in `map_status` with the raw JSON as its text.

No producer in-tree emits that shape today, so this is hardening rather than a live bug. But `#[serde(default)]` here costs one line and makes `nested_codeless_envelope_keeps_its_message` pass for its own reason, rather than because the fixture happens to carry `"type":"HttpError"`.

src/boxlite/src/rest/client.rs
```diff
@@ -22,6 +22,30 @@ const REFRESH_LEEWAY: Duration = Duration::from_secs(60);
...
+/// would be trusting it for length. Past the limit the status speaks alone.
+const CONNECT_REFUSAL_MAX_BYTES: usize = 512;
+
+/// Read the sentence a peer sent with its non-2xx CONNECT reply.
+async fn read_connect_refusal(response: hyper::Response<hyper::body::Incoming>) -> String {
+    use http_body_util::{BodyExt, Limited};
+
+    const FALLBACK: &str = "CONNECT proxy rejected tunnel";
+    let body = Limited::new(response.into_body(), CONNECT_REFUSAL_MAX_BYTES);
+    let Ok(Ok(collected)) = tokio::time::timeout(TUNNEL_SETUP_TIMEOUT, body.collect()).await else {
```
DefeatMan: `TUNNEL_SETUP_TIMEOUT` is 30s, and this is its third independent use on one call path: the connector `call` gets 30s, `send_request` gets 30s, and now reading at most 512 bytes off an already-arrived response gets another 30. Worst case a *refused* tunnel takes 90s to report, up from 60.

`apps/proxy` writes every refusal with `http.Error`, so the body lands in the same flush as the headers and this read is normally instant; the budget only matters for a peer that sends headers and then stalls. That argues for a short dedicated bound rather than reusing the setup budget — a refusal should fail fast:

```rust
/// Reading the sentence off an already-arrived refusal is instant against
/// `apps/proxy` (`http.Error` flushes the body with the headers); this bound
/// only covers a peer that stalls mid-body, and a refusal should fail fast.
const CONNECT_REFUSAL_READ_TIMEOUT: Duration = Duration::from_secs(2);
```

DefeatMan (commented):
Read this against the code on both sides of the wire rather than against the description; the core fix holds up. Notes on what I checked, then five suggestions inline.

## Verified

- `AllExceptionsFilter` (`apps/api/src/filters/all-exceptions.filter.ts:80`) really does emit `{path, timestamp, statusCode, error, message}` and spreads the code only as `...(code ? { code } : {})` — so the codeless flat body this PR fixes is the *normal* shape, not an edge case. The `ErrorResponse`-before-`FlatErrorResponse` ordering in `map_http_body` is right for it too: top-level `error` is a status string there, so the nested shape correctly fails first.
- Every status given a baseline arm has a real codeless producer. 408 via `RequestTimeoutException` (`box-state-waiter.service.ts:127`, `volume.service.ts:103`, `box-auto-resume.service.ts:48`); 428 via `HttpStatus.PRECONDITION_REQUIRED` (`runner.service.ts:326,336`, `region.service.ts:235`). The 410/422 omission is consistent with that rule.
- All 21 codes in `box.openapi.yaml`'s `code` enum have an arm in `refine_by_code`.
- The CONNECT peer really is `apps/proxy`, and `handleTunnelConnect` (`apps/proxy/pkg/proxy/tunnel.go:29-56`) answers 405/400/403/502 through `http.Error` — plain text — so `map_plain_reply` is the right shape for it. The runner's gin-JSON tunnel handler is only ever read by the proxy, never by this client.
- No stale `map_http_error` / `map_http_status` references left anywhere in the tree.

## One thing outside the diff

`openapi/box.openapi.yaml` is the normative document for this contract, and it still states the opposite of what this PR implements:

- line 1533 — `required: [message, type, code]`
- line 1525 — "Clients SHOULD pattern-match on `code` for typed error handling."

The reference server's docstring, `respondError`'s comment and the Python test docstring were all updated to "`code` refines the status-derived variant". The spec those three describe was not — and `required: [code]` is now contradicted by the API's own exception filter, which provably omits the field. Suggest dropping `code` from `required` and rewording that description the same way `errors.py` was; otherwise the next reader of the spec writes `code.unwrap_or("internal")` again for the same reason.

## Nit, no action needed if deliberate

`read_connect_refusal` drops the *entire* sentence once the body passes 512 bytes, because `Limited::collect()` errors rather than truncating — so a 600-byte refusal from a future proxy version reports only "CONNECT proxy rejected tunnel". The comment says as much ("past the limit the status speaks alone"), so this reads as a choice rather than an oversight; keeping the first 512 bytes would preserve the useful prefix without trusting the peer for length any more than the current code already does.

src/boxlite/src/rest/error.rs
```diff
@@ -21,23 +27,52 @@ use super::types::{ErrorModel, ErrorResponse, FlatErrorResponse};
...
+/// status depends on what the body left out, and the two cases are not the
+/// same failure.
+fn map_enveloped(status: StatusCode, body: &ErrorModel) -> BoxliteError {
     let msg = body.message.clone();
-    match body.code.as_str() {
+    if let Some(err) = refine_by_code(&body.code, &msg) {
+        return err;
+    }
+    if let Some(err) = refine_by_status(status, &msg) {
+        return err;
```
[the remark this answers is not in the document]
Monsteer714: thanks for leaving comment. clarify that status code should always doing the fallback. problem was in the commit description. let me fix this.

DefeatMan (approved):
LGTM

PR #1434, opened 2026-09-03
feat(volumes): back managed volumes with a local named-volume store

Monsteer714:
## Call graph
```text
Before
  VolumeHandle::create|list|get|remove   (VolumeHandle · src/boxlite/src/runtime/volumes.rs:59-74)            — public facade
    └─ <LocalRuntime as VolumeBackend>   (LocalRuntime · src/boxlite/src/runtime/rt_impl.rs:1871 @main)       — every method Err(Unsupported)
  VolumeFlags::apply_to                  (VolumeFlags · src/cli/src/cli.rs:1348 @main)                         — parses -v
    ├─ Anonymous → mkdir {home}/volumes/anonymous/<ulid> → VolumeSpec::bind_mount   (cli.rs:1375-1383 @main)  — the CLI creates the directory itself
    └─ ManagedVolume(v) → VolumeSpec::managed_volume(v)
         └─ sanitize_local_options       (rt_impl.rs:1769 @main)                                               — any managed_volume → Err(Unsupported "only supported by REST runtimes")
  build_box_options                      (serve · src/cli/src/commands/serve/mod.rs:1131,1199 @main)           — non-empty volumes[] → Err(InvalidArgument "not supported by boxlite serve")

After
  VolumeHandle::create|list|get|remove   (VolumeHandle · src/boxlite/src/runtime/volumes.rs:59-74)
    └─ <LocalRuntime as VolumeBackend>   (LocalRuntime · src/boxlite/src/runtime/rt_impl.rs:2076)              — one store call per method
         ├─ run_volume_op                (rt_impl.rs:2051)                                                     — shutdown check → LocalVolumeStore::new → spawn_blocking
         ├─ LocalVolumeStore::create|list|get|remove   (LocalVolumeStore · src/boxlite/src/volumes/store.rs:88,151,193,213)
         │    └─ locate                  (store.rs:261)                                                        — validate_reference → stat {id}/_data → find_by_name over sidecars; unknown → NotFound
         └─ remove_volume                (rt_impl.rs:2094)                                                     — a volume any persisted box mounts → Err(InvalidState), force included
  VolumeFlags::apply_to                  (VolumeFlags · src/cli/src/cli.rs:1310)
    ├─ ManagedVolume(v) → VolumeSpec::managed_volume(v)   (cli.rs:1316)                                        — read-only refused
    └─ Anonymous → VolumeSpec::anonymous_volume           (cli.rs:1332)                                        — no mint, no disk IO; read-only refused
         └─ BoxliteRuntime::create       (BoxliteRuntime · src/boxlite/src/runtime/core.rs:291,307)            — validate_in_request: exactly one origin, or none for an anonymous mount
              └─ <LocalRuntime as RuntimeBackend>::create   (rt_impl.rs:1967) → sanitize_local_options (rt_impl.rs:1946)
                   └─ create_inner       (rt_impl.rs:432)                                                      — lookup_box first: reusing a box resolves nothing
                        ├─ resolve_managed_volumes   (rt_impl.rs:1839,1882)                                   — pass 1 store.get every reference (unknown → NotFound, nothing created); pass 2 create(None) for an anonymous mount without id, then canonical id + payload path on every mount; a pass that fails removes what it created
                        ├─ allocate → add_box                                                                  — on failure rollback_created_volumes (rt_impl.rs:1913) removes the ids resolution returned, then the name-race recovery runs
                        └─ boot: resolve_user_volumes   (src/boxlite/src/litebox/init/types.rs:50,64)          — a reference with no path is Internal; a resolved mount is a directory share
                             └─ build_path_access          (src/boxlite/src/jailer/mod.rs:213,343)             — grants the payload directory to the VMM
  build_box_options                      (serve · src/cli/src/commands/serve/mod.rs:1131,1235,1264)           — CreateVolumeMount (serve/types.rs:185) → VolumeSpec::managed_volume; read_only:true → Err(Unsupported), per openapi enum:[false]
  <RestRuntime as RuntimeBackend>::create → sanitize_remote   (src/boxlite/src/rest/runtime.rs:103,170,179)   — a host bind or an anonymous mount → Err(Unsupported) before network IO
  import_box                             (src/boxlite/src/runtime/import.rs:25,86,108)                        — options_from_manifest → sanitize_import (options.rs:674): resolved-shape mounts, kernel source checked as create does
    └─ provision_box                     (rt_impl.rs:1261,1270)                                                — shared with clone (src/boxlite/src/litebox/clone_export.rs:128): resolve_managed_volumes before mint/move; rollback when the record does not commit
```

## Summary

Adds the local implementation of `VolumeBackend`, so `boxlite volume create|ls|get|rm`, `boxlite run -v <id-or-name>:/path`, and the `/v1/volumes` routes and mount requests of `boxlite serve` work without a hosted API. A volume is one directory, `{home}/volumes/{id}/`: name and creation time in `{id}/.metadata.json`, payload in `{id}/_data/`, which is the only part a box mounts. Every entry point accepts the server-minted id or the unique name; an unknown reference is refused with not-found, as the hosted API does, never created on first use.

## Changes

- `LocalVolumeStore` (`src/boxlite/src/volumes/store.rs`): create by optional unique name, list, get, remove, and `payload_dir` for the runtime. A reference is validated before any filesystem access; ids take one stat, names a scan of the sidecars. `create` and `remove` hold an exclusive flock on the volumes directory so a name cannot be created twice across processes; `payload_dir` only reads and takes no lock. Directories without a `_data/` payload are not volumes (an upgrade leaves `volumes/anonymous/` behind), an undecodable sidecar demotes only its own volume to id-only, and a name is capped at the id length.
- Persisted mounts keep their identity: resolution writes the canonical volume id into `managed_volume` next to the payload path in `host_path`, so `remove_volume` refuses a volume any persisted box mounts, `force` included, and a persisted mount re-resolves to the same volume on clone or import.
- Anonymous mounts (`-v /data`) carry an explicit `anonymous` marker; the runtime creates the volume at box creation. The CLI mints nothing and touches no disk; its orphaned `ulid` dependency goes.
- Resolution (`resolve_managed_volumes`) is all-or-nothing: references are checked before anything is created, a pass that fails part-way removes what it created, and the ids it returns are what `create_inner` and `provision_box` remove when the box record does not commit, so create, import and clone leave no orphan volume behind.
- Request and persisted shapes are validated separately: requests keep the one-origin rule (`validate_in_request`) at the runtime facade for both backends; persisted options (`sanitize_persisted`) and archives (`sanitize_import`) accept the resolved shape, and an archive additionally checks a custom kernel source the way create does, since export bundles neither the kernel file nor the staged copy.
- `boxlite serve` accepts `volumes[]` on `POST /v1/boxes` as `{managed_volume, guest_path, read_only}` and rejects unknown fields; `read_only: true` is refused per the OpenAPI contract. The REST client refuses host binds and anonymous mounts before any network IO. `/v1/volumes` CRUD runs against the local store.
- Boot and the jailer treat a resolved mount as the directory share it is and refuse only a reference with no path.
- `LocalNamedVolumeStore` is renamed `LocalVolumeStore`; `VolumeID::parse`, `short` and `SHORT_LENGTH` are removed (no callers); `rederive_fsize_limit` is split out of `sanitize`; module docs and SDK comments describe the local backend where they still denied it.

## How to verify

- `cargo test -p boxlite --lib -- volume resolve_ rt_impl::tests import::tests options::tests jailer::tests id::tests` and `cargo test -p boxlite --doc -- runtime::id`.
- `cargo test -p boxlite-cli --bin boxlite -- volume serve::`; `cargo test -p boxlite --features rest --lib rest::runtime::`.
- `make fmt:check:rust`; `cargo clippy -p boxlite -p boxlite-cli --all-targets --all-features -- -D warnings`.
- Local, with a real VM (fresh `--home` under `/tmp`): `boxlite volume create --name smoke` prints a 12-char id; `boxlite run --rm -v smoke:/data alpine:latest sh -c "echo hi > /data/f"` then `boxlite run --rm -v smoke:/data alpine:latest cat /data/f` prints `hi` (also by id); `boxlite run --rm -v fresh:/data alpine:latest true` fails with not found until `boxlite volume create --name fresh`; `boxlite run --rm -v /data alpine:latest sh -c "echo anon > /data/x && cat /data/x"` prints `anon` and leaves one volume in `boxlite volume ls`; `boxlite volume rm smoke` is refused while a box mounts it and succeeds after that box is removed.
- Over REST: `boxlite serve --host 127.0.0.1 --port 18131`, then `curl -X POST :18131/v1/volumes -d "{\"name\":\"srv\"}"`, mount `srv` from a box created against `--url http://127.0.0.1:18131`; `curl -X DELETE :18131/v1/volumes/srv` answers 409 while a box mounts it and 204 after the box is gone. A mount with `host_path` or `read_only: true` in the body is refused.
- Not run here: `make test:integration:cli`, the full `make test`, SDK volume paths.

## Risks / rollout

- Behaviour change against the earlier revision of this PR: an unknown `-v name:/path` is refused with not found instead of being created on first use; create the volume first with `boxlite volume create --name`.
- Anonymous volumes are managed volumes: they appear in `volume ls` under their id and are not removed with their box yet (`boxlite volume rm <id>` reclaims them once no box holds them). A clone, or an import on the same machine, shares the anonymous volume of the source box.
- Read-only managed mounts are refused everywhere (CLI, REST client, `serve`) until implemented; the OpenAPI document already pins `read_only` to `false`.
- A trusted import of an archive whose custom kernel source is missing on this host is refused at import rather than failing at first boot.
- Boxes persisted by earlier revisions of this branch recorded only the payload path and are not recognised as volume holders; nothing released did.

## New Features
- Local runtimes now support creating, listing, resolving, and removing named volumes.
- Anonymous mounts create managed volumes automatically during box creation.
- Volume references support validated identifiers and lookup by name or ID.
- CLI and serve requests support structured managed-volume mounts.

## Bug Fixes
- Resolved volumes receive correct writable access, with rollback on provisioning failures.
- Mounted volumes cannot be removed, including with force.
- Improved sandbox access to certificate bundles and volume paths.
- Invalid network rate limits are rejected when networking is disabled.

## Documentation
- Updated volume behavior and storage documentation for local and REST runtimes.


src/boxlite/src/volumes/store.rs
```diff
@@ -30,3 +59,586 @@ pub struct VolumeInfo {
...
+    fn create_locked(&self, name: Option<&str>) -> BoxliteResult<VolumeInfo> {
+        let id = VolumeIDMint::mint().to_string();
+        let name = match name {
+            Some(name) => {
+                validate_volume_name(name)?;
+                if self.find_by_name(name)?.is_some() {
+                    return Err(BoxliteError::AlreadyExists(format!(
+                        "a volume named {name:?} already exists"
+                    )));
+                }
```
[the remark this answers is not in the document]
Monsteer714: thanks for review. this should be a problem. let me fix this.

src/boxlite/src/volumes/store.rs
```diff
@@ -30,3 +59,586 @@ pub struct VolumeInfo {
...
+            ))
+        })?;
+
+        let metadata = VolumeMetadata {
+            name,
+            created_at: Utc::now(),
+        };
+        self.write_metadata(&id, &metadata)?;
+        Ok(VolumeInfo {
+            id,
```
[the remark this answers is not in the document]
Monsteer714: thanks 4 review. should be a problem. let me fix this

src/boxlite/src/runtime/rt_impl.rs
```diff
@@ -1797,7 +1814,12 @@ impl super::backend::RuntimeBackend for LocalRuntime {
         options: BoxOptions,
         name: Option<String>,
     ) -> BoxliteResult<(LiteBox, bool)> {
-        let options = sanitize_local_options(&self.0.experimental_features, options).await?;
+        let options = sanitize_local_options(
+            &self.0.experimental_features,
+            crate::volumes::NamedVolumeStore::new(self.0.layout.home_dir()),
+            options,
+        )
+        .await?;
```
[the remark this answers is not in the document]
Monsteer714: thx 4 review. should be a problem creating orphan resources. let me fix this.

src/boxlite/src/volumes/store.rs
```diff
@@ -30,3 +59,586 @@ pub struct VolumeInfo {
...
+        };
+
+        // The sidecar lives inside `dir`, so this frees the name too.
+        fs::remove_dir_all(&dir).map_err(|e| {
+            BoxliteError::Storage(format!(
+                "failed to remove directory {}: {}",
+                dir.display(),
+                e,
+            ))
+        })
```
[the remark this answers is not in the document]
Monsteer714: thx 4 review. should be a concurrency problem. let me fix this. might lock all the write ops and be free to read ops

src/boxlite/src/runtime/rt_impl.rs
```diff
@@ -1749,32 +1749,44 @@ fn reject_local_unsupported_options(options: &BoxOptions) -> BoxliteResult<()> {
...
 
+/// Replace every managed-volume reference with the directory backing it.
+///
+/// The reference reaches the runtime as `managed_volume` — an id or a name —
+/// and only the store can say where that volume's payload lives. A name the
+/// store has never seen is created here, so `-v my-data:/data` works on first
+/// use the way it does with docker. Resolving at create rather than at boot
+/// keeps any failure where the caller is still holding the error, instead of
+/// part-way through `resolve_user_volumes`.
+fn resolve_managed_volumes(
```
DorianZheng, line 1776: seems it should be replaced in `serve`

src/boxlite/src/volumes/store.rs
```diff
@@ -30,3 +59,586 @@ pub struct VolumeInfo {
...
+
+/// What the store persists about a volume, inside its directory.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct VolumeMetadata {
+    name: String,
+    created_at: DateTime<Utc>,
+}
+
+#[derive(Debug, Clone)]
+pub struct NamedVolumeStore {
```
DorianZheng: seems `LocalVolumeStore` is better

src/boxlite/src/volumes/store.rs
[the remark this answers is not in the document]
Monsteer714: check again plz. should be resolved now.

src/boxlite/src/volumes/store.rs
Monsteer714: resolved

src/boxlite/src/runtime/rt_impl.rs
```diff
@@ -1749,20 +1760,32 @@ fn reject_local_unsupported_options(options: &BoxOptions) -> BoxliteResult<()> {
...
+    options: &mut BoxOptions,
+) -> BoxliteResult<()> {
+    for volume in &mut options.volumes {
+        let Some(reference) = volume.managed_volume.take() else {
+            continue;
+        };
+        volume.host_path = store
+            .payload_dir(&reference)?
+            .to_string_lossy()
+            .into_owned();
```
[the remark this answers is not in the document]
Monsteer714: thx 4 review. should be a problem but out of the scope of this pull request. will be fix in further pr when remove_volume() applying more options including checking in-use volumes.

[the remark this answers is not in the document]
Monsteer714: thats unnecessary. it will be included in a issue that has more completing scope. resolve this.

src/boxlite/src/volumes/store.rs
```diff
@@ -30,3 +59,586 @@ pub struct VolumeInfo {
...
+        }
+    }
+
+    /// Create a volume, naming it after its id when the caller supplies none.
+    ///
+    /// The name is what makes `-v my-data:/data` work without knowing the id,
+    /// so it has to be unique: a duplicate would make that reference ambiguous
+    /// and silently pick one of two volumes.
+    pub fn create(&self, name: Option<&str>) -> BoxliteResult<VolumeInfo> {
+        let _lock = self.lock_volumes_dir()?;
```
DorianZheng, line 89: lock is not necessary here if you can make sure volume id is unique

Monsteer714: seem to make sense

src/boxlite/src/runtime/rt_impl.rs
Monsteer714: vague description to me. would you clarify a little more details?

src/boxlite/src/volumes/store.rs
Monsteer714: after review, this lock actually guards the name of volume. (e.g., two concurrent `boxlite volume create --name foo` produce two volumes named `foo` and it might be an undefined behaviour)

src/boxlite/src/runtime/rt_impl.rs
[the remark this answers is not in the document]
Monsteer714: thx for review. should be a problem causing orphan resources. let me fix this but by validating the reference of the volume instead of adding roll back.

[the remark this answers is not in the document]
Monsteer714: thx for comment. let me fix it by splitting resolving and creating volumes into two funtions. or mounting a non-existing volume should cause error instead of create it.

src/boxlite/src/volumes/store.rs
```diff
@@ -30,3 +59,736 @@ pub struct VolumeInfo {
...
+        for entry in entries {
+            let entry = entry.map_err(|e| {
+                BoxliteError::Storage(format!(
+                    "failed to read an entry of volume dir {}: {}",
+                    self.volumes_dir.display(),
+                    e
+                ))
+            })?;
+            let path = entry.path();
+            if !path.is_dir() {
```
DorianZheng, line 174: Every directory under `volumes/` counts as a volume here. Anyone upgrading already has `~/.boxlite/volumes/anonymous/` sitting there from the old `-v /path` code, so it lists as a volume named `anonymous` (no sidecar, so it falls back to id-as-name) — and `volume rm anonymous` would `remove_dir_all` all their old anonymous data in one go. `-v anonymous:/data` also resolves, then fails at boot because `_data/` isn't there.

Filtering on `VolumeID::is_valid` before treating an entry as a volume would keep strays out.

src/boxlite/src/volumes/store.rs
```diff
@@ -30,3 +59,736 @@ pub struct VolumeInfo {
...
+            Err(e) if e.kind() == io::ErrorKind::NotFound => return Ok(None),
+            Err(e) => {
+                return Err(BoxliteError::Storage(format!(
+                    "failed to read volume metadata {}: {}",
+                    path.display(),
+                    e
+                )));
+            }
+        };
+        serde_json::from_slice(&body).map(Some).map_err(|e| {
```
DorianZheng, line 358: One corrupt sidecar fails the whole listing. This propagates through `volume_info` → `list` → `find_by_name`, so a single bad `.metadata.json` breaks name lookup for *every* volume, not just its own — create, get, rm and first-use mount all start failing.

Skipping the entry with a warn would keep the blast radius at one volume.

src/boxlite/src/volumes/store.rs
```diff
@@ -30,3 +59,736 @@ pub struct VolumeInfo {
...
+    ///
+    /// A reference the store has never seen becomes a new volume of that name,
+    /// so `-v my-data:/data` works on first use as it does with docker. Only
+    /// the mount path does this; [`Self::get`] and [`Self::remove`] stay strict.
+    pub(crate) fn payload_dir(&self, reference: &str) -> BoxliteResult<PathBuf> {
+        let _lock = self.lock_volumes_dir()?;
+        let dir = match self.locate(reference)? {
+            Some((_, dir)) => dir,
+            None => {
+                let created = self.create_locked(Some(reference))?;
```
DorianZheng, line 242: Auto-create is reachable from the REST surface, not just the CLI: `serve` builds a `managed_volume` spec straight out of the request body (`serve/mod.rs:1242`) and it lands here, so a typo'd name in `POST /v1/boxes` quietly creates an empty volume server-side. Intentional?

src/boxlite/src/volumes/store.rs
```diff
@@ -30,3 +59,736 @@ pub struct VolumeInfo {
...
+/// directory is always the id — but it is user-facing, so it stays printable
+/// and shell-safe.
+fn validate_volume_name(name: &str) -> BoxliteResult<()> {
+    let invalid = || {
+        BoxliteError::InvalidArgument(format!(
+            "invalid volume name {name:?}: at least two characters of \
+             [a-zA-Z0-9][a-zA-Z0-9_.-]"
+        ))
+    };
+    if name.len() < 2 {
```
DorianZheng, line 461: Character set is bounded but length isn't, while ids cap at 128 for exactly this reason. A multi-megabyte `--name` is accepted and written into the sidecar.

src/boxlite/src/runtime/rt_impl.rs
```diff
@@ -457,6 +457,17 @@ impl RuntimeImpl {
             };
         }
 
+        let store = crate::volumes::LocalNamedVolumeStore::new(self.layout.home_dir());
```
DorianZheng, line 460: `run_volume_op` (line 1897) already wraps store construction + `spawn_blocking` + join mapping. Reusing it here would stop the two paths drifting — they already disagree on the error text, "failed to join volume creation task" vs "failed to join volume store task".

src/cli/src/cli.rs
```diff
@@ -1350,14 +1330,8 @@ impl VolumeFlags {
...
                 crate::volumespec::MountOrigin::Anonymous => {
-                    // Random id for the directory name (same approach as Podman:
-                    // cryptographically random to avoid collisions under any load).
-                    let unique = ulid::Ulid::new().to_string();
-                    let dir = base.join("volumes").join("anonymous").join(unique);
-                    std::fs::create_dir_all(&dir).map_err(|e| {
-                        anyhow::anyhow!("failed to create anonymous volume dir {:?}: {}", dir, e)
-                    })?;
-                    VolumeSpec::bind_mount(dir.to_string_lossy().into_owned(), mount.guest_path)
+                    let id = VolumeIDMint::mint();
```
DorianZheng, line 1333: The store never adopts this id. `payload_dir` finds no directory for it, falls through to `create_locked`, which mints a **second** id and keeps this one as the volume's *name* — so `volume ls` shows two different random strings for one anonymous mount.

It also means every `-v /data` now leaves a permanent named volume behind, where before it was a scratch dir under `volumes/anonymous/`. The comment above says "the runtime creates the volume when the box is created", which is true but hides the double mint.

src/cli/src/volumespec.rs
```diff
@@ -13,10 +13,11 @@
 //!   `~`, a `\\` prefix, or `X:` mean a host path. Anything else is a managed
 //!   volume, addressed by id or by name.
 //!
-//! Docker's own failure mode does not carry over. There, a mistyped source
-//! silently creates an empty volume and the caller's data appears to vanish;
-//! boxlite never auto-creates, so an unknown reference is a loud "not found"
-//! from the server (`VolumeService.validateVolumes`).
+//! What an unknown reference does depends on the runtime. The local runtime
+//! creates the volume on first use, as docker does, so `-v data:/app` works
+//! without a prior `volume create`; a REST server answers "not found"
```
DorianZheng, line 18: `boxlite serve` is a REST server sitting on the local runtime, so it *does* auto-create on an unknown reference. This only holds for the hosted API.

DorianZheng (commented):
Read the store end to end. The shape is good — one directory per volume, sidecar beside the payload rather than inside it, lock held across scan+mkdir. Notes inline. Three things outside the diff that this PR makes wrong:

- `src/cli/Cargo.toml:54` — `ulid = "1.1"` is orphaned now, the anonymous-volume branch was its only user.
- `src/cli/src/commands/volume/mod.rs:8` and `src/cli/src/commands/serve/handlers/volumes.rs:3` still say the local runtime has no volume backend and every op answers `400 UnsupportedError`.
- `docs/reference/cli/README.md:694,863,874,881` — same sentence, plus "boxlite never auto-creates" and the `volumes/anonymous/<ulid>` layout.

Nit: `VolumeID::parse`, `short()` and `SHORT_LENGTH` have no caller outside id.rs's own tests.

src/boxlite/src/volumes/store.rs
```diff
@@ -30,3 +59,736 @@ pub struct VolumeInfo {
...
+        let _lock = self.lock_volumes_dir()?;
+        let Some((_, dir)) = self.locate(reference)? else {
+            if force {
+                return Ok(());
+            }
+            return Err(not_found(reference));
+        };
+
+        // The sidecar lives inside `dir`, so this frees the name too.
+        fs::remove_dir_all(&dir).map_err(|e| {
```
DorianZheng, line 215: No in-use check before `remove_dir_all`. `volume rm` (and `DELETE /v1/volumes/{id}`) will delete a directory that's currently virtiofs-shared into a running box, out from under the guest. This was unreachable before — `remove_volume` was an unconditional `Unsupported` — so the PR is what makes it possible.

Worse, `resolve_managed_volumes` does `managed_volume.take()` and writes the resolved path into `host_path`, so the persisted config no longer records *which* volume a box mounts. There's nothing left to check against without reverse-mapping host paths. Docker refuses removal while a container references the volume.

src/cli/src/cli.rs
DorianZheng, line 1333: Separate from the double-mint: this arm walks straight past the read-only guard twelve lines up. `-v my-data:/data:ro` bails, but `-v /cache:ro` lands here, becomes a managed volume, and then picks up `read_only: mount.read_only` on the way into `opts.volumes` — exactly the mount the guard exists to prevent.

It didn't apply before because the anonymous arm produced a *bind* mount, where read-only is fine.

DorianZheng (commented):
Second pass, mostly on the paths that changed behaviour rather than the store internals. The `read_only` one is the most concrete: the PR's own new test asserts a value the OpenAPI document forbids.

Two smaller things I didn't inline:

- `remove` holds the exclusive flock across `remove_dir_all` of the whole payload, so a slow delete blocks every unrelated create and first-use mount on the machine.
- `payload_dir` takes the exclusive lock unconditionally, so mounting an *existing* volume — the common case — serialises against every other mutation.

src/cli/src/commands/serve/mod.rs
```diff
@@ -1237,8 +1229,30 @@ fn build_box_options(req: &CreateBoxRequest) -> Result<BoxOptions, boxlite::Boxl
...
+        .as_ref()
+        .map(|mounts| {
+            mounts
+                .iter()
+                .map(|mount| {
+                    let mut spec = boxlite::runtime::options::VolumeSpec::managed_volume(
+                        mount.managed_volume.clone(),
+                        mount.guest_path.clone(),
+                    );
+                    spec.read_only = mount.read_only;
```
DorianZheng, line 1246: `read_only` goes straight from the wire onto a managed mount, but `openapi/box.openapi.yaml:1974` pins that field to `enum: [false]` with "rejected rather than silently downgraded", and the REST client's `sanitize_remote` refuses to send it.

The new test below locks the mismatch in: `build_box_options_carries_volume_mounts_from_the_wire` posts `"read_only":true` and asserts `options.volumes[0].read_only`. Either the contract moves or this needs to reject.

src/boxlite/src/runtime/rt_impl.rs
```diff
@@ -1749,20 +1760,32 @@ fn reject_local_unsupported_options(options: &BoxOptions) -> BoxliteResult<()> {
...
-    }
+    Ok(())
+}
 
+/// Replace every managed-volume reference with the directory backing it.
+///
+/// The reference reaches the runtime as `managed_volume` — an id or a name —
+/// and only the store can say where that volume's payload lives. A name the
+/// store has never seen is created here, so `-v my-data:/data` works on first
+/// use the way it does with docker. Runs in `create_inner` only once a new box
```
DorianZheng, line 1771: "only once a new box is certain" doesn't hold. The `lookup_box` check is before this, but `add_box` further down can still fail — the TOCTOU comment right there says so and has a recovery branch for it — and by then the volumes are already on disk. The losing racer creates volumes for a box it never gets.

`spawn_blocking` also doesn't cancel, so a dropped create future (client disconnect on `serve`) still creates the volume.

src/boxlite/src/litebox/init/types.rs
```diff
@@ -51,15 +51,17 @@ pub fn resolve_user_volumes(volumes: &[VolumeSpec]) -> BoxliteResult<Vec<Resolve
...
-        // points the caller at their own filesystem instead of at the
-        // runtime they used.
+        // A reference should already be a directory by now: the local runtime
+        // resolves it against its store at create (`resolve_managed_volumes`,
+        // in `create_inner`), and a REST runtime never sends one this far.
+        // Reaching here means a spec skipped that step, so say that rather
+        // than let it fall through to the host-path checks below and surface
+        // as "host path does not exist: my-data", which points the caller at
+        // their own filesystem.
         if let Some(reference) = &vol.managed_volume {
```
DorianZheng, line 61: Good trap, but only `create_inner` establishes the invariant it enforces. `import_box` is the other create path and never calls `resolve_managed_volumes` — and its volume gate is untrusted-only (`import.rs:101` returns early for `Trusted`), so a trusted archive carrying a `managed_volume` reaches here unresolved and boots into an `Internal`, i.e. a 500.

Pre-existing gap, but this PR is what makes the reference resolvable everywhere else.

src/boxlite/src/volumes/store.rs
Monsteer714: thx 4 review. should be a problem. let me fix it.

src/cli/src/volumespec.rs
Monsteer714: thx for review. modified to
`//! What an unknown reference does depends on who resolves it. The local`
`//! runtime creates the volume on first use, as docker does, so -v data:/app`
`//! works without a prior volume create; boxlite serve sits on the local`
`//! runtime and inherits that. The hosted API answers "not found"`

src/boxlite/src/volumes/store.rs
Monsteer714: thx 4 review. intended. but after review it may return `NotFound` as hosted API did.

src/boxlite/src/volumes/store.rs
Monsteer714: thx for review. should be a problem. let me fix that.

src/boxlite/src/runtime/rt_impl.rs
Monsteer714: thx for review. should be a problem. let me fix that.

src/cli/src/cli.rs
Monsteer714: thx for review. should be a problem. let me fix that.

src/boxlite/src/volumes/store.rs
Monsteer714: thx for review. should be a problem. let me fix that.

src/boxlite/src/volumes/store.rs
Monsteer714: thx for review. should be a problem. considering if i should fix this in this scope.

src/boxlite/src/volumes/store.rs
Monsteer714: mark as resolved in my fork.

src/boxlite/src/volumes/store.rs
Monsteer714: mark as resolved in my fork.

src/boxlite/src/volumes/store.rs
Monsteer714: mark as resolved in my fork.

src/boxlite/src/runtime/rt_impl.rs
Monsteer714: mark as resolved in my fork.

src/cli/src/volumespec.rs
Monsteer714: mark as resolved in my fork.

src/cli/src/cli.rs
Monsteer714: thx for review. anonymous volumes are always created by store backend, so it should become a managed volume. mark as resolved in my fork.

src/boxlite/src/volumes/store.rs
Monsteer714: after review, `resolve_managed_volumes` may not do the `take()` so persisted config will have `VolumeID` as `managed_volume` OR `remove_volume` check the volume in-use by the host_path.

src/boxlite/src/litebox/init/types.rs
Monsteer714: thx for review. should be a problem. the coupling problem here is complex. let me fix this.

Monsteer714: mark as resolved in local fork. also `clone()` is the other create path that should carry the persisted `VolumeSpec`.

src/boxlite/src/volumes/store.rs
Monsteer714: mark as resolved in local fork.

src/cli/src/commands/serve/mod.rs
Monsteer714: mark as resolved in my local fork.

src/boxlite/src/runtime/rt_impl.rs
Monsteer714: thx for review; should be a problem and have already resolved by rollback the volumes created; also do this at `provision_box()` that import and clone will call.

src/boxlite/src/litebox/init/types.rs
Monsteer714: resolved.

src/cli/src/commands/serve/mod.rs
Monsteer714: resolved.

src/cli/src/cli.rs
Monsteer714: resolved.

src/boxlite/src/volumes/store.rs
Monsteer714: resolved.

src/cli/src/volumespec.rs
Monsteer714: resolved.

src/cli/src/cli.rs
Monsteer714: resolved.

src/boxlite/src/runtime/rt_impl.rs
Monsteer714: resolved.

src/boxlite/src/volumes/store.rs
Monsteer714: resolved.

src/boxlite/src/volumes/store.rs
Monsteer714: resolved.

src/boxlite/src/runtime/options.rs
```diff
@@ -784,6 +877,24 @@ impl VolumeSpec {
...
+    /// with this rule and not the request one.
+    pub fn validate_in_persisted(&self) -> BoxliteResult<()> {
+        let managed_volume = &self.managed_volume;
+        let guest_path = &self.guest_path;
+        if self.host_path.is_empty() {
+            return Err(boxlite_shared::errors::BoxliteError::InvalidArgument(
+                format!("volume mount {managed_volume:?} to {guest_path:?} needs a host_path"),
+            ));
+        }
+        Ok(())
```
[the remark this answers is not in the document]
Monsteer714: thx for review. should be a problem and fix by requiring `managed_volume` and `host_path` are both empty in `resolve_managed_volume` when creating anonymous volume. check if this is feasible.

src/boxlite/src/runtime/rt_impl.rs
```diff
@@ -1864,35 +2041,100 @@ impl super::images::ImageBackend for LocalRuntime {
...
+            "Cannot {what}: runtime has been shut down"
+        )));
+    }
+    let store = crate::volumes::LocalVolumeStore::new(runtime.layout.home_dir());
+    tokio::task::spawn_blocking(move || op(store))
+        .await
+        .map_err(|error| {
+            BoxliteError::Internal(format!("failed to join volume store task: {error}"))
+        })?
+}
```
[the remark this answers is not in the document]
Monsteer714: thx for review. should be a problem. let me fix this.

src/boxlite/src/volumes/store.rs
```diff
@@ -30,3 +59,940 @@ pub struct VolumeInfo {
...
```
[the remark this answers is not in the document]
Monsteer714: thx for review. should be a problem. let me fix this.

src/cli/src/commands/serve/types.rs
```diff
@@ -175,6 +175,21 @@ pub(super) struct ListBoxesResponse {
...
 }
 
+/// One mount in a create-box request: which volume, and where in the box.
+///
+/// Field names are the REST wire contract every client already speaks —
+/// `CreateBoxVolumeSpec` in `boxlite::rest`, the dashboard, the SDK docs —
+/// so `managed_volume`, not a shorter local spelling.
+#[derive(Deserialize)]
+#[serde(deny_unknown_fields)]
+pub(super) struct CreateVolumeMount {
```
DorianZheng, line 185: naming should be consistent with `SecretSpecRequest`

src/boxlite/src/runtime/rt_impl.rs
```diff
@@ -1864,35 +2041,100 @@ impl super::images::ImageBackend for LocalRuntime {
...
+        let rt = self.0.clone();
+        // A volume a persisted box mounts is in use whatever the box's state:
+        // its directory is what virtiofs shares, and `force` does not change
+        // that, as it does not in docker (`volume/service/store.go`, the
+        // `hasRef` refusal). Holders are found by the canonical id resolution
+        // wrote into each mount, so a box that mounted by name is recognised
+        // too. The scan runs outside the store's lock: a create that
+        // references this volume between the scan and the `remove` below is a
+        // window this accepts, rather than holding every create for the
+        // length of a `remove_dir_all`.
```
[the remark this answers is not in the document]
Monsteer714: thx for review. this is an existing problem that already mentioned. let me fix this.

src/boxlite/src/runtime/options.rs
```diff
@@ -752,16 +823,38 @@ impl VolumeSpec {
...
+    /// Persisted mounts are judged by [`Self::validate_in_persisted`] instead:
+    /// after resolution a managed mount legitimately carries both fields.
+    ///
+    /// An anonymous mount that also carries an origin is not refused here;
+    /// what resolution does with that origin is `resolve_managed_volumes`'s
+    /// call, and that is where the behaviour is tested.
+    pub fn validate_in_request(&self) -> BoxliteResult<()> {
         let guest_path = &self.guest_path;
         match &self.managed_volume {
+            None if (self.anonymous && self.host_path.is_empty()) => Ok(()),
```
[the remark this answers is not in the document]
Monsteer714: thx for review. this is already fixed in my local fork.

src/boxlite/src/runtime/rt_impl.rs
```diff
@@ -1228,48 +1267,65 @@ impl RuntimeImpl {
...
-        state.set_lock_id(lock_id);
+            let mut state = BoxState::new();
+            state.set_status(initial_status);
 
-        if let Err(e) = self.box_manager.add_box(&config, &state) {
-            let _ = self.lock_manager.free(lock_id);
-            let _ = std::fs::remove_dir_all(&config.box_home);
-            config.sockets().remove();
-            return Err(e);
+            let lock_id = self.lock_manager.allocate()?;
```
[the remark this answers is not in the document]
Monsteer714: thx for review. this is already fixed in my local fork.

src/boxlite/src/volumes/store.rs
```diff
@@ -30,3 +59,940 @@ pub struct VolumeInfo {
...
+                        "a volume named {name:?} already exists"
+                    )));
+                }
+                name.to_string()
+            }
+            None => id.clone(),
+        };
+
+        let payload = self.volumes_dir.join(&id).join(PAYLOAD_DIR);
+        fs::create_dir_all(&payload).map_err(|e| {
```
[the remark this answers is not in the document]
Monsteer714: thx for review. should be a problem. never consider of this since probability of same id is low. let me fix this to improve the robustness.

src/boxlite/src/runtime/rt_impl.rs
```diff
@@ -1864,35 +2041,100 @@ impl super::images::ImageBackend for LocalRuntime {
...
+    F: FnOnce(crate::volumes::LocalVolumeStore) -> BoxliteResult<T> + Send + 'static,
+    T: Send + 'static,
+{
+    if runtime.shutdown_token.is_cancelled() {
+        return Err(BoxliteError::Stopped(format!(
+            "Cannot {what}: runtime has been shut down"
+        )));
+    }
+    let store = crate::volumes::LocalVolumeStore::new(runtime.layout.home_dir());
+    tokio::task::spawn_blocking(move || op(store))
```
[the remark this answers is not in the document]
Monsteer714: thx for review. this is an existing problem that already mentioned. let me fix this.

src/cli/src/commands/serve/types.rs
Monsteer714: thx for review. how about `MountSpecRequest` ?

src/cli/src/commands/volume/mod.rs
```diff
@@ -5,9 +5,11 @@
...
 //!
 //! Each leaf module owns its own `Args` struct and `run()`; this module holds
-//! the subcommand enum and dispatches. Volumes are a REST-runtime capability:
-//! the local runtime has no volume backend and every command returns
-//! "not supported" against it.
+//! the subcommand enum and dispatches. Against a REST runtime the commands
+//! call `/v1/volumes`; against the local runtime they work on the store under
+//! `{home}/volumes/`, one directory per volume. Either way an unknown id or
+//! name is "not found": only `create` and an anonymous `-v /path` mount make
+//! a volume.
```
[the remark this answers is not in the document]
Monsteer714: thx for review. this is a doc problem. let me fix this.

src/boxlite/src/runtime/rt_impl.rs
```diff
@@ -1864,35 +2041,100 @@ impl super::images::ImageBackend for LocalRuntime {
...
+                .map(|(cfg, _)| cfg.id.to_string())
+                .collect();
+            if !holders.is_empty() {
+                return Err(BoxliteError::InvalidState(format!(
+                    "failed to remove volume: {}, it is in-use by boxes: [{}]",
+                    reference,
+                    holders.join(", "),
+                )));
+            }
+            store.remove(&reference, force)
```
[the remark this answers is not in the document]
Monsteer714: thx for review. this should be a problem. let me fix this.

src/boxlite/src/runtime/rt_impl.rs
```diff
@@ -1749,20 +1805,141 @@ fn reject_local_unsupported_options(options: &BoxOptions) -> BoxliteResult<()> {
...
+            match volume.managed_volume.clone() {
+                Some(reference) => reference,
+                None => continue,
+            }
+        };
+        volume.managed_volume = Some(store.get(&reference)?.id);
+        volume.host_path = store
+            .payload_dir(&reference)?
+            .to_string_lossy()
+            .into_owned();
```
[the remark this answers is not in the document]
Monsteer714: thx for review. marked as resolved in mny local fork.

Monsteer714:
@coderabbitai review again when this is ready, since several commits have pushed.

Monsteer714:
@coderabbitai you might focus on the orphaned `Configured` box that will be created when the `create_box()` was interrupted by its caller e.g. `boxlite serve`. add it to comment if you find it is true.

src/boxlite/src/volumes/store.rs
```diff
@@ -30,3 +59,1025 @@ pub struct VolumeInfo {
...
+        match serde_json::from_slice(&body) {
+            Ok(metadata) => Ok(Some(metadata)),
+            Err(e) => {
+                tracing::warn!(
+                    volume = %id,
+                    path = %path.display(),
+                    error = %e,
+                    "ignoring undecodeable volume sidecar"
+                );
+                Ok(None)
```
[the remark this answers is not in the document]
Monsteer714: thx for review. should be a problem.

PR #1435, opened 2026-09-03
feat(net): cap per-box network bandwidth

ltstriker:
## Call graph

```text
Before
  NetworkFlags::apply_to      (CLI · src/cli/src/cli.rs:1185)                     — mode + allowlist only
    └─ build_network_backend  (spawn · src/boxlite/src/litebox/init/tasks/vmm_spawn.rs:347)
         └─ GvproxyBackend::spec  (backend · src/boxlite/src/net/gvproxy/services.rs:336)
              └─ GvproxyInstance::from_config (shim · src/boxlite/src/net/gvproxy/instance.rs:86)
                   └─ create_instance (ffi · src/boxlite/src/net/gvproxy/ffi.rs:22)
                        └─ gvproxy_create (bridge · gvproxy-bridge/main.go:304)
                             └─ vn.AcceptQemu(ctx, acceptedConn)  (main.go:570)   — raw conn, unshaped

After
  NetworkFlags::apply_to      (CLI · src/cli/src/cli.rs:1184)                     — also sets advanced.network_rate_limit
    └─ build_network_backend  (spawn · vmm_spawn.rs:347)                          — carries it into NetworkBackendConfig
         └─ GvproxyBackend::spec  (backend · services.rs:336)                     — onto NetworkBackendSpec
              └─ GvproxyInstance::from_config (shim · instance.rs:86)             — GvproxyConfig::with_rate_limit
                   └─ bucket_from_kbps (config · config.rs:149)                   — kbit/s to a token bucket
                        └─ create_instance (ffi · ffi.rs:22)                      — rate_limit on the wire
                             └─ gvproxy_create (bridge · main.go:304)             — RateLimitConfig.Validate
                                  └─ wrapConn (shaper · gvproxy-bridge/shaper.go:413)   — NEW, nil config returns the conn as-is
                                       └─ vn.AcceptQemu(ctx, shapedConn)  (main.go:570)
                                            ├─ shapedConn.Read  (shaper.go:504)   — TX: charge after read, sleep off the deficit
                                            └─ shapedConn.Write (shaper.go:542)   — RX: bounded queue, paced by a goroutine
```

## Summary

Caps a box's network bandwidth per direction through `advanced.network_rate_limit` (`NetworkRateLimit`) and the `--net-tx-kbps`/`--net-rx-kbps` flags. The limit is enforced in the gvproxy bridge by shaping the guest's own link, so it covers TCP, UDP, ICMP and ARP with one budget per direction and cannot be turned off from inside the guest. The cap threads from the CLI to the bridge; before, nothing carried it and the guest link was handed to the switch unwrapped.

## Changes

- The cap is `advanced.network_rate_limit`, typed `NetworkRateLimit`, grouped with the other local-only expert knobs (`security`, `isolate_mounts`, `privileged`) rather than placed inside `NetworkConfig`: those object-shaped types are the REST wire form and deny unknown fields. `sanitize_remote` refuses it for remote runtimes and the wire type has no field to carry it, the same treatment `SecurityOptions` gets. An unlimited value is omitted from the serialized form, so an ordinary box's persisted options and exported manifest keep their previous shape.
- Internally the value travels as `rate_limit` on `NetworkBackendConfig` and `NetworkBackendSpec`, matching `GvproxyConfig::rate_limit`; validation errors name `advanced.network_rate_limit.<field>`.
- Directions follow Firecracker's net device and are named from the box's point of view — `tx` is what the box sends. The cap is on the interface, so inbound port-forward traffic counts against the same budget as a reply to an outbound request.
- TX is charged after the read, so the balance goes negative and the next read sleeps off the overshoot; declining to read is itself the backpressure. RX cannot do the same — `txPkt` holds `writeLock` and `connLock` across `conn.Write` and `txBuf` disconnects the VM on any non-`ENOBUFS` error — so it queues and drops instead, paced by a goroutine.
- A passive framer tracks frame boundaries so the buckets charge frame bytes rather than socket bytes. Billing the length prefix would cost Linux four bytes per frame and macOS none: 0.26% on a 1514-byte frame but 5.7% on a stream of ACKs, and inconsistent across platforms.
- `GvproxyInstance::new` took five positional arguments and would have taken six; it now takes the config it was assembling.

## How to verify

```bash
cd src/deps/libgvproxy-sys/gvproxy-bridge && go test -race ./...
cargo test -p boxlite --no-default-features --features rest --lib -- net::gvproxy runtime::options rest::
cargo test -p boxlite-cli --bins
cargo check --workspace --all-targets
```

End to end, against a host serving a 4 MiB file:

```bash
boxlite run --rm --net-rx-kbps 2000 alpine \
  sh -c 'wget -q -O /tmp/p http://host.boxlite.internal:PORT/payload.bin && wc -c < /tmp/p'
```

Measured 242 and 483 kB/s against 250 and 500 configured, and 253.6 kB/s on the tx side against 250 — measured at the receiver, since libkrun's 16 MiB `SO_SNDBUF` lets the guest finish writing long before the bytes are on the wire.

## Risks / rollout

- Default is uncapped, and an uncapped box gets the connection back unwrapped, so the shaper is not in the path at all.
- No automated test drives the CLI-to-bridge chain: a bandwidth assertion needs a real VM and tens of seconds of wall clock. The measurement above is manual.
- Shaper counters reach no operator surface. The live `/stats` path serves `vn.ServicesMux()` directly with no bridge-owned mux to inject into, so an RX queue overflow shows up only in a `Debug` log at close.
- macOS ships the tx cap enabled with no runtime guard. Its guest link is a datagram socket whose sender behaviour under a full receive buffer is unverified and may drop frames rather than shape them; the flag help and all three docs say so.

## Summary by CodeRabbit

* **New Features**
  * Added inbound and outbound network rate limits through `boxlite run` and `boxlite create`.
  * Limits cover TCP, UDP, ICMP, and ARP traffic; unset or `0` means unlimited.
  * Added rate-limited networking support on Linux and macOS, including connection deadline handling.

* **Bug Fixes**
  * Remote runtimes now reject unsupported non-unlimited network limits before connecting.
  * Improved traffic shaping under backpressure and read timeouts.

* **Documentation**
  * Updated CLI and Rust API references with rate-limit settings and platform limitations.

### DIFF
--- a/src/boxlite/src/net/gvproxy/config.rs
+++ b/src/boxlite/src/net/gvproxy/config.rs
@@ -97,6 +97,64 @@ pub struct GvproxyConfig {
     /// PEM-encoded MITM CA private key (PKCS8 format, consumed by Go).
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub ca_key_pem: Option<String>,
+
+    /// Per-direction bandwidth cap for the guest link. Absent means unlimited.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub rate_limit: Option<GvproxyRateLimit>,
+}
+
+/// Token buckets handed to the Go bridge, one per direction. Field names match
+/// `RateLimitConfig` in gvproxy-bridge/shaper.go; the two must stay in sync.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct GvproxyRateLimit {
+    /// Internet to guest.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub rx: Option<GvproxyTokenBucket>,
+    /// Guest to internet.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub tx: Option<GvproxyTokenBucket>,
+}
+
+/// Firecracker's bucket shape: a capacity in bytes and the time to refill it.
+/// Sustained rate is `size * 1000 / refill_time_ms` bytes per second.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct GvproxyTokenBucket {
+    pub size: u64,
+    pub refill_time_ms: u64,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub one_time_burst: Option<u64>,
+}
+
+/// Refill period for a derived bucket. Fixed at 100ms: short enough to keep the
+/// burst small, long enough that the shaper is not waking on every frame.
+const RATE_LIMIT_REFILL_MS: u64 = 100;
+
+impl From<crate::runtime::options::NetBandwidth> for GvproxyRateLimit {
+    fn from(bw: crate::runtime::options::NetBandwidth) -> Self {
+        Self {
+            rx: bucket_from_kbps(bw.rx_kbps),
+            tx: bucket_from_kbps(bw.tx_kbps),
+        }
+    }
+}
+
+/// Kilobits per second to a token bucket. `None` and `0` both mean unlimited,
+/// so this is the one place the user-facing unit is converted.
+///
+/// `size` is not floored here even though the bridge floors bucket *capacity* at
+/// one maximum frame. The bridge derives the sustained rate from this exact
+/// (size, refill_time_ms) pair and keeps the capacity floor in a separate field,
+/// so raising `size` to that floor would not enlarge the burst — it would raise
+/// the rate itself, silently lifting every cap below ~5.2 Mbit/s.
+fn bucket_from_kbps(kbps: Option<u64>) -> Option<GvproxyTokenBucket> {
+    let kbps = kbps.filter(|k| *k > 0)?;
+    // kbit/s -> byte/s is * 1000 / 8, i.e. * 125.
+    let bytes_per_sec = kbps.saturating_mul(125);
+    Some(GvproxyTokenBucket {
+        size: bytes_per_sec / (1000 / RATE_LIMIT_REFILL_MS),
+        refill_time_ms: RATE_LIMIT_REFILL_MS,
+        one_time_burst: None,
+    })
 }
 
 /// Secret configuration for gvproxy MITM proxy.
@@ -139,6 +197,7 @@ impl std::fmt::Debug for GvproxyConfig {
                 "ca_key_pem",
                 &self.ca_key_pem.as_ref().map(|_| "[REDACTED]"),
             )
+            .field("rate_limit", &self.rate_limit)
             .finish()
     }
 }
@@ -187,6 +246,7 @@ fn defaults_with_socket_path(socket_path: PathBuf) -> GvproxyConfig {
         secrets: Vec::new(),
         ca_cert_pem: None,
         ca_key_pem: None,
+        rate_limit: None,
     }
 }
 
@@ -290,6 +350,17 @@ impl GvproxyConfig {
     /// Set secrets for MITM proxy injection.
     pub fn with_secrets(mut self, secrets: Vec<GvproxySecretConfig>) -> Self {
         self.secrets = secrets;
+        self
+    }
+
+    /// Set the per-direction bandwidth cap. An unlimited value is left off the
+    /// wire entirely rather than serialized as an empty object.
+    pub fn with_rate_limit(mut self, bandwidth: crate::runtime::options::NetBandwidth) -> Self {
+        self.rate_limit = if bandwidth.is_unlimited() {
+            None
+        } else {
+            Some(bandwidth.into())
+        };
         self
     }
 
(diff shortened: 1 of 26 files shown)


src/boxlite/src/net/gvproxy/config.rs
```diff
@@ -97,6 +97,64 @@ pub struct GvproxyConfig {
...
+/// the rate itself, silently lifting every cap below ~5.2 Mbit/s.
+fn bucket_from_kbps(kbps: Option<u64>) -> Option<GvproxyTokenBucket> {
+    let kbps = kbps.filter(|k| *k > 0)?;
+    // kbit/s -> byte/s is * 1000 / 8, i.e. * 125.
+    let bytes_per_sec = kbps.saturating_mul(125);
+    Some(GvproxyTokenBucket {
+        size: bytes_per_sec / (1000 / RATE_LIMIT_REFILL_MS),
+        refill_time_ms: RATE_LIMIT_REFILL_MS,
+        one_time_burst: None,
+    })
```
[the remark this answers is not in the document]
ltstriker: fixed

src/boxlite/src/runtime/options.rs
```diff
@@ -343,6 +343,15 @@ pub struct BoxOptions {
...
     pub ports: Vec<PortSpec>,
+    /// Bandwidth cap for the box's network interface.
+    ///
+    /// A sibling of `network` rather than a field inside it, for the same
+    /// reason `ports` is: the object-shaped `NetworkConfig` types are the REST
+    /// wire form and deny unknown fields, while this is local-only. It is also
+    /// bidirectional, so it belongs to neither the outbound nor the inbound
+    /// half.
+    #[serde(default)]
+    pub net_bandwidth: NetBandwidth,
```
DorianZheng: move to advanced options, better name required

DorianZheng: how about `NetworkRateLimiter`?

ltstriker: change to `advanced.network_rate_limit`.  `limiter` give me a sense like some instance but `limit` like a constraint which i think better.

src/boxlite/src/net/mod.rs
```diff
@@ -91,6 +94,10 @@ pub struct NetworkBackendSpec {
     /// PEM-encoded MITM CA private key (PKCS8, minted when secrets are configured).
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub ca_key_pem: Option<String>,
+    /// Per-direction rate limit for the guest link. Defaulted so a spec
+    /// written by an older core still deserializes.
+    #[serde(default, skip_serializing_if = "NetworkRateLimit::is_unlimited")]
+    pub rate_limit: NetworkRateLimit,
```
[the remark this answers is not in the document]
ltstriker: Not applicable: the `net_bandwidth` key never shipped. It exists only in this branch's own commits (825f311b..cc3a0b83) and is absent from `main` and every tag. The spec is also not persisted: `ShimController::start` rebuilds `InstanceSpec` from `BoxOptions` on every spawn and hands the JSON to a shim from the same build (`src/boxlite/src/vmm/controller/shim.rs:322`), and reattach produces no spec at all. There is no path where an older writer's `net_bandwidth` is read by this code, so the alias would be dead compatibility code.

src/boxlite/src/runtime/options.rs
```diff
@@ -1258,6 +1268,128 @@ mod tests {
         );
     }
 
+    fn rate_limited(tx_kbps: Option<u64>, rx_kbps: Option<u64>) -> AdvancedBoxOptions {
+        let mut advanced = AdvancedBoxOptions::default();
+        advanced.network_rate_limit = NetworkRateLimit { tx_kbps, rx_kbps };
+        advanced
+    }
```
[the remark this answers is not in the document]
ltstriker: Not applicable: no released build ever wrote a top-level `net_bandwidth` into `box_options`. The key exists only in this unmerged branch's commits (825f311b..cc3a0b83) and is absent from `main` and every tag, so there are no legacy archives to migrate. The only affected case is a developer who exported a box from an earlier commit of this branch; that box imports uncapped. The separate forward-compatibility question (an older importer silently dropping `advanced.network_rate_limit` because `archive_version_for_options` does not bump for it) predates this rename and is independent of it.

src/boxlite/src/net/mod.rs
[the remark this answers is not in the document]
ltstriker: never show in main

src/boxlite/src/runtime/options.rs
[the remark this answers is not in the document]
ltstriker: never show in main

src/deps/libgvproxy-sys/gvproxy-bridge/shaper.go
```diff
@@ -0,0 +1,753 @@
...
+
+// ---------------------------------------------------------------------------
+// shaped connection
+// ---------------------------------------------------------------------------
+
+type shapedConn struct {
+	net.Conn
+
+	hdrLen int
+	tx     *tokenBucket
```
DorianZheng: `txTokenBucket` for readable?

ltstriker: fixed

src/deps/libgvproxy-sys/gvproxy-bridge/shaper.go
```diff
@@ -0,0 +1,767 @@
...
+}
+
+// writeFrame reports whether the pacer should keep running.
+func (c *shapedConn) writeFrame(frame []byte) bool {
+	for {
+		if _, err := c.Conn.Write(frame); err != nil {
+			if errors.Is(err, syscall.ENOBUFS) {
+				// The same retry upstream performs (gvisor-tap-vsock#367), but
+				// paced: a bare continue spins a core while the buffer drains.
+				if !c.sleep(200 * time.Microsecond) {
```
DorianZheng, line 704: Could you elaborate a little bit on how this happen and why do we sleep for 200ns

DorianZheng: I think we should pass through the error, let the upstream caller handles it instead of cover the error

src/deps/libgvproxy-sys/gvproxy-bridge/shaper.go
```diff
@@ -0,0 +1,767 @@
...
+// frame is fine without a pool: the queue is bounded and drains at the
+// configured rate, so churn is capped by the limit the user asked for.
+func (c *shapedConn) Write(p []byte) (int, error) {
+	if c.rxTokenBucket == nil {
+		return c.Conn.Write(p)
+	}
+
+	size := int64(len(p))
+	if c.queueBytes.Load()+size > c.maxQueueBytes {
+		c.dropped(size)
```
DorianZheng, line 640: any explicit error to let upstream wait?

DorianZheng, line 640: any explicit error to let upstream wait?

DorianZheng: ENOBUFS?

ltstriker: upstream will never wait but just retry with writelock, which will make it [spin lock](https://github.com/containers/gvisor-tap-vsock/blob/main/pkg/tap/switch.go#L196)

drop it just make the box found that the network is congested and pkt was dropped, then it can slow down the send rate. I think it is just fit our need.

ltstriker: upstream will never wait but just retry with writelock, which will make it [spin lock](https://github.com/containers/gvisor-tap-vsock/blob/main/pkg/tap/switch.go#L196)

src/deps/libgvproxy-sys/gvproxy-bridge/shaper.go
ltstriker: we use shaping model to limit net rate. ( use a queue to store the recv data when netstack transfer to conn. and a coroutine `pace` to pull the queue within the rate limit and write that to the box.   It can make the limit more smooth, less drop).

the `ENOBUFS ` happened when the `pace ` write the box net stack but its buff is full.  At that time, we should just wait the box free.  I think it is received when the pkt is queued, and no upstream is exist at that time.

ori: look the before comment,  the `ENOBUFS `  can be perform when `conn.write`,  return `ENOBUFS `  will make a spin in lock.

ps. there is 200ms. maybe can be more considered, but here just let it try.

src/deps/libgvproxy-sys/gvproxy-bridge/shaper.go
DorianZheng: From the link you paste here, seems upstream never read returned `len(p)`

src/deps/libgvproxy-sys/gvproxy-bridge/shaper.go
DorianZheng: `200 * time.Microsecond` why this is 200ms?

DorianZheng: > It can make the limit more smooth, less drop).

I don't think this is correct. If the sender keep flooding the queue, you have to drop anyway unless you have unlimited queue which is infeasible

src/deps/libgvproxy-sys/gvproxy-bridge/shaper.go
ltstriker: it is, in here.

src/deps/libgvproxy-sys/gvproxy-bridge/shaper.go
ltstriker: > `200 * time.Microsecond` why this is 200ms?

200us,  type error.  currently just a magic number. maybe some fix policy will be better in the future.

ltstriker: > > It can make the limit more smooth, less drop).
> 
> I don't think this is correct. If the sender keep flooding the queue, you have to drop anyway unless you have unlimited queue which is infeasible

if it is raw udp, it will be real. But in tcp or udp with congest control,  the shaping model give the congest control a chance to fit the rate limit. the queue make the transfer rate fix in the rate limit which means more latency =》 bigger `bandwidth x latency` =>  sender get the latency signal and slow down the send, and sending rate is convergence to the rate limit finnaly.

It is relate to the policer model, which pass the packet to box just depend on the free token. It just have a decision to drop or not.  When drop , rate will be hard to convergence (etc. tcp will half the rate).

DorianZheng (approved):
LGTM

PR #1440, opened 2026-09-04
ci(node): commit the SDK lockfile and install it with npm ci

ltstriker:
## Summary

`sdks/node` had no committed lockfile, so every CI leg re-resolved its dependency tree from the registry at job time. Commit the lockfile and install it with `npm ci`, so all legs test one tree and the install stops depending on what the registry serves that hour.

## Call graph

Before
```
node                       (Job · .github/workflows/test.yml:264)
  └─ Set up Node.js        (setup-node · .github/workflows/test.yml:279)  — no cache, no lock path
       └─ npm install      (step · .github/workflows/test.yml:286)        ← re-resolves per leg: vite 8 on Node 22, vite 6 on Node 18; npm 10.x's arborist crashes walking the Node 22 peer set
            └─ npm run test (step · .github/workflows/test.yml:290)       — tests a tree nobody pinned

node                       (Job · .github/workflows/lint.yml:188)
  └─ npm install           (step · .github/workflows/lint.yml:203)        ← same re-resolution, uncached
       └─ npx tsc --noEmit (step · .github/workflows/lint.yml:213)
```

After
```
sdks/node/package-lock.json (new · 216 entries)                           — resolved under Node 18

node                       (Job · .github/workflows/test.yml:264)
  └─ Set up Node.js        (setup-node · .github/workflows/test.yml:279)  — cache: npm, keyed on the lockfile
       └─ npm ci           (step · .github/workflows/test.yml:298)        — installs the committed tree verbatim
            └─ npm run test (step · .github/workflows/test.yml:300)       — all legs test the same tree

node                       (Job · .github/workflows/lint.yml:188)
  └─ Set up Node.js        (setup-node · .github/workflows/lint.yml:196)  — same cache key
       └─ npm ci           (step · .github/workflows/lint.yml:206)
            └─ npx tsc --noEmit (step · .github/workflows/lint.yml:216)
```

## Changes

- Commit `sdks/node/package-lock.json`, un-ignoring it the way `apps/infra/package-lock.json` already is.
- `test.yml` and `lint.yml` install with `npm ci` and cache through `setup-node`.
- `lint.yml`'s `node` paths-filter gains `package-lock.json`, so a lock-only change still reaches the job that installs it.
- `sdks/node/CLAUDE.md` records the two constraints the lockfile carries.

**The lockfile is resolved under Node 18**, the floor of the SDK's `engines.node`. That direction is load-bearing: npm picks manifests by `engines`, so a tree resolved under Node 22 pins vite 8, whose `rolldown` dependency imports `styleText` from `node:util`, and the Node 18 leg then dies at vitest startup:

```
SyntaxError: The requested module 'node:util' does not provide an export named 'styleText'
```

Resolved under Node 18, vite stays on 6.4.3 and the suite runs on 18, 20 and 22.

## How to verify

```
cd sdks/node && npm ci && npm run test
```

Ran locally against the committed lockfile on all three matrix versions — `npm ci` exits 0 and the unit suite reports 92 passed on each:

| Node | npm | `npm ci` | `npm run test` |
| --- | --- | --- | --- |
| 18.20.8 | 10.8.2 | 0 | 92 passed |
| 20.19.5 | 10.8.2 | 0 | 92 passed |
| 22.23.1 | 10.9.8 | 0 | 92 passed |

`make lint:node`, `make fmt:check:node` and `npx tsc --noEmit` also pass on that tree.

## Risks / rollout

- **`build-node.yml`'s test job cannot use `npm ci`.** It extracts platform packages into `sdks/node/npm/` before installing, and `workspaces: ["npm/*"]` makes them part of the tree, so `npm ci` reports the lock out of sync — `--workspaces=false` does not exempt them. That job keeps `npm install`. Its build, publish and upload-to-release jobs, and `e2e-cloud.yml`, install before those directories exist and could convert too; all run on Node 20 or 24, away from the arborist crash, so they are left to a change that can exercise the release and e2e flows.
- **The local build flow dirties the lockfile.** `npm run artifacts` creates `npm/<target>/`, which the workspaces glob pulls in, so a later `npm install` adds those entries. Committing them would break `npm ci` on a fresh checkout. The guard is the `git checkout -- sdks/node/package-lock.json` documented in `sdks/node/CLAUDE.md`; nothing in CI enforces it.
- **Node 18 keeps warning `EBADENGINE`** for vitest 4.1.11 and ~34 other locked dev packages whose `engines` exclude it. Pre-existing — `npm install` resolves the same vitest there today — but this lockfile pins it in place.

## Summary by CodeRabbit

- **Chores**
  - Improved Node SDK dependency installation and caching for more consistent linting and test runs across supported Node.js versions.
  - Added automated checks to detect inconsistent or generated dependency lockfile entries.
  - Expanded formatting and quality checks to cover Node.js module scripts.

- **Documentation**
  - Documented the required workflow for regenerating and restoring the Node SDK dependency lockfile during local development.

### DIFF
--- a/sdks/node/CLAUDE.md
+++ b/sdks/node/CLAUDE.md
@@ -1,0 +1,16 @@
+# Node SDK agent instructions
+
+## package-lock.json
+
+Committed, and CI installs it with `npm ci`. Two constraints when it changes:
+
+- **Regenerate under Node 18**, the floor of `engines.node`. npm selects
+  manifests by `engines`, so Node 22 resolves vite 8 (which pulls rolldown,
+  importing `styleText` from `node:util`) and the Node 18 CI leg then dies at
+  vitest startup. Resolved under Node 18, vite stays on 6 and the unit suite
+  passes on 18, 20 and 22. Node 18 still warns EBADENGINE for vitest itself,
+  whose `engines` exclude it — that predates the lockfile and is unchanged.
+- **The local build flow dirties it.** `npm run artifacts` creates `npm/<target>/`,
+  which `workspaces: ["npm/*"]` pulls in, so a later `npm install` adds those
+  entries. Restore the file (`git checkout -- sdks/node/package-lock.json`)
+  before committing; only dependency changes belong in the diff.
--- a/sdks/node/CLAUDE.md
+++ b/sdks/node/CLAUDE.md
@@ -2,15 +2,15 @@
 
 ## package-lock.json
 
-Committed, and CI installs it with `npm ci`. Two constraints when it changes:
+Committed, and CI installs it with `npm ci`. `make lint:node` — which the
+pre-commit hook and the lint workflow both reach — checks it against
+`package.json` and rejects generated `npm/*` workspace entries, so run
+`npm install` here whenever you change a dependency.
 
-- **Regenerate under Node 18**, the floor of `engines.node`. npm selects
-  manifests by `engines`, so Node 22 resolves vite 8 (which pulls rolldown,
-  importing `styleText` from `node:util`) and the Node 18 CI leg then dies at
-  vitest startup. Resolved under Node 18, vite stays on 6 and the unit suite
-  passes on 18, 20 and 22. Node 18 still warns EBADENGINE for vitest itself,
-  whose `engines` exclude it — that predates the lockfile and is unchanged.
-- **The local build flow dirties it.** `npm run artifacts` creates `npm/<target>/`,
-  which `workspaces: ["npm/*"]` pulls in, so a later `npm install` adds those
-  entries. Restore the file (`git checkout -- sdks/node/package-lock.json`)
-  before committing; only dependency changes belong in the diff.
+**Regenerate under Node 18**, the floor of `engines.node`; nothing enforces
+this, so it is on you. npm selects manifests by `engines`, so Node 22 resolves
+vite 8 (which pulls rolldown, importing `styleText` from `node:util`) and the
+Node 18 CI leg then dies at vitest startup. Resolved under Node 18, vite stays
+on 6 and the unit suite passes on 18, 20 and 22. Node 18 still warns
+EBADENGINE for vitest itself, whose `engines` exclude it — that predates the
+lockfile and is unchanged.
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -68,6 +68,7 @@ jobs:
               - 'sdks/python/pyproject.toml'
             node:
               - 'sdks/node/**/*.ts'
+              - 'sdks/node/**/*.mjs'
               - 'sdks/node/package.json'
               - 'sdks/node/package-lock.json'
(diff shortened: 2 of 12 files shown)


sdks/node/CLAUDE.md
```diff
@@ -0,0 +1,16 @@
+# Node SDK agent instructions
+
+## package-lock.json
```
DorianZheng, line 3: add this to pre-commit hook instead of CLAUDE.md

DorianZheng (dismissed):
left a comment, rest LGTM

sdks/node/CLAUDE.md
ltstriker: Done in ac276028 — both constraints are now checked, not documented.

Put the check in `make lint:node` rather than its own hook entry, because `.pre-commit-config.yaml` already routes `lint-fix` → `make lint:fix` → `make lint`, and `detect_changes` (make/changes.mk:19) already maps `sdks/node/**` to the `node` component. A make target also means the lint workflow enforces it, which a pre-commit-only hook would not — it already runs `make lint:node`.

One gap that made this necessary: the hook's `files:` allowlist listed `sdks/node/package.json` but not the lock, so a lock-only commit skipped `lint-fix` entirely. Added, along with `.mjs`.

What it rejects:
- lock disagreeing with package.json on name/version/dependency maps (key-order-insensitive, so a freshly generated lock is not falsely rejected)
- committed `npm/<target>` workspace entries

Kept the Node 18 resolution note in CLAUDE.md: 30 locked packages already declare `engines` excluding Node 18, vitest among them, so there is no honest mechanical test for the resolution direction.

sdks/node/CLAUDE.md
```diff
@@ -0,0 +1,16 @@
...
+
+- **Regenerate under Node 18**, the floor of `engines.node`. npm selects
+  manifests by `engines`, so Node 22 resolves vite 8 (which pulls rolldown,
+  importing `styleText` from `node:util`) and the Node 18 CI leg then dies at
+  vitest startup. Resolved under Node 18, vite stays on 6 and the unit suite
+  passes on 18, 20 and 22. Node 18 still warns EBADENGINE for vitest itself,
+  whose `engines` exclude it — that predates the lockfile and is unchanged.
+- **The local build flow dirties it.** `npm run artifacts` creates `npm/<target>/`,
+  which `workspaces: ["npm/*"]` pulls in, so a later `npm install` adds those
+  entries. Restore the file (`git checkout -- sdks/node/package-lock.json`)
```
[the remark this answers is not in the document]
ltstriker: Valid — the guidance contradicted itself. It said to restore the whole file, then said dependency changes belong in the diff; `git checkout --` would have discarded exactly those.

Fixed in ac276028 by removing the advice rather than qualifying it. `make lint:node` now names the offending `npm/<target>` entries instead, so nothing suggests a blanket restore. Regenerating the lock is the remedy in both cases, which keeps real dependency updates.

DorianZheng (approved):

PR #1453, opened 2026-09-06
feat(infra): deploy through mstage, mbuild and mdeploy, on AWS or GCP

DefeatMan:
## Summary

Two files totalling 900 lines — `deployment/sst.ts` and `stack/deploy.ts` — become three tools with one config file each, twelve stack modules that name no cloud, and two engines that apply the same twelve. Built beside the incumbent, which still runs unchanged.

## Call graph

```text
Before
  sst.ts                  (driver · apps/infra/deployment/sst.ts:1)              — 583 lines: signals, .env, SSM, store, preflights, spawn
    ├─ readStoredStageConfig (deployment/stage-config.ts:215)                    — `sst secret list`, manifest + digest
    ├─ hydrateStageConfig    (deployment/stage-config.ts:167)                    — into this process's own env
    └─ spawn sst                                                                 — AWS only; no other engine reachable
         ├─ configureApp   (stack/app.ts:6)                                      — home: 'aws', hard-coded
         └─ deployStack    (stack/deploy.ts:17)                                  — 325 lines, every module inline, no seam

After
  run                     (CLI · apps/infra/mdeploy/src/run.ts:47)               — parse, mstage login, protect/confirm guards
    ├─ resolveScope       (Scope · apps/infra/mstage/src/aws/precedence.ts:122)  — folds stages.<n>.home over the repo default
    ├─ resolveDeployTarget (DeployTarget · mdeploy/src/deploy.ts:104)            — switch(scope.home) → store + identity + engine
    │    ├─ awsTarget     (mdeploy/src/sst.ts:59)                                — spawn sst → mdeploy/sst.config.ts
    │    └─ gcpTarget     (mdeploy/src/pulumi.ts:237)                            — inline Pulumi → mdeploy/pulumi/program.ts
    ├─ fetchStageEnvironment (mdeploy/src/env.ts:234)                            — one store read, narrowed per declared group
    └─ readStackEnvironment  (mdeploy/src/stack-env.ts:227)                      — one assembly, read by both engines
         └─ deployStack   (mdeploy/stack/index.ts:176)                           — 12 modules; the order lives only here
              ├─ stack/*.ts                                                      — requests and handles, no cloud named
              ├─ stack/providers/aws/*.ts                                        — sst.aws.* components
              └─ stack/providers/gcp/*.ts                                        — gcp.* / random.* / cloudflare.*
```

A runner reaching that control plane, which is the second path this PR now
changes:

```text
Before
  boxlite-runner        (systemd unit · the host)                            — BOXLITE_API_URL, frozen at first boot (stack/runner-boot.ts:117)
    └─ api.<domain>     (cloudflare.Record · gcp/api.ts:472)                 — the only record there is, a public anycast address
         └─ Cloud NAT   (gcp.compute.RouterNat · gcp/network.ts:137)         — a VM with no address of its own, out to the internet
              └─ ApiForwardingRule (gcp/api.ts:444)                          — EXTERNAL_MANAGED
                   └─ Api (gcp.cloudrunv2.Service · gcp/api.ts:241)          — ingress: INTERNAL_LOAD_BALANCER

After
  boxlite-runner        (systemd unit · the host)                            — the same value, unchanged
    └─ api.<domain>     (gcp.dns.ManagedZone · gcp/api.ts:627)               — private, this VPC alone; the public record is untouched
         └─ ApiInternalRecord (gcp.dns.RecordSet · gcp/api.ts:635)           — A, an address out of the workload subnet
              └─ ApiInternalForwardingRule (gcp/api.ts:598)                  — INTERNAL_MANAGED, never leaves the VPC
                   └─ ApiInternalHttpsProxy (gcp/api.ts:574)                 — a regional Certificate Manager certificate
                        └─ Api (gcp.cloudrunv2.Service · gcp/api.ts:241)     — the same service, the same ingress
  …
  RunnerDenyPublicApiFirewall (gcp/api.ts:667)                               — EGRESS deny of that one public address on 443
  ManagedProxySubnetwork      (gcp/network.ts:108)                           — REGIONAL_MANAGED_PROXY, where the Envoys run
```

## Changes

- **`mdeploy`** — twelve cloud-free module contracts, an AWS provider bundle and a GCP one, and the two engines that apply them. The AWS providers keep the incumbent's logical resource names, so the new path adopts existing state rather than building a second set beside it.
- **`home` is per stage, not per repository** — the one deliberate departure from the design this borrows. The AWS stages stay in service while a GCP stage comes up beside them; that one field picks the store backend, the identity, the provider bundle, the engine and the registry kind.
- **Asymmetries written down rather than smoothed over** — the box proxy cannot be a Cloud Run service (it reads the SNI name itself, so it needs a layer-4 path Cloud Run cannot back, and runs on a managed instance group instead); a runner needs a nesting-capable machine family, `minCpuPlatform` and `enableNestedVirtualization` on GCP, plus KVM group permissions in the guest — the three things `scripts/deploy/gcp/create-instance.sh` and `setup-kvm.sh` have been doing by hand.
- **`mdeploy` is inside the typecheck** — `globals.d.ts` declares what both engines inject, so `tsc` runs without `sst install`. Resource-argument spelling is still unchecked, and the file says so.
- **A GCP runner reaches the control plane without leaving the network** — a regional internal balancer in front of the same Cloud Run service, and a private zone answering for `api.<domain>` inside the VPC alone. The hostname does not change, which is the whole reason it is built this way: `BOXLITE_API_URL` is frozen in a host's systemd unit at first boot, so a name this stack does not own could never be told to a host that already exists. An egress rule denies the runner the public address so the internal route is a guarantee rather than a default — that one address, not the internet, since image pulls come over the same NAT.
- **The image scan gate no longer sits inside a retry loop** — it exits 78, which the workflow recognises instead of spending two more attempts re-reading the same findings.

## How to verify

```
cd apps/infra && npm test          # 1211 assertions across the five suites
make test:apps:infra               # the same gate CI runs
npm run mdeploy -- --plan          # the module graph, derived, no cloud needed
```

## Risks / rollout

**GCP `dev` is applied through this path; AWS is not.** Nothing here is wired into an automatic trigger — both new workflows are `workflow_dispatch` only, and the incumbents still own every AWS deploy.

Known blockers before this can deploy AWS `dev`:

| | |
|---|---|
| `env.selectGroup.deploy` demands all 55 keys, including per-deploy values (`BOXLITE_IMAGE_TAG`, `BOXLITE_RUNNER_BINARY_*`) that cannot live in a store | `mstage.config.json` |
| `vars.AWS_DEPLOY_ROLE_ARN` / `AWS_ECR_PUSH_ROLE_ARN` do not exist in this repository; `deploy-infra.yml` composes the ARN from `vars.AWS_ACCOUNT_ID` | `mdeploy.yml`, `mbuild.yml` |
| the digest gate rejects existing stages — same formula, different key set | `mstage/src/env/digest.ts` vs `deployment/stage-config.ts` |
| `prod` has never been bootstrapped; `gcp-dev` has no project, identity pool or billing | — |

The GCP providers are driven by fake-provider wiring tests and by applies against `dev` in `us-east5`, which is where every `fix(infra)` commit on this branch came from. The internal path above was verified from inside the VPC: the private zone answers `10.20.x.x`, and HTTPS to `api.<domain>` returns 200 against the regional certificate without leaving the network.

## Summary by CodeRabbit

- **New Features**
  - Added cloud deployment tooling supporting AWS and GCP stages, including environment management, build/publish/promote workflows, deployment previews, and protected-stage controls.
  - Added manually triggered artifact publishing and infrastructure deployment workflows with validation, retries, scan gates, and summaries.
  - Added runner artifact building, upgrades, registration, checksum verification, and rollback safeguards.
  - Added AWS and GCP bootstrap reconciliation for deployment roles, permissions, registries, buckets, identity federation, and required services.
  - Added Auth0 template-only email configuration and file-based proxy API key support.

- **Documentation**
  - Expanded deployment, bootstrap, environment, and migration guidance.


silverzki:
Merged this branch locally to verify the GCS volume path end to end against #1468, and hit two things worth passing back.

### 1. `make lint` fails on this branch, and no CI job runs that lint

```
apps/infra/mdeploy/stack/providers/aws/clickhouse.ts
  25:1  error  Projects cannot be imported by a relative or absolute path,
               and must begin with a npm scope   @nx/enforce-module-boundaries
```

`mdeploy/package.json` makes `mdeploy` its own nx project, so

```ts
import { CLICKHOUSE_IMAGE, ... } from '../../../../scripts/clickhouse-host.js'
```

escapes into `apps/infra`. `apps/infra/stack/clickhouse.ts:119` reaches the same module, but as `await import(...)`, which the rule does not see.

Checks are green because nothing in CI runs it:

- `lint:apps` appears only at `make/quality.mk:123` and in `make/help.mk` — no workflow invokes it.
- `lint.yml`'s `node` filter is `sdks/node/**`, so "Node.js lint and format checks" skips for `apps/**` changes.
- The `infra` job runs `make test:apps:infra` and `make test:apps:infra-config` — tests and type-check, not eslint.

So `make lint:apps` currently runs only in the local pre-commit hook. That gap is probably worth its own issue regardless of this PR.

Moving the script under `mdeploy`, or exporting it from the `apps/infra` package, both fix it properly. I only suppressed it locally to get the merge committed; that patch is not going anywhere near a PR.

### 2. `VOLUME_STORAGE_BACKEND` is never set, so the gcsfuse install cannot be reached

`gcp/runners.ts:135-139` installs gcsfuse and `gcp/storage.ts:78-80` scopes the volume-bucket IAM condition exactly the way the runner needs — but grepping `apps/infra` on this branch, `VOLUME_STORAGE_BACKEND` has zero hits. `gcp/runners.ts`'s `unitEnvironment` carries `CLOUDSDK_CORE_PROJECT` and the OTEL settings; `api-environment.ts` carries `S3_REGION` and no `GCS_LOCATION`.

That variable is what #1468 adds to select the backend (runner side `default:"s3"`, so unset means s3). Without it a GCP runner has gcsfuse installed and still execs `mount-s3`, which is not on those hosts.

Three values close it, all in files this PR owns:

- `gcp/runners.ts` `unitEnvironment`: `VOLUME_STORAGE_BACKEND: 'gcs'`
- `api-environment.ts`: `VOLUME_STORAGE_BACKEND` and `GCS_LOCATION`

Happy to send that as a PR against this branch if you'd rather not carry it — it is small, and #1468 is useless without it.

One more datum since you scoped the IAM condition yourself: `storage.ts:94`'s `roles/storage.legacyBucketReader` matches what I measured separately. `roles/storage.objectUser` alone is not enough, because every gcsfuse mount issues `GetStorageLayout` and that role lacks `storage.buckets.get`.

.gitignore
```diff
@@ -27,7 +27,12 @@ __pycache__/
 .eggs/
 .venv/
 venv/
-env/
+# Anchored to the repository root. Unanchored, this matched every directory
+# named `env` at any depth — including `apps/infra/mstage/src/env/`, whose seven
+# TypeScript sources were silently untracked while everything importing them was
+# committed. A Python virtualenv lives at the root or beside its own project, so
+# the anchor costs nothing.
+/env/
```
DorianZheng, line 35: this is error prone.

apps/proxy/cmd/proxy/main.go
```diff
@@ -51,6 +51,10 @@ func run() int {
 			ServiceName:    "boxlite-proxy",
 			ServiceVersion: internal.Version,
 			Environment:    cfg.Environment,
+			// The endpoint doubles as the audience: Cloud Run validates a
```
DorianZheng: does proxy need this?

apps/runner/cmd/runner/main.go
```diff
@@ -64,6 +64,10 @@ func run() int {
 			ServiceName:    "boxlite-runner",
 			ServiceVersion: internal.Version,
 			Environment:    cfg.Environment,
+			// The endpoint doubles as the audience: Cloud Run validates a
+			// token's `aud` against the service's own address. Empty unless a
+			// GCP stage asked for it, and empty leaves the transport alone.
+			GoogleIDTokenAudience: googleIDTokenAudience(cfg),
```
DorianZheng: ditto

.github/workflows/mdeploy.yml
```diff
@@ -0,0 +1,240 @@
...
+
+      # ----------------------------------------------------------- 1/5
+      # Which cloud this stage deploys to, and where in it. One file says both,
+      # and nothing after this point knows the answer in advance.
+      - name: Resolve the stage
+        id: config
+        working-directory: apps/infra
+        run: |
+          set -Eeuo pipefail
+          if ! output="$(npm run --silent mbuild inspect -- --stage "${{ inputs.stage }}" 2>&1)"; then
```
DorianZheng, line 97: `.mstage.config.json` is gitignored and nothing writes it in CI, so `mbuild inspect` throws here.

.github/workflows/mdeploy.yml
```diff
@@ -0,0 +1,240 @@
...
+        working-directory: apps/infra
+        run: npm run --silent mbuild verify -- --tag "$BOXLITE_IMAGE_TAG" --stage "${{ inputs.stage }}"
+
+      # ----------------------------------------------------------- 5/5
+      - name: Confirm a protected stage
+        working-directory: apps/infra
+        run: |
+          set -Eeuo pipefail
+          protected=$(node -e '
+            const config = require("./mstage.config.json")
```
DorianZheng, line 166: `mstage.config.json` is gone since 1c0a7e1.

.github/workflows/mbuild.yml
```diff
@@ -0,0 +1,248 @@
...
+      # --------------------------------------------------------------- 1/5
+      # Which cloud this stage publishes to. Nothing after this point knows the
+      # answer in advance, which is the point: adding a stage on another cloud
+      # is an edit to mbuild.config.json, not to this file.
+      - name: Read mbuild.config.json
+        id: config
+        working-directory: apps/infra
+        run: |
+          set -Eeuo pipefail
+          if ! output="$(npm run --silent mbuild inspect -- --stage "$TARGET_STAGE" 2>&1)"; then
```
DorianZheng, line 127: ditto

.github/workflows/mbuild.yml
```diff
@@ -0,0 +1,248 @@
...
+            echo "::error title=promote::promote needs both from and to"
+            exit 1
+          fi
+          for attempt in 1 2 3; do
+            if npm run --silent mbuild promote -- --tag "$IMAGE_TAG" --from "$FROM_STAGE" --to "$TARGET_STAGE"; then
+              exit 0
+            fi
+            echo "Promotion failed (attempt $attempt/3); retrying in $((attempt * 15))s"
+            sleep $((attempt * 15))
+          done
```
DorianZheng, line 228: promote can exit 78 too, why retry?

apps/infra/DEPLOY.md
```diff
@@ -0,0 +1,303 @@
...
+    src/deploy.ts          which engine — resolved once, from one field
+    src/plan.ts            what each module builds, and what it needs first
+    src/stack-env.ts       what one deploy reads out of the environment, for both engines
+    src/api-environment.ts what the control plane container reads
+    stack/                 what each module needs, described without a cloud
+    stack/providers/aws/   how AWS answers it
+    stack/providers/gcp/   how GCP answers it
+  mstage.config.json       which stages exist, where they live, and what the store may hand out
+  mbuild.config.json       what to build, and which repository receives it
+  mdeploy.config.json      what shape to deploy into
```
DorianZheng, line 34: stale, it's `mstage.env.json` + `.mstage.config.json` now.

apps/infra/DEPLOY.md
```diff
@@ -0,0 +1,303 @@
...
+stage's own artifacts bucket — S3 on AWS, Cloud Storage on GCP — under the one
+key `runner/<commit>/`, which is also the address the deploy resolves and the
+only prefix the hosts are let read. `npm run bootstrap -- --stage <stage>`
+creates the bucket; a stage without one installs a published release.
+
+The channel differs per cloud and needs one prerequisite each:
+
+| | AWS | GCP |
+|---|---|---|
+| transport | `ssm send-command`, polled to a terminal status | `gcloud compute ssh --tunnel-through-iap` |
```
DorianZheng, line 146: stale, gcp goes through os policy now.

apps/infra/DEPLOY.md
```diff
@@ -0,0 +1,303 @@
...
+| CLI on the deployer | `aws` | `gcloud` |
+
+Neither opens a way in for a person: the GCP rule admits Google's IAP forwarding
+range alone, and reaching that tunnel is an IAM question `bootstrap/gcp.ts`
+answers for the deployer and nobody else.
+
+## Two clouds, one stage at a time
+
+`mstage.config.json` declares `home` for the repository, and per stage where one
+differs:
```
DorianZheng, line 158: stale, no repo default anymore. `home` is required per stage.

apps/infra/mdeploy/src/upgrade-runners.ts
```diff
@@ -0,0 +1,554 @@
...
+
+/**
+ * IAP: one tunnelled ssh, whose remote exit status is the verdict.
+ *
+ * `--quiet` so gcloud mints a key rather than asking, and no host key is
+ * pinned — the tunnel is the authenticated channel and there is no long-lived
+ * host identity to trust: a replaced host would present a new one and the roll
+ * would stop on a prompt nobody can answer.
+ */
+const upgradeOverIap = ({
```
DorianZheng, line 396: dead code? nothing routes gcp here anymore. header above is stale too.

apps/runner/cmd/runner/main.go
```diff
@@ -308,3 +308,14 @@ func migrationWorkDir(cfg *config.Config, logger *slog.Logger) string {
 
 	return filepath.Join(home, "migrate")
 }
+
+/*
```
DorianZheng, line 312: remove redundant comments

apps/proxy/cmd/proxy/main.go
```diff
@@ -120,3 +120,14 @@ func initLogger(
 		telemetry.ShutdownLogger(newLogger, provider)
 	}, nil
 }
+
+/*
```
DorianZheng, line 124: remove redundant comments

apps/proxy/cmd/proxy/config/config.go
```diff
@@ -148,6 +156,33 @@ func GetConfig() (*Config, error) {
 	return config, nil
 }
 
+// loadProxyAPIKey keeps the two supported delivery channels mutually
+// exclusive. GKE mounts Secret Manager into a file; local and AWS deployments
+// continue to supply the same value directly. A terminal line ending is
+// ignored because secrets are commonly created from a line-oriented CLI, while
+// all other bytes (including spaces) remain part of the key.
+func loadProxyAPIKey(value, path string) (string, error) {
```
DorianZheng, line 164: seems we can resolve key or file outside of proxy and pass as key anyways

apps/libs/common-go/pkg/telemetry/common.go
```diff
@@ -21,6 +21,15 @@ type Config struct {
 	Environment    string
 	ExtraLabels    map[string]string
 	TLSConfig      *tls.Config
+	/*
```
DorianZheng, line 24: remove redundant comments

.github/workflows/mbuild.yml
DefeatMan: accepted, only retry network timeout error

DorianZheng (dismissed):

PR #1454, opened 2026-09-06
fix(jailer): skip symlinked CA paths

### LINKED ISSUE #1421: jailer: bwrap >= 0.12 fails boot when a system CA path is a symlink (Can't mount on symlink destination)
### Summary

`build_path_access()` grants host CA paths read-only to the shim sandbox with only an `exists()` check. On hosts where `/etc/ssl/certs/ca-certificates.crt` is a **symlink** (e.g. extracted-style cert layouts where it points to `../../ca-certificates/extracted/tls-ca-bundle.pem`), bubblewrap **0.12+** refuses the bind with:

```
bwrap: Can't mount on symlink destination /etc/ssl/certs/ca-certificates.crt
```

…which fails every sandboxed boot.

### Why v0.12 specifically

boxlite prefers the system `bwrap` (`Command::new("bwrap")` in `jailer/bwrap.rs`). bubblewrap added the symlink-destination refusal after 0.11 (the vendored copy in `bubblewrap-sys/vendor` has no such check), so this only fires on hosts with a newer system bwrap.

### Repro

* Host: Linux, bwrap 0.12.0, `/etc/ssl/certs/ca-certificates.crt` → symlink
* Any sandboxed box start (`boxlite run …`) with a bwrap jailer

### Fix

Remove these two redundant entries from `system_ca_paths()`:

* `/etc/ssl/certs/ca-certificates.crt`
* `/etc/pki/tls/certs/ca-bundle.crt`

Their parent directories are already mounted read-only. Keep those directory mounts, including symlinked source directories, and the existing entries that expose the certificate targets. This avoids a second bind over symlinks brought into the sandbox by the directory mounts.

These host certificates are used by the shim's network backend for upstream TLS verification.

### Implementation and validation

boxlite-ai/boxlite#1461 is open for review (commit `e9e8f256`).

* With unchanged production code, the real-bwrap 0.12.0 regression fails on both symlinked bundle destinations.
* With the fix, regular files, relative links, absolute links, and symlinked certificate directories all pass in the no-default-features and REST configurations.
* Rust formatting passes. The existing jailer suite reports 115 passed and 2 failed in each configuration; the two seccomp tests require mode 0, while the test environment already has mode 2.
* VM end-to-end tests were not run because `/dev/kvm` is unavailable.

atirna:
## Summary

Skip symlinked system CA paths when building jailer access rules so bubblewrap does not receive a bind whose destination is a symlink.

## Call graph

Before
  Jailer::context (Jailer · src/boxlite/src/jailer/mod.rs:568)
    └─ build_path_access (jailer · src/boxlite/src/jailer/mod.rs:213)
         └─ BwrapSandbox::apply (BwrapSandbox · src/boxlite/src/jailer/sandbox/bwrap.rs:120)  ← BUG: emits a direct `--ro-bind` for a symlinked CA destination

After
  Jailer::context (Jailer · src/boxlite/src/jailer/mod.rs:568)
    └─ build_path_access (jailer · src/boxlite/src/jailer/mod.rs:213)
         └─ system_ca_path_access (jailer · src/boxlite/src/jailer/mod.rs:359), keeps existing regular CA paths and excludes symlinks before bwrap receives them

Fixes #1421

## Changes

- Keep regular system CA paths read-only and omit symlinked CA entries.
- Add a regression fixture covering a regular CA file beside a symlinked CA file.

## How to verify

- `make lint && make fmt:check && make test`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved system certificate handling by granting read-only access only to existing, non-symlinked certificate paths.
  * Symlinked certificate files are now excluded from path access grants.

### DIFF
--- a/src/boxlite/src/jailer/mod.rs
+++ b/src/boxlite/src/jailer/mod.rs
@@ -321,14 +321,7 @@ fn build_path_access(layout: &BoxFilesystemLayout, volumes: &[VolumeSpec]) -> Ve
     // The in-shim network backend may validate upstream TLS certificates
     // (for example secret-substitution MITM forwarding). Keep host trust
     // stores readable inside the sandbox without granting broader /etc access.
-    for path in system_ca_paths() {
-        if path.exists() {
-            paths.push(PathAccess {
-                path,
-                writable: false,
-            });
-        }
-    }
+    paths.extend(system_ca_path_access(system_ca_paths()));
 
     // User volumes. Directories are shared directly, so grant the VMM access.
     // Single files are staged under shared_dir (granted above), so they need no
@@ -361,6 +354,17 @@ fn system_ca_paths() -> [PathBuf; 7] {
         PathBuf::from("/etc/pki/tls/certs/ca-bundle.crt"),
         PathBuf::from("/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"),
     ]
+}
+
+fn system_ca_path_access(paths: impl IntoIterator<Item = PathBuf>) -> Vec<PathAccess> {
+    paths
+        .into_iter()
+        .filter(|path| path.exists() && !path.is_symlink())
+        .map(|path| PathAccess {
+            path,
+            writable: false,
+        })
+        .collect()
 }
 
 /// Jailer provides process isolation for boxlite-shim.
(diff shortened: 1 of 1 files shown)


ltstriker:
this pr may cause side effect with refuse all src symlink.

fixed with https://github.com/boxlite-ai/boxlite/pull/1461.

PR #1462, opened 2026-09-08
fix(vmm): move box teardown off Tokio worker threads

### LINKED ISSUE #1242: Box teardown blocks a Tokio worker for seconds, and the timeout guarding it cannot fire
Running at rev `8048500dd5b0c8572b34dcb61e41ad0d56b0da60`.

## What happens

`ShimHandler::stop` (`vmm/controller/shim.rs:175`) blocks the calling thread
with `std::thread::sleep` while it waits for the box to exit, and every caller
reaches it from an async context:

```
BoxImpl::stop                 (async fn · litebox/box_impl.rs:630)
└─ live.handler.lock()        (std::sync::Mutex · box_impl.rs:55)
   └─ ShimHandler::stop       (shim.rs:175)
      └─ graceful_stop        (shim.rs:86)  thread::sleep(50ms) poll, up to 2000ms

callers, both async:
  rt_impl.rs:800   tokio::time::timeout(duration, box_impl.stop()).await
  rt_impl.rs:989   handler.stop()          — force-removal path, no timeout at all
```

Three consequences, in decreasing severity:

1. **A Tokio worker is parked for the duration.** `rt_impl.rs:795` stops all
   boxes concurrently, so stopping N boxes parks N workers.
2. **The `tokio::time::timeout` at `rt_impl.rs:800` cannot interrupt it.**
   `timeout` can only cancel a future at an await point, and a synchronous
   sleep never yields — the timer fires but cannot preempt the thread. The
   timeout looks like a guard and is not one.
3. **The lock held across the wait is a `std::sync::Mutex`**
   (`box_impl.rs:55`), so any other async task contending for that handler
   blocks synchronously too.

## Why file it separately

This is pre-existing, not introduced by #1240. On `main` the same path already
sleeps up to 2 s inside `graceful_stop` (`shim.rs:116,161`,
`GRACEFUL_SHUTDOWN_TIMEOUT_MS = 2000`). #1240 adds a second bounded wait
(`REAP_GRACE = 2000`) so the worst case roughly doubles to ~4 s, which is what
surfaced the issue in review — but removing that wait is not an option: it is
the SIGTERM grace period that lets libkrun flush virtio-blk before the hard
kill, without which the VM dies mid-flush and the qcow2 can be corrupted.

## Options

- Wrap the handler call in `spawn_blocking` (or an equivalent blocking context).
  Smallest change; needs the `std::sync::Mutex` guard to not cross the await,
  and `VmmHandler` to be `Send + 'static`.
- Make `VmmHandlerTrait::stop` async and use `tokio::time::sleep`. Cleanest
  result, but it changes a trait signature and every implementor and call site.
- Shorten the grace periods. Treats the symptom and directly weakens the flush
  guarantee — not recommended.

Reported by CodeRabbit on #1240:
"Move synchronous teardown off Tokio worker threads. `BoxImpl::stop()` calls
`VmmHandler::stop()` from an async method, and force removal reaches the same
blocking handler through `RuntimeInnerImpl::remove_box()`. `ShimHandler::stop()`
can block for about four seconds, so runtime shutdown and async removal can
stall Tokio workers. Use `spawn_blocking` or an equivalent blocking context."

sparkzky:
## Summary
Move the ~2 s synchronous graceful-shutdown poll (`ShimHandler::stop` / `graceful_stop`, `thread::sleep(50ms) × ≤40`) off Tokio worker threads at every async-reachable call site, so the worker is no longer parked and the `tokio::time::timeout` guarding `stop()` can actually fire.

## Call graph

Before
```
BoxImpl::stop                       (async fn · src/boxlite/src/litebox/box_impl.rs:729)
  └─ live.handler.lock()           (std::Mutex · box_impl.rs:56)
     └─ ShimHandler::stop          (VmmHandler · src/boxlite/src/vmm/controller/shim.rs:175)
        └─ graceful_stop           (fn · shim.rs:86)  thread::sleep(50ms) × ≤40  ← BUG: parks the Tokio worker ~2s; tokio::time::timeout at rt_impl.rs:822 / serve/mod.rs:917 cannot preempt a sync sleep
LocalRuntime::remove                (async · src/boxlite/src/runtime/rt_impl.rs:1824)
  └─ RuntimeImpl::remove            (sync fn · rt_impl.rs:638)
     └─ remove_box(force=true)     (sync fn · rt_impl.rs:1011)
        └─ ShimHandler::stop        (VmmHandler · shim.rs:175)  ← BUG: same ~2s sync block, no timeout at all
CleanupGuard::drop                  (Drop · src/boxlite/src/litebox/init/types.rs:255)
  └─ handler.stop()                 (VmmHandler · shim.rs:175)  ← BUG: same ~2s sync block on the init-failure path
```

After
```
BoxImpl::stop                       (async fn · src/boxlite/src/litebox/box_impl.rs:729)
  └─ spawn_blocking(|| { live.handler.lock(); handler.stop() }).await  (tokio task · box_impl.rs:738)
     └─ ShimHandler::stop          (VmmHandler · shim.rs:175)
        └─ graceful_stop           (fn · shim.rs:86)  — thread::sleep on the blocking pool; outer timeout now fires at the await (detached task still completes SIGTERM→SIGKILL→reap)
LocalRuntime::remove                (async · src/boxlite/src/runtime/rt_impl.rs:1824)
  └─ spawn_blocking(|| this.remove(...)).await  (tokio task · rt_impl.rs:1830)
     └─ RuntimeImpl::remove → remove_box(force=true) → ShimHandler::stop
CleanupGuard::drop                  (Drop · src/boxlite/src/litebox/init/types.rs:255)
  └─ Handle::try_current().spawn_blocking(|| handler.stop())  (fire-and-forget; keepalive Drop is the abandonment safety net)
     └─ ShimHandler::stop
```

Fixes #1242

## Changes
- `LiveState::handler` (`box_impl.rs:56`) wrapped in `Arc` so the `spawn_blocking` closure holds a `Send + 'static` ref to lock, without borrowing across the await. `metrics()`/`pid()` lockers unchanged — `Arc` derefs to `Mutex`.
- `BoxImpl::stop` runs `handler.stop()` inside `spawn_blocking`; the await is the cancellation point the outer `tokio::time::timeout` (shutdown / AutoStop) can actually fire on. Dropping the `JoinHandle` on cancel detaches — the kill still runs to completion in the background.
- `LocalRuntime::remove` wraps the entire sync `RuntimeImpl::remove` in `spawn_blocking` at the async boundary. Sync callers (`BoxImpl::stop`'s `remove_box(force=false)` at `box_impl.rs:818`, sync unit tests) are untouched — they don't go through this wrapper.
- `CleanupGuard::drop` fire-and-forgets `handler.stop()` via `Handle::try_current()`; falls back to blocking in place when no runtime context. The `ShimHandler` keepalive (dropped with the handler) is the kernel-level safety net for the case where the task never starts.

## How to verify
- `BOXLITE_DEPS_STUB=1 cargo clippy -p boxlite --lib --tests --all-features -- -D warnings`
- `BOXLITE_DEPS_STUB=1 cargo test -p boxlite --lib box_impl::tests::stop_does_not_park` — new reproducer; passes with the fix, fails ("stop() parked the Tokio worker...") with Change A reverted (two-sided verified).
- KVM-equipped CI: the existing `test_stop_recovered_box_kills_orphan_process`, `remove_box` / `shutdown_sync` suites, and `cleanup_guard_drop_persists_failed_state_and_keeps_record`.

## Risks / rollout
- On `tokio::time::timeout` cancel of `box_impl.stop()`, the box's post-kill state transition (mark `Stopped`, remove PID file, DB save) is skipped for that call; the detached `spawn_blocking` still completes the kill. The box self-heals on the next `stop()` (idempotent: `graceful_stop` takes `process` → `None` the second time → attached-mode `waitpid`/`kill` on the dead pid) and via `recover_boxes` at next runtime start.
- `metrics()` vs `stop()` on multi-threaded runtimes has a residual microsecond TOCTOU race (token check vs lock acquire) that can block a worker for ~2 s. Issue option 1 does not cover this; the token check (`box_impl.rs:651`, cancelled before `stop()` locks at `705`) handles the realistic case. Leaving it avoids an untested `try_lock` branch or a trait-wide `tokio::sync::Mutex` change with larger blast radius.
- Grace periods and the `std::sync::Mutex` are unchanged; the qcow2 flush guarantee is preserved.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented box shutdown and removal operations from blocking asynchronous workers during graceful cleanup.
  - Timeouts and other concurrent asynchronous tasks now remain responsive while shutdown work is in progress.
  - Improved automatic cleanup behavior to avoid duplicate shutdown attempts and safely handle cleanup with or without an active runtime.

- **Documentation**
  - Added investigation and implementation guidance for shutdown-related worker blocking and timeout behavior.

### DIFF
--- a/src/boxlite/src/litebox/box_impl.rs
+++ b/src/boxlite/src/litebox/box_impl.rs
@@ -53,7 +53,7 @@ pub type SharedBoxImpl = Arc<BoxImpl>;
 /// Separated from BoxImpl to allow operations like `info()` without initializing LiveState.
 pub(crate) struct LiveState {
     // VM process control
-    handler: std::sync::Mutex<Box<dyn VmmHandler>>,
+    handler: Arc<std::sync::Mutex<Box<dyn VmmHandler>>>,
     guest_session: GuestSession,
 
     /// Host-side network control backend (gvproxy ServicesMux client), owned
@@ -94,7 +94,7 @@ impl LiveState {
         #[cfg(target_os = "linux")] bind_mount: Option<BindMountHandle>,
     ) -> Self {
         Self {
-            handler: std::sync::Mutex::new(handler),
+            handler: Arc::new(std::sync::Mutex::new(handler)),
             guest_session,
             network: network.map(Arc::from),
             published_ports,
@@ -725,10 +725,25 @@ impl BoxImpl {
                 tracing::warn!(box_id = %self.config.id, "Guest shutdown timed out after 10s");
             }
 
-            // Stop handler
-            if let Ok(mut handler) = live.handler.lock() {
-                handler.stop()?;
-            }
+            // Stop handler. `ShimHandler::stop` polls graceful shutdown
+            // with `std::thread::sleep` for up to `GRACEFUL_SHUTDOWN_TIMEOUT_MS`,
+            // so it must NOT run on this Tokio worker — it would park the
+            // worker and the `tokio::time::timeout` guarding `stop()`
+            // (runtime shutdown / AutoStop sweeper) cannot preempt a
+            // synchronous sleep. Run it on a blocking thread; the await
+            // here is the cancellation point the outer timeout can actually
+            // fire on, and dropping the JoinHandle on cancel detaches —
+            // the task still runs to completion, killing the VM in the
+            // background. See docs/investigations/issue-1242-teardown-blocks-worker.md.
+            let handler = Arc::clone(&live.handler);
+            tokio::task::spawn_blocking(move || {
+                let mut handler = handler
+                    .lock()
+                    .map_err(|e| BoxliteError::Internal(format!("handler lock poisoned: {}", e)))?;
+                handler.stop()
+            })
+            .await
+            .map_err(|e| BoxliteError::Internal(format!("spawn_blocking failed: {}", e)))??;
         }
         // If live_state() failed (vmm_attach said Absent — shim is gone),
         // or status wasn't Running, fall through to cleanup.
@@ -2125,6 +2140,128 @@ mod tests {
         assert!(
             !is_process_alive(pid),
             "stop() must kill the recovered shim process when self.live is None"
+        );
+
+        // ChildGuard's Drop reaps the (now-dead) child.
+        drop(child);
+    }
+
+    // Reproducer for issue #1242: `ShimHandler::stop` polls graceful
+    // shutdown with `std::thread::sleep` for up to
+    // `GRACEFUL_SHUTDOWN_TIMEOUT_MS` (2 s). Before the fix this sleep ran
+    // on the calling Tokio worker, parking it for the whole window — the
+    // `tokio::time::timeout` guarding `stop()` (runtime shutdown / REST
+    // AutoStop) could not preempt it. This test proves the graceful
+    // teardown now runs on a blocking thread: a concurrent async task on
+    // the *same* (current-thread, single-worker) runtime makes progress
+    // while `stop()` is inside the graceful window.
+    //
+    // The stand-in must ignore SIGTERM — otherwise `graceful_stop`'s
+    // `waitpid(WNOHANG)` reaps the process on the first poll and the 2 s
+    // `thread::sleep` loop is never entered, so the bug would not
+    // manifest and the test would pass both with and without the fix
+    // (`sleep 300` dies on SIGTERM, so it cannot be the stand-in). `sh -c
+    // "trap '' TERM; while true; do sleep 1; done"` ignores SIGTERM and
+    // loops forever (the `while` prevents an `exec` optimization that
+    // would replace the shell and drop the trap), forcing `graceful_stop`
+    // into the full 2 s poll loop until SIGKILL lands.
+    #[tokio::test]
+    async fn stop_does_not_park_tokio_worker_during_graceful_shutdown() {
+        let temp_dir = TempDir::new_in("/tmp").expect("create temp dir");
+        // `new_for_test` skips the host KVM preflight — this test uses a
+        // stand-in process (not a real VM), so no hypervisor is needed.
+        let runtime = RuntimeImpl::new_for_test(BoxliteOptions {
+            home_dir: temp_dir.path().to_path_buf(),
+            image_registries: vec![],
+        })
+        .expect("create runtime");
+
+        let child = ChildGuard(
+            std::process::Command::new("sh")
+                .arg("-c")
+                .arg("trap '' TERM; while true; do sleep 1; done")
+                .spawn()
+                .expect("spawn SIGTERM-ignoring stand-in"),
+        );
+        let pid = child.0.id();
+
+        let id = BoxIDMint::mint();
+        let box_home = runtime.layout.boxes_dir().join(id.as_str());
+        let config = BoxConfig {
+            id: id.clone(),
+            name: None,
+            created_at: Utc::now(),
+            container: ContainerRuntimeConfig {
+                id: ContainerID::new(),
+            },
+            options: BoxOptions {
+                rootfs: RootfsSpec::Image("alpine:latest".into()),
+                detach: false,
+                auto_delete: Some(0),
+                ..Default::default()
+            },
+            engine_kind: VmmKind::Libkrun,
+            box_home: box_home.clone(),
+        };
+
+        let mut state = BoxState::new();
+        state.status = BoxStatus::Running;
+        state.pid = Some(pid);
+        let lock_id = runtime.lock_manager.allocate().expect("allocate lock");
+        state.set_lock_id(lock_id);
+
+        std::fs::create_dir_all(&box_home).expect("create box dir");
+        let st = crate::util::process_start_time(pid).expect("OS reports start_time");
+        let layout = runtime
+            .layout
+            .box_layout(config.id.as_str(), false)
+            .expect("box_layout is infallible");
+        let pid_file = layout.pid_file_path();
+        std::fs::write(&pid_file, format!("{pid}\n{st}\n")).expect("write pid file");
+
+        runtime
+            .box_manager
+            .add_box(&config, &state)
+            .expect("add box to manager");
+
+        let litebox = runtime
+            .get(config.id.as_str())
+            .await
+            .expect("get box")
+            .expect("box exists");
+
+        // A 50 ms beacon on the *same* current-thread runtime. With the
+        // bug, `stop()`'s sync `thread::sleep` parks the worker for ~2 s,
+        // the beacon is never polled, and `stop()` wins the select at
+        // ~2 s. With the fix, `stop()` yields at the `spawn_blocking`
+        // join, the worker polls the beacon, and the beacon wins at
+        // ~50 ms.
+        let beacon = tokio::time::sleep(Duration::from_millis(50));
+        tokio::pin!(beacon);
+        let stop_fut = litebox.stop();
+        tokio::pin!(stop_fut);
+
+        let beacon_won = tokio::select! {
+            _ = &mut beacon => true,
+            _ = &mut stop_fut => false,
+        };
+
+        assert!(
+            beacon_won,
+            "stop() parked the Tokio worker during graceful shutdown; the graceful \
+             teardown must run on a blocking thread, not the worker"
+        );
+
+        // Let the (still running) graceful teardown complete for cleanup —
+        // the SIGKILL lands at the 2 s graceful timeout.
+        stop_fut.await.expect("stop should complete after beacon");
+
+        // Give SIGKILL a moment to land and be reaped.
+        tokio::time::sleep(Duration::from_millis(200)).await;
+        assert!(
+            !is_process_alive(pid),
+            "stop() must still kill the stand-in (via SIGKILL at the graceful timeout) \
+             even though it ran on a blocking thread"
         );
 
         // ChildGuard's Drop reaps the (now-dead) child.
--- a/src/boxlite/src/litebox/init/types.rs
+++ b/src/boxlite/src/litebox/init/types.rs
@@ -250,11 +250,38 @@ impl Drop for CleanupGuard {
 
         tracing::warn!(box_id = %self.box_id, reason = %reason, "Box initialization failed, cleaning up");
 
-        // Stop handler if started
-        if let Some(ref mut handler) = self.handler
-            && let Err(e) = handler.stop()
-        {
-            tracing::warn!("Failed to stop handler during cleanup: {}", e);
+        // Stop handler if started. `ShimHandler::stop` polls graceful
+        // shutdown with `std::thread::sleep` (≤ `GRACEFUL_SHUTDOWN_TIMEOUT_MS`),
+        // so it must NOT run on the Tokio worker this `Drop` runs on. A
+        // `Drop` cannot `.await`, so fire-and-forget on a blocking thread.
+        // This is safe here specifically: init-failure cleanup is
+        // best-effort, and the `ShimHandler`'s `keepalive` (dropped with
+        // the handler when the task is dropped, whether it ran or was
+        // abandoned) closes the watchdog pipe, delivering POLLHUP to the
+        // shim and triggering its self-shutdown — the kernel-level safety
+        // net that covers the case where the task never starts.
+        // See docs/investigations/issue-1242-teardown-blocks-worker.md.
+        if let Some(mut handler) = self.handler.take() {
+            let box_id = self.box_id.clone();
+            if let Ok(handle) = tokio::runtime::Handle::try_current() {
+                handle.spawn_blocking(move || {
+                    if let Err(e) = handler.stop() {
+                        tracing::warn!(
+                            box_id = %box_id,
+                            "Failed to stop handler during cleanup: {}", e
+                        );
+                    }
+                });
+            } else {
+                // No Tokio runtime context (e.g. a `Drop` outside any
+                // runtime) — block in place; there is no worker to park.
+                if let Err(e) = handler.stop() {
+                    tracing::warn!(
+                        box_id = %box_id,
+                        "Failed to stop handler during cleanup: {}", e
+                    );
+                }
+            }
         }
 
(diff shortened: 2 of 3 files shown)


docs/investigations/issue-1242-teardown-blocks-worker.md
```diff
@@ -0,0 +1,568 @@
+# Issue #1242 — Box teardown blocks a Tokio worker, timeout cannot fire
```
ltstriker: no need?

src/boxlite/src/litebox/box_impl.rs
```diff
@@ -725,10 +725,25 @@ impl BoxImpl {
...
+            // so it must NOT run on this Tokio worker — it would park the
+            // worker and the `tokio::time::timeout` guarding `stop()`
+            // (runtime shutdown / AutoStop sweeper) cannot preempt a
+            // synchronous sleep. Run it on a blocking thread; the await
+            // here is the cancellation point the outer timeout can actually
+            // fire on, and dropping the JoinHandle on cancel detaches —
+            // the task still runs to completion, killing the VM in the
+            // background. See docs/investigations/issue-1242-teardown-blocks-worker.md.
+            let handler = Arc::clone(&live.handler);
+            tokio::task::spawn_blocking(move || {
```
ltstriker, line 739: not likely a good idea to use too much blocking thread to sleep?

src/boxlite/src/litebox/box_impl.rs
```diff
@@ -725,10 +725,25 @@ impl BoxImpl {
...
+                let mut handler = handler
+                    .lock()
+                    .map_err(|e| BoxliteError::Internal(format!("handler lock poisoned: {}", e)))?;
+                handler.stop()
+            })
+            .await
```
DorianZheng, line 745: This `.await` is a new place where `stop()` can be cancelled. If the caller's timeout fires here (`runtime.shutdown(timeout)`, serve's AutoStop, or a client that drops `POST /stop`), the kill keeps going in the background, but everything after this line is skipped: pid file removal, the Stopped state, `invalidate_box_impl`, `on_box_stopped`, remove-on-stop. So the VM is dead but the box still says Running.

I checked with a small test (a Running box whose shim ignores SIGTERM): `timeout(500ms, stop())` returns `Elapsed`, and after the shim dies the box is still Running in memory and in the DB. `runtime.shutdown(Some(1))` ends the same way, and `shutdown_sync()` doesn't fix it because `shutdown()` already cancelled the token. Before this PR both cases end in Stopped.

It also makes a retry unsafe. The box is still Running, so the next `stop()` calls `handler.stop()` again on the same handler. Its process is already gone, so `graceful_stop` just does `kill(self.pid)` with no identity check. If that pid got reused, we kill someone else's process. I forced a pid reuse in a test and the retry did kill the other process. Rare (it needs a pid wrap), but possible on a long running serve.

Maybe spawn the kill and the state update as one task and await that, so a cancel can't split them.

src/boxlite/src/litebox/box_impl.rs
```diff
@@ -2131,6 +2146,128 @@ mod tests {
...
+        let runtime = RuntimeImpl::new_for_test(BoxliteOptions {
+            home_dir: temp_dir.path().to_path_buf(),
+            image_registries: vec![],
+        })
+        .expect("create runtime");
+
+        let child = ChildGuard(
+            std::process::Command::new("sh")
+                .arg("-c")
+                .arg("trap '' TERM; while true; do sleep 1; done")
```
DorianZheng, line 2182: This test can be flaky even with the fix. Nothing waits for `trap '' TERM` to be set up, and `stop()` sends SIGTERM only a few ms after the spawn. If the signal lands first, `sh` dies right away, `stop()` finishes in about 50ms, and it races the 50ms beacon in `select!`.

I looped it: 2 fails out of 120 runs on an idle machine, 9 out of 150 with 6 runs in parallel (the "stop() parked the Tokio worker" assert, in a run that took 0.07s). The default nextest profile retries once, so CI would mostly show it as flaky.

Simple fix: `sh -c "trap '' TERM; echo ready; exec sleep 300"` with stdout piped, and read the `ready` line before using the pid. An ignored signal stays ignored across `exec`, so the `while` loop isn't needed. That also gets rid of the LEAK nextest reports now (the leftover `sleep 1` keeps stdout open after `sh` gets SIGKILL).

Btw `#[tokio::test]` is already current_thread by default, so the runtime flavor is fine here.

src/boxlite/src/litebox/box_impl.rs
```diff
@@ -2131,6 +2146,128 @@ mod tests {
...
+            "stop() parked the Tokio worker during graceful shutdown; the graceful \
+             teardown must run on a blocking thread, not the worker"
+        );
+
+        // Let the (still running) graceful teardown complete for cleanup —
+        // the SIGKILL lands at the 2 s graceful timeout.
+        stop_fut.await.expect("stop should complete after beacon");
+
+        // Give SIGKILL a moment to land and be reaped.
+        tokio::time::sleep(Duration::from_millis(200)).await;
```
DorianZheng, line 2260: nit: this uses a fixed sleep to wait for a process we own. `child.0.wait()` would be exact. Also nothing reaps it here (the attached path in `graceful_stop` returns right after SIGKILL); the assert passes because `is_process_alive` counts zombies as dead.

src/boxlite/src/litebox/box_impl.rs
```diff
@@ -2131,6 +2146,128 @@ mod tests {
...
+            .layout
+            .box_layout(config.id.as_str(), false)
+            .expect("box_layout is infallible");
+        let pid_file = layout.pid_file_path();
+        std::fs::write(&pid_file, format!("{pid}\n{st}\n")).expect("write pid file");
+
+        runtime
+            .box_manager
+            .add_box(&config, &state)
+            .expect("add box to manager");
```
DorianZheng, line 2225: nit: this block is a copy of the setup in `test_stop_recovered_box_kills_orphan_process` above (same 38 lines). `running_box_with_standin()` below does the same job; if the stand-in command were a parameter, both tests could use it.

DorianZheng (commented):
Moving the kill off the worker makes sense, but the new `.await` in `stop()` is a new place where `stop()` can be cancelled, and a cancel there leaves a dead VM marked Running (details inline). The `CleanupGuard` change also lets the box lock go before the VM is dead.

I checked each point with small tests on this branch and on the base commit (fb41cf8): the stop / remove / cleanup cases fail here and pass before this PR. Happy to share the tests.

Not from this PR, but the same blocking pattern is still in a few places: `remove_box` at the end of `stop()` (remove-on-stop), `BoxWatcher`'s `remove_box`, and `shutdown_sync`'s 50ms sleep loop when `RuntimeImpl` is dropped inside a task (a 50ms timer fired after ~5s in a test).

src/boxlite/src/litebox/init/types.rs
```diff
@@ -250,11 +250,38 @@ impl Drop for CleanupGuard {
...
+        // best-effort, and the `ShimHandler`'s `keepalive` (dropped with
+        // the handler when the task is dropped, whether it ran or was
+        // abandoned) closes the watchdog pipe, delivering POLLHUP to the
+        // shim and triggering its self-shutdown — the kernel-level safety
+        // net that covers the case where the task never starts.
+        // See docs/investigations/issue-1242-teardown-blocks-worker.md.
+        if let Some(mut handler) = self.handler.take() {
+            let box_id = self.box_id.clone();
+            if let Ok(handle) = tokio::runtime::Handle::try_current() {
+                handle.spawn_blocking(move || {
```
DorianZheng, line 267: This changes the order of things. `Drop` now returns right away while the VM is still in its 2s SIGTERM grace, and `init_live_state` releases the box lock as soon as `build()` returns. So a quick retry of `start()` (or an exec racing start) can boot a second VM for the same box while the old one is still alive. Both would have the same `container.qcow2` open, and on Linux the old task's `reap_box` can cgroup-kill the new VM.

In a test, `drop()` returned in about 0.3ms with the record already marked Failed and the VM still running. Before this PR `drop()` took about 2s and the VM was dead when it returned.

Idea: do the stop in `BoxBuilder::build`'s Err arm instead. It's already async and runs before the lock is released, so it can take the handler out of the guard and `spawn_blocking(...).await` it there. `Drop` stays as the fallback for when the future is cancelled.

Also the doc on `CleanupGuard` (lines 168-173) still says Drop stops the VM before marking the box Failed, which isn't true anymore.

src/boxlite/src/litebox/init/types.rs
```diff
@@ -250,11 +250,38 @@ impl Drop for CleanupGuard {
...
+        // Stop handler if started. `ShimHandler::stop` polls graceful
+        // shutdown with `std::thread::sleep` (≤ `GRACEFUL_SHUTDOWN_TIMEOUT_MS`),
+        // so it must NOT run on the Tokio worker this `Drop` runs on. A
+        // `Drop` cannot `.await`, so fire-and-forget on a blocking thread.
+        // This is safe here specifically: init-failure cleanup is
+        // best-effort, and the `ShimHandler`'s `keepalive` (dropped with
+        // the handler when the task is dropped, whether it ran or was
+        // abandoned) closes the watchdog pipe, delivering POLLHUP to the
+        // shim and triggering its self-shutdown — the kernel-level safety
+        // net that covers the case where the task never starts.
```
DorianZheng, line 262: The keepalive only exists for normal spawned shims. Detached spawns (`spawn.rs:108`) and `from_pid` handlers (`shim.rs:76`) have `keepalive: None`, so if this task never runs, nothing stops the shim. And `mark_failed` has already cleared the pid, so `shutdown_sync` and `remove(force)` can't find it later.

The task not running can really happen: tokio drops `spawn_blocking` tasks that are submitted after the blocking pool starts shutting down. I dropped a multi-thread runtime while a start was in flight (like `boxlite_runtime_free` does) and the stop never ran. The shim stayed alive, and the record said Failed with no pid. Before this PR the stop ran inside `Drop`.

src/boxlite/src/litebox/init/types.rs
```diff
@@ -250,11 +250,38 @@ impl Drop for CleanupGuard {
...
+                });
+            } else {
+                // No Tokio runtime context (e.g. a `Drop` outside any
+                // runtime) — block in place; there is no worker to park.
+                if let Err(e) = handler.stop() {
+                    tracing::warn!(
+                        box_id = %box_id,
+                        "Failed to stop handler during cleanup: {}", e
+                    );
+                }
```
DorianZheng, line 283: nit: same block as in the branch above. One closure used by both branches would avoid the copy, e.g. `let mut stop = move || { if let Err(e) = handler.stop() { warn!(..) } };` and then `handle.spawn_blocking(stop)` or `stop()`.

src/boxlite/src/runtime/rt_impl.rs
```diff
@@ -1822,7 +1822,19 @@ impl super::backend::RuntimeBackend for LocalRuntime {
...
+        // dependency checks, `ShimHandler::from_pid(pid).stop()` on the
+        // force path with its ≤2 s graceful-shutdown poll, DB removal).
+        // Running it on a Tokio worker parks the worker for that whole
+        // window — see issue #1242. Move the entire sync call onto a
+        // blocking thread; the `.await` here yields, the runtime stays
+        // responsive, and the inner `GRACEFUL_SHUTDOWN_TIMEOUT_MS` still
+        // bounds the kill.
+        let this = Arc::clone(&self.0);
+        let id_or_name = id_or_name.to_string();
+        tokio::task::spawn_blocking(move || this.remove(&id_or_name, force))
```
DorianZheng, line 1835: Before this, `remove()` ran to the end in one poll, so it couldn't be cut in half. Now if the caller goes away at this await (a client disconnecting from `DELETE /boxes/{id}`; hyper drops the handler on EOF), the remove still finishes in the background but the caller never sees it. So serve's `forget_box()` right after the await is skipped and the lifecycle entry stays in the map. Small leak, but easy to miss. In a test, `timeout(500ms, remove(id, true))` returned `Elapsed` and the row was deleted anyway.

Also nothing tests this change or the `CleanupGuard` one: I reverted both and the full unit suite still passes.

src/boxlite/src/litebox/box_impl.rs
```diff
@@ -725,10 +725,25 @@ impl BoxImpl {
...
+            // here is the cancellation point the outer timeout can actually
+            // fire on, and dropping the JoinHandle on cancel detaches —
+            // the task still runs to completion, killing the VM in the
+            // background. See docs/investigations/issue-1242-teardown-blocks-worker.md.
+            let handler = Arc::clone(&live.handler);
+            tokio::task::spawn_blocking(move || {
+                let mut handler = handler
+                    .lock()
+                    .map_err(|e| BoxliteError::Internal(format!("handler lock poisoned: {}", e)))?;
+                handler.stop()
```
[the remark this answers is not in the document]
DorianZheng: Confirmed with a test: after the lock is poisoned, both `stop()` calls fail with "handler lock poisoned" and the box stays Running for good. Before this PR the poisoned lock was skipped and `stop()` finished. `unwrap_or_else(PoisonError::into_inner)` would keep the old behavior.

docs/investigations/issue-1242-teardown-blocks-worker.md
DorianZheng: +1. Also some of it isn't true for the final code:
- "shutdown_sync mops up any survivors at atexit" (line 354, and the snippet at 292): `shutdown()` cancels the token first, so `shutdown_sync` returns right away (`rt_impl.rs:871`).
- the keepalive as a safety net "independent of whether the task started" (256): not for detached or `from_pid` handlers, they have no keepalive.
- "no data dependency on the kill" (268): the box lock release depends on it.
- "N boxes park N workers" (40): `join_all` polls all the stops in one task, so they ran one after another on a single worker.
- the verification part mentions `try_lock` changes that aren't in the PR.

It's also tied to a local worktree (`~/boxlite-1242`) and the line numbers have already moved. I'd drop it, or keep just the final design in the PR description.

PR #1465, opened 2026-09-10
test(e2e): pin what survives a stop/start — declared service vs exec-started

G4614:
## Why

`auto_resume` returns a running **box**, not a running **service**. Stopping a box kills every process in it; `start` reboots the VM, rebuilds the container and re-runs its init, so the only command replayed on the user's behalf is `entrypoint`/`cmd`. A service an agent started with `exec` + `nohup` — the natural thing to do mid-session — is gone, and no amount of auto-resume brings it back.

That asymmetry is the contract users have to design around, and it is about to be written into the docs as the supported way to keep a service across a resume. Nothing in the suite pins it today.

## What

Two cases in `apps/e2e/cases/test_service_survives_resume.py`, same image and same service (`python3 -m http.server 3000`), differing only in how it is launched:

- declared via `entrypoint`/`cmd` → **must** be serving again after stop→start, with nothing restarted by hand
- started with `exec` + `nohup` → **must not** be

The declared case is checked **twice** after the resume: from inside the box, which separates "did the process come back" from proxy wiring, and **over the box's own network tunnel**, which is how a user actually reaches it. Only the second answers the question users ask — a service that is listening but unreachable through its exposed port is not back as far as they are concerned. Inbound is enabled on that box for the same reason.

The negative case is as load-bearing as the positive one: if exec-started processes ever do survive, the guidance built on this has to change, and the failure message says so rather than just reporting a bad assert.

Two details that are deliberate, not incidental:

- **Readiness is polled, not slept on.** A cold boot re-runs init and how long the service needs to bind is the box's business, not a constant we get to pick. The negative case polls for a bounded window too, so a pass means "still absent", not "we did not wait long enough".
- **A failing exec counts as "not serving".** Right after a start a box can briefly refuse execs, which is the same answer as an empty port for this purpose; treating it as an error would make the test flaky for a reason unrelated to what it measures.

`stop`→`start` goes through a small retry rather than a fixed sleep, since a `start` issued while the stop is still settling is refused.

## Test plan

Verified end to end against **dev** over raw REST before writing the case (the pytest suite needs the bootstrapped Linux stack, which I do not have locally — CI or a reviewer on a test host should run the file itself):

**1. In-box, both cases** — two boxes, `ghcr.io/boxlite-ai/boxlite-agent-base:v0.1.0`, same service:

```
before stop    entrypoint serving=YES     exec serving=YES
after resume   entrypoint serving=YES     exec serving=NO (curl exit 7, connection refused)
               ^ nothing restarted by hand; exec box probed 3x, 6s apart
```

**2. Through the exposed port** — public box, service declared as `entrypoint`/`cmd`, fetched over the preview URL (the same path the tunnel probe takes):

```
before stop    200 / 187 bytes
while stopped  502
after resume   200 / 187 bytes   <- no manual restart, URL not re-issued
```

The URL is unchanged across the cycle, since it is a pure function of the box id.

API surface used by the case was checked against the bindings rather than assumed: `BoxOptions(entrypoint=…, cmd=…)` (`sdks/python/src/options.rs:615`) and `box.start()` / `box.stop()` / `box.exec()` (`sdks/python/src/box_handle.rs:58,136,146`).

**Not run here:** `apps/e2e` itself. The scenario is verified; the harness wiring of this file is not.

## Summary by CodeRabbit

- **Tests**
  - Added end-to-end coverage for service behavior when a box is stopped and restarted.
  - Verified that services configured through the container entrypoint and command are restored and reachable afterward.
  - Verified that services launched directly through execution commands remain stopped after restart.
  - Added readiness and network-response checks to confirm restored services are usable.
  - Improved test reliability with bounded execution times and stronger readiness validation.

### DIFF
--- a/apps/e2e/cases/test_service_survives_resume.py
+++ b/apps/e2e/cases/test_service_survives_resume.py
@@ -1,0 +1,125 @@
+"""What survives a stop→start: a declared service vs. an exec-started one.
+
+A stopped box keeps its disk but not its processes — `start` reboots the VM,
+rebuilds the container and re-runs its init. So the only thing that brings a
+user's service back on its own is the one command the platform replays:
+`entrypoint`/`cmd`. A service launched with `exec` (the natural thing for an
+agent to do, `nohup … &`) is gone, and no amount of auto-resume brings it back.
+
+That asymmetry is a contract users have to design around — auto_resume returns
+a running *box*, not a running *service* — so it is pinned here in both
+directions. The negative case is as load-bearing as the positive one: if
+exec-started processes ever do survive, the guidance built on this changes.
+"""
+from __future__ import annotations
+
+import asyncio
+
+import boxlite
+import pytest
+
+from conftest import drain
+
+PORT = 3000
+SERVICE_ARGV = ["-m", "http.server", str(PORT), "--bind", "0.0.0.0"]
+
+
+async def _run(box, script: str, timeout: int = 30) -> int:
+    ex = await box.exec("sh", ["-c", script])
+    await drain(ex)
+    rc = await asyncio.wait_for(ex.wait(), timeout=timeout)
+    return rc.exit_code
+
+
+async def _is_serving(box) -> bool:
+    """True when something answers on PORT from inside the box.
+
+    A failing exec counts as "not serving": right after a start the box can
+    briefly refuse execs, and that is the same answer as an empty port for
+    our purposes — the caller polls.
+    """
+    try:
+        code = await _run(box, f"curl -sf --max-time 5 http://127.0.0.1:{PORT}/ >/dev/null")
+    except Exception:
+        return False
+    return code == 0
+
+
+async def _restart(box) -> None:
+    """stop→start, tolerating the window where the state change is still in
+    flight — `start` right after `stop` can be refused, and the retry is the
+    point of the test, not something to paper over with a fixed sleep."""
+    await box.stop()
+    deadline = asyncio.get_running_loop().time() + 60
+    last: Exception | None = None
+    while asyncio.get_running_loop().time() < deadline:
+        try:
+            await box.start()
+            return
+        except Exception as exc:  # state change in progress
+            last = exc
+            await asyncio.sleep(2)
+    raise AssertionError(f"could not start the box again: {last}")
+
+
+async def _wait_serving(box, timeout: float = 45.0) -> bool:
+    """Poll rather than sleep a fixed amount: a cold boot re-runs init, and how
+    long the service needs is the box's business, not a constant we can pick."""
+    deadline = asyncio.get_running_loop().time() + timeout
+    while asyncio.get_running_loop().time() < deadline:
+        if await _is_serving(box):
+            return True
+        await asyncio.sleep(2)
+    return False
+
+
+@pytest.mark.asyncio
+async def test_entrypoint_declared_service_survives_stop_start(rt, image):
+    """A service declared as entrypoint/cmd is serving again after a resume,
+    with nothing restarted by hand."""
+    box = await rt.create(
+        boxlite.BoxOptions(
+            image=image,
+            auto_remove=False,
+            entrypoint=["python3"],
+            cmd=SERVICE_ARGV,
+        )
+    )
+    try:
+        assert await _wait_serving(box), "declared service never came up before the stop"
+
+        await _restart(box)
+
+        assert await _wait_serving(box), (
+            "a service declared via entrypoint/cmd did not come back after "
+            "stop→start; container init is expected to replay it"
+        )
+    finally:
+        await rt.remove(box.id, force=True)
+
+
+@pytest.mark.asyncio
+async def test_exec_started_service_does_not_survive_stop_start(rt, image):
+    """The contrast: the same service started with exec is gone after a resume.
+
+    Guards the guidance, not a desired behaviour — stopping a box kills every
+    process in it, so a service the platform was never told about cannot be
+    replayed.
+    """
+    box = await rt.create(boxlite.BoxOptions(image=image, auto_remove=False))
+    try:
+        argv = " ".join(["python3", *SERVICE_ARGV])
+        await _run(box, f"nohup {argv} >/tmp/service.log 2>&1 &")
+        assert await _wait_serving(box), "exec-started service never came up before the stop"
+
+        await _restart(box)
+
+        # Give it at least as long as the positive case gets, so a pass here
+        # means "still absent", not "we did not wait long enough".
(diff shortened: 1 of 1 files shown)


apps/e2e/cases/test_service_survives_resume.py
```diff
@@ -0,0 +1,125 @@
...
+    """True when something answers on PORT from inside the box.
+
+    A failing exec counts as "not serving": right after a start the box can
+    briefly refuse execs, and that is the same answer as an empty port for
+    our purposes — the caller polls.
+    """
+    try:
+        code = await _run(box, f"curl -sf --max-time 5 http://127.0.0.1:{PORT}/ >/dev/null")
+    except Exception:
+        return False
```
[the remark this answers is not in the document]
G4614: Valid, and it was the worst failure mode this file could have — fixed in 76d2f4f07.

Added `_wait_exec_ready`, which polls a trivial `true` exec until the box accepts one, asserted right after `_restart` and before the port probe. Until that passes, "no service" and "no box" are the same observation, so the absence assertion proved nothing.

Left `_is_serving`'s behaviour alone: swallowing exec failures is correct for a poll (a freshly started box briefly refuses execs, which is the same answer as an empty port), it was only wrong as the basis for the negative conclusion. The positive case needs no equivalent gate — it asserts the service *is* reachable, which cannot pass without exec working.

apps/e2e/cases/test_service_survives_resume.py
```diff
@@ -0,0 +1,202 @@
...
+from conftest import drain
+
+PORT = 3000
+SERVICE_ARGV = ["-m", "http.server", str(PORT), "--bind", "0.0.0.0"]
+
+
+async def _run(box, script: str, timeout: int = 30) -> int:
+    ex = await box.exec("sh", ["-c", script])
+    await drain(ex)
+    rc = await asyncio.wait_for(ex.wait(), timeout=timeout)
```
DefeatMan: _run has no timeout on exec or drain, which defeats every deadline in the file.

async def _run(box, script, timeout=30):
    ex = await box.exec("sh", ["-c", script])
    await drain(ex)                     # ← unbounded
    rc = await asyncio.wait_for(ex.wait(), timeout=timeout)

_is_serving's docstring says a failing exec counts as "not serving… the caller polls" — but a hanging exec never returns, and _wait_serving only checks its deadline between iterations, so it hangs forever. This is exactly the failure test_exec_timeout.py is ignored for on cloud (e2e-cloud.yml:276-281: orphan grandchild keeps the stdout pipe open, ex.wait() hangs). test_sdk_tunnel.py:47 already wraps drain — match it:

ex = await asyncio.wait_for(box.exec("sh", ["-c", script]), timeout=timeout)
await asyncio.wait_for(drain(ex), timeout=timeout)

G4614: You're right, and it's worse than a missing timeout — fixed in 6be61b8eb.

The pollers only check their deadline *between* iterations, so one unbounded drain hangs the whole test rather than counting as "not serving" the way the docstring claims. And the shape you point at is exactly what the negative case launches: `nohup … &` is the orphaned-grandchild case from #910, so this file was set up to hit the hang that exec-timeout is ignored for on cloud, not to ride it out.

Wrapped the exec and the drain as in `test_sdk_tunnel.py`, and dropped the two probes to a 15s budget — below the 2s poll interval's window — so a stuck call costs one iteration instead of the entire wait. Docstring now says why the bound is there rather than just that it is.

apps/e2e/cases/test_service_survives_resume.py
```diff
@@ -0,0 +1,210 @@
...
+    """Run a command in the box, with every step bounded.
+
+    `drain` in particular: an exec whose orphaned grandchild still holds the
+    stdout pipe never EOFs and `wait()` hangs (guest-side root cause in #910,
+    which is why exec-timeout is ignored on cloud). An unbounded drain here
+    would hang inside a single poll iteration and defeat every deadline in
+    this file, since the callers only check theirs between iterations.
+    """
+    ex = await asyncio.wait_for(box.exec("sh", ["-c", script]), timeout=timeout)
+    await asyncio.wait_for(drain(ex), timeout=timeout)
```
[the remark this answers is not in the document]
G4614: Fixed in e4a8fc5e5 — one deadline computed before the exec, each stage getting only what's left.

apps/e2e/cases/test_service_survives_resume.py
```diff
@@ -0,0 +1,218 @@
...
+        # anything into a failed probe — otherwise "no service" and "no box"
+        # are the same observation and the assertion below proves nothing.
+        assert await _wait_exec_ready(box), (
+            "box never accepted an exec after the restart, so its service "
+            "state could not be measured"
+        )
+
+        # Give it at least as long as the positive case gets, so a pass here
+        # means "still absent", not "we did not wait long enough".
+        assert not await _wait_serving(box, timeout=20.0), (
```
DefeatMan, line 212: The comment says the negative case waits at least as long as the positive case, but this uses 20 seconds while `_wait_serving` defaults to 45. If an exec-started service were restored after 20–45 seconds, this test would incorrectly pass. Could both cases use the same shared timeout so the negative assertion cannot miss a slower revival?

apps/e2e/cases/test_service_survives_resume.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+async def _fetch_over_tunnel(box, timeout: float = 45.0) -> bytes:
+    """Reach the service the way a user does: through the box's network tunnel.
+
+    Retries for the same reason the in-box probe polls — the port is not
+    listening the instant the box reports running, and opening a tunnel to a
+    box that just came up can lose the race.
+    """
+    deadline = asyncio.get_running_loop().time() + timeout
```
DefeatMan, line 123: The 45-second deadline is still checked only between attempts. Tunnel creation, connection, write, and close are not bounded by the remaining budget, and a peer that continually returns small chunks within each five-second read timeout can keep the read loop alive well past the deadline. Could this use one end-to-end timeout budget, as `_run` now does? This was raised previously, but it remains unresolved on the current head.

DefeatMan:
Since this PR’s only deliverable is an E2E test, could we get an actual run of `test_service_survives_resume.py` before merging? The PR says the pytest case was not run, the repository’s E2E stack workflow is currently disabled, and the ordinary Python/API test jobs were skipped on this commit. The raw REST experiment verifies the scenario, but not the pytest harness or cleanup/timeout behavior.

Suggested command:

```bash
pytest apps/e2e/cases/test_service_survives_resume.py -v --timeout=180 --timeout-method=thread
```

PR #1468, opened 2026-09-10
feat(volumes): add a GCS/gcsfuse backend alongside S3/mount-s3

silverzki:
## Summary

Adds a GCS/gcsfuse backend for managed volumes alongside the existing S3/mount-s3
one. `VOLUME_STORAGE_BACKEND` selects it and defaults to `s3`, so a deployment
that does not set it keeps today's behaviour byte for byte.

## Call graph

```text
Before
  create                   (VolumeService · apps/api/src/box/services/volume.service.ts:51)  — gates on s3.endpoint
  handlePendingCreate      (VolumeManager · apps/api/src/box/managers/volume.manager.ts:199)
    ├─ s3Client.send(CreateBucketCommand)                                     — client assembled inline in the constructor
    └─ s3Client.send(PutBucketTaggingCommand)
  handlePendingDelete      (VolumeManager · apps/api/src/box/managers/volume.manager.ts:261)
    └─ deleteS3Bucket      (util · apps/api/src/common/utils/delete-s3-bucket.ts:15)

  ensureVolumeFuseMounted  (Client · apps/runner/pkg/boxlite/volumes.go:122)   — unbounded exec
    └─ getMountCmd         (Client · apps/runner/pkg/boxlite/volumes.go:215)   — argv, credential env and the systemd wrapper in one function
         └─ exec.Command("mount-s3", …)                                       — the only backend

After
  create                   (VolumeService · apps/api/src/box/services/volume.service.ts:52)
    └─ isVolumeStorageConfigured  (fn · apps/api/src/box/stores/volume-bucket.store.ts:43)  — one predicate, shared with the reconciler
  handlePendingCreate      (VolumeManager · apps/api/src/box/managers/volume.manager.ts:161)
    └─ VolumeBucketStore.create   (iface · apps/api/src/box/stores/volume-bucket.store.ts:33)
         ├─ S3VolumeBucketStore   — CreateBucket + PutBucketTagging, unchanged
         └─ GcsVolumeBucketStore  — createBucket + setLabels
  handlePendingDelete      (VolumeManager · apps/api/src/box/managers/volume.manager.ts:200)
    └─ VolumeBucketStore.destroy  (iface · apps/api/src/box/stores/volume-bucket.store.ts:35)
         ├─ S3VolumeBucketStore   — deleteS3Bucket, unchanged
         └─ GcsVolumeBucketStore  — deleteFiles (array rejection) then delete

  ensureVolumeFuseMounted  (Client · apps/runner/pkg/boxlite/volumes.go:132)   — volumeMountTimeout; unmounts before rmdir on failure
    ├─ getMountCmd         (Client · apps/runner/pkg/boxlite/volumes.go:273)   — selects the backend, nothing else
    │    ├─ mountS3Spec      (Client · apps/runner/pkg/boxlite/volumes.go:283) — existing argv and AWS_* env, verbatim
    │    ├─ gcsfuseMountSpec (func   · apps/runner/pkg/boxlite/volumes.go:323) — gcsfuse argv; no credential env, resolves ADC
    │    └─ wrapMountCmd     (Client · apps/runner/pkg/boxlite/volumes.go:346) — shared systemd --scope wrapper, context-bound
    ├─ isDirectoryMounted  (Client · apps/runner/pkg/boxlite/volumes.go:230)   — now bounded; stat on a dead mount blocks
    └─ umountPath          (Client · apps/runner/pkg/boxlite/volumes.go:242)   — new, same bound
```

## Changes

- **Runner** — `getMountCmd` split into backend selection, two argv builders and a
  shared wrapper. The gcsfuse argv is not a one-to-one translation:
  `--allow-delete` / `--allow-overwrite` have no counterpart (gcsfuse permits both
  by default and rejects the flags), `--implicit-dirs` is required for parity, and
  `--metadata-cache-ttl-secs` is pinned to `0` because gcsfuse otherwise caches
  metadata for 60s where mount-s3 defaults to strong read-after-write consistency.
- **Runner** — the mount runs under `exec.CommandContext` in both branches. gcsfuse
  retries an unreachable endpoint indefinitely, so the previously unbounded exec
  would wedge box creation.
- **Runner** — `systemd-run --scope` is kept for both backends. A FUSE daemon started
  as a plain child of the runner service lands in that service's cgroup and dies with
  the next `systemctl restart`, leaving mounted boxes on a dead mountpoint.
- **API** — `VolumeManager` no longer builds an S3 client; it depends on a
  `VolumeBucketStore` with `probe` / `create` / `destroy`. The S3 implementation is
  the existing code moved verbatim, including the both-or-neither key validation and
  the scoped `ListObjectsV2` probe.
- **API** — the GCS implementation normalises two backend differences that would
  otherwise break silently: `BucketNotEmpty` vs HTTP 409 (callers branch on the
  shared `VolumeBucketNotEmptyError`), and GCS labels rejecting the uppercase S3 tag
  keys.
- **Credentials** — the GCS paths pass none. gcsfuse and `@google-cloud/storage` both
  resolve Application Default Credentials, mirroring how the S3 paths fall through to
  the instance and task roles. That also keeps secrets out of the `systemd-run` argv.
- Adds `@google-cloud/storage` to the apps workspace.

## How to verify

```bash
# Runner: the three mount-s3 cases pin today's argv and env; they pass unchanged
# against the pre-refactor code as well.
cd apps/runner && go test ./pkg/boxlite/ -run 'TestMountS3|TestGcsfuse|TestUnsetBackend' -v

# API
cd apps && npx jest --config api/jest.config.ts \
  --testPathPatterns 'volume-bucket.store.spec|volume.manager.spec'
```

## Risks / rollout

- Default is `s3`; an unset `VOLUME_STORAGE_BACKEND` selects mount-s3, covered by a test.
- `exec.Command` → `exec.CommandContext` also affects the S3 path, but only on hosts
  without systemd; the systemd branch was already context-bound.
- Selecting `gcs` requires the API and the runner to agree, and the runner host to have
  `gcsfuse` installed. Mixing backends across the two would create buckets one side
  cannot mount.
- **`apps/infra` is deliberately untouched, so no deployment this repository builds can
  select the new backend.** That stack provisions AWS EC2 runners, which run the s3
  backend; installing gcsfuse into its AMI or wiring a GCS-only variable into its systemd
  unit would ship configuration those hosts can never use. A gcs-capable host is
  provisioned outside this repository. Flagging it explicitly because it is a scope
  boundary, not an omission — say the word and the plumbing goes in.
- Selecting a backend does not install its mount binary: the host must already provide
  mount-s3, or gcsfuse 2.0+ (where `--metadata-cache-ttl-secs` replaced `--stat-cache-ttl`).
- **Readiness timing changes for both backends.** `waitForMountReady` moves from 50 fixed
  attempts (~5s) to a 30s deadline, because each probe can now spend up to
  `volumeProbeTimeout` against a wedged mount and counting attempts would multiply into
  minutes of holding the per-volume mutex. A slow mount that used to fail at ~5s now
  succeeds — user-visible, and the reason the bound is a deadline rather than a count.
- `TestMountProbeSeparatesCannotRunFromNotMounted` needs the `mountpoint` binary
  (util-linux) on the test host. Present on ubuntu-latest; on a host without it the case
  fails rather than skips.

## Verified by hand

- **gcsfuse** against a real GCS bucket with this exact argv: mount, daemonisation,
  `--implicit-dirs`, `--only-dir` tenant isolation, read/write, file rename. Directory
  rename fails on a flat bucket (`--rename-dir-limit` defaults to 0) exactly as it does
  on mount-s3, and reports `EMFILE`, which is worth knowing when triaging.
- **mount-s3 1.20.0** against MinIO with the argv this PR must not change: mount, read,
  write, delete, unmount — and the same two cases through `ensureVolumeFuseMounted`
  itself, covering the new timeout, the context-bound probes and the failure cleanup.
- Not exercised: the virtiofs hop into a box, and the 90s timeout firing (mount-s3 does
  not hang the way gcsfuse does, so it was not reproducible here).

## Summary by CodeRabbit

- **New Features**
  - Added Google Cloud Storage support for volume buckets and mounting.
  - Added configurable volume storage backend selection, with S3 remaining the default.
  - Added GCS location and project configuration options.

- **Bug Fixes**
  - Improved volume mount readiness checks, timeouts, cleanup, and unmount handling.
  - Improved handling of missing, incomplete, or non-empty storage buckets.
  - Prevented storage credentials from appearing in mount command arguments.

- **Documentation**
  - Added architecture documentation for the GCS volume storage backend.

- **Tests**
  - Added coverage for S3 and GCS bucket operations and volume mounting scenarios.

### DIFF
--- a/apps/api/src/box/managers/volume.manager.ts
+++ b/apps/api/src/box/managers/volume.manager.ts
@@ -10,12 +10,11 @@ import { Repository, In } from 'typeorm'
 import { Volume } from '../entities/volume.entity'
 import { VolumeState } from '../enums/volume-state.enum'
 import { Cron, CronExpression, SchedulerRegistry } from '@nestjs/schedule'
-import { S3Client, CreateBucketCommand, ListObjectsV2Command, PutBucketTaggingCommand } from '@aws-sdk/client-s3'
 import { InjectRedis } from '@nestjs-modules/ioredis'
 import { Redis } from 'ioredis'
 import { RedisLockProvider } from '../common/redis-lock.provider'
 import { TypedConfigService } from '../../config/typed-config.service'
-import { deleteS3Bucket } from '../../common/utils/delete-s3-bucket'
+import { createVolumeBucketStore, VolumeBucketNotEmptyError, VolumeBucketStore } from '../stores/volume-bucket.store'
 
 import { TrackableJobExecutions } from '../../common/interfaces/trackable-job-executions'
 import { TrackJobExecution } from '../../common/decorators/track-job-execution.decorator'
@@ -34,7 +33,7 @@ export class VolumeManager
   private readonly logger = new Logger(VolumeManager.name)
   private processingVolumes: Set<string> = new Set()
   private skipTestConnection = false
-  private s3Client: S3Client | null = null
+  private bucketStore: VolumeBucketStore | null = null
 
   constructor(
     @InjectRepository(Volume)
@@ -44,44 +43,17 @@ export class VolumeManager
     private readonly redisLockProvider: RedisLockProvider,
     private readonly schedulerRegistry: SchedulerRegistry,
   ) {
-    if (!this.configService.get('s3.endpoint')) {
-      return
-    }
-
-    const endpoint = this.configService.getOrThrow('s3.endpoint')
-    const region = this.configService.getOrThrow('s3.region')
-    const accessKeyId = this.configService.get('s3.accessKey')
-    const secretAccessKey = this.configService.get('s3.secretKey')
     this.skipTestConnection = this.configService.get('skipConnections')
-
-    // Both-or-neither: a lone key is a typo'd pair, and silently falling back
-    // to the SDK default chain would mask the misconfig.
-    if ((accessKeyId && !secretAccessKey) || (!accessKeyId && secretAccessKey)) {
-      throw new Error('S3_ACCESS_KEY and S3_SECRET_KEY must be set together')
-    }
-    // MinIO cannot use the SDK default chain — fail fast at boot with a clear
-    // message instead of a generic auth error from the connection probe.
-    if (endpoint.includes('minio') && !accessKeyId) {
-      throw new Error('MinIO requires S3_ACCESS_KEY and S3_SECRET_KEY to be configured')
-    }
-
-    this.s3Client = new S3Client({
-      endpoint: endpoint.startsWith('http') ? endpoint : `http://${endpoint}`,
-      region,
-      // Static keys for S3-compatible deployments (MinIO); unset on AWS,
-      // where the SDK default chain supplies the ECS task-role credentials.
-      ...(accessKeyId && secretAccessKey ? { credentials: { accessKeyId, secretAccessKey } } : {}),
-      forcePathStyle: true,
-    })
+    this.bucketStore = createVolumeBucketStore(this.configService)
   }
 
   async onModuleInit() {
-    if (!this.s3Client) {
+    if (!this.bucketStore) {
       return
     }
 
     if (this.skipTestConnection) {
-      this.logger.debug('Skipping S3 connection test')
+      this.logger.debug('Skipping volume bucket store connection test')
       return
     }
 
@@ -89,7 +61,7 @@ export class VolumeManager
   }
 
   onApplicationBootstrap() {
-    if (!this.s3Client) {
+    if (!this.bucketStore) {
       return
     }
 
@@ -105,21 +77,11 @@ export class VolumeManager
   }
 
   private async testConnection() {
-    // Probe a bucket we already know instead of ListBuckets: same
-    // connectivity+auth signal, but needs no account-wide
-    // s3:ListAllMyBuckets grant on the task role.
-    const bucket = this.configService.get('s3.defaultBucket')
-    if (!bucket) {
-      this.logger.debug('No default bucket configured; skipping S3 connection test')
-      return
-    }
-
     try {
-      const command = new ListObjectsV2Command({ Bucket: bucket, MaxKeys: 1 })
-      await this.s3Client.send(command)
-      this.logger.debug('Successfully connected to S3')
+      await this.bucketStore.probe()
+      this.logger.debug('Successfully connected to the volume bucket store')
     } catch (error) {
-      this.logger.error('Failed to connect to S3:', error)
+      this.logger.error('Failed to connect to the volume bucket store:', error)
       throw error
     }
   }
@@ -129,7 +91,7 @@ export class VolumeManager
   @LogExecution('process-pending-volumes')
   @WithInstrumentation()
   async processPendingVolumes() {
-    if (!this.s3Client) {
+    if (!this.bucketStore) {
       return
     }
 
@@ -207,37 +169,14 @@ export class VolumeManager
         state: VolumeState.CREATING,
       })
 
-      // Refresh lock before S3 operation
+      // Refresh lock before the object-store operation
       await this.redis.setex(lockKey, 30, '1')
 
-      // Create bucket in Minio/S3
-      const createBucketCommand = new CreateBucketCommand({
-        Bucket: volume.getBucketName(),
+      await this.bucketStore.create(volume.getBucketName(), {
+        VolumeId: volume.id,
+        OrganizationId: volume.organizationId,
+        Environment: this.configService.get('environment'),
       })
-
-      await this.s3Client.send(createBucketCommand)
-
-      await this.s3Client.send(
-        new PutBucketTaggingCommand({
-          Bucket: volume.getBucketName(),
-          Tagging: {
-            TagSet: [
-              {
-                Key: 'VolumeId',
-                Value: volume.id,
-              },
-              {
-                Key: 'OrganizationId',
-                Value: volume.organizationId,
-              },
-              {
-                Key: 'Environment',
-                Value: this.configService.get('environment'),
-              },
-            ],
-          },
-        }),
-      )
 
       // Refresh lock before final state update
       await this.redis.setex(lockKey, 30, '1')
@@ -269,20 +208,17 @@ export class VolumeManager
         state: VolumeState.DELETING,
       })
 
-      // Refresh lock before S3 operation
+      // Refresh lock before the object-store operation
       await this.redis.setex(lockKey, 30, '1')
 
-      // Delete bucket from Minio/S3
+      // A bucket that is already gone is absorbed by the store as success.
       try {
-        await deleteS3Bucket(this.s3Client, volume.getBucketName())
+        await this.bucketStore.destroy(volume.getBucketName())
       } catch (error) {
-        if (error.name === 'NoSuchBucket') {
-          this.logger.warn(`Bucket for volume ${volume.id} does not exist, treating as already deleted`)
-        } else if (error.name === 'BucketNotEmpty') {
+        if (error instanceof VolumeBucketNotEmptyError) {
           throw new Error('Volume deletion failed because the bucket is not empty. You may retry deletion.')
-        } else {
-          throw error
         }
+        throw error
       }
 
       // Refresh lock before final state update
--- a/apps/api/src/box/stores/volume-bucket.store.ts
+++ b/apps/api/src/box/stores/volume-bucket.store.ts
@@ -1,0 +1,244 @@
+/*
+ * Copyright 2026 BoxLite AI
+ * SPDX-License-Identifier: AGPL-3.0
+ */
+
+import { Logger } from '@nestjs/common'
+import { S3Client, CreateBucketCommand, PutBucketTaggingCommand, ListObjectsV2Command } from '@aws-sdk/client-s3'
+import { Storage } from '@google-cloud/storage'
+import { TypedConfigService } from '../../config/typed-config.service'
+import { deleteS3Bucket } from '../../common/utils/delete-s3-bucket'
+
+/**
+ * Raised by `destroy` when the bucket still holds objects. The two backends
+ * signal this differently — S3 with a `BucketNotEmpty` error name, GCS with a
+ * 409 — so it is normalised here; a caller that branched on either backend's
+ * native shape would silently stop handling the case on the other.
+ */
+export class VolumeBucketNotEmptyError extends Error {
+  constructor(bucket: string) {
+    super(`Volume bucket ${bucket} is not empty`)
+    this.name = 'VolumeBucketNotEmptyError'
+  }
+}
+
+/**
+ * The object-store operations a volume's lifecycle needs. One volume is one
+ * bucket, so the whole surface is create/destroy plus a boot-time reachability
+ * probe.
+ */
+export interface VolumeBucketStore {
+  /** Fail fast at boot rather than on the first volume a user creates. */
+  probe(): Promise<void>
+  create(bucket: string, labels: Record<string, string>): Promise<void>
+  /** Idempotent: a bucket that is already gone resolves rather than throwing. */
+  destroy(bucket: string): Promise<void>
+}
+
+/**
+ * Whether volumes can be served at all. Both the request path and the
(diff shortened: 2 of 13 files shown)


apps/runner/pkg/boxlite/volumes.go
```diff
@@ -230,17 +407,65 @@ func (c *Client) getMountCmd(ctx context.Context, volume string, path string) *e
...
+// service account. The runner's own environment is still inherited (cmd.Env is
+// nil), which is what lets a GOOGLE_APPLICATION_CREDENTIALS path work off-GCE.
+// What this does buy is that no secret is placed on the systemd-run argv, where
+// /proc/<pid>/cmdline exposes it to every local user.
+func gcsfuseMountSpec(bucket string, path string) mountSpec {
+	return mountSpec{
+		bin: "gcsfuse",
+		args: []string{
+			"-o", "allow_other",
+			"--file-mode", "0666",
```
DorianZheng, line 437: extract common constants for both s3 and gcs

apps/runner/pkg/boxlite/volumes.go
```diff
@@ -230,17 +407,65 @@ func (c *Client) getMountCmd(ctx context.Context, volume string, path string) *e
...
+//
+// The spec contributes no credential environment of its own; gcsfuse resolves
+// Application Default Credentials, which on GCE is the attached instance
+// service account. The runner's own environment is still inherited (cmd.Env is
+// nil), which is what lets a GOOGLE_APPLICATION_CREDENTIALS path work off-GCE.
+// What this does buy is that no secret is placed on the systemd-run argv, where
+// /proc/<pid>/cmdline exposes it to every local user.
+func gcsfuseMountSpec(bucket string, path string) mountSpec {
+	return mountSpec{
+		bin: "gcsfuse",
```
DorianZheng, line 434: where is this binary got installed

DorianZheng:
```
Call graph (After)

  constructor (VolumeManager · volume.manager.ts:38) — at boot · CHANGED: no inline S3Client
    └─ createVolumeBucketStore (fn · volume-bucket.store.ts:55) — ADDED: VOLUME_STORAGE_BACKEND picks S3 or GCS
  create_volume (RestRuntime · runtime.rs:44) — SDK: POST /v1/volumes
    └─ create (BoxliteVolumeController · boxlite-volume.controller.ts:35)
      ├─ create (VolumeService · volume.service.ts:51) — inserts pending_create · CHANGED gate
      │  └─ isVolumeStorageConfigured (fn · volume-bucket.store.ts:43) — ADDED: gcs needs no S3_ENDPOINT
      └─ waitForReady (VolumeService · volume.service.ts:87) — polls every 500 ms, 30 s max
  processPendingVolumes (VolumeManager · volume.manager.ts:93) — cron, every 5 s
    └─ processVolumeState (:140)
      └─ handlePendingCreate (:161) — creating → bucket → ready · CHANGED: calls the store
        └─ create (VolumeBucketStore · volume-bucket.store.ts:33) — ADDED interface
          ├─ create (S3VolumeBucketStore · :129) — same S3 calls, moved here
          └─ create (GcsVolumeBucketStore · :181) — ADDED
  create (RestRuntime · runtime.rs:199) — SDK: POST /v1/boxes
    └─ createBox (BoxliteBoxController · boxlite-box.controller.ts:112)
      └─ create (BoxService · box.service.ts:187) — emits BoxEvents.CREATED
        └─ validateVolumes (VolumeService · volume.service.ts:215) — volume must be ready
  run (BoxStartAction · box-start.action.ts:37) — after BoxEvents.CREATED
    └─ handleRunnerBoxUnknownStateOnDesiredStateStart (:59)
      └─ createBox (RunnerAdapterV2 · runnerAdapter.v2.ts:116) — enqueues the CREATE_BOX job
  createBox (Executor · executor/box.go:17) — runner, the job it polled
    └─ Create (BoxliteAdapter · boxlite_adapter.go:29)
      └─ Create (Client · client.go:240)
        ├─ getVolumeMounts (volumes.go:68) — mounts at /mnt/boxlite-volume-{id}
        │  └─ ensureVolumeFuseMounted (:139) — CHANGED: 90 s limit on the mount
        │     ├─ isDirectoryMounted (:262) — CHANGED: 10 s limit
        │     ├─ getMountCmd (:382) — CHANGED: only picks the backend
        │     │  ├─ mountS3Spec (:392) — existing mount-s3 args and AWS env
        │     │  ├─ gcsfuseMountSpec (:432) — ADDED
        │     │  └─ wrapMountCmd (:458) — systemd-run --scope
        │     ├─ waitForMountReady (:331) — CHANGED: 30 s deadline
        │     └─ cleanUpFailedMount (:221) — ADDED: unmount, then remove the dir
        └─ WithBindMount (fn · sdks/go/options.go:299) — into the box at mountPath
```

apps/runner/pkg/boxlite/volumes.go
silverzki: apps/infra/stack/runners.ts 的 gcp stack里，这部分代码等新宇的代码合并

apps/runner/pkg/boxlite/volumes.go
silverzki: OK

silverzki:
Thanks — status on all seven.

### Fixed

**Secret in the `systemd-run` argv** — `42e16409`. Different route than the suggested diff: `--setenv` is gone entirely rather than reduced to `--setenv=NAME`. Under `--scope`, systemd-run execs the tool as its own child, so `cmd.Env` already reaches it. Both branches now share one `cmd.Env = append(os.Environ(), spec.env...)`, which also restores the `PATH`, proxy and `AWS_SESSION_TOKEN` that the old non-systemd branch silently dropped when credentials were configured. `TestMountS3KeepsTheSecretOutOfTheCommandLine` scans the whole argv.

**Probe hard-depends on `mountpoint`** — `42e16409`, `volumes.go:284`. Instead of a Go-side fallback, the probe now names the cause: `errors.Is(err, exec.ErrNotFound)` returns *"mountpoint is not installed on this host"* rather than an inconclusive failure. Provisioning still owns util-linux; this makes its absence legible instead of silent.

**Test depends on `mountpoint`** — `volumes_test.go:394` guards the assertion with `exec.LookPath("mountpoint")`, and `TestMountProbeNamesAMissingBinary` (`:418`) covers the missing-binary path via `t.Setenv("PATH", "")`.

**GCS soft delete** — `ab05ef1c`: `softDeletePolicy: { retentionDurationSeconds: '0' }` at `volume-bucket.store.ts:194`. `cf29f31e` adds a post-create check that the bucket came back with the settings we asked for.

### Not changing

**Force-delete of GCS bucket contents** — the premise doesn't hold. S3 does not raise `VolumeBucketNotEmptyError` for a non-empty bucket: `delete-s3-bucket.ts` pages `ListObjectVersions` and `ListObjectsV2`, `DeleteObjects` everything, then deletes *"the (now-empty) bucket"* (`:66-67`). Its `BucketNotEmpty` branch is reachable only on a write race. The two backends already agree.

Making GCS refuse would have inverted that, not aligned it — and the consequence is worse than it looks: `volume.manager.ts` maps the refusal to `ERROR` with "you may retry", while every retry re-issues the same bare `delete()`. A non-empty GCS volume would have had no path to deletion at all — live and billable indefinitely, with soft delete now disabled.

**`CreateBucketConfiguration.LocationConstraint`** — real for an AWS deployment outside `us-east-1`, but pre-existing rather than introduced here. The same bare `CreateBucketCommand({ Bucket })` sat at `volume.manager.ts:214` before this PR (comment: *"Create bucket in Minio/S3"*); this PR relocates it unchanged into the new store. Tracked separately rather than widening this PR.

**Pin gcsfuse ≥ 2.0** — correct that `--metadata-cache-ttl-secs` is rejected by 1.x, and correct that nothing in this repository provisions gcsfuse. That makes the version floor a property of whatever installs the binary, which this PR does not add. The measured behaviour is recorded in `docs/architecture/volume-gcs-adapter.md` (gcsfuse 3.8.4) so the constraint is written down where provisioning will need it.

docs/architecture/volume-gcs-adapter.md
```diff
@@ -0,0 +1,276 @@
+# Volume mounts on GCS (keeping S3 compatibility)
```
DorianZheng: seems it's a temporary file

silverzki: removed in dc16209f

DorianZheng (approved):
LGTM

apps/api/src/box/stores/volume-bucket.store.ts
```diff
@@ -0,0 +1,244 @@
...
+    // connectivity+auth signal, but needs no account-wide
+    // s3:ListAllMyBuckets grant on the task role.
+    if (!this.probeBucket) {
+      return
+    }
+    await this.client.send(new ListObjectsV2Command({ Bucket: this.probeBucket, MaxKeys: 1 }))
+  }
+
+  async create(bucket: string, labels: Record<string, string>): Promise<void> {
+    await this.client.send(new CreateBucketCommand({ Bucket: bucket }))
```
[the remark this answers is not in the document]
silverzki: it was pre-existing rather than introduced here, which is true but understates it. It is not a defect at all.

@aws-sdk/client-s3 does this automatically. @aws-sdk/middleware-location-constraint is a direct dependency of the client (client-s3@3.1048.0 requires ^3.972.10, 3.972.10 installed), and its plugin is attached to CreateBucketCommand specifically — client-s3/dist-cjs/index.js:229, in the plugin list immediately alongside .s("AmazonS3", "CreateBucket"). The middleware body:

const { CreateBucketConfiguration } = args.input;
const region = await options.region();
if (!CreateBucketConfiguration?.LocationConstraint && !CreateBucketConfiguration?.Location) {
    if (region !== "us-east-1") {
        args.input.CreateBucketConfiguration = args.input.CreateBucketConfiguration ?? {};
        args.input.CreateBucketConfiguration.LocationConstraint = region;
    }
}

That covers all three points in the comment: the constraint is added when the region is not us-east-1, it is taken from the client's configured region, and it is omitted for us-east-1. So the bare CreateBucketCommand({ Bucket }) at volume-bucket.store.ts:130 is already complete by the time it leaves the SDK, and PutBucketTaggingCommand is never reached with a rejected bucket creation.

Worth noting the middleware only fills the field in when it is unset. Writing it by hand would be redundant, and a hand-written version sourcing the region from anywhere other than the client config would override a value the SDK had already gotten right — so the suggested change is a small net negative rather than a no-op.

The S3-compatible endpoint configuration is untouched either way: the middleware runs at the initialize step on the input shape and does not interact with endpoint or forcePathStyle.

apps/api/src/box/managers/volume.manager.ts
```diff
@@ -269,20 +208,17 @@ export class VolumeManager
...
 
-      // Refresh lock before S3 operation
+      // Refresh lock before the object-store operation
       await this.redis.setex(lockKey, 30, '1')
 
-      // Delete bucket from Minio/S3
+      // A bucket that is already gone is absorbed by the store as success.
       try {
-        await deleteS3Bucket(this.s3Client, volume.getBucketName())
+        await this.bucketStore.destroy(volume.getBucketName())
```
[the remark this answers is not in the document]
silverzki: Force-delete of GCS bucket contents — the premise doesn't hold. S3 does not raise VolumeBucketNotEmptyError for a non-empty bucket: delete-s3-bucket.ts pages ListObjectVersions and ListObjectsV2, DeleteObjects everything, then deletes "the (now-empty) bucket" (:66-67). Its BucketNotEmpty branch is reachable only on a write race. The two backends already agree.

Making GCS refuse would have inverted that, not aligned it — and the consequence is worse than it looks: volume.manager.ts maps the refusal to ERROR with "you may retry", while every retry re-issues the same bare delete(). A non-empty GCS volume would have had no path to deletion at all — live and billable indefinitely, with soft delete now disabled.

https://linear.app/polygala/issue/POL-585/deleting-a-volume-destroys-a-non-empty-bucket-with-no-confirmation

will add confirmation

PR #1474, opened 2026-09-11
fix(runtime): propagate default runtime errors (#1444)

### LINKED ISSUE #1444: SDK bindings abort instead of raising when the database schema is newer than the process
`Database::ensure_schema` already produces a proper, recoverable error when the on-disk schema is newer than the process expects. `BoxliteRuntime::default_runtime()` then converts that error into a `panic!`, which crosses the Node FFI boundary as a process abort. SDK callers cannot catch it, log it, or degrade — the process dies.

The CLI handles the identical error correctly, which is what makes this look like an oversight rather than a design choice.

## The two code paths

`src/boxlite/src/db/mod.rs:118` — returns a recoverable error:

```rust
Some(v) if v > schema::SCHEMA_VERSION => {
    return Err(BoxliteError::Database(format!(
        "Schema version mismatch: database has v{}, process expects v{}. \
         Upgrade boxlite to a newer version.",
        v, schema::SCHEMA_VERSION
    )));
}
```

`src/boxlite/src/runtime/core.rs:197` — turns it into a panic:

```rust
.unwrap_or_else(|e| panic!("Failed to initialize BoxliteRuntime:\n\n{e}"))
```

## Reproduction

Any version skew across SDKs on one machine reaches this. Verified on Darwin arm64 (macOS 26.1).

```bash
# 1. Take the shared database to the current schema with a current SDK
pip install -U boxlite                        # 0.10.0
python3 -c "
import asyncio, boxlite
async def m():
    async with boxlite.SimpleBox(image='alpine:latest') as b:
        print(b.id)
asyncio.run(m())"

# 2. Install the newest Node package that can actually load (see #1443)
D=$(mktemp -d) && cd "$D" && npm init -y >/dev/null
npm install @boxlite-ai/boxlite@0.9.7

# 3. Wrap everything in try/catch and run it
cat > t.mjs <<'JS'
try {
  const box = new SimpleBox({ image: "alpine:latest" });
  await box.start();
  console.log("started");
} catch (err) {
  console.log("CAUGHT:", err.message);
}
console.log("still alive after try/catch");
JS
node t.mjs
```

Actual:

```
thread '<unnamed>' (23402646) panicked at src/boxlite/src/runtime/core.rs:173:37:
Failed to initialize BoxliteRuntime:

storage error: Failed to initialize database at /Users/<user>/.boxlite/db/boxlite.db: database error: Schema version mismatch: database has v10, process expects v8. Upgrade boxlite to a newer version.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 5, aborting
```

**Neither `CAUGHT:` nor `still alive after try/catch` is ever printed.** The process aborts inside `box.start()`.

## Contrast: the CLI gets this right

Same machine, same database, same underlying error:

```
$ boxlite ls
Error: storage error: Failed to initialize database at /Users/<user>/.boxlite/db/boxlite.db: database error: Schema version mismatch: database has v10, process expects v8. Upgrade boxlite to a newer version.
```

Clean message, normal exit, no panic. The error is recoverable — only the SDK entry point discards that.

## Impact

Every SDK example in the documentation teaches `try`/`catch` (Python) or `try`/`catch`/`finally` (Node) around box creation. That pattern silently does not apply to this failure. A service that creates boxes cannot log the reason, fall back, or return a 500 — the worker process is gone.

The trigger is narrow (version skew), but the failure mode is the worst available: an uncatchable abort in a library.

## Suggested fix

Propagate the `Result` to the binding layer rather than unwrapping it. `core.rs:149` already documents a variant that "returns a `Result` instead of panicking" — the SDK entry points should be using that path, so `BoxliteError::Database` arrives in Python as an exception and in Node as a rejected promise.

hangzws:
## Summary

Make default runtime initialization failures catchable in the Python and Node SDKs.

`default_runtime_with_result()` preserves the global default-runtime singleton while returning recoverable initialization errors. Failed initialization is not cached, so callers can retry after correcting the underlying condition.

## Call graph

Before

  Boxlite.default / JsBoxlite.withDefaultConfig
    └─ PyBoxlite::default / JsBoxlite::with_default_config
         (SDK binding · sdks/python/src/runtime.rs:34 / sdks/node/src/runtime.rs:58)
         └─ BoxliteRuntime::default_runtime
              (BoxliteRuntime · src/boxlite/src/runtime/core.rs:233)
              └─ OnceLock::get_or_init → BoxliteRuntime::with_defaults().unwrap_or_else(...)
                   ← BUG: an initialization error panics across the PyO3 or napi-rs FFI boundary

After

  Boxlite.default / JsBoxlite.withDefaultConfig
    └─ PyBoxlite::default / JsBoxlite::with_default_config
         (SDK binding · sdks/python/src/runtime.rs:34 / sdks/node/src/runtime.rs:58)
         └─ BoxliteRuntime::default_runtime_with_result
              (BoxliteRuntime · src/boxlite/src/runtime/core.rs:180)
              └─ initialization mutex → BoxliteRuntime::with_defaults()?
                   └─ successful runtime stored in DEFAULT_RUNTIME
                   └─ initialization error returned through map_err as RuntimeError / Error

Fixes #1444

## Changes

- Add `BoxliteRuntime::default_runtime_with_result()` for fallible default-runtime initialization.
- Serialize default-runtime initialization without caching failures, allowing a later retry after the environment is corrected.
- Keep `default_runtime()` as the backward-compatible panic-on-error Rust API.
- Route Python and Node default runtime creation through the fallible API so errors remain catchable at the FFI boundary.
- Clarify schema mismatch diagnostics with compatible-SDK guidance and the isolated `BOXLITE_HOME` fallback.
- Add isolated Rust, Python, and Node coverage for a newer database schema, catchable errors, and successful retry after removing the incompatible database.

## How to verify

```bash
make test:unit:rust FILTER=default_runtime_schema_mismatch_is_recoverable_and_retriable
make test:unit:rust FILTER=init_default_runtime_rejects_a_second_configuration_before_initializing_it
make test:integration:python FILTER=schema_mismatch_is_catchable_and_retriable
make test:integration:node FILTER="schema mismatch is catchable and retriable"

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Bug Fixes**
  * Default runtime initialization errors in the Node.js and Python SDKs are now reported as catchable errors instead of causing fatal runtime failures.
  * Schema mismatches provide clearer guidance, including upgrading the SDK or using a new `BOXLITE_HOME`.
  * Runtime initialization can be retried successfully after resolving an incompatible database.

* **API Updates**
  * Added a distinct “already initialized” error for runtime initialization conflicts across supported APIs.

* **Tests**
  * Added coverage for schema mismatch recovery, repeated initialization, error classification, and prevention of panic output.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### CODE REVIEW

File: src/boxlite/src/runtime/core.rs [hunk 1]
```diff
@@ -163,6 +177,37 @@ impl BoxliteRuntime {
         Self::new(BoxliteOptions::default())
     }
 
+    /// Get or initialize the default global runtime without panicking.
+    ///
+    /// The first successful call stores one shared runtime and installs its exit
+    /// cleanup handlers. Initialization errors are returned without populating the
+    /// singleton, so a later caller can retry after correcting the underlying
+    /// condition.
+    pub fn default_runtime_with_result() -> BoxliteResult<&'static Self> {
```
DorianZheng, line 186: if the goal is to catch the failure of default runtime initialization, just use `init_default_runtime`. this interface seems duplicated

hangzws: Yes, it can receive BoxliteOptions::default(), but init_default_runtime is a one-shot explicit initializer: calling it a second time returns an “already initialized” error.
The SDK default entry points need lazy, repeatable access to the shared runtime, so default_runtime_with_result() is still needed.

DorianZheng: I think default runtime only throw exception when initializing. We can catch and ignore `already initialized` error

DorianZheng: and we can call `init_default_runtime` only for once when start the app

hangzws: I do not think so, init_default_runtime(BoxliteOptions::default()) works for a one-time startup flow, but SDK default entry points may be called repeatedly in a long-lived process. A second call returns already initialized, so the SDK must catch that error and then call default_runtime() to obtain the existing instance. The home lock only coordinates processes and does not guarantee one call per process. Therefore, default_runtime_with_result() better matches the SDK’s lazy, reusable, and fallible semantics.

hangzws: The SDK could catch the already initialized error and then call default_runtime() to obtain the existing instance. However, that requires classifying this error separately from genuine initialization failures and coordinating two API calls. Given the current repeatable default-entry-point semantics, I prefer keeping default_runtime_with_result(), but I’m happy to follow the project’s preferred initialization contract.

If the intended contract is startup-only initialization, I can switch the SDK bindings to this approach.

hangzws: > and we can call `init_default_runtime` only for once when start the app

I have changed the PR to call `init_default_runtime` in python/nodejs sdk,  Note: The two function init_default_runtime, default_runtime lack lock protection and are susceptible to race conditions under extreme conditions. please check it, if you have any questions please reply.

PR #1480, opened 2026-09-11
feat(net): merge Secret.hosts into a non-empty allow_net

ltstriker:
> **Stacked on #1481 (POL-563).** That PR fixes pre-existing defects in the DNS
> zone builder which merging hostname rules makes reachable without the caller
> writing them. Review #1481 first; this PR's own diff touches no DNS builder
> code. Base is `tingliang/pol-563-wildcard-allow-net-dns`, not `main`.

## Summary

A host named by `Secret.hosts` is now reachable without repeating it in `allow_net`, and the MITM proxy stops acting as an egress authorization of its own. Secrets merge only into a **non-empty** `allow_net`: empty already means unrestricted, so merging there would turn an open box into a restricted one.

## Call graph

```text
Before
  BoxOptions{network.allow_net, secrets}
    ├─ build_network_backend   (·  src/boxlite/src/litebox/init/tasks/vmm_spawn.rs:353)  — allow_net passed through raw
    │    └─ buildAllowNet      (·  src/deps/libgvproxy-sys/gvproxy-bridge/dns_filter.go:48)
    │         └─ DNS zones     — ← BUG (POL-483): a Secret.hosts entry is in no zone, so it
    │                              resolves to 0.0.0.0 and the guest never gets an address
    ├─ inspectAndForward       (·  src/deps/libgvproxy-sys/gvproxy-bridge/forked_tcp.go:190)
    │    └─ mitmAndForward     (·  .../mitm_proxy.go:19)   — ← BUG: returns before the allowlist
    │                                                          check, so a credential authorizes
    │                                                          egress on its own; the same order
    │                                                          skips MatchesIP, disabling
    │                                                          allow_net's IP/CIDR rules on :443
    └─ BoxInfo::new            (·  src/boxlite/src/runtime/types.rs:558)  — reports only the raw list

After
  BoxOptions::effective_allow_net   (BoxOptions · src/boxlite/src/runtime/options.rs:611)  — one policy, recomputed
    ├─ build_network_backend        (· vmm_spawn.rs:365)   — merged list reaches the backend
    │    └─ buildAllowNet           (· dns_filter.go:48)   — unchanged here; see #1481
    ├─ VmmAttachTask::run           (· vmm_attach.rs:40)   — same policy on the reattach path
    ├─ decideTCPInspectRoute        (· forked_tcp.go:97)   — one gate, three outcomes
    │    └─ inspectedDestinationAllowed (· forked_tcp.go:118)
    │         ├─ MatchesIP          (AllowNetFilter · allow_net_filter.go:111)
    │         └─ AllowHostToIP      (AllowNetFilter · allow_net_filter.go:169)
    │              └─ mitmAndForward (· mitm_proxy.go:19)  — only on an already-allowed destination
    └─ BoxInfo::new                 (· types.rs:558)       — reports allow_net and the enforced set
```

Refs POL-483

## Changes

- **`BoxOptions::effective_allow_net()`** sits beside `effective_auto_delete`, recomputed rather than written back, and is used by both `NetworkBackendConfig` construction sites so the spawn and reattach paths cannot drift.
- **Only exact hostnames merge.** A wildcard grants more as an allowlist rule (any depth) than as a secret host (one label); an address can never match an SNI, and merged it would become real egress — `0.0.0.0/0` would open all IPv4, and address rules open UDP too; a padded or dotted entry is unmatchable by the secret matcher; a single-label name has no reachable zone. Each would report egress the box does not have.
- **`inspectAndForward` decides once, before acting.** MITM became a treatment applied to an allowed destination. This also restores the IP/CIDR term the pre-peek path skips whenever secrets exist — configuring a credential no longer silently disables `allow_net`'s address rules on :443.
- **No DNS zone builder changes here.** Merging hostname rules automatically does make several pre-existing gaps in that builder reachable without the caller writing them; those are POL-563 and live in #1481, which this PR is stacked on.
- **`network.outbound.effective_allow_net`** on `BoxInfo`, mirrored in Python, Node, Go and C. The difference from `allow_net` is the provenance, so no per-entry source is stored. Pre-merge JSON folds to `effective == allow_net` rather than defaulting to empty, which would read as unrestricted.

## How to verify

```bash
cargo nextest run -p boxlite --features krun,gvproxy --test network_spec --profile vm --run-ignored all -E 'test(~secret_hosts) or test(~unlisted_host_stays) or test(~empty_allowlist_with) or test(~info_reports)'
```

```bash
make test:unit:gvproxy GOTEST_FILTER='-run TestInspectTCP|TestBuildAllowNet'
```

Two-side verification, with every production change reverted and only the tests kept:

| Reproducer | Pre-fix signal |
|---|---|
| `secret_hosts_join_a_restrictive_allowlist` | `httpbingo.org was sinkholed to 0.0.0.0, got answers ["0.0.0.0"]` |
| `TestInspectTCP_MitmDoesNotAuthorizeUnlistedDestination` | handshake succeeded, leaf issued by `"BoxLite Test CA"` — MITM authorized the egress |
| `TestInspectTCP_CIDRRuleStillAllowsHTTPSWhenSecretsConfigured` | `allowNet TCP: blocked (hostname not in allowlist)` |

Their positive controls pass on both sides.

## Risks / rollout

- **Behavior widens, and only for a non-empty `allow_net`.** A box with an empty `allow_net` is unaffected — that is the case the emptiness gate protects, and `empty_allowlist_with_secrets_stays_unrestricted` guards it.
- **A secret is now an egress grant as well as a credential.** `allow_net` is no longer the only gate; the reference docs say so explicitly.
- **`effective_allow_net` is computed from stored options, not read back from the gateway.** A box already running when this ships reports the merged set while its live gateway still enforces the pre-merge one; a restart re-applies policy. Documented on the field.
- **C consumers recompile**: `COutboundNetworkInfo` gains two trailing fields, so `CNetworkInfo.inbound` moves. Fields before it keep their offsets.
- The DNS-builder work this depends on is reviewed separately in #1481; landing this without it leaves the merge unreliable for configs that combine wildcards or repeat a host.
- Not verifiable on Cloud: the create DTO still rejects `secrets`, so the whole issue is unacceptable there until that lands.
- `docs.boxlite.ai/agent-tools/github-operations` is not in this repo; its two examples and the Troubleshooting row telling users to list a host in both places still need updating.


ltstriker:
Closing: #1495 delivers what this PR set out to do, and the remaining parts of this diff would now work against it.

Measured on current main, `allow_net=["pypi.org"]` plus `Secret(hosts=["httpbingo.org"])`, nothing else:

```
secret host :443, not in allow_net   EXIT:0   reachable
pypi.org (listed)                    EXIT:0   reachable
example.org (neither)                EXIT:1   blocked
secret host :80                      EXIT:1   blocked (MITM is 443 only)
```

So the redundancy this PR removed — writing the same host into `allow_net` and `Secret.hosts` — is already gone for HTTPS, which is the only transport a secret is substituted on.

Part by part:

- **Gating MITM behind the allowlist** would reverse a deliberate decision. `decideTCPInspectRoute` on main takes the MITM branch first precisely so a secret host needs no `allow_net` rule of its own. Restoring the gate re-creates the problem POL-483 filed.
- **Merging `Secret.hosts` into `allow_net`** is no longer needed for the stated use case. It would only add non-443 and UDP egress to a credential's host, which is more than declaring the credential implies.
- **`effective_allow_net` on `BoxInfo` and the four SDKs** reported the merged set; with no merge there is nothing extra to report.

The two defects this PR cited are also handled on main by other means. Domain fronting is no longer possible because a secret host is dialed by name and the guest's chosen address is discarded. IP/CIDR rules no longer die on :443 when a secret is configured, because `MatchesIP` is consulted on the non-secret path.

What does survive is documentation: neither the `secrets` section nor the `allow_net` section says that declaring a secret grants HTTPS egress to its host, or that port 80 to that host stays blocked. That is being handled separately.

PR #1481, opened 2026-09-11
fix(net): make allow_net wildcard rules actually resolve

ltstriker:
## Summary

A wildcard rule in `allow_net` has never worked. Its catch-all DNS record was installed with no address, so every subdomain under the wildcard resolved to nothing and the guest could not reach any of them — while the egress filter went on reporting them as allowed. This has been true since `allow_net` was introduced.

## Call graph

```text
Before
  buildAllowNetWithResolver      (· src/deps/libgvproxy-sys/gvproxy-bridge/dns_filter.go:48)
    ├─ {Regexp: ".*"}            (· dns_filter.go:76)   ← BUG (POL-563): no IP, and first in the
    │                                                      zone's record list
    │    └─ addLocalAnswers      (dnsHandler · gvisor-tap-vsock dns.go:57)
    │         └─ A: record.IP    — nil; matched everything, answered with nothing, returned
    │                              before the real record and before the zone sinkhole
    ├─ zones from a Go map       (· dns_filter.go:102)  ← BUG: emitted in map order, so a less
    │                                                      specific zone could swallow a query
    └─ exactIPs[key] = …         (· dns_filter.go:89)   ← BUG: overwritten, so two spellings of
                                                           one host left the pin behind DNS

After
  buildAllowNetWithResolver      (· dns_filter.go:48)
    ├─ resolved[…]               (· dns_filter.go:56)   — one resolution per canonical name
    ├─ {Regexp: ".*", IP: ip}    (· dns_filter.go:94)   — the base domain's own addresses, the
    │                                                      set AllowHostToIP pins the wildcard to
    ├─ inherited coverage        (· dns_filter.go:120)  — a zone under a wildcard keeps it
    ├─ records: exact first      (· dns_filter.go:158)  — a catch-all cannot mask an exact rule
    ├─ zones: most specific first(· dns_filter.go:190)  — deterministic zone selection
    └─ exactIPs/suffixIPs union  (· dns_filter.go:89)   — the pin permits whatever DNS serves
```

Refs POL-563

## Changes

- **The catch-all carries the base domain's addresses.** That is the same set `AllowHostToIP` already pins the wildcard to, so DNS and the egress pin agree instead of contradicting each other.
- **Zone selection is deterministic and most-specific-first.** The resolver answers from the first zone whose suffix matches, sinkhole included; with `com.` and `example.com.` both present, map order decided whether an allowed host resolved or was sinkholed.
- **Within a zone, exact records precede the catch-all**, so `*.x.com` cannot mask the resolution `api.x.com` pinned for itself.
- **A zone created beneath a wildcard inherits its coverage**, skipping a covering wildcard whose own lookup failed. Without this, adding one exact rule silently sinkholed every sibling the wildcard still allows.
- **Each canonical hostname resolves once, and the pins union.** `"x.com:443"` beside `"x.com"` resolved twice: both answers became zone records, the resolver serves the first, and the pin kept only the last — so a round-robin host resolved to an address `AllowHostToIP` then rejected.

## How to verify

```bash
make test:unit:gvproxy GOTEST_FILTER='-run TestBuildAllowNet'
```

Two-side verification — revert `dns_filter.go` alone, keep the tests:

| Reproducer | Pre-fix signal |
|---|---|
| `…_WildcardAloneResolvesSubdomains` | `foo.example.com resolved to "<nil>"` |
| `…_WildcardAndExactRulesShareAZone` | `api.example.com resolved to "<nil>"` |
| `…_DeeperZoneInheritsParentWildcard` | `a sibling still covered by *.example.test was sinkholed: got "<nil>"` |
| `…_NestedWildcardInheritanceIsDeterministic` | sibling inherits the wrong wildcard |
| `…_InheritanceSkipsAnUnresolvedWildcard` | sibling left on a bare sinkhole |
| `…_OverlappingZonesResolveTheMoreSpecificRule` | `api.example.com resolved to "0.0.0.0"` |
| `…_DuplicateHostResolvesOnce` | `resolved github.test 2 times, want 1` |

`…_BothSpellingsOfAHostStayReachable` covers the pin union specifically; reverting just the two `append`s back to assignment fails it with `DNS served "203.0.113.61" for GitHub.test but the pin permits [203.0.113.62]`.

The pre-existing `TestBuildAllowNetDNSZones_PerTLDZonesHaveSinkholeDefaultIP` fails on this machine on baseline too — it resolves `example.com` over real DNS, which this sandbox cannot reach.

## Risks / rollout

- **Behavior widens for wildcard users**, from "no subdomain resolves" to "subdomains resolve to the base domain's address". The egress filter already permitted these hosts, so this makes DNS agree with the policy rather than granting anything new.
- A wildcard whose base domain fails to resolve still yields no address — unchanged, and now diagnosable from the zone it leaves behind.
- Why the existing tests missed it: they asserted zone counts and sinkhole defaults but never resolved a name. The new tests serve the built zones and query them.

### DIFF
--- a/src/deps/libgvproxy-sys/gvproxy-bridge/dns_filter.go
+++ b/src/deps/libgvproxy-sys/gvproxy-bridge/dns_filter.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"net"
 	"regexp"
+	"sort"
 	"strings"
 
 	"github.com/containers/gvisor-tap-vsock/pkg/types"
@@ -48,6 +49,12 @@ func buildAllowNetWithResolver(allowNet []string, lookup func(context.Context, s
 	zoneRecords := make(map[string][]types.Record)
 	exactIPs := make(map[string][]net.IP)
 	suffixIPs := make(map[string][]net.IP)
+	// One resolution per canonical hostname. Two rules naming the same host
+	// ("x.com:443" beside "x.com", say) would otherwise be looked up twice: both answers become zone records,
+	// the resolver serves the first, and the pin map keeps only the last — so a
+	// round-robin host resolves to an address AllowHostToIP rejects. That is
+	// the coupling contract this type documents.
+	resolved := make(map[string]bool)
 
 	for _, rule := range allowNet {
 		rule = strings.TrimSpace(rule)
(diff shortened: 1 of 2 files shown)


ltstriker:
Re-verified the premise at the VM level — real box, real guest resolver (busybox `nslookup`), `allow_net=["*.httpbingo.org"]`, no secrets:

```
before (main)
  nslookup sub.httpbingo.org   exit=1   (no Name:/Address: line at all)
  nslookup httpbingo.org       exit=0   Address: 0.0.0.0

after
  nslookup sub.httpbingo.org   exit=0   Address: 66.241.125.232
  nslookup httpbingo.org       exit=0   Address: 0.0.0.0
```

Two refinements to the description:

**Symptom.** The `"<nil>"` in the table above is the Go test client rendering a zero-length A record. A real guest resolver does not read it as an empty address — it rejects the answer and the lookup simply fails. So the user-visible symptom is a failed name resolution, not a bogus `0.0.0.0`.

**The apex is out of scope, and that part of the original logic was always self-consistent.** `httpbingo.org` stays sinkholed before and after. `*.example.com` does not cover `example.com` in either layer:

- DNS: the zone is `example.com.` and suffix matching uses `.example.com.`; `HasSuffix("example.com.", ".example.com.")` is false, so the apex never enters the zone and falls to the root sinkhole.
- Filter: `MatchesHostname("example.com")` uses `HasSuffix("example.com", ".example.com")` — also false.

So "never worked" is precisely: the wildcard's coverage of **subdomains** never worked, because the two layers disagreed there. Its refusal of the **apex** was always correct, both layers agree, and this PR deliberately leaves it alone. Allowing an apex still needs its own exact rule.

ltstriker:
fix with a bug, not so good. fix with dynamic resolve

Issue #1482, opened 2026-09-11
`make setup` cannot reinstall Git hooks in linked worktrees


hangzws:
# `make setup` cannot reinstall Git hooks in linked worktrees

## Summary

Running `make setup` in a linked Git worktree can leave the repository with stale or missing pre-commit/pre-push hooks.

The setup script clears `core.hooksPath` from the regular local Git configuration, but linked worktrees may store it in worktree-specific configuration. In that case, `prek install` refuses to reinstall the hooks.

## Environment

- macOS or Linux
- Git linked worktree
- `extensions.worktreeConfig=true`
- BoxLite repository using `prek` and agent-tooling hooks

## Steps to reproduce

The two `git config` commands below emulate the worktree-local state normally
created by agent-tooling setup; users do not normally need to run them manually.

```bash
git worktree add ../boxlite-worktree <branch>
cd ../boxlite-worktree
make setup
```

Run `make setup` again after the hooks have been configured.

## Actual behavior

`prek` refuses to install the hooks because `core.hooksPath` is still set in the worktree configuration:

```text
Cowardly refusing to install hooks with core.hooksPath set
```

Because hook installation is best-effort, setup continues and may appear successful. Subsequent commits or pushes can then use stale hooks, or the expected lint/test hooks may not run.

## Expected behavior

`make setup` should be idempotent and reinstall the current pre-commit and pre-push hooks in every worktree.

## Root cause

The setup script only clears the repository-local configuration:

```bash
git config --unset core.hooksPath
```

It does not clear the worktree-local configuration created by:

```bash
git config --worktree core.hooksPath ...
```

## Proposed fix

Clear both configuration scopes before invoking `prek install`:

```bash
git config --local --unset-all core.hooksPath
git config --worktree --unset-all core.hooksPath
```

Add a regression test that verifies hook installation succeeds when `core.hooksPath` is configured at the worktree scope.

## Acceptance criteria

- `make setup` works in a linked worktree.
- Re-running `make setup` is idempotent.
- `prek install` no longer refuses because of a stale worktree-level `core.hooksPath`.
- The pre-commit and pre-push hooks are installed in the shared Git hooks directory.
- A regression test covers worktree-local hook configuration.

PR #1484, opened 2026-09-11
fix(setup): clear worktree-local core.hooksPath before prek install

ZxlDragonDoctor:
### Problem
`make setup` only cleared repository-local `core.hooksPath` before `prek install`. In a linked worktree with `extensions.worktreeConfig=true`, agent-tooling stores the redirect in worktree-local config, so `prek` refused to reinstall hooks (`Cowardly refusing to install hooks with core.hooksPath set`) and re-runs were not idempotent.

### Solution
Unset `core.hooksPath` from both `--local` and `--worktree` scopes before `prek install`, matching the issue's proposed fix. Add `scripts/test/test-setup-worktree-hookspath.sh` that seeds a worktree-local value, applies the same two git commands, asserts the value is gone, and asserts a second unset is a no-op.

### Testing
- Command: `bash scripts/test/test-setup-worktree-hookspath.sh`
- Result: `test-setup-worktree-hookspath: ok`
- Also: `bash -n scripts/setup/setup-common.sh` (syntax clean)

### Issue alignment
Addresses maintainer-proposed fix in #1482 (clear both config scopes; add regression test). Issue has 0 comments / no claimed assignee at PR time.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Git hook setup for linked worktrees by correctly clearing worktree-specific hook path settings.
  - Prevented stale hook configuration from interfering with the setup process.
  - Improved reliability when hook setup is run repeatedly, including when configuration is already cleared.

- **Tests**
  - Added regression coverage to verify worktree-local hook settings are cleared reliably and setup remains safe to run repeatedly.


DorianZheng:
Hi @ZxlDragonDoctor. Thanks for the contribution. Merged

PR #1504, opened 2026-09-16
fix(cli): disable auto-stop for foreground --rm

mamamiyear:
## Summary

Send `auto_stop=0` for foreground `boxlite run --rm` so `auto_delete=1` passes cloud lifecycle validation instead of conflicting with the default 900-second auto-stop interval. Preserve the existing `--detach --rm` behavior.

Refs [POL-587](https://linear.app/polygala/issue/POL-587/website-cli-quickstart-fails-with-auto-delete-interval-validation).

## Call graph

```text
Before
build_options (CLI · src/cli/src/commands/run.rs:90) — auto_delete=1, auto_stop unset
└─ create (RestRuntime · src/boxlite/src/rest/runtime.rs:208) — POST /boxes omits auto_stop
   └─ resolveLifecyclePolicy (BoxService · apps/api/src/box/services/box.service.ts:1471) — defaults auto_stop to 900 ← BUG: rejects auto_delete=1

After
build_options (CLI · src/cli/src/commands/run.rs:90) — foreground --rm sets auto_delete=1, auto_stop=0
└─ create (RestRuntime · src/boxlite/src/rest/runtime.rs:208) — POST /boxes includes auto_stop=0
   └─ resolveLifecyclePolicy (BoxService · apps/api/src/box/services/box.service.ts:1471) — accepts 1 > 0
```

## How to verify

- `git diff --check origin/main...HEAD` — passed.
- Rust formatting, build, and tests were not run: `cargo` is unavailable locally.

## Summary by CodeRabbit

- **Bug Fixes**
  - Foreground runs using automatic cleanup now stop and remove as expected.
  - Detached runs are no longer configured with foreground cleanup behavior.


DefeatMan (approved):
LGTM

PR #1507, opened 2026-09-16
feat(ssh): support explicit host keys and full guest SSH restarts

lfld287:
## Summary

Enable guest SSH through the existing guest gRPC control service with caller-provided host private keys and authentication using CA-signed certificates or authorized public keys.

This PR changes only the runtime layer, including guest SSH, its shared gRPC protocol, tests, and documentation. CLI, SDKs, and apps are unchanged.

## Call graph

```text
Before (main)
  configure (GuestServer · src/guest/src/service/ssh/control.rs:12)
    └─ configure (SshManager · src/guest/src/service/ssh/mod.rs:212)
         └─ prepare_listener (src/guest/src/service/ssh/mod.rs:504)
              — host identity unavailable; SSH cannot start

After
  configure (GuestServer · src/guest/src/service/ssh/control.rs:10)
    └─ configure (SshManager · src/guest/src/service/ssh/mod.rs:127)
         ├─ parse (SshConfig · src/guest/src/service/ssh/mod.rs:41)
         │    — validate authentication and host identity
         ├─ stop (SshManager · src/guest/src/service/ssh/mod.rs:187)
         │    — disconnect clients and drain SSH work
         └─ accept_loop (src/guest/src/service/ssh/mod.rs:264)
              — accept connections with the new configuration
```

## Changes

- Replace flat `SshConfigureRequest` fields with a nested `SshConfig`.
- Validate host private keys, including signing support and matching public/private key halves.
- Include the host public key and fingerprint in SSH status.
- Fully restart SSH on every valid configuration, including identical configurations: disconnect clients, drain execution and forwarding tasks, then bind a new listener.
- Keep SSH configuration and status in guest memory.
- Add SSH control documentation and unit/integration coverage.

## How to verify

Completed:

- `git diff --check main...HEAD` and `git diff --check origin/main...HEAD`.
- `make fmt:check:rust`.
- Pre-push test matrix: 1,094 core unit tests, 65 shared unit tests, and 99 REST tests passed.
- VM integration suite: 267 passed, 12 skipped, including a passing `guest_ssh_rpc_exec_pty_sftp_reconnect_and_vm_restart` test.

Exclusions:

- Known failing `rest::litebox::tests::ws_watchdog_fires_when_idle` was explicitly excluded; all other push checks remained enabled.
- `make test:unit:guest` was skipped on macOS because guest unit tests require Linux.

To verify on Linux:

```sh
make test:unit:guest
```

To rerun SSH integration coverage with a VM environment:

```sh
make test:integration:rust FILTER=guest_ssh
```

## Risks / rollout

- Compatibility: callers using the old Configure request must update to the nested `SshConfig` schema.
- SSH starts disabled and requires configuration after each VM restart.
- Every valid Configure disconnects existing SSH sessions. Invalid configuration leaves the current service running.
- Listener bind failure leaves SSH disabled without restoring the previous configuration.
- No new CLI or SDK control interface is introduced; control uses the existing guest gRPC connection.

## Summary by CodeRabbit

* **New Features**
  * Added guest SSH configuration and control through the guest control socket.
  * SSH now supports multiple named accounts with individual public keys or certificate authorities.
  * Added host-key reporting, status updates, enable/disable controls, public-key authentication, PTY sessions, SFTP, and port forwarding.
  * SSH sessions respect the configured execution user, environment, home directory, and permissions.
  * SSH configuration is held in memory and resets when the virtual machine restarts.

* **Documentation**
  * Documented SSH configuration, account requirements, authentication, status behavior, and lifecycle rules.


docs/guides/README.md
```diff
@@ -4,6 +4,10 @@
 
 - [AutoStop、AutoResume 与 AutoDelete](./auto-stop-resume.md)
 
+## Guest SSH
+
+See [Guest SSH control](ssh.md) for in-memory configuration through `box.sock`.
```
ltstriker, line 9: look at the login account in container, it should not be the fixed `root`.  Consider about the default user for container.  Key is diff between the auth account and exec account.

ps. original impl is used to debug, so root is acceptable, but not now.

src/shared/proto/boxlite/v1/service.proto
```diff
@@ -47,18 +47,10 @@ service Guest {
   rpc Thaw(ThawRequest) returns (ThawResponse);
 }
 
-// Live control for the embedded SSH listener. SSH is deliberately separate
-// from Guest.Init so credentials and listener policy can be rotated without
-// rebuilding or restarting the guest.
+// In-memory SSH control; requires successful Guest.Init.
 service Ssh {
-  // Enable SSH or replace policy for subsequent authentication attempts.
```
ltstriker, line 54: delete for what?

src/shared/proto/boxlite/v1/service.proto
```diff
@@ -102,6 +94,9 @@ message GuestInitRequest {
 
   // Network configuration (optional)
   NetworkInit network = 2;
+
+  reserved 3;
```
ltstriker: reserved for what？ seems just add a reserved here.

docs/guides/README.md
lfld287: Would it work if multiple authentication accounts can be configured (several login names), while the execution account stays a single one — always the container's default user?

src/shared/proto/boxlite/v1/service.proto
lfld287: Restored the RPC and status-field comments to match the current restart and cleanup behavior. Fixed in 26d4f628.

src/shared/proto/boxlite/v1/service.proto
lfld287: Removed the intermediate GuestInit reserved fields, restoring the PR baseline definitions. Fixed in 26d4f628.

docs/guides/README.md
lfld287: SSH now supports multiple accounts for auth and uses the container’s default user for execution.

ltstriker: ref the `exec` logic

src/guest/src/service/ssh/auth.rs
```diff
@@ -30,19 +40,118 @@ impl std::fmt::Display for AuthorizerError {
...
+
+impl SshAuthorizer {
+    pub(crate) fn new(config: &boxlite_shared::SshConfig) -> Result<Self, AuthorizerError> {
+        if config.accounts.is_empty() {
+            return Err(AuthorizerError::MissingAccounts);
+        }
+        let mut accounts = HashMap::new();
+        for account in &config.accounts {
+            if account.login.is_empty()
+                || account.login.len() > 128
```
ltstriker: do not use magic number in code.  maybe a const?

ltstriker: and why 128.

lfld287: I used 128 to match the existing certificate principal length limit on main.
This PR applies the same bound to configurable login names. Previously, main only allowed root as the login, so the login-length check itself is new.

src/shared/src/lib.rs
```diff
@@ -61,11 +58,102 @@ pub use generated::execution_server::{Execution, ExecutionServer};
...
+
 // All generated types
 pub use generated::*;
 
+impl std::fmt::Debug for generated::SshConfig {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("SshConfig")
+            .field("listen_address", &self.listen_address)
+            .field("host_private_key", &"[REDACTED]")
+            .field("accounts", &self.accounts)
```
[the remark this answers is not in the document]
lfld287: fixed

src/guest/src/service/ssh/server.rs
```diff
@@ -131,57 +141,31 @@ impl russh::server::Handler for SshConnection {
...
-        } else {
-            Auth::reject()
-        })
+        Ok(
+            if self.authorizer.has_account(user) && !self.tasks.is_cancelled() {
+                Auth::Accept
+            } else {
+                Auth::reject()
+            },
+        )
```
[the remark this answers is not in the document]
lfld287: fixed

src/guest/src/service/ssh/auth.rs
lfld287: fixed

ltstriker:
**[P1] SSH stdin backpressure prevents Disable and subsequent Configure from completing**

Reproduced on the current head, `1cae661a5d84437992ba7e7688541e56d0b1b9e7`, using a real Linux KVM VM, Alpine, and OpenSSH. Production code was unchanged; the reproducer was added to `src/boxlite/tests/guest_ssh.rs`.

Reproduction:

1. Configure SSH with an explicit host key and an authorized client key, then open an authenticated exec session running `echo $$ > /tmp/review-ssh-pid; echo ready; exec sleep 300`.
2. After receiving `ready`, write 16 MiB to the OpenSSH client's stdin. The command never reads stdin, and the upload remains blocked after 3 seconds.
3. Call the guest SSH `Disable` RPC, then `Configure` with the same valid configuration.

| Input | Disable | Subsequent Configure |
|---|---|---|
| 16 KiB control | Success, 1.004 s | Success, 8 ms |
| 16 MiB | `DeadlineExceeded`, 10.004 s | `DeadlineExceeded`, 10.011 s |

Both failures report `timed out waiting for SSH sessions to stop`. The SSH-launched process is still alive after both calls. Killing that process through the regular box exec API allows Configure and a fresh SSH command to succeed.

The blocking path is [`SshConnection::data` awaiting `bridge.stdin(...)`](https://github.com/boxlite-ai/boxlite/blob/1cae661a5d84437992ba7e7688541e56d0b1b9e7/src/guest/src/service/ssh/server.rs#L570-L585). When the execution's stdin pipe and forwarding queue fill, that callback cannot return. [Shutdown disconnects the transport and awaits the running session](https://github.com/boxlite-ai/boxlite/blob/1cae661a5d84437992ba7e7688541e56d0b1b9e7/src/guest/src/service/ssh/mod.rs#L349-L353), but process termination depends on channel cleanup or dropping the handler. [The generation drain consequently times out](https://github.com/boxlite-ai/boxlite/blob/1cae661a5d84437992ba7e7688541e56d0b1b9e7/src/guest/src/service/ssh/mod.rs#L187-L212); the retained task group also blocks the following Configure.

Please make blocked stdin/EOF forwarding cancellation-aware and ensure shutdown can terminate executions independently of handler completion. Add VM regression coverage for this case before merging: one blocked session currently prevents the documented full SSH restart.

Validation: `make test:integration:rust SETUP_DONE=1 FILTER=guest_ssh` ran 4 tests: the PR's 3 existing SSH tests passed, and the added backpressure regression failed at its final Disable/Configure success assertion. The test verified recovery and stopped the VM before failing.

lfld287:
> **[P1] SSH stdin backpressure prevents Disable and subsequent Configure from completing**
> 
> Reproduced on the current head, `1cae661a5d84437992ba7e7688541e56d0b1b9e7`, using a real Linux KVM VM, Alpine, and OpenSSH. Production code was unchanged; the reproducer was added to `src/boxlite/tests/guest_ssh.rs`.
> 
> Reproduction:
> 
> 1. Configure SSH with an explicit host key and an authorized client key, then open an authenticated exec session running `echo $$ > /tmp/review-ssh-pid; echo ready; exec sleep 300`.
> 2. After receiving `ready`, write 16 MiB to the OpenSSH client's stdin. The command never reads stdin, and the upload remains blocked after 3 seconds.
> 3. Call the guest SSH `Disable` RPC, then `Configure` with the same valid configuration.
> 
> Input	Disable	Subsequent Configure
> 16 KiB control	Success, 1.004 s	Success, 8 ms
> 16 MiB	`DeadlineExceeded`, 10.004 s	`DeadlineExceeded`, 10.011 s
> Both failures report `timed out waiting for SSH sessions to stop`. The SSH-launched process is still alive after both calls. Killing that process through the regular box exec API allows Configure and a fresh SSH command to succeed.
> 
> The blocking path is [`SshConnection::data` awaiting `bridge.stdin(...)`](https://github.com/boxlite-ai/boxlite/blob/1cae661a5d84437992ba7e7688541e56d0b1b9e7/src/guest/src/service/ssh/server.rs#L570-L585). When the execution's stdin pipe and forwarding queue fill, that callback cannot return. [Shutdown disconnects the transport and awaits the running session](https://github.com/boxlite-ai/boxlite/blob/1cae661a5d84437992ba7e7688541e56d0b1b9e7/src/guest/src/service/ssh/mod.rs#L349-L353), but process termination depends on channel cleanup or dropping the handler. [The generation drain consequently times out](https://github.com/boxlite-ai/boxlite/blob/1cae661a5d84437992ba7e7688541e56d0b1b9e7/src/guest/src/service/ssh/mod.rs#L187-L212); the retained task group also blocks the following Configure.
> 
> Please make blocked stdin/EOF forwarding cancellation-aware and ensure shutdown can terminate executions independently of handler completion. Add VM regression coverage for this case before merging: one blocked session currently prevents the documented full SSH restart.
> 
> Validation: `make test:integration:rust SETUP_DONE=1 FILTER=guest_ssh` ran 4 tests: the PR's 3 existing SSH tests passed, and the added backpressure regression failed at its final Disable/Configure success assertion. The test verified recovery and stopped the VM before failing.

fixed

lfld287:
• I’m aware of a few pre-existing issues that still need follow-up:

  - Late zygote results: Cancelling a launch before the zygote returns its PID may leave the resulting process without cleanup.
  - Process-group cleanup: Once the group leader has exited and been reaped, the identity check may prevent signals from reaching surviving descendants.
  - Full stdin queues: A handler blocked on stdin backpressure may delay processing a disconnect or channel close, preventing timely cleanup.
  - Reverse Unix forwarding shutdown: The helper may keep waiting for a half-closed connection to drain after service cancellation, potentially exceeding the SSH shutdown timeout. Service shutdown should explicitly terminate these relays while preserving half-close behavior during normal forwarding.

  These concerns predate this PR, and the exact failure cases still need reproduction. I’d prefer to verify and address them individually in separate follow-up PRs rather than expand this PR’s scope further.

ltstriker:
LGTM.

ltstriker (approved):

Issue #1491, opened 2026-09-14
Quickstart handoff mints two API keys when opened by URL on an empty account


Mandalorian-Wang:
Found in review of #1466 and deliberately not fixed there, so the PR could merge on its existing approval. This is the most serious of the deferred findings.

## What happens

On an account with no boxes, opening quickstart from the sidebar (`?onboarding=1`) mounts **two** `QuickstartAgentHandoff` instances at once:

- `apps/dashboard/src/pages/Boxes.tsx:711` — `requestedByUrl ||` opens `OnboardingGuideDialog`, whose `publish` scenario renders one
- `apps/dashboard/src/pages/Boxes.tsx:769` — `{firstRun && …}` renders the inline first-run quickstart, which renders another

The issuance lock is per-instance:

```
apps/dashboard/src/components/QuickstartAgentHandoff.tsx:144
  const issuingFor = useRef(new Set<string>())
…:183
  alreadyIssuingForKey: issuingFor.current.has(identity)
```

`handoffStore` is shared, but `issuingFor` is a `useRef`, so each instance holds its own. Both mount in the same render pass, both see an empty store and an empty lock, and both pass `shouldIssueKey`. Two `createApiKey` calls go out; the second `handoffStore.set` overwrites the first.

## Why it matters

The plaintext key value is returned exactly once. The overwritten key is live, scoped to boxes, valid for 7 days, and the user is never shown it and has no way to recover it — it simply sits in the account until it expires. #1466 exists to stop exactly this class of outcome, and this is the most common path a brand-new user takes into it.

## Fix sketch

Either render one handoff in this state (the dialog and the inline section are redundant on first run), or lift the issuance lock out of the component so it is shared the way `handoffStore` already is. The second is more robust — the first still leaves two instances able to collide if another surface ever renders one.

## Notes

`Boxes.tsx:711`'s `requestedByUrl ||` predates #1466 (it came from #1337, merged 2026-08-25). The collision is new because #1466 adds the second, inline instance.

PR #1515, opened 2026-09-17
refactor(dashboard): remove console search, and fix what it was hiding

Mandalorian-Wang:
Five changes to the console, each answering something that was wrong on screen. They ride together because they overlap in the same files.

## Boxes: an empty state, not a takeover

An account with no boxes had the Boxes page replaced wholesale by an inline copy of the quickstart's agent handoff — a second entrance to the guide the header already links, non-dismissible, offering one of the guide's two jobs with no way to reach the other. A page called Boxes said nothing about boxes, not even "you have none".

The table's own empty state does that job and now carries a button into the same dialog the header opens.

Mounting that inline handoff also minted an API key on arrival, before any click. That goes with it, and so does **#1491**: on `?onboarding=1` two handoff instances mounted together, each holding its own `useRef` issuance lock, and each minted a live key. One instance cannot race itself, so the bug is gone structurally rather than patched.

## Search: removed

Search covered one of the console's three resources. Boxes had a real one — debounced, server-filtered through `listBoxesPaginated`'s `name`. Volumes had only actions, and the command group was called "Volume actions" because that is what it was: `listVolumes` takes no name, and the implementation stopped where the API stopped. API keys had nothing.

A search that finds one kind of thing teaches people to stop pressing ⌘K, and nothing said where its edge was — `No results found for "subtitle"` reads as *there is no subtitle*, not as *I did not look for one*. Removed rather than completed, per product decision.

Two capabilities had no other entrance and are handled rather than dropped:

| Capability | Before | Now |
|---|---|---|
| Bulk-delete volumes | `SelectionToast`'s "Actions" opened the palette so the command could be found there | Deletes directly from the toast |
| Copy Organization ID | Palette only | Already on the organization settings page, with the ID visible — the palette entry was the redundant one |

`Create Volume` also existed as a command; `Volumes.tsx` has a visible **New Volume** button, so nothing is lost. Pylon's floating widget is untouched.

## Quickstart: say what it does

| | Before | After |
|---|---|---|
| Headline | Hand the job to your coding agent. | **Put your agent's app online.** |
| Prompt label | Prompt · key quickstart-77ff · boxes only · 7d · this tab only | **Copy this prompt to your agent** |

The old headline instructed the reader to do something that happens elsewhere, named a job the screen had not yet introduced, and left BoxLite out of its own console. Not "deploy", either: preview traffic keeps a running box alive but [cannot wake a stopped one](https://docs.boxlite.ai/cloud/box-lifecycle), and boxes stop after 15 idle minutes — "online" is the weaker word and the true one.

The old prompt label was a section heading followed by four telegraphic notes about a credential the reader never asked for, at the moment they were trying to copy something.

Selection on the two scenario cards was a plain rule, the one piece of that screen with nothing of the console's own language on it — corner marks now. The unchosen card's drawing goes `grayscale`: `text-muted-foreground` only reaches strokes drawn in `currentColor`, so the brand accent each drawing spends once stayed fully saturated on the card meant to recede.

## Billing: two money histories on tabs

Billing history and credit activity were stacked, 25 rows a page each — roughly 1100px — so whichever came first pushed the other off screen with its pager. Credit activity was underneath: last on the page, folded, and with no count, because the query producing the count was gated on the fold. Nothing could suggest opening it.

The fold guarded something real — a credit grant is quota, not a payment, and grants beside charges at equal weight is what once let a prorated upgrade's granted quota read as the amount billed for it.

Tabs serve both. Each label carries its size, so neither history must be opened to be sized up, and only one table is ever on screen — a stronger separation than a fold, which could be opened into exactly the side-by-side it existed to prevent. Both tables are bounded at 420px with headers stuck inside the scroller, so a pager stays reachable and no header drifts out of alignment by a scrollbar's width.

## Mock preview

`resolveEnvironment` reads the OIDC issuer, and mock's was a placeholder, so on localhost the quickstart fell through to `config.apiUrl` — a host `start-dashboard.mjs` injects purely as an MSW matching prefix. The preview advertised a different endpoint than production, which reads as "the change did not take". Mock identifies as production now and resolves every environment-derived value the way production will.

## Call graph

```text
Before
  Boxes                          (page      · apps/dashboard/src/pages/Boxes.tsx:672)
    ├─ firstRun                  (branch    · Boxes.tsx:769)                    ← BUG: replaced the list with a second copy of the guide
    │    └─ QuickstartAgentHandoff                                              ← BUG: minted a key on mount; a second instance minted another (#1491)
    └─ BoxTable                  (component · BoxTable/index.tsx)               — hidden while firstRun
  Sidebar                        (component · Sidebar.tsx:259)
    └─ Search → CommandPalette   (component · CommandPalette.tsx:297)           ← BUG: found boxes only; said "no results" for everything else
  WalletSection                  (component · billing/WalletSection.tsx)
    ├─ BillingHistorySection     (component · :127)                             — 25 rows, unbounded
    └─ WalletTransactionsSection (component · :158)                             ← BUG: folded, last, and countless — its count came from a query the fold gated

After
  Boxes                          (page      · Boxes.tsx)
    └─ BoxTable                  (component · BoxTable/index.tsx:273)           — always rendered
         └─ empty state          (render    · BoxTable/index.tsx:276)           — "No boxes yet." + Open Quickstart → the header's own dialog
  Sidebar                        (component · Sidebar.tsx)                      — no palette, no ⌘K
  SelectionToast                 (component · SelectionToast.tsx:42)            — deletes directly, no palette hop
  MoneyHistorySection            (component · billing/WalletSection.tsx:118)
    ├─ tab: Billing history 149  (panel)                                        — bounded 420px, header sticky inside
    └─ tab: Credit activity 146  (panel)                                        — same, and never on screen beside the charges
```

## Verification

- `vitest run` — 49 files / 320 tests pass
- eslint reports 9 errors, the same 9 as `main`; none new. tsc clean on the touched files
- every claim above checked in the running console (`dashboard-mock`): the empty state and its button, `?onboarding=1` mounting exactly one handoff, ⌘K doing nothing and the header button gone, both billing tabs at 420px with 1172px of content scrolling inside, and the sticky header holding position at `scrollTop: 600`
- the three billing tests that encoded the fold were rewritten to the tab model rather than deleted, and now assert something stricter than before — that a grant and a charge are never on screen together

## Note on the endpoint

`REST_API_URL_BY_ENV` already dropped its `/api` on main in #1453, with `development` pointing at `api.dev.boxlite.ai`. I had changed it to `dev.boxlite.ai`, which is wrong — that host answers `/v1/boxes` with 200 and no auth because it serves the dashboard SPA, not the API. Main's values are kept.

## Summary by CodeRabbit

- **New Features**
  - Added an “Open Quickstart” action to empty box lists.
  - Billing history and credit activity are now available in separate tabs with counts and pagination.
  - Bulk volume deletion can be initiated directly from the selection action.

- **Improvements**
  - Invoice and wallet transaction tables now scroll within a fixed-height area with sticky headers.
  - Updated onboarding scenario visuals and clearer prompt instructions.

- **Removed**
  - Removed the dashboard command palette and its search, navigation, and support actions.


DefeatMan (approved):

Mandalorian-Wang:
Superseded by #1588 — same commit `26f3fd7`, same diff byte for byte, only the head branch moved from my fork into this repository.

This one could not go green. The diff touches only `apps/dashboard/**`, which produces no coverage, so the required `codecov/patch` check can only be satisfied by `test.yml`'s `coverage-empty` job (`codecov empty-upload`). That job authenticates with `use_oidc: true`, and a pull request from a fork is not granted `id-token: write`:

```
-> Token length: 0
Empty Upload failed: {"message":"Token required because branch is protected"}
```

which then failed `Test (conclusion)`. Merge queue entry is gated on the pull request's own required checks, so `merge_group` could not resolve it either.

From a branch in this repository the identical job succeeds — `Token length: 1995`, `Empty Upload complete` — so #1588 is unblocked.

Worth noting separately: `apps/dashboard/**` appears in none of `test.yml`'s path filters, so the dashboard's own vitest suites never run on a dashboard-only pull request.

PR #1523, opened 2026-09-17
fix(infra): pin refs to their branch, and drop a deploy's spare approval

DefeatMan:
## Call graph

```text
Before
  Actions ▸ mdeploy-all (stage dev, ref <sha or tag>)                      — dispatched by a maintainer
  └─ resolve            (.github/actions/resolve-ref/action.yml:53)        ← BUG: a 40-character string is taken on trust, and no branch is asked
  ├─ plan               (mdeploy-all.yml · environment <stage>)            — prompt 1
  ├─ build / promote    (mbuild.yml, mrunner.yml · environment <stage>)    — prompt 2
  ├─ preflight          (mdeploy.yml:105 · environment <stage>)            ← BUG: a second job on the same Environment, so a second approval to run checks that change nothing
  └─ deploy             (mdeploy.yml:248 · environment <stage>)            — prompt 4
       └─ pulumiDeploy  (apps/infra/mdeploy/src/pulumi.ts:158)             ← BUG: an interrupted run leaves pending operations and no intent clears them

After
  Actions ▸ mdeploy-all (stage dev, ref <sha or tag>)                      — dispatched by a maintainer
  └─ resolve            (.github/actions/resolve-ref/action.yml:56)        — the commit must exist here and the dispatched branch must contain it
  ├─ plan               (mdeploy-all.yml · environment <stage>)            — prompt 1
  ├─ build / promote    (mbuild.yml, mrunner.yml · environment <stage>)    — prompt 2
  └─ deploy             (mdeploy.yml:105 · environment <stage>)            — prompt 3; guarded by mdeploy/test/workflow.test.ts
       ├─ the checks preflight used to make                                — every one still before the apply it guards
       └─ pulumiDeploy  (apps/infra/mdeploy/src/pulumi.ts:158)             — `mdeploy --refresh` clears what an interrupted run left
```

## Why

- **A ref was taken on trust.** `resolve-ref` matched a 40-character shape and made it the deploy's commit without asking whether this repository had ever held it, so a typo travelled as far as the registry, which answered that the images were missing — true, and about the wrong commit. Existing is not enough either: a commit on an abandoned or force-pushed branch still resolves and still names bytes, and deploying it puts a stage on something no branch here will produce again. Every deploy-path ref is now resolved against the branch it was dispatched from.
- **That makes dev off-main safe to allow.** dev is shaken out from the branch that is changing it; every stage that reaches a protected Environment still runs from main alone. The guard is an allow-list naming dev rather than a deny-list naming prod, so a stage added to the choice list later is main-only until an edit says otherwise.
- **`preflight` cost an approval to perform reads.** It bound the same Environment the apply binds, so a dispatch waited on the stage's reviewers twice — the second time to answer which commit the refs name, whether the stage holds those images, whether its configuration matches its own fingerprint, and whether a protected stage was confirmed. Merging it into the apply removes that wait and the duplication under it: one checkout, one `setup-infra`, one federation.
- **An interrupted run could not be repaired from here.** Pulumi records an operation before starting it, so a killed driver leaves the record and every later run opens with `pending operations from previous deployment`. The engine's advice is a refresh, and taking it meant a script outside the repository holding the backend URL and the passphrase by hand. `mdeploy --refresh` is that intent, on both engines.
- **Rejected: a separate unreviewed Environment with a read-only identity.** It reaches two prompts instead of three, and it was built and then withdrawn. Every value those jobs read is scoped to an Environment — `BOXLITE_MSTAGE_<APP>_CONFIG` is written per stage by `mstage config put` — so a second Environment has to carry a copy of each, and the copy goes stale the next time an operator updates the original. A stale stage configuration is worse than a missing one. This keeps three prompts and nothing to keep in sync.

## User-facing change

A maintainer dispatching `mdeploy-all` approves three times rather than four, and can dispatch dev from a feature branch. A ref that the dispatched branch does not contain is refused before anything is built.

## Verification

- `npm test` in `apps/infra`: 1315 pass, 0 fail. `npm run typecheck`: clean.
- Every guard demonstrated both ways, each run: the branch-pin guard fails naming the call site when `branch` is dropped; the prod invariant fails when a guard is written as `!= 'prod'`; the job-count guard fails when a second job is split back out; `--refresh` fails when wired to `preview`.
- `resolve-ref`'s script was exercised against this repository: a commit on main passes, a commit on a feature branch is refused against `main`, a well-formed but absent SHA is refused (it used to pass straight through), and no branch input skips the check.
- The merge dropped `Install gcloud` with the deploy half's duplicated federation, and the existing test for it caught that before the commit; it is restored above the checks, which shell out to the same CLI.
- Not exercised by a real dispatch. The four workflows parse, and the job and step ordering are asserted from source, but a merged job's behaviour is what the next dispatch shows.


DefeatMan:
Superseded by #1570, which carries these three commits unchanged (`git range-diff` reports them identical in patch and message) alongside the rollout-path consolidation that builds on them.

PR #1528, opened 2026-09-18
feat(box): report the main command exit code on box info

silverzki:
> **Updated after review.** The first version stored the code in a control-plane
> column. @DorianZheng asked why it needed persisting at all, and the answer was
> that it should not — the runner is asked on demand instead. The second commit
> is that rework; the discussion is in the review threads below.

## Call graph

```text
Before
  GET /v1/boxes/{id}
  └─ boxToBoxResponse (box-to-box.mapper.ts:18)     — nothing to map: no FFI field,
                                                      no runner field, no DTO field

After
  GET /v1/boxes/{id}
  └─ BoxService.toBoxDtoWithExitCode (box.service.ts:1120)   — read path only
     └─ BoxExitCodeService.getExitCode (box-exit-code.service.ts:66)
        └─ GET {runner}/boxes/{id} (box.go:246)              — BoxInfoResponse.exitCode
           └─ Client.GetBoxInfo (client.go:452)              — state and code, one snapshot
              └─ BoxInfo.ExitCode *int  ←  CBoxInfo.exit_code (int *, null = absent)
                 └─ BoxState.exit_code  — recorded by the runtime, already persisted
     └─ boxToBoxResponse (box-to-box.mapper.ts:35)           — publishes exit_code
```

## Why

A box that stopped because its main command exited and one whose command crashed
both end up as `status: stopped`, with nothing in `GET /v1/boxes/{id}` telling
them apart. For an agent runtime the box *is* the process, so that is the whole
outcome of the workload — and it is unreadable on Cloud today.

This is not a missing capability. The core has recorded the code all along:
`record_main_command_exit()` (`src/boxlite/src/runtime/rt_impl.rs:51`) reads the
guest's exit file into `BoxState.exit_code`, local `boxlite serve` publishes it
on the same REST shape, and the Rust REST client already parses it with a
compatibility branch for servers that omit it. Only the cloud path never carried
it — the C FFI and Go SDK had no field, so the runner could not read one.

A same-shape comparison isolates it: identical bare-REST sequence, identical
`cmd`, only the base URL differs.

| Target | Main command | `exit_code` |
| --- | --- | --- |
| local `boxlite serve` 0.10.2 | `exit 0` / `exit 42` | `0` / `42` |
| Cloud | same | field absent |

## Design decisions

- **The runner answers; we keep no copy.** A copy can only ever be as good as
  the one delivery that fills it, and that delivery is conditional: the runner
  reports box state only when its view differs from ours, so a box whose state
  we already record as `stopped` never reports its exit code at all — the
  condition POL-606 describes. The runtime keeps the code in the box's own
  record and can answer for as long as the box lives there.
- **The read sits on a tenant's read of one box, and nowhere else.** The
  conversion it would otherwise share is also the event path —
  `NotificationService` converts through it for every state update, as does
  every `BoxStateWaiterService` resolution — and a box reaching `stopped` is
  what fires those. Reading there would put a runner round trip in front of
  every stop notification and stall it for the timeout precisely when the
  runner is what failed. Two tests hold that line.
- **Listing boxes omits the field.** It would fan out to one runner call per
  box. It is optional in the spec, so omitting it is conformant, and the schema
  now says absence in a list means nothing either way.
- **The value is how the command ended, not whether it ended by itself.**
  Stopping a box signals the main command — `SIGTERM`, then `SIGKILL` after a
  grace period — and the guest records the result the same way it records a
  self-chosen exit. Verified against a local `boxlite serve`: a command
  trapping `SIGTERM` and exiting `7` reads back as `7`, while
  `sh -c 'sleep 600'` reads back as `137`. Those are the only two shapes, since
  the main command is the container's init and an unhandled `SIGTERM` to PID 1
  is dropped. Suppressing the code for a stop we asked for was rejected: the
  answer would then depend on the control plane's belief about why a box
  stopped, which POL-606 shows can be wrong, and the failure mode is hiding a
  real kill — an OOM also reports `137`.
- **`0` is a value, never a sentinel.** `CBoxInfo` spells absence as `0` for
  `pid` and `started_at`; an exit code cannot, since `0` is exactly what
  separates a command that succeeded from one that did not. The FFI carries an
  owned `int *`, null for absent, freed by `free_box_info` — the shape
  `CBoxInfo.network` already uses. A pointer beats a value plus a flag because
  it leaves nothing to misread.
- **Clearing is the runtime's own rule.** `init_live_state` clears `exit_code`
  when a box starts, the way docker clears `ExitCode`
  (`src/boxlite/src/litebox/box_impl.rs:1282`). The stored-copy design had to
  reimplement that across all three writers of box state; this one inherits it.

**Two costs, both deliberate.** An unreachable runner reads as "no exit code
recorded" — the read degrades to absent rather than failing the box response
around it. And a box that has migrated has none at all, because migration
recreates it with a fresh state and destroys the source; that is a pre-existing
gap which loses `started_at` and `error_reason` the same way, filed as #1551.

**Rejected: `pid`.** The same comparison showed Cloud also omits it. It is the
VMM subprocess's id on the runner *host* — a tenant cannot act on it, and the
spec already leaves it out of `Box.required`, so omitting it is conformant.
An exit code describes the tenant's workload; a host PID describes ours.

## User-facing change

`GET /v1/boxes/{id}` and `GET /api/box/{id}` gain an optional `exit_code` /
`exitCode`: how the main command ended, present once the box has stopped and
the runtime recorded it, absent otherwise. `0` is a real value, so clients must
distinguish the missing field from a zero one. Listing boxes does not carry it.
Adding an optional field is backward compatible, and the Rust SDK already
tolerates servers that omit it.

No database change. The reverted first version's column and pre-deploy migration
net out to nothing — neither exists on `main`, and nothing is orphaned.

## Verification

Each test was watched failing with the change it guards removed, then passing
with it restored:

- **FFI encode/decode** — `box_info_carries_exit_code_as_owned_pointer_null_when_absent`
  (Rust) and `TestCBoxInfoToGoCarriesTheMainCommandExitCode` (Go). `Some(0)` is
  what separates a pointer from a sentinel, so it is the case that fails first
  if absence ever becomes a value again.
- **Runner wire shape** — `TestBoxInfoResponseDistinguishesZeroExitCodeFromNone`:
  `0` reaches the wire as a value, "not recorded" leaves the key out.
- **API read** — `box-exit-code.service.spec.ts`: `0` and `137` returned, none
  returned as absent, an unreachable runner degraded to absent, and no call made
  at all for a box that cannot hold one. Replacing `?? undefined` with a
  truthiness check fails exactly the clean-exit case and nothing else.
- **The split** — `box.service.dto.spec.ts`: the event path makes no runner
  call and the read path does, so moving the call back into the shared
  conversion fails the suite rather than quietly slowing every stop notification.
- **Publication** — `box-to-box.mapper.spec.ts` pins `0` as a value against
  `null`/`undefined` as absent, and `box.dto.spec.ts` pins `0` against nothing
  at all. Passing the value through instead of testing it is what the second
  one holds: a truthiness check there fails both clean-exit cases.

Suites run locally on macOS arm64:

| | |
| --- | --- |
| cbindgen-regenerated `boxlite.h` | matches the committed header |
| `boxlite-c` | 90 pass |
| Go SDK · runner (`./pkg/...`) | pass · pass |
| API unit suites | 99 suites / 861 tests |
| `tsc` app + spec projects | clean |
| `gofmt` | clean |

**End to end, against a local control plane** — real runner, real API, real
Postgres, box image `ghcr.io/boxlite-ai/boxlite-agent-base:v0.1.0`:

```text
[exit 0]   15s after stop   status=stopped   exit_code=0    ✅
[exit 42]  15s after stop   status=stopped   exit_code=42   ✅
[restart]                   status=running   (field absent) ✅
```

Confirmed in the same run that the control plane's `box` table has no `exitCode`
column, so those values came from the runner on each read. Re-run after every
change to the read path rather than carried over.

<details>
<summary>Substituted dependencies, and why none weakens the result</summary>

The subject under test — runner, API, their contract, Postgres — ran for real.
Four dependencies around it were substituted, none of which the exit code
passes through:

- **OIDC** — a discovery + JWKS stub, because the API refuses to boot without a
  reachable issuer. The test authenticates with an API key, which does not touch
  OIDC.
- **S3** — disabled through the code's own `createVolumeBucketStore` opt-out. No
  volumes are used.
- **Org, key and runner rows** — seeded directly, rather than through the admin
  endpoints that need a feature flag. They are inputs to the test, not the
  subject of it.
- **`BOXLITE_RUNTIME_DIR`** — pointed at runtime artifacts from a non-stub
  build, since the library was built with `BOXLITE_DEPS_STUB=1`.
</details>

## Note on the diff

Regenerating the runner's client to pick up the field brings in drift the
committed copy had accumulated: a `secrets` DTO that had landed without a
regeneration, a manifest line the current generator no longer writes, and two
reordered definitions in the swagger. None of it belongs to this change, but a
generator rewrites whole files, so trimming it back by hand would only make the
drift check fail.

Closes POL-603.

## Summary by CodeRabbit

* **New Features**
  * Single-box details now include the main command’s optional exit code after termination.
  * Exit codes preserve successful completion (`0`) and signal-based termination values.
  * Exit-code information is available across REST, Go, Node.js, Python, and C interfaces.
  * Running boxes and list or state-change responses omit unavailable exit codes.

* **Documentation**
  * API and SDK documentation now describe exit-code behavior, signal handling, and absent values.

* **Tests**
  * Added coverage for successful, failed, signal-terminated, and unavailable exit codes.


silverzki:
## End-to-end on a local control plane

The verification in the description covers each layer in isolation — the runner
tests stub the API with `httptest`, the API tests mock the runner. Both sides
being right is not the same as the seam between them being right, and that seam
was the one thing left unproven.

So I stood the control plane up locally (`apps/infra-local`, L2 layer) and ran
the whole chain for real:

```
02:21:19  [exit 0]   9s  after stop  status=stopped  exit_code=0    ✅
02:21:29  [exit 42]  6s  after stop  status=stopped  exit_code=42   ✅
02:21:32  [restart]                  status=running  (field absent) ✅ cleared
```

Every hop was the real thing — a runner binary linked against the new
`libboxlite.a`, booting a real microVM whose main command actually exited,
reporting over HTTP to the API from this branch, persisting into a real
Postgres 16 with the migration applied through the normal startup path, and
read back out of `GET /v1/boxes/{id}`.

That closes the three items the description had listed as "confirm after
release":

- `exit 0` publishes `exit_code: 0` — a clean exit is not swallowed as absent
- `exit 42` publishes `exit_code: 42`
- **restarting clears it** — the case that depends on all three writers of box
  state agreeing, and the one I was least willing to take on unit tests alone

Repro script: `pol603-e2e-local-stack.mjs` (bare REST, no SDK).

<details>
<summary>What had to be stubbed, and why none of it weakens the result</summary>

The L1 layer (12 BoxLite boxes: postgres, redis, minio, dex, …) could not come
up here — every one of those images is on Docker Hub and anonymous pulls are
refused in this environment. L1 only supplies dependencies, so I substituted
them and left the subject under test untouched:

| Dependency | Substituted with | Bearing on the result |
| --- | --- | --- |
| Postgres / Redis | already-running native instances | none — same versions, migration applied the same way |
| Dex (OIDC; the API hard-fails boot without discovery) | a minimal discovery document on `:25556` | none — the whole run authenticates with an API key; no JWT path is reached |
| MinIO (S3; also fatal at boot) | `S3_ENDPOINT` emptied, so `createVolumeBucketStore` returns nothing and the probe is skipped | none — no volumes involved. This is the code's own opt-out, not a bypass |
| tenant API key | minted locally (`sha256(value)` into `api_key`) | none — local dev database |

The runner, the API, the database and the REST surface — everything the change
actually touches — ran unmodified from this branch.
</details>

apps/api/src/migrations/pre-deploy/1787100000000-add-box-exit-code-migration.ts
```diff
@@ -0,0 +1,13 @@
+import { MigrationInterface, QueryRunner } from 'typeorm'
+
+export class AddBoxExitCode1787100000000 implements MigrationInterface {
+  name = 'AddBoxExitCode1787100000000'
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "box" ADD "exitCode" integer`)
```
DorianZheng: No need to persist it

sdks/c/include/boxlite.h
```diff
@@ -388,6 +388,17 @@ typedef struct CBoxInfo {
...
   // is always the case for local runtimes.
   int64_t last_activity_at;
+  // The main command's exit code. Read it only when
+  // [`Self::has_exit_code`] is nonzero.
+  //
+  // Absence cannot be a sentinel the way it is for [`Self::pid`] and
+  // [`Self::started_at`]: `0` is the exit code of every command that
+  // succeeded, so the flag below is the only thing separating "exited
+  // cleanly" from "no exit code recorded".
+  int exit_code;
```
DorianZheng: how about define it as a pointer so no `has_exit_code` needed

DorianZheng (dismissed):
left some comments. rest LGTM

apps/api/src/migrations/pre-deploy/1787100000000-add-box-exit-code-migration.ts
silverzki: So the real choice is storing it versus asking the runner on demand:

|  | column (this PR) | ask the runner |
| --- | --- | --- |
| read path | one local row | a runner hop on every `GET /v1/boxes/{id}` |
| runner replaced or unreachable | unaffected | no answer |
| box already reaped on the runner | still answers | no answer |
| schema | one nullable column | none |

openapi/box.openapi.yaml
```diff
@@ -1791,6 +1791,19 @@ components:
           type: boolean
           default: true
           description: Whether the box automatically resumes when accessed after AutoStop
+        exit_code:
+          type: integer
+          description: |
+            Exit code of the box's main command, present once the box has
+            stopped because that command exited. Absent while the box is
+            running, and when it stopped for any other reason — a stop
+            request, AutoStop, or a failure before the command ran.
```
zombee0: FYR, Why not record exit code when box is stopped manually or automatically？

apps/api/src/migrations/pre-deploy/1787100000000-add-box-exit-code-migration.ts
silverzki: Going with this — the column comes out and the API reads the code from the runner instead.

What changed my mind is a hole I found after writing the comparison above. `UpdateBoxState` has exactly one call site (`box_sync.go:181`), reached only when `local.state != convertedRemoteState` (`box_sync.go:210`). So if the control plane's state is already `stopped` when the main command exits — which is the condition POL-606 reports, a box serving traffic while the API shows `stopped` — the states match, nothing is reported, and the exit code never leaves the runner at all. A stored copy cannot be more reliable than the one delivery that fills it, and reading from the runner does not depend on that delivery.

Two things worth flagging, both consequences of the pull design rather than objections to it:

- **The list endpoint will not carry `exit_code`.** `Box` is shared by `GET /boxes/{box_id}` and `ListBoxesResponse`, and pulling per box would make a list N runner round trips. The field is optional, so omitting it there is conformant; I'll say so in the schema description. A batch box-info endpoint on the runner would close that later.
- **When the runner is unreachable the field is absent**, which by our own contract reads as "no exit code recorded". I'd rather degrade that way than fail the whole box read, but it is a wrong answer in that window, not a missing one.

Separately, migration drops the box's whole `BoxState` — `exit_code`, `started_at`, `error_reason` all reset, and the source copy is deleted — so the pull path returns nothing for a box that has moved. Filed as #1551; it is worth fixing on its own account, since `started_at` and `error_reason` are lost the same way today.

sdks/c/include/boxlite.h
silverzki: Adopting this. It matches what `CBoxInfo` already does for `network` — an owned pointer, null for absent, released in `free_box_info` — and it makes the misuse impossible rather than merely documented: with a value plus a flag, a caller that ignores the flag reads a plausible `0`.

Cost is one allocation per box in `CBoxInfoList`, which is worth it for a field whose whole point is that `0` and absent are different answers.

silverzki:
@DorianZheng both taken — pushed as a second commit rather than an amend, so
what changed since your approval is readable on its own.

**No stored copy.** The column, its pre-deploy migration, the three writers that
had to clear it and the sync report that fed it are all gone. `GET /v1/boxes/{id}`
asks the box's runner instead. What decided it was not the column's cost: a copy
can only ever be as good as the one delivery that fills it, and `UpdateBoxState`
fires only when the runner's state differs from ours — so a box we already record
as `stopped` never reports its exit code at all, which is exactly the state
POL-606 describes. Reading on demand does not depend on that delivery. As a
bonus, the "clear it when a box starts" rule is now the runtime's own
(`box_impl.rs:1282`, mirroring docker) instead of being reimplemented across
three writers.

**Pointer instead of the flag.** `CBoxInfo.exit_code` is an owned `int *`, null
for absent, freed in `free_box_info` — the shape `network` already uses.

Two things worth your eye, both consequences of the pull design:

1. **The call is on a tenant's read of one box only.** My first pass put it in
   `toBoxDto`, which is also the event path — `NotificationService` converts
   through it for every state update, and a box reaching `stopped` is what fires
   that. That would have put a runner round trip in front of every stop
   notification, and stalled it for the timeout exactly when the runner is what
   failed. It now lives in `toBoxDtoWithExitCode`, used by the two read
   endpoints, with tests on both sides of the line.
2. **Listing boxes omits the field**, since a list would fan out to one runner
   call per box. It is optional in the spec, so that is conformant, and the
   schema now says absence in a list means nothing either way.

Two costs come with dropping the copy, both deliberate and documented: an
unreachable runner reads as "no exit code recorded", and a migrated box has none
at all — migration recreates the box with a fresh state and destroys the source,
which loses `started_at` and `error_reason` the same way. That one is
pre-existing and filed as #1551.

Verified end to end against a local control plane (real runner, real API, real
Postgres): `exit 0` reads back `0`, `exit 42` reads back `42`, and a restarted
box reports none — with the `box` table confirmed to have no `exitCode` column,
so each value came from the runner on the read.

silverzki:
> @DorianZheng both taken — pushed as a second commit rather than an amend, so what changed since your approval is readable on its own.
> 
> **No stored copy.** The column, its pre-deploy migration, the three writers that had to clear it and the sync report that fed it are all gone. `GET /v1/boxes/{id}` asks the box's runner instead. What decided it was not the column's cost: a copy can only ever be as good as the one delivery that fills it, and `UpdateBoxState` fires only when the runner's state differs from ours — so a box we already record as `stopped` never reports its exit code at all, which is exactly the state POL-606 describes. Reading on demand does not depend on that delivery. As a bonus, the "clear it when a box starts" rule is now the runtime's own (`box_impl.rs:1282`, mirroring docker) instead of being reimplemented across three writers.
> 
> **Pointer instead of the flag.** `CBoxInfo.exit_code` is an owned `int *`, null for absent, freed in `free_box_info` — the shape `network` already uses.
> 
> Two things worth your eye, both consequences of the pull design:
> 
> 1. **The call is on a tenant's read of one box only.** My first pass put it in
>    `toBoxDto`, which is also the event path — `NotificationService` converts
>    through it for every state update, and a box reaching `stopped` is what fires
>    that. That would have put a runner round trip in front of every stop
>    notification, and stalled it for the timeout exactly when the runner is what
>    failed. It now lives in `toBoxDtoWithExitCode`, used by the two read
>    endpoints, with tests on both sides of the line.
> 2. **Listing boxes omits the field**, since a list would fan out to one runner
>    call per box. It is optional in the spec, so that is conformant, and the
>    schema now says absence in a list means nothing either way.
> 
> Two costs come with dropping the copy, both deliberate and documented: an unreachable runner reads as "no exit code recorded", and a migrated box has none at all — migration recreates the box with a fresh state and destroys the source, which loses `started_at` and `error_reason` the same way. That one is pre-existing and filed as #1551.
> 
> Verified end to end against a local control plane (real runner, real API, real Postgres): `exit 0` reads back `0`, `exit 42` reads back `42`, and a restarted box reports none — with the `box` table confirmed to have no `exitCode` column, so each value came from the runner on the read.

openapi/box.openapi.yaml
silverzki: Good catch — it does record one, and the description was wrong. Thanks.

Stopping a box signals its main command: `SIGTERM`, then `SIGKILL` after a
grace period (`src/guest/src/container/lifecycle.rs`). The guest records the
result the same way it records a self-chosen exit, folding a signal into
`128 + n` through `ExitStatus::shell_code()`. Nothing special happens on a
stop; only my prose claimed otherwise.

Your question also caught me over-correcting. My first fix said a stop always
reads back as `137`, which is only true when the command ignores `SIGTERM`.
Both shapes, against a local `boxlite serve`:

```text
cmd = sh -c 'trap "exit 7" TERM; while :; do sleep 1; done'
  POST /stop  ->  stopped  exit_code=7      # handled the signal

cmd = sh -c 'sleep 600'
  POST /stop  ->  stopped  exit_code=137    # ignores SIGTERM as PID 1,
                                            # killed after the grace period
```

So the field is "how the main command ended", not "did it exit by itself",
and `137` is not a fixed answer for a stop. Every copy of the old claim now
says that — the spec, both API schemas, the runner's response and model
fields, the C/Go/Node/Python field docs, the two SDK references, and two test
names that called `137` a crash.

I did consider suppressing the code for a stop we asked for, and decided
against it. The answer would then depend on the control plane's belief about
*why* a box stopped, and POL-606 is a live example of that belief being wrong;
the failure mode would be hiding a real kill — an OOM also reports `137` — to
remove some expected ones. It would also diverge from docker, which this field
otherwise follows: `docker stop` leaves `ExitCode: 137`, and the runtime
already clears the code on start "the way docker clears `ExitCode`"
(`src/boxlite/src/litebox/box_impl.rs:1282`).

A caller that wants "did my workload fail" has what it needs to disambiguate:
it knows whether it requested the stop, and `auto_stop` is its own setting.

silverzki:
@DorianZheng ready for another look when you have a moment.

Your approval was dismissed by the two follow-up commits (`af4ffb93`, `c87a08c0`), and as an outside contributor I can't re-request a review through the API — hence the ping.

Nothing new since my comment above: the diff to re-read is the pull-design rework (no stored copy; `GET /v1/boxes/{id}` asks the runner) plus the `*mut c_int` pointer change, both described there.

On CI, in case it looks blocking: `Update author review status` is showing a stale failure from an earlier `pull_request_target` run. The gate it feeds, `Author reviewed the PR`, is green on `c87a08c0` — the ack ran on `issue_comment`, which only writes the commit status and does not update the older check run. That job is not itself a required check, so the stale red is cosmetic and clears on the next push.

DorianZheng (approved):
LGTM

PR #1534, opened 2026-09-18
fix(infra): say every bootstrap binding carries no condition

### LINKED ISSUE #1530: bootstrap: the promotion bucket grant is refused unwritten, so a prod promote fails on storage.objects.get
`npm run bootstrap -- --stage <destination> --confirm` reports the bucket half of
its promotion grants as refused on every run, and nothing reaches Cloud Storage.

`grantPromotionReads` makes two grants in the source project. The registry one
carries `--condition=None`; the bucket one does not. A source stage that stages its
runner binary rather than installing a release holds a condition on the very bucket
a promotion reads from — the deploy attaches `runner-artifacts-only` to it
(`apps/infra/mdeploy/stack/providers/gcp/runners.ts:237`, gated on
`transport === 'gcs'`) — and `gcloud … add-iam-policy-binding`, which reads the
policy before editing it, refuses to edit in a binding carrying no condition when
the policy it read holds one:

```
ERROR: (gcloud.storage.buckets.add-iam-policy-binding) Adding a binding without
specifying a condition to a policy containing conditions is prohibited in
non-interactive mode. Run the command again with `--condition=None`
```

The refusal is non-fatal by design, so the bootstrap still exits 0 and the line is
easy to miss. Because no policy is written, the source project's admin-activity log
holds nothing either — the grant looks like it was never made rather than refused.

The failure surfaces later, in `mdeploy-all`'s promote runner job:

```
listing what is staged under runner/<sha>/ failed: ERROR: (gcloud.storage.ls)
HTTPError 403: bl-app-prod-deploy@<destination>.iam.gserviceaccount.com does not
have storage.objects.get access to the Google Cloud Storage object.
```

The command printed for an operator to run by hand also omits the flag, so it only
works where a prompt can be answered.

`allowImpersonation` (`apps/infra/bootstrap/gcp.ts`) has the same omission. Service
account policies carry no conditions today, so that site is latent rather than
broken.

DefeatMan:
## Call graph

```text
Before
  npm run bootstrap -- --stage prod --confirm                                      — operator, once per destination stage
  └─ bootstrapGcp             (Bootstrap · apps/infra/bootstrap/gcp.ts:1103)
       ├─ allowImpersonation  (Bootstrap · apps/infra/bootstrap/gcp.ts:757)        ← BUG: binds without saying it carries no condition
       └─ grantPromotionReads (Bootstrap · apps/infra/bootstrap/gcp.ts:943)
            ├─ projects add-iam-policy-binding                                     — --condition=None; applied
            └─ storage buckets add-iam-policy-binding                              ← BUG: same omission, on a policy that holds a condition whenever the source stages its runner binary
                 …                                                                 — refused between the read and the write; printed, exit 0

  gh workflow run mdeploy-all.yml -f stage=prod -f auto_promote_from=dev           — days later
  └─ promoteRunner            (Command · apps/infra/mdeploy/src/runner-promote.ts:84)
       └─ staged              (Destination · apps/infra/mdeploy/src/runner-build.ts:370)
            …                                                                      — 403 storage.objects.get; nothing promoted

After
  npm run bootstrap -- --stage prod --confirm                                      — operator, once per destination stage
  └─ bootstrapGcp             (Bootstrap · apps/infra/bootstrap/gcp.ts:1103)
       ├─ allowImpersonation  (Bootstrap · apps/infra/bootstrap/gcp.ts:757)        — UNCONDITIONAL; guarded by "a deployer whose policy already holds a condition still takes its impersonation binding"
       └─ grantPromotionReads (Bootstrap · apps/infra/bootstrap/gcp.ts:943)
            └─ storage buckets add-iam-policy-binding                              — UNCONDITIONAL; guarded by "a source bucket whose policy already holds a condition still takes the grant"
                 …                                                                 — both grants applied

  gh workflow run mdeploy-all.yml -f stage=prod -f auto_promote_from=dev
  └─ promoteRunner            (Command · apps/infra/mdeploy/src/runner-promote.ts:84)
       └─ staged              (Destination · apps/infra/mdeploy/src/runner-build.ts:370)
            …                                                                      — lists gs://<source>/runner/<sha>/ under objectViewer
```

Fixes #1530

## Why

- `add-iam-policy-binding` reads the policy, edits it and writes it back, and when the policy it read already holds a condition it refuses to edit in a binding carrying none unless the command says that is what it means — so it never reaches the write. A terminal prompts for the condition instead; `--quiet` over the closed stdin `execRun` gives every call is the error.
- Two of the four bindings this bootstrap makes said nothing, and nothing here can tell which policies hold a condition: this run never reads one before writing to it. One of the two does hold a condition whenever its stage stages a runner binary rather than installing a release — the deploy attaches `runner-artifacts-only` to the artifacts bucket of that stage, which is the bucket a promotion reads from. Against such a source the grant was refused every time, reported rather than fatal, and absent from the admin-activity log of the source project because no policy was ever written.
- The rule now lives in one named constant with the reason next to it, applied at all four call sites. The two that already carried the flag carried the same comment twice, which is the shape that let the other two be written without it.
- The flag stays in the argv of each call rather than moving into `applyPolicy`: `grantPromotionReads` prints the grants it could not make as the commands an operator runs by hand, and a flag hidden in the caller would fix the machine path while leaving the human one refused the same way.
- `allowImpersonation` had the same omission. Service account policies carry no conditions today, so that site is latent rather than broken; it is fixed in the same pass because it is the same defect, and pinned by its own test.
- The refusal also printed `granting X: Could not granting X: ERROR …`, the line prefix and the error message repeating each other. It now prints the error once.
- Rejected: modelling the rule inside `applyPolicy`. Fewer repetitions, but the printed fallback commands are built from the same argv, so an operator would be handed a command that only runs where a prompt can be answered.
- Rejected: leaving `allowImpersonation` until it fails. The binding that decides who may act as the deployer is a poor place to learn that someone added a condition upstream.

## User-facing change

NONE — bootstrap wiring. A GCP stage that is promoted into no longer needs its bucket grant run by hand.

## Verification

- Two-side, both runs manual. With every non-test change reverted and only the tests present, `npx tsx --test bootstrap/gcp.test.ts` fails 3: `a source bucket whose policy already holds a condition still takes the grant` reports `1 of 2 refused` carrying the wording gcloud itself prints, `Adding a binding without specifying a condition …`; `a deployer whose policy already holds a condition still takes its impersonation binding` never reaches the line `bootstrapGcp` prints after that grant, because `allowImpersonation` throws the same message; and the printed bucket command lacks the flag. With the change restored, 23 pass.
- `npm test` in `apps/infra`: 1321 pass, 0 fail across all five suites — the 1319 that passed before, plus the two new tests. `npm run typecheck` reports nothing.
- Against the real projects: the exact argv the bootstrap sends, run with `--quiet` and stdin closed, prints that refusal without the flag and exits 0 with it. The binding it makes is the one the failed prod promotion was missing.
- Not reproduced end to end: that needs a prod dispatch with the grant absent again.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved deployment reliability by ensuring IAM policy updates complete correctly when existing policies include conditions.
  - Prevented automated deployments from waiting for interactive condition prompts.

- **Documentation**
  - Updated deployment guidance to explain IAM condition handling and the required configuration.
  - Added a review-status note for tracking changes that have not yet been reviewed.


DefeatMan:
Superseded by #1570. The production change in `bootstrap/gcp.ts` and `bootstrap/gcp.test.ts` is carried over line-for-line; `apps/infra/DEPLOY.md` differs in two ways, both deliberate: one paragraph was re-wrapped, and the `promoteFrom` paragraph this PR restated was rewritten because #1570 removes the `auto_promote_from` input it described.

PR #1541, opened 2026-09-19
chore(infra): correct stale mstage and mbuild prose, drop a dead export

DefeatMan:
## Call graph

```text
Before
  npm run mbuild inspect -- --stage dev                                        — user command
  └─ main                (bin · apps/infra/mbuild/bin/mbuild.ts:116)
       └─ regionOf       (bin · apps/infra/mbuild/bin/mbuild.ts:77)            — reads the region from mstage
  …
  regionFor              (mbuild · apps/infra/mbuild/src/config.ts:483)        — exported, reached by nothing

  npm run mstage login -- --help                                               — user command
  └─ moduleUsage         (help · apps/infra/mstage/src/cli/help.ts:75)
       └─ LOGIN_COMMANDS (run · apps/infra/mstage/src/cli/run.ts:143)          — renders all four providers
  …
  help.test.ts:46        (test · apps/infra/mstage/test/help.test.ts:46)       — checks three of them, from a list of its own

After
  npm run mbuild inspect -- --stage dev                                        — user command
  └─ main                (bin · apps/infra/mbuild/bin/mbuild.ts:116)
       └─ regionOf       (bin · apps/infra/mbuild/bin/mbuild.ts:77)            — unchanged, the only region path
  …
  (regionFor deleted; stageIn stays, registryFor is what calls it)

  npm run mstage login -- --help                                               — user command
  └─ moduleUsage         (help · apps/infra/mstage/src/cli/help.ts:75)
       └─ LOGIN_COMMANDS (run · apps/infra/mstage/src/cli/run.ts:143)          — unchanged
  …
  help.test.ts:46        (test · apps/infra/mstage/test/help.test.ts:46)       — reads LOGIN_PROVIDER_NAMES (run.ts:137)
```

## Why

- mstage and mbuild are vendored from a shared upstream, and several artifacts
  in them outlived the change that invalidated them.
- The two-step gcloud sign-in (`gcloud auth login` then `gcloud auth
  application-default login`) landed without the prose describing the old one:
  the README, `cli/handlers/login.ts` and `checkGcp` still named `gcloud auth
  login --update-adc` as the command mstage runs, and `checkGcp` carried both
  the new comment and the one that change had disproved. Prose that contradicts
  the code is read as current, so it is worse than none. The `--update-adc`
  mentions that remain are deliberate — they say why that flag is the wrong fix.
- The `audit` block in `mbuild` documented the layout of this repository (images
  built from `apps/`, a Yarn 4 lockfile, no lockfile at the root) inside a tool
  other repositories vendor by copying, where every one of those facts is false.
  Restated as a property of the repository mbuild was written in.
- `regionFor` has had no caller since the bin started reading the region of a
  stage from `mstage/config`. Rejected: keeping it as a convenience export — nothing
  imports `mbuild/config` for it, and a second way to answer a question one
  function already answers is how the two drift apart.
- Three places said mstage has three login providers where `LOGIN_PROVIDERS`
  holds four. One of the three was `help.test.ts`, whose written-out
  list naming aws, github and auth0 sat under the name "login lists every
  provider mstage can check" — the only guard for the README sentence, and blind
  to the provider it omitted. Rejected: adding gcp to that list, which fixes today
  and leaves the next provider to be forgotten the same way; it now reads
  `LOGIN_PROVIDER_NAMES`, exported the way `MODULE_NAMES` beside it already is.

## User-facing change

NONE — comments, one export nothing imported, and a test that now reads its list
from the registry.

## Verification

- `npm run typecheck:mstage`, `npm run typecheck:mbuild`: clean.
- `npm run test:mstage`: 381/381. `npm run test:mbuild`: 78/78.
- The strengthened help test, both ways round: with `LOGIN_COMMANDS` temporarily
  filtered to drop `gcp`, it fails with `gcp is missing from login help`, while
  the written-out list it replaces passes 11/11 against the same defect. Both
  runs were on `help.test.ts` alone; the filter was reverted and the full suite
  re-run green.
- `git grep -w regionFor`: no matches after the deletion, and `stageIn`,
  `registryFor` and `BuildConfigError` all keep live callers.
- No reproducer for the prose: nothing there changes behavior, so there is no
  failing-then-passing test to show.


DefeatMan:
Superseded by #1570, which carries both commits unchanged (`git range-diff` reports them identical in patch and message). The bot `UNREVIEWED.md` marker was deliberately not carried over — it is per-PR CI state.

PR #1547, opened 2026-09-20
feat!: start a box from any public image

silverzki:
## Call graph

```text
Before
  POST /v1/boxes {"image":"quay.io/acme/app"}                                                                  — API request
  └─ create                 (BoxService · apps/api/src/box/services/box.service.ts:231)                        — curated-only
       └─ assertSupportedImage (curated · apps/api/src/box/constants/curated-images.constant.ts:95)             — 400; the runner is never reached

After
  POST /v1/boxes {"image":"quay.io/acme/app"}                                                                  — API request
  └─ create                 (BoxService · apps/api/src/box/services/box.service.ts:231)                        — admit, resolve, dispatch
       ├─ assert            (ImageAdmissionService · apps/api/src/image/services/image-admission.service.ts:48) — allowlist, per-org catalog limit, cold-pull budget
       ├─ resolve           (ImageResolverService · apps/api/src/image/services/image-resolver.service.ts:53)   — a catalog hit returns <repo>@sha256:…, a miss passes the ref through
       └─ createBox         (RunnerAdapterV2 · apps/api/src/box/runner-adapter/runnerAdapter.v2.ts:117)         — + anonymousImagePull, imageRevalidate
            └─ Create       (Client · apps/runner/pkg/boxlite/client.go:272)                                    — runner job
                 ├─ pull    (ImageStore · src/boxlite/src/images/store.rs:181)                                  — anonymous for a tenant ref
                 │    └─ assert_registry_is_public (store.rs:606)                                               — refuses a registry that resolves to a private address
                 └─ recordPulledImage (Client · apps/runner/pkg/boxlite/client.go:401)                          — after Start, so the digest exists; guarded by create_invariant_test.go
  PUT /box/:id/state {"imageDigest":"sha256:…"}                                                                — runner sync
  └─ updateState            (BoxService · apps/api/src/box/services/box.service.ts:1383)                       — recorded before the state machine's early returns
       └─ onBoxStarted      (ImageRegistrarService · apps/api/src/image/services/image-registrar.service.ts:48) — writes image/image_version/image_tag; guarded by image-registrar.service.integration.spec.ts
```

## Why

A box could only boot from three curated selectors. Everything else was
refused at the API, so a tenant with an ordinary public image had nowhere to
put it. This opens that path and keeps the two properties that made the
curated restriction safe in the first place.

**The registry credential stays out of tenant reach.** A runner holds the
operator's `ghcr.io` and Docker Hub credentials. Admitting tenant references
without changing anything else would let `ghcr.io/<someone-else>/<private>`
be fetched with the operator's token and handed to whoever asked — a confused
deputy. `PullPolicy { anonymous }` decides by whether the ref is curated, not
by host, because an operator may point a curated entry at a private package
and those must keep their credentials. `docker.io` and `ghcr.io` are
deliberately absent from the default allowlist until every runner in a fleet
carries this change.

**A ref resolves to one build, once.** A runner's image cache is keyed by the
ref string and never re-checked, so an unpinned hit would make which build a
box gets depend on which runner it landed on. The catalog records the digest
a reference first resolved to and hands it back pinned; tags are not followed
when they move upstream. That applies to a bare repository too, which a
registry serves as `latest`.

**The pull cannot be aimed inward.** A registry hostname is resolved before
`oci_client` is called and refused if any address is private, loopback,
link-local or carrier-grade NAT — including the IPv6 shapes that a naive
`to_ipv4_mapped` check misses. Two residues are documented at the point of
risk in `store.rs` and are not closed here: `oci-client` re-resolves the
hostname per connection, and a blob request's redirects are followed. Both
need control of an allowlisted registry or its DNS, and closing either means
vendoring or a pull gateway. They are tracked as one ticket.

**Scope, decided deliberately.** Two commits are not about booting from a
public image and were kept here anyway, knowing the cost:
`feat(runner): expose the core runtime metrics on /metrics` would split for
almost nothing, and `chore(api)!: drop the retired template deactivation
column` would cost two client regenerations, a hand-separated
`organization.entity.ts` hunk and a guaranteed post-rebase conflict on
generated files, because one regeneration commit carries both that removal
and this feature's two new fields. The consequence a reviewer should not have
to discover: the change most consequential for downstream consumers — a
required `Organization` property disappearing — arrives as commit 6 of 15 in
a pull request about image admission.

## User-facing change

`POST /v1/boxes` accepts any image on the registry allowlist (`quay.io`, `gcr.io`, `public.ecr.aws` by default) instead of only the three curated selectors, and BREAKING: `Organization` responses no longer carry `templateDeactivationTimeoutMinutes`, a property the specification declared required.

## Verification

- `npx jest --config apps/api/jest.config.ts` against a real Postgres and Redis — 429 tests across 62 suites covering `image/`, `box/`, `migrations/` and `organization/`; the `*.integration.spec.ts` suites really run rather than skip, which is what CI does too (`test.yml` sets `DB_HOST`).
- `go test -tags boxlite_dev ./pkg/boxlite/ ./pkg/services/` — both packages pass; `gofmt -l` and `go vet` clean.
- `cargo test` for the `images` and `litebox` modules; `node apps/node_modules/typescript/bin/tsc -p apps/api/tsconfig.app.json --noEmit` clean.
- Clients regenerated with openapi-generator 7.23.0, the version `openapitools.json` pins and the drift check uses, so `api-client-drift.yml` has nothing to report.

Two-side verification on the reproducer tests, each run both ways rather than argued:

- **A bare repository never pinned.** With `image-ref.util.ts`, `image-resolver.service.ts` and `image-registrar.service.ts` reverted and only the new integration test present, it fails at the defect: `Expected: "latest" / Received: undefined`, no tag row written. Restored, it passes. Separately, mutating the shared `IMPLICIT_TAG` fails exactly two tests, one per side — which proves neither assertion is inert, and does not prove the fix, since a mutation moves both sides together.
- **A destroyed box's image report leaked.** Removing `delete(c.pendingImageReports, boxId)` from `forgetBox` fails both subtests, at `forget_box_test.go:38` and `:61`; restored, both pass.
- **The digest report was inert.** The runner read `PulledImage()` immediately after `GetOrCreate`, but core does not start the VM until `start()`, so the `OnceLock` was always empty and every test still passed because both sides were stubbed. The read moved after `Start`, and an AST test now fails if a fallible step is added between them.
- **`imageSizeBytes` silently rounded.** Declared without a type, the spec said `number` and the Go generator chose `float32`; 987654321 arrived as 987654340, still an integer and still passing validation into a `bigint` column. The property now declares `integer`/`int64` and the cast at the call site is gone.

## Summary by CodeRabbit

* **New Features**
  * Added organization-scoped image catalogs with version and tag tracking.
  * Added registry allowlisting, image-count limits, and per-organization pull rate limits.
  * Added anonymous image pulls, image revalidation controls, and pulled-image digest/size reporting across APIs and SDKs.
  * Added runtime metrics for running boxes, resource counters, and sampling failures.
* **Bug Fixes**
  * Improved warm-pool matching to respect requested GPU counts and prevent cross-organization image reuse.
* **Breaking Changes**
  * Removed the deprecated organization template deactivation timeout field and Images permission group.


src/boxlite/src/images/store.rs
```diff
@@ -51,6 +51,32 @@ impl ImageStoreInner {
     }
 }
 
+/// How a pull authenticates.
```
DorianZheng: image registry and credential should be configured just once. we don't need to set it every time. `PullPolicy` confused me, i think we don't need it

silverzki: You're right that registries and credentials are configured once, and nothing here changes that. `PullPolicy` doesn't configure anything — it carries no credential and no registry. It's two booleans that *select* among the credentials the runtime was already given at `NewClient`.

The reason the selection can't be runtime-level is that one runner serves two kinds of pull **from the same hosts**:

- curated images are the operator's and must keep the operator's ghcr / Docker Hub credential — `apps/infra/.env.example:127` documents pointing a curated entry at a private package, and that credential exists for exactly that;
- a ref a tenant named must not reach that credential, or `ghcr.io/<another-org>/<private-image>` gets fetched with the operator's token and handed to whoever asked for it.

Both are `ghcr.io`. `registry_auth_for` matches by host, so before this PR the host match alone decided it — which was safe only because the API refused every non-curated ref. This PR removes that refusal, so the decision has to move to the pull.

It also can't be decided inside the runner: *which* refs are the operator's own is the control plane's knowledge — it's the curated set, env-driven, and the API is what resolves against it. So it arrives as a bit on the dispatch, beside the ref.

Deciding by host instead is the hole coming back either way: strip credentials from every `ghcr.io` pull and curated private packages stop working; keep them and the confused deputy returns.

`revalidate` is per-pull for a different reason — it depends on whether *this* ref is already digest-pinned, which is a property of the ref rather than of the runtime.

That said, the doc comment is what confused you, and that's on it: it opens with "How a pull authenticates", which reads like per-pull auth configuration. I'm rewording it to lead with what it actually is — a selection among credentials that are still configured exactly once — and will push that with the rest of the review fixes.

If you'd still rather see a different shape — two plain arguments instead of a struct, or the flag folded into the existing `BoxOptions` the payload already carries — say which and I'll change it.

silverzki: Pushed as 9989a1a3. One note, because it delivers more than I said it would: I promised to lead with "a selection among credentials", but the new summary line covers both fields —

> What one pull may do that the runtime it runs on cannot decide for it.

`PullPolicy` grew a second field (`revalidate`, whether the ref-keyed cache may answer) when tag revalidation landed, and the old summary — and the "covers credentials only" paragraph under it — still described only `anonymous`. Narrowing the line to credentials would have left the same gap you ran into, in a type whose reason for existing is what you were questioning. So the summary names both, and "covers credentials only" is now scoped to `anonymous`, which is the field it was always about. The sentence that credentials are matched by host is back too, as the reason the choice cannot be per runtime rather than as an aside.

The offer stands: if you'd rather see a different shape — two plain arguments instead of a struct, or the flag carried some other way — say which and I'll change it.

apps/api/src/image/services/image-registrar.service.ts
```diff
@@ -0,0 +1,140 @@
...
+   * Record what a box booted from.
+   *
+   * Curated images are skipped: they are the operator's, shared by every
+   * organization, and putting them in a tenant's catalog would count them
+   * against that tenant's limit and let one tenant delete what everyone uses.
+   */
+  async onBoxStarted(organizationId: string, imageRef: string, reported: ReportedImage): Promise<void> {
+    if (isCuratedSelector(imageRef)) {
+      return
+    }
```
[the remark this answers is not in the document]
silverzki: Fixed in 84e25966 — the ownership decision is now persisted on the box row and `onBoxStarted` is told rather than recomputing.

Two notes on how it differs from the suggestion.

The registrar takes `{ ref, isOrgOwned }` instead of a `Box`, so the image module stays free of the box entity. The caller reads the row through `boxImageIsOrgOwned(box)`.

And the same defect had two more sites, which are fixed with it: both runner adapters computed `anonymousImagePull` and `imageRevalidate` from the same stale recomputation. That one has teeth — after a curated rotation a box whose curated image points at a private package (`apps/infra/.env.example:127` documents doing exactly that) would start pulling anonymously and stop pulling at all.

The column is nullable and deliberately not backfilled: what the curated set held when an existing box was created is not knowable now, so a null means "not recorded" and the reader falls back to the recomputation those rows have always had. New rows are decided once, at resolve time, where the answer is already known.

silverzki:
## 本地端到端验证：通过

在本机起了完整控制平面（Postgres + Redis + API `:3000` + runner `:8080`），走真实 REST 链路，不是单测。

`POST /v1/{prefix}/boxes {"image":"docker.io/library/alpine:3.21"}` → `HTTP 201`，`status: running`，无任何前置调用。

镜像身份在 VM 内用 exit code 判定：`grep -q "^3\.21" /etc/alpine-release` 得 0，反向对照 `^9\.99` 得 1，`exit $(wc -c < /etc/alpine-release)` 得 7（`3.21.8\n` 正好 7 字节）。

目录三张表写入正确，`box.imageIsOrgOwned = t`，`image_version.storageRef` 钉到 digest，`sizeBytes 3974501`、`state ready`、`sourceKind pull`。

**解析出的 digest 在三条独立路径上完全一致**：Rust SDK 集成测试、Go SDK 集成测试、以及这次的完整控制平面。

拒绝路径同时验到：默认允许表下发 `docker.io` 的 ref 得 400，错误信息把允许的 registry 列了出来。

### 与验收标准 B1 的一处出入

B1 写的是 `POST /boxes {image:"docker.io/library/alpine:3.20"}`，但 `docker.io` 按 I3 不在默认允许表内（默认只有 `quay.io` / `gcr.io` / `public.ecr.aws`），照字面发必然 400。我是把 `BOXLITE_IMAGE_REGISTRY_ALLOWLIST` 显式加了 `docker.io` 之后才跑通的。B1 的措辞需要跟下面的放开步骤对齐。

---

## allowlist 放开步骤

### 第一层：`docker.io` / `ghcr.io` 加回默认表

**不能和本 PR 同一次发布。** 前置是本 PR 铺满整个 fleet，然后另起一次配置变更。理由见内部设计文档 R16（部署顺序）与 `apps/infra/.env.example` 中该 key 上方的说明。

1. 本 PR 合并、部署
2. 确认全 fleet runner 都带上匿名拉取
3. 另起一次配置变更，加 `docker.io` / `ghcr.io`

**第 3 步目前做不到**，`BOXLITE_IMAGE_REGISTRY_ALLOWLIST` 在两条部署路径上都到不了 API：

| 位置 | 状态 |
| -- | -- |
| `apps/infra/mdeploy/src/api-environment.ts` passthrough | 有 |
| `apps/infra/.env.example` | 有 |
| **`apps/infra/mstage.env.json` 的 deploy 组** | **没有** ← 断点 |
| `apps/infra/stack/api.ts`（SST 路径） | 没引用 |

mstage 只取声明过的 key 注入部署进程；没声明就取不到，`passthrough` 得空串，API 落回内置表。兄弟 key `BOXLITE_SYSTEM_IMAGES`、`BOXLITE_SYSTEM_BASE_IMAGE`、`BOXLITE_SYSTEM_SOURCE_REGISTRY_URL` 全都声明了。

`mdeploy/test/api-environment.test.ts:189` 有一个「读了就必须有人 fetch」的测试，但只覆盖 `BILLING_KEYS` 与 `STATUS_SYNC_KEYS`，所以没拦住。

补法：加进 `mstage.env.json` deploy 组的 optional；SST 路径照 `BOXLITE_SYSTEM_IMAGES` 加一行 `envOr`；把那个测试扩到 `systemImagesFrom` 读的所有 key。**这三处不在本 PR 范围内**，会另开工单。

### 第二层：允许表整个去掉（任意 registry）

**当前没有已规划的阶段能做到。** 允许表本身还承担着 S1 明确记录并接受的若干残余风险的兜底作用；S1 把这些的归宿写成 S2 的网关，而 S2 现有设计里公开镜像不经网关，两者需要单独裁决。细节见内部设计文档 R2 / R5 / R15 与 S2 的 V5。

src/boxlite/src/runtime/options.rs
```diff
@@ -393,6 +424,30 @@ pub struct BoxOptions {
...
+
+    /// Pull this box's image without the runtime's registry credentials.
+    ///
+    /// Those credentials are matched by host and belong to whoever configured
+    /// the runtime. A caller that hands over an image ref it did not choose —
+    /// a multi-tenant control plane passing a tenant's ref — sets this so the
+    /// ref cannot reach them. Default `false`: a caller pulling its own images
+    /// keeps the registries it configured.
+    #[serde(default)]
+    pub anonymous_image_pull: bool,
```
DorianZheng, line 449: don't define these here

src/boxlite/src/litebox/mod.rs
```diff
@@ -86,6 +86,15 @@ impl LiteBox {
         &self.id
     }
 
+    /// What this box's image resolved to, when this process resolved it.
+    ///
+    /// For a caller that handed over a mutable tag and needs to know which
+    /// build it got. Read it after starting the box — nothing is resolved until
+    /// then — and expect `None` for a box this process only reattached to.
+    pub fn pulled_image(&self) -> Option<crate::images::PulledImage> {
```
DorianZheng, line 94: don't define this API

apps/api/src/migrations/pre-deploy/1787100000000-add-image-catalog-migration.spec.ts
```diff
@@ -0,0 +1,67 @@
+import { QueryRunner } from 'typeorm'
```
DorianZheng, line 1: why so many migrations for 1 PR?

src/boxlite/src/runtime/options.rs
silverzki: Agreed — they were one concept stored twice: core already had `PullPolicy { anonymous, revalidate }`, and `BoxOptions` flattened it into two fields.

81eb524e makes them one field, `image_pull: ImagePullOptions`, and deletes `PullPolicy`; the store and image manager take the same type. `anonymous` persists with the box because a restart reads the image again; `revalidate` stays `serde(skip)`. C gets one setter, Go gets `WithImagePull(...)`.

The choice still arrives per box rather than per runtime: curated and tenant images share `ghcr.io`, and the runtime matches credentials by host. If you would rather it live elsewhere, say where.

src/boxlite/src/litebox/mod.rs
silverzki: Removed in ddc6dd69. It duplicated `ImageObject::manifest_digest()` / `total_layer_size()`, exposed a second time only because the pull happens inside `start()`.

The digest now lives with the box: `BoxState::resolved_image`, persisted in the state JSON like `exit_code` (no migration, same save that start already makes), and read through the existing `BoxInfo`. No new API in Rust, C or Go; the runner calls `Info` after start.

Only the start that builds the disk records it. A restart keeps the record, and a clone inherits its source's.

apps/api/src/migrations/pre-deploy/1787100000000-add-image-catalog-migration.spec.ts
silverzki: Down to one.

- `box.imageIsOrgOwned` is folded into this migration (093ae1a1); both were pre-deploy and additive.
- The post-deploy drop of `organization.template_deactivation_timeout_minutes` is gone (33dd49b7).

That drop could not be merged in: old API instances still `SELECT` the column during the rollout. But nothing acts on its value, so this PR only unmaps it. It is `NOT NULL DEFAULT 20160`, so inserts that omit it still work.

The column is listed in `apps/api/src/migrations/README.md` with the ssh-gateway leftovers, to be dropped together later.

src/boxlite/src/images/store.rs
silverzki: `PullPolicy` is gone in 81eb524e: the two choices are one `ImagePullOptions` field on `BoxOptions` now. Details on the `options.rs` thread.

silverzki:
This replaces the "allowlist rollout" steps in my earlier comment:

- **docker.io is in the default list (8dcce4ac).** No repo-managed runner holds Docker Hub credentials, so old runners pull it anonymously too. B1 now passes as written.
- **ghcr.io still waits.** The SST stack gives runners `GHCR_TOKEN`, and an old-build runner still sends it. Once every runner serves this build, set `BOXLITE_IMAGE_REGISTRY_ALLOWLIST=docker.io,ghcr.io,quay.io,gcr.io,public.ecr.aws` and redeploy the API. The value replaces the default, so list all five.
- **No separate ticket.** POL-659 is fixed here (a718530f): the value now reaches the API on both deploy paths.

silverzki:
Call graph, before and after:

```text
Before
POST /v1/boxes
└─ BoxService.create               box.service.ts:190
   └─ assertSupportedImage         curated-images.constant.ts:93   400 unless curated

After
POST /v1/boxes
└─ BoxService.create               box.service.ts:215
   ├─ ImageAdmissionService.assert image-admission.service.ts:48   allowlist, limits
   ├─ ImageResolverService.resolve image-resolver.service.ts:53    pin to digest
   └─ RunnerAdapterV2.createBox    runnerAdapter.v2.ts:117         anonymous pull flag
      └─ Client.Create             client.go:288
         ├─ ImageStore.pull        store.rs:157                    no credentials
         │  └─ assert_registry_is_public store.rs:1105             public hosts only
         └─ recordPulledImage      client.go:421                   digest after Start
PUT /box/:id/state
└─ BoxService.updateState          box.service.ts:1400
   └─ ImageRegistrarService.onBoxStarted image-registrar.service.ts:69 catalog rows
```

apps/api/src/image/utils/image-ref.util.ts
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  const segments = rest.split('/')
+  // A first segment carrying a dot, a port, or the literal `localhost` is a
+  // registry host; otherwise the ref is a Docker Hub short form.
+  const hasHost =
+    segments.length > 1 && (segments[0].includes('.') || segments[0].includes(':') || segments[0] === 'localhost')
+  const host = hasHost ? segments[0] : 'docker.io'
+  const pathSegments = hasHost ? segments.slice(1) : segments
+  const repository = pathSegments.length === 1 && !hasHost ? `library/${pathSegments[0]}` : pathSegments.join('/')
```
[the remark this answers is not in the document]
silverzki: Fixed in 505e7a6. The `library/` namespace now follows the host: a single-segment repository gets it whenever the host is docker.io, so `python` and `docker.io/python` share one catalog key. The parse table gains `docker.io/python:3.12`, which failed before the fix with repository `python`, and a single-segment ref on another registry, which stays unprefixed.

silverzki:
Update in cfbc3f23 and 067a1c35 (design: #1680):

- Runners hold no registry credential; ImagePullOptions and the SDK setters are gone.
- Core refreshes a tag only when it builds a new disk, falls back to the cache when no request is answered, and restarts on the build its disk was made from.
- Curated tags are pinned per runner at dispatch, so curated boxes stay on the cache.
- The cold-pull budget is spent only on catalog misses.

Verified on a local control plane against docker.io, quay.io, gcr.io, ghcr.io and public.ecr.aws, and through the Python REST SDK.

PR #1549, opened 2026-09-20
fix(infra): admit Cloud Run to a VM by network tag, not by identity

### LINKED ISSUE #1548: gcp: every /v1/boxes/* route 504s — the runner firewall admits the API by service account, which direct VPC egress never matches
On a GCP stage every control-plane route that proxies to a runner — `exec`,
`files`, `metrics`, the tunnel CONNECT — returns 504 after ~127 s, against a
runner that is `state: ready`, reports fresh metrics, and answers the same port
from a GKE proxy Pod.

```
boxlite list                          → API → Postgres            ok
boxlite exec <box> -- echo hello      → API → runner 10.x.x.x:3003  504, latency 127.03 s
boxlite stats <box>                   → API → runner 10.x.x.x:3003  hangs, then 504
```

127 s is a Linux TCP connect timeout: the SYN is dropped, so nothing reaches
the runner and nothing is logged there.

`stack/providers/gcp/network.ts` admits the control plane to the runner with
`sourceServiceAccounts: [accounts.api.email]`. Source service accounts match
traffic from VM instances; the API is a Cloud Run service reaching the network
through direct VPC egress, and Google attributes those packets to no service
account at all — so the rule admits nothing and the explicit deny at 65534
takes the packet. Network tags and IP ranges are the only sources that match
serverless traffic
(https://cloud.google.com/run/docs/configuring/vpc-direct-vpc).

A Network Intelligence connectivity test from the API revision to the runner
confirms it:

```
START_FROM_CLOUD_RUN_REVISION
DIRECT_VPC_EGRESS_CONNECTION
APPLY_EGRESS_FIREWALL_RULE    fw=default-allow-egress  action=ALLOW
ARRIVE_AT_INSTANCE
APPLY_INGRESS_FIREWALL_RULE   fw=<app>-<stage>-deny-ingress  action=DENY
DROP                          cause=FIREWALL_RULE
```

The same defect has a second site: `ClickHouseFirewall` in
`stack/providers/gcp/clickhouse.ts` admits the API and the collector the same
way, and a connectivity test from the collector revision to the ClickHouse host
on 8123 drops at the same rule. So telemetry writes and the analytics reads are
blackholed on a GCP stage too.

Box lifecycle is unaffected, which is why the stage looks healthy: a runner
polls the control plane for its work and pushes its own health, so nothing in
that direction needs this rule.

AWS is unaffected — the security-group pair expresses the same reachability by
position, and an ECS task is a real network interface.

DefeatMan:
## Call graph

```text
Before
  boxlite exec <box> -- echo hello                                                              — user command
  └─ proxyExec           (BoxliteProxyController · apps/api/src/boxlite-rest/boxlite-proxy.controller.ts:76)
       └─ proxyToRunner  (BoxliteProxyController · apps/api/src/boxlite-rest/boxlite-proxy.controller.ts:287)  — dials runner.apiUrl over direct VPC egress
            └─ RunnerFirewall (gcp.compute.Firewall · apps/infra/mdeploy/stack/providers/gcp/network.ts:241)   ← BUG: sourceServiceAccounts matches no serverless packet, so the deny at 65534 takes the SYN
                 └─ BoxliteExec (gin route · apps/runner/pkg/api/server.go:151)                                — never reached; 504 after a 127 s connect timeout
  …
  collector → ClickHouseFirewall (gcp.compute.Firewall · apps/infra/mdeploy/stack/providers/gcp/clickhouse.ts:411)  ← BUG: same rule shape, same drop, on 8123

After
  boxlite exec <box> -- echo hello                                                              — user command
  └─ proxyExec           (BoxliteProxyController · apps/api/src/boxlite-rest/boxlite-proxy.controller.ts:76)
       └─ proxyToRunner  (BoxliteProxyController · apps/api/src/boxlite-rest/boxlite-proxy.controller.ts:287)
            └─ RunnerFirewall (gcp.compute.Firewall · apps/infra/mdeploy/stack/providers/gcp/network.ts:265)   — sourceTags api → targetTags runner; guarded by "a Cloud Run service reaches a VM by tag"
                 └─ BoxliteExec (gin route · apps/runner/pkg/api/server.go:151)                                — reached; exec returns its exit status
  …
  collector → ClickHouseFirewall (gcp.compute.Firewall · apps/infra/mdeploy/stack/providers/gcp/clickhouse.ts:439)  — sourceTags callerTags → targetTags hostName
```

Fixes #1548

## Why

- `RunnerFirewall` and `ClickHouseFirewall` admitted their callers with `sourceServiceAccounts`. That matches traffic from VM instances; a Cloud Run service reaching the network through direct VPC egress is attributed to no service account, so both rules admitted nothing and the explicit deny at 65534 dropped the SYN — 504 after a full TCP connect timeout against a runner that is healthy, answers the GKE proxy on the same port, and logs nothing because nothing arrived.
- A placement now carries a network tag beside its account, the two Cloud Run services stamp it on the interface their packets leave through, and both rules are keyed on it. Network tags and IP ranges are the only sources Google matches serverless traffic by.
- The hosts are named by tag as well because Google refuses a rule that pairs a source tag with a target service account.
- The accounts stay where they grant: ClickHouse still hands its callers the password by identity, so `callers` and `callerTags` are both passed rather than one replacing the other.
- Rejected: keying the rules on the subnet range — the other source that works, and the one Google's own guidance names. It would admit every workload that ever lands in `10.20.0.0/20`, the runners included, which is the widening `InternalFirewall`'s own comment argues against. A dedicated subnet per caller would make ranges as narrow as tags, and is a larger change than this defect warrants.

Two corrections this branch makes to its own earlier claims, since the squash merge keeps only this section:

- Network Intelligence connectivity tests report `DROP: INGRESS_FIREWALL_TAGS_UNSUPPORTED_BY_DIRECT_VPC_EGRESS` against the tag-keyed rule, and the Cloud Run docs list "network tags or service identity in ingress firewall rules" as unsupported. Measured against a live stage, the data plane honours the rule: that limitation is about naming a Cloud Run workload as the *target* of an ingress rule, not as the *source* of one admitting it to a VM.
- A source-tag rule is not applied when it is written. After an unrelated deny on the same tag was added and removed, the runner's rule admitted nothing for forty minutes; rewriting the rule is what ended it. One trial, and the removed deny is a confounder that cannot be ruled out from it — so the note on `Placement.networkTag` records the observation and the repair, not a mechanism.

## User-facing change

On a GCP stage, `boxlite exec`, `cp`, `stats` and box tunnels reach the runner instead of timing out with 504.

## Deploying this

The first apply rewrites both rules. A rewritten source-tag rule may not serve immediately — run `boxlite exec <box> -- echo hello` against the stage afterwards, and if it times out, write the rule again rather than waiting it out.

## Verification

- `make test:apps:infra`: 557 + 7 + 381 + 78 + 297 tests, 0 failures; `typecheck:tooling` clean.
- Two-sided, both run by hand: with every non-test change reverted, `mdeploy/test/gcp-pitfalls.test.ts` fails 2 — `admitted: [] !== ['api','otel-collector']`, and no match for `/sourceTags: \[tagFor\('api'\)\]/`. With the change restored, 61 pass.
- Applied to the dev stage by hand and exercised end to end: `boxlite exec` returns in ~3 s and `boxlite stats` serves, where both previously returned 504 after 127 s. The tag-keyed rule is the only one that can admit the control plane to `:3003` — `*-internal` targets the three Cloud Run service accounts and `*-proxy-to-runner` sources the GKE Pod range.
- Falsified rather than assumed, since "it works" would also fit "the source is never evaluated", which would be a hole rather than a fix: a deny at priority 900 keyed on the same source tag stops `boxlite exec`; removing it restores service. The deny matched, so the source tag is what the rule is keyed on.


apps/infra/mdeploy/stack/providers/gcp/network.ts
```diff
@@ -229,17 +250,27 @@ export const gcpNetworkProvider =
...
     const runnerIngress = new gcp.compute.Firewall('RunnerFirewall', {
       name: instanceFor({ app: $app.name, stage: $app.stage, artifact: 'runner' }),
       project,
       network: network.id,
       direction: 'INGRESS',
       priority: 1000,
       allows: [{ protocol: 'tcp', ports: [String(RUNNER_PORT)] }],
-      sourceServiceAccounts: [accounts.api.email],
-      targetServiceAccounts: [accounts.runner.email],
+      sourceTags: [tagFor('api')],
```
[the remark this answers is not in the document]
DefeatMan: Measured against the live stage before answering, and the finding does not hold: source tags **are** evaluated for Direct VPC egress traffic.

The dev stage now runs exactly this shape — `boxlite-app-dev-runner` at priority 1000, `sourceTags: [boxlite-app-dev-api]`, `targetTags: [boxlite-app-dev-runner]`, tcp:3003 — and it is the only rule that can admit the control plane to that port (`*-internal` targets the three Cloud Run service accounts, `*-proxy-to-runner` sources the GKE Pod range `10.20.32.0/19`). With it in place `boxlite exec` returns in ~3s where it previously returned 504 after a 127s connect timeout.

Falsification, since "it works" alone could also mean "the source is not evaluated at all", which would be a hole rather than a fix:

```
+ deny  priority 900  sourceTags=[boxlite-app-dev-api] targetTags=[boxlite-app-dev-runner] tcp:3003
  boxlite exec  -> fails
- deny
  boxlite exec  -> serves again
```

The deny matched, so the source tag is what the rule is keyed on.

Two things worth recording for anyone who reads this later:

- Network Intelligence connectivity tests report `DROP: INGRESS_FIREWALL_TAGS_UNSUPPORTED_BY_DIRECT_VPC_EGRESS` on the very rule the data plane honours. The documented limitation is about using a tag as the *target* of an ingress rule to a Cloud Run workload, not as the *source* of an ingress rule to a VM.
- Source tags do carry a real cost, and it is now in the comment beside the rule: an ingress rule keyed on a source tag is the one kind Google does not apply immediately. Editing it left the stage serving the old answer for minutes.

`sourceRanges: [10.20.0.0/20]` remains the documented alternative and is rejected on purpose: that subnet holds the runners, the ClickHouse host, the GKE nodes and every other Cloud Run service placed in it, so it admits far more than the control plane. A dedicated subnet per caller would fix that, and is a larger change than this defect warrants.

DorianZheng:
@boxliteai how GCP recommend to connect Cloud Run and Compute Engine network

DorianZheng:
@claude how GCP recommend to connect Cloud Run and Compute Engine network

boxliteai:
@DorianZheng sorry, I couldn't finish this one — the run failed on my side. Please try again in a bit.

<sub>📦 Ran in an isolated <a href="https://boxlite.ai">BoxLite</a> microVM · reply with a mention to follow up</sub>

apps/infra/mdeploy/stack/providers/gcp/runners.ts
DorianZheng, line 373: @boxliteai is this the best practice? runner will access external network frequently

boxliteai: @DorianZheng **yes, private runners with Cloud NAT are a good default even for frequent internet access.** Google’s [Terraform example](https://github.com/terraform-google-modules/terraform-google-cloud-nat/blob/main/examples/nat_with_compute_engine/main.tf#L35-L95) uses this layout.

The two traffic paths are separate:

```text
Cloud Run → Direct VPC egress → runner private IP
Runner    → Cloud NAT        → internet
```

Google [recommends Direct VPC egress](https://cloud.google.com/run/docs/configuring/connecting-vpc#direct-vpc) for the first path. Image pulls use the second.

**The throughput explanation in this comment needs correcting:**

- **Bandwidth:** Google explicitly states [“Cloud NAT does not reduce the network bandwidth per VM.”](https://cloud.google.com/nat/docs/overview#benefits) It is distributed; adding NAT addresses increases port capacity, not VM bandwidth.
- **Concurrent connections:** Our `network.ts:193` sets `AUTO_ONLY`, which scales NAT addresses. It does **not** enable dynamic per-VM port allocation. For busy runners, measure port usage and dropped connections, then size [port allocations](https://cloud.google.com/nat/docs/ports-and-addresses#ports) accordingly.
- **Cost:** [Public NAT charges $0.045/GiB processed](https://cloud.google.com/nat/pricing), including downloaded image bytes, plus gateway/address charges and applicable network transfer charges. High download volume makes this worth measuring.

I’d keep the private interface and replace the comment’s throughput claim with:

```ts
// Private runners use Cloud NAT for outbound internet access.
// Cloud NAT preserves per-VM bandwidth. For heavy outbound workloads,
// size NAT port allocations and account for per-GiB processing charges.
```

The similar “queue behind a shared NAT” explanation in `network.ts` should also be corrected.

Checked the PR diff with `git diff`, inspected configuration with `rg`/`sed`, and fetched Google’s documentation with `curl`. No code changed or live GCP benchmark run.

<sub>📦 Ran in an isolated <a href="https://boxlite.ai">BoxLite</a> microVM · reply with a mention to follow up</sub>

DorianZheng:
@copilot resolve the merge conflicts in this pull request

